PR #2003, opened 2026-05-10
feat(wam-r): lmdb('path') external fact-source backend (step 1, load-everything)

s243a:
## Summary

Closes step 1 of handoff item #1 (LMDB backend). Adds `lmdb('path')` as a third backend to `r_fact_sources`, alongside `file('data.csv')` and `grouped_by_first('data.tsv')`.

- **Step-1 semantics: load-everything.** Treats LMDB as a serialization format -- the runtime reads all key/value pairs from the LMDB env at program-load time and feeds them through the same `build_fact_indexes` + `fact_table_dispatch` pipeline as the other backends. Per-arg indexing, multi-solution backtracking, `iterate_goal` integration, and `fact_in_range/5` all work the same way regardless of backend.
- **Step-2 (probe-on-demand)** remains a follow-up. The Scala precedent's `lookupByArg1` / `streamAll` split is the memory-efficiency win for huge tables that don't fit in RAM; step-1 keeps the PR contained and lets users with already-LMDB-shaped data benefit from on-disk persistence today.

### Codegen
- New `fact_source_loader_call(lmdb(Path), ...)` clause emits a one-liner that calls the new runtime helper at program init.

### Runtime
- `WamRuntime$read_facts_lmdb(path, intern_table)` opens the env via `thor` (preferred) or `lmdbr` and decodes each value as a TAB-separated string of `tag:payload` fields (`a:string`, `i:int`, `f:float`). Returns an empty fact table with a stderr message if no R LMDB binding is installed -- downstream queries fail (no solutions) rather than erroring out, matching the absent-CSV-file degradation pattern.
- `WamRuntime$lmdb_write_facts(path, facts, intern_table)` is a setup-time helper for tests / users who want to convert a list of facts (or a CSV) into an LMDB env once at project init. Keys are `sprintf("%010d", i)` for natural cursor-walk ordering.

### E2E test
- `external_fact_source_lmdb_e2e_rscript` -- auto-skips when no `thor` / `lmdbr` is installed in the dev env (mirrors the `rscript_available` skip pattern). Seeds a 5-tuple LMDB env via a small thor-using Rscript at setup time, generates a project pointing to it, and asserts on direct queries, backtracking-must-fail, and two findall shapes. Mirrors the CSV / grouped-TSV e2e structure.

### Docs
- `docs/WAM_R_TARGET.md` "External fact sources" section gets the new backend, a value-encoding subsection, and an install-steps subsection (`apt install liblmdb-dev`, `R -e 'install.packages("thor")'`).
- Test catalog gets the new entry.

## Test plan

- [x] Full WAM-R generator suite: **76/76 pass** (75 prior + 1 new, which auto-skips since no R LMDB pkg is installed in the dev env).
- [x] Codegen path inspected via the new test's `sub_string` assertions on `pred_lpedge_facts <- WamRuntime$read_facts_lmdb(...)` and the `lowered_dispatch` registration line.
- [ ] **Not exercisable here.** Live LMDB read/write path is gated on `thor` / `lmdbr` being installed (neither is available in this sandbox -- no system liblmdb and no network access to install the R pkg). The test auto-skip cleanly hides this; for full e2e coverage, run on a machine with `apt install liblmdb-dev && R -e 'install.packages("thor")'`.

## Risks / follow-ups

- Step-2 probe-on-demand (`WamRuntime$lmdb_fact_dispatch` that skips the in-memory tuple list and probes by arg1 via LMDB key lookup) remains item #1's open follow-up.
- The runtime writer (`lmdb_write_facts`) and reader pair use a simple text encoding. If an atom name contains a literal TAB or newline the round-trip breaks; the CSV / grouped-TSV backends share the same implicit ASCII-safe-atom assumption, so this is documented-as-shipped rather than a regression.

https://claude.ai/code/session_01QLSbofTYaePTdPbhSkE6Sd

---
_Generated by [Claude Code](https://claude.ai/code/session_01QLSbofTYaePTdPbhSkE6Sd)_

### DIFF
--- a/docs/WAM_R_TARGET.md
+++ b/docs/WAM_R_TARGET.md
@@ -329,7 +329,7 @@ with the parameterized C# query runtime -- see
 ### External fact sources
 
 Mirrors the Scala target's `scala_fact_sources` option. Users can
-declare a predicate's facts as a separate file. Two backends are
+declare a predicate's facts as a separate file. Three backends are
 supported today:
 
 - `file('data.csv')` -- comma-separated rows, one fact per row.
@@ -337,6 +337,12 @@ supported today:
 - `grouped_by_first('data.tsv')` -- tab-separated rows shaped as
   `key<TAB>v1<TAB>v2<TAB>...`; the loader explodes each row into
   separate `(key, vK)` tuples. Arity-2 only.
+- `lmdb('path.lmdb')` -- on-disk LMDB env containing one fact per
+  key, value encoded as TAB-separated `tag:payload` fields. Any
+  arity. Requires an R LMDB binding (see *LMDB install* below);
+  loader returns an empty fact table with a warning if the binding
+  is absent. Step-1 semantics: load-everything (treats LMDB as a
+  serialization format); step-2 probe-on-demand is a follow-up.
 
 ```prolog
 :- write_wam_r_project(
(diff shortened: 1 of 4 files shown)


s243a:
**Update: live LMDB path now exercised, not just the auto-skip codepath.**

The sandbox blocks CRAN (403) but allows GitHub (200) and has `liblmdb-dev` in the local apt mirror, so I was able to install `thor` from source:

```sh
apt install liblmdb-dev
git clone https://github.com/richfitz/storr.git && R CMD INSTALL storr
git clone https://github.com/richfitz/thor.git  && R CMD INSTALL thor
```

With `thor` available:

- `external_fact_source_lmdb_e2e_rscript` alone: **passed in 6.5s** -- exercises the real `thor::mdb_env` seed write, the runtime `WamRuntime$read_facts_lmdb` read, plus all five assertion sub-cases (direct lookup, backtracking branch, must-fail, findall of partial bindings, findall of all tuples).
- Full WAM-R suite (76 tests) with `thor` installed: **passed in 663s**, no regressions.

So both paths are now verified:
- **With** `thor` / `lmdbr`: actual LMDB read/write works end-to-end.
- **Without**: test cleanly auto-skips via the `r_lmdb_pkg_available` guard (re-confirmed in the original 76/76 run before I installed thor).

No code change needed; this comment just adjusts the "not exercisable here" caveat in the original test plan.

---

PR #2019, opened 2026-05-11
feat(wam-r): mode-analysis phase 4 -- multi_clause_n + lowered_dispatch self-jump + TCO trampoline

s243a:
## Summary

Generalises phase-1's `multi_clause_1` to lower every clause inline when each is individually `wam_r_lowerable`, plus makes recursive self-calls jump straight back into the lowered fn through a tail-call trampoline. Compounds phase 3's `is/2` inline win on recursive arith-heavy workloads.

### (a) `multi_clause_n` lowered emission

When every clause of a multi-clause predicate is line-supported, the emitter drops all of them inline within one lowered function, with state snapshot/restore between attempts. The emitter helper is recursive over the clause list (base case: no more clauses, fall through to `return(FALSE)`; recursive case: emit inline-try, then state-restore, then recurse). When only clause 1 is lowerable, the emitter degrades gracefully to `multi_clause_1`.

### (b) `lowered_dispatch` self-jump

Phase-1 lowered fns were deliberately not registered in `program$lowered_dispatch`, so internal `call`/`execute` of a lowered pred went through the WAM array via `step`. Phase 4 registers `multi_clause_n` preds in `lowered_dispatch`, and the lowered emitter's `emit_call`/`emit_execute` consult it before the array path. `deterministic` and `multi_clause_1` preds stay off `lowered_dispatch` because their failure paths depend on `state$pc` pointing into the WAM array.

### (c) Tail-call trampoline (correctness)

First-cut phase 4 hit an R C-stack overflow at pn depth 2000 because each `execute pn/1` recursed *into* `lowered_pn_1`. Fixed by trampolining the self-jump through a state-borne signal:

- `state$tail_call` slot added to `WamRuntime$new_state`.
- `emit_execute` (the Prolog tail-call WAM op) sets `state$tail_call <- "X/Y"` and returns `TRUE` instead of invoking the lowered fn directly.
- `WamRuntime$invoke_lowered_with_tco(program, state, key)` is the trampoline -- a `repeat` loop that calls the fn, inspects `state$tail_call`, and dispatches the next iteration in-place (no R frame added).
- All entry points routed through the helper: runtime `Call`/`Execute`/`dispatch_call` handlers, lowered emitter's `emit_call`, and the per-pred wrapper.

pn at depth 2000 now runs to completion in bounded R stack.

### (d) Frame-shape fix (latent bug surfaced by (a))

Inline `allocate`/`deallocate` emission used field name `locals` while the runtime's `Allocate`/`Deallocate` handlers create/read `ys` + `cps_barrier`. Never surfaced under `multi_clause_1` because allocate-bearing clauses stayed in the array. Phase 4 puts them inline, so the inline ops now match the runtime exactly, including `state$shadow_frame` retention for post-`deallocate` Y reads.

## Measured impact

Alternating phase-3-baseline vs phase-4 runs on pn n=200 depth=2000:

| Run | PHASE3 baseline | PHASE4 |
|-----|-----------------|--------|
| 1 | 161.5s (warmup outlier) | 139.1s |
| 2 | 142.1s | 136.2s |
| 3 | 142.5s | 135.9s |
| 4 | 139.6s | 135.7s |

Steady-state means (dropping run-1 warmup): PHASE3 141.4s, PHASE4 135.9s. **Phase 4 is ~3.9% faster** with non-overlapping distributions (max PHASE4 136.2s < min PHASE3 139.6s).

Compounds with phase 3's gain on the same workload: cumulative **~7% wall-clock improvement** from phase 2 baseline to phase 4 end-to-end.

## Test plan

- [x] `mode_analysis_phase4_multi_clause_n_inlined` -- structural: 2-clause recursive pred emits BOTH clauses inline; lowered_dispatch registration present.
- [x] `mode_analysis_phase4_multi_clause_n_e2e_rscript` -- e2e: pn at depths 0/5/50 runs correctly via Rscript (exercises the recursive self-jump + trampoline).
- [x] **Full WAM-R suite: 83/83 pass** (81 prior + 2 new), ~614s wall-clock.
- [x] pn at depth 2000 runs to completion (validates the C-stack overflow fix end-to-end at the bench depth).
- [x] Updated `lowered_emitter_phase3` + `emit_mode_functions_lowers_multi_clause` to expect the new multi_clause_n shape (3-clause fact pred wam_r_choice_fact/1 now classifies as `multi_clause_n(3)`).

## What's next (phase 5)

`docs/design/WAM_R_MODE_ANALYSIS_PLAN.md` lists phase-5 candidates: `get_value` head-match specialisation, inlining more step-delegated ops in multi_clause_n bodies (the pn body still calls `WamRuntime$step` for `BuiltinCall(">/2", 2)` etc.), and whole-program mode inference.

https://claude.ai/code/session_01QLSbofTYaePTdPbhSkE6Sd

---
_Generated by [Claude Code](https://claude.ai/code/session_01QLSbofTYaePTdPbhSkE6Sd)_

### DIFF
--- a/docs/WAM_R_TARGET.md
+++ b/docs/WAM_R_TARGET.md
@@ -534,6 +534,27 @@ shorthand (input / output / any), the same convention
 [`design/WAM_R_MODE_ANALYSIS_PLAN.md`](design/WAM_R_MODE_ANALYSIS_PLAN.md)
 for the phase roadmap and measured impact.
 
+### `multi_clause_n` lowering + self-jump (phase 4)
+
+Generalises phase-1's `multi_clause_1` (clause 1 inline, clauses 2+
+via the WAM array): when every clause of a multi-clause predicate
+is individually `wam_r_lowerable`, all clauses are emitted inline
+within one lowered function, with state snapshot/restore between
+attempts. The emitter helper is recursive over the clause list.
+
+Additionally, `multi_clause_n` preds are registered in
+`program$lowered_dispatch` so internal `call`/`execute` (including
+recursive self-calls) re-enter the lowered fn directly instead of
+iterating the WAM array via `step`. The lowered emitter's
+`emit_call` / `emit_execute` consult `lowered_dispatch` before
+falling back to the array path. `deterministic` and
+`multi_clause_1` preds stay off `lowered_dispatch` because their
+failure paths assume `state$pc` points into the WAM array.
+
+This compounds the phase-3 `is/2` inline: every recursive call of
+an arith-heavy predicate now runs entirely on the lowered path,
+hitting the inline `is/2` emission on each iteration.
+
 ### `is/2` specialisation (phase 3)
 
 Two complementary specialisations targeting the `is/2` arithmetic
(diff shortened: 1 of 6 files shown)


s243a:
Closing as obsolete during stale-PR triage (2026-06-10).

WAM-R mode-analysis phase 4 has already been delivered on `main`, in a newer form than this branch:

- `src/unifyweaver/targets/wam_r_lowered_emitter.pl` on `main` implements `multi_clause_n` lowered emission.
- `docs/design/WAM_R_MODE_ANALYSIS_PLAN.md` marks phase 4 as delivered — `multi_clause_n`, the `get_value` head-match specialisation, and the measurement harness (`tests/benchmarks/wam_r_mode_analysis_bench.pl`).
- The deep-recursion safety problem this PR solved with a `state$tail_call` trampoline is handled on `main` with a different design (iter-style retry choice points), so the trampoline code would conflict with rather than complement the current implementation.

The branch also shares no merge base with current `main` (history restructured in May 2026), so it cannot be rebased mechanically. Nothing remains to salvage.

---

PR #2023, opened 2026-05-11
docs: scan-strategy + algorithm-manifest design (4 docs)

s243a:
## Summary

Four design docs capturing the design discussion that followed Phase 2c's cache-cost-model arc. Doc-only PR — no code changes.

The scan-strategy work is the next layer above the cost model (`CACHE_COST_MODEL_PHILOSOPHY`). The algorithm-manifest doc abstracts a meta-concern that surfaced during the discussion: the declarative split between *algorithm* (the logical problem specification) and *optimization manifest* (the operational completion).

| doc | lines | role |
| --- | --- | --- |
| `ALGORITHM_MANIFEST_SPECIFICATION.md` | 355 | new — `algorithm/2` + `algorithm_optimization/2` API |
| `SCAN_STRATEGY_PHILOSOPHY.md` | 259 | what / why / when; three retention modes |
| `SCAN_STRATEGY_SPECIFICATION.md` | 442 | data structures, formulas, options, composition |
| `SCAN_STRATEGY_IMPLEMENTATION_PLAN.md` | 457 | phased rollout P0–P7 |

Total: 1,513 lines of docs.

## Algorithm-manifest abstraction

Workload-author API (extends the existing `user:demand_filter_spec/2` pattern to cover all optimizations):

```prolog
:- algorithm(effective_distance, [
       kernel(category_ancestor/4),
       seeds(article/1),
       roots(root_category/1),
       max_depth(10),
       dimension_n(5)
   ]).

:- algorithm_optimization(effective_distance, [
       cache_strategy(auto),
       working_set_fraction(0.001),
       scan_strategy(auto),
       tree_cost_function(flux, [iterations(1), parent_decay(0.5), child_decay(0.3)]),
       tree_retention(snapshot_only),
       demand_filter_spec(hop_limit, [max_hops(10)])
   ]).
```

Codegen runs `load_algorithm_manifest/2` before any resolver. Caller-provided options win on conflict. Backwards-compatible.

**Terminology note** (the docs flag this explicitly): "algorithm" in these docs means the *logical* specification of a graph problem — closer to the Datalog/SQL sense of "a query" than to the algorithms-textbook sense of "a step-by-step procedure". The optimization manifest *completes* the definition by specifying how to compile it.

## Scan-strategy core insight

Three phases, three data structures:

| phase | structure | optimised for |
| --- | --- | --- |
| warm-build | sorted heap + parallel index | O(log N) ranked construction |
| snapshot | one-shot conversion | one-time |
| steady-state | fixed-bucket hashtable, overwrite-on-collision | O(1) lookup, miss-memoised |

Cache is a real cache: misses cursor to LMDB and memoise into the hashtable. Snapshot provides the *initial state*, not a frozen lookup table. Eviction policy is overwrite-on-collision — same mechanism as the existing L1/L2 caches in the codebase.

## Three tree retention modes

| mode | tree post-snapshot | use case |
| --- | --- | --- |
| `discard` | GC'd | default; minimum memory |
| `snapshot_only` | frozen ranked list retained | spark routing, static prioritisation |
| `live` | full heap + index kept | adaptive re-ranking, query-history-driven updates |

`live` mode has no current consumer; the data structures we pick during warm support it without architectural changes, so the door stays open for query-history-driven re-ranking, adaptive cost-function switching, etc.

## Cost-function strategy slot

Pluggable: `flux`, `hop_distance`, `semantic_similarity`. Each has its own parameter list. **Flux specifically** uses separate parent/child decay constants because category-graph branching is asymmetric (few parents per node, many children).

`iterations(N)` unified across access patterns: 1 = local one-hop (cursor mode) / one full pass (scan mode), N → ∞ = converged scoring.

## Stage-1 → Stage-2 crossover

Same crossover formula as the cost model:

```
K_cross = W_remaining / (bandwidth_eff * latency_eff)
```

When unvisited frontier exceeds the crossover, switch from cursor seeks to a single scan pass. Fixed-point loop bounded to ≤ 2 scans for the first cut.

## Implementation plan (P0–P7)

| phase | size | gates |
| --- | --- | --- |
| P0: Algorithm-manifest abstraction | small | prerequisite for everything below |
| P1: Pluggable cost-function slot | small | P0 |
| P2: At-scale validation re-ingest | small (tooling) | needs P0+P1 in for matrix bench |
| P3: Warm-build + snapshot core | medium | P1; P2 data informs tuning |
| P4: `scan_strategy(auto)` resolver + matrix-bench mode | small-medium | P3 |
| P5: Spark routing via retained ranking (`snapshot_only`) | medium | P3 |
| P6: Adaptive fixed-point iteration | small | P3, P4; data-driven |
| P7+: Live-mode tree updates | medium per-algorithm | research-y, deferred |

P0–P4 are the core deliverable. P5–P7 are extensions; treat as optional based on P3/P4 measurements.

## Composition with existing arcs

The scan-strategy options compose with the cost-model resolvers landed in Phase 2c:

```
load_algorithm_manifest         (new in P0)
  → resolve_auto_use_lmdb       (existing)
  → resolve_auto_cache_strategy (Phase 2c)
  → resolve_auto_lmdb_cache_mode (Phase 2c+)
  → resolve_auto_scan_strategy  (new in P4)
  → codegen
```

Each layer reads the merged option list and refines its own slice of the decision space.

## Open questions filed for measurement

- Default `warm_budget_nodes(N)` fraction at each scale.
- Default `parent_decay` / `child_decay` for flux (waiting on the branching-factor distribution from P2).
- `iterations(1)` vs `iterations(3)` accuracy/cost curve.
- `stage2_max_scans(N)` bound — does 2 capture the gains?
- Whether `semantic_similarity` is worth supporting at our scale.

## What this isn't

- Not a runtime auto-tuner — warm parameters are picked at codegen / workload-author time.
- Not a re-warming mechanism — steady-state cache learns via miss-memoisation but the tree doesn't re-build mid-workload (that's `live` mode, deferred).
- Not a correctness device — misses fall back to LMDB cursor and return correct results; the strategy affects performance only.

## See also

- `docs/design/CACHE_COST_MODEL_PHILOSOPHY.md` — the cost model this layer builds on.
- `docs/design/WAM_PERF_OPTIMIZATION_LOG.md` — Phase M / Phase L measurements that motivate the regimes.
- Existing patterns this work generalises: `user:demand_filter_spec/2`, `user:max_depth/1`, `statistics:declare_cache_hints/1`.

### DIFF
--- a/docs/design/ALGORITHM_MANIFEST_SPECIFICATION.md
+++ b/docs/design/ALGORITHM_MANIFEST_SPECIFICATION.md
@@ -1,0 +1,355 @@
+# Algorithm Optimization Manifest — Specification
+
+## Terminology note
+
+In this document and the scan-strategy docs that reference it,
+**"algorithm" means the *logical specification* of the graph
+problem** — the declarative *what*, not the operational *how*.
+This is closer to the Datalog / SQL sense (a query is an
+algorithm; its execution plan is a separate artefact) than the
+algorithms-textbook sense (an algorithm is a step-by-step
+procedure).
+
+An algorithm in our sense may *hint* at how it's computed (e.g. a
+recursive kernel suggests BFS-style expansion) but it doesn't
+fully specify the operational behaviour. The **optimization
+specification** completes the definition by saying which cache
+mode, scan strategy, cost function, etc. to use. Together,
+algorithm + optimization manifest = executable artefact.
+
+This split matters because the two halves have different owners
+(workload author vs perf engineer), different lifecycles (stable
+vs measurement-driven), and different concerns (correctness vs
+throughput). Mechanically the split is small; conceptually it's
+the central point of this doc.
+
+## What this is
+
+A workload-author-level mechanism for separating two concerns:
+
+1. **The algorithm** (logical specification): *what* is being
+   computed (a kernel, its inputs, its output shape).
+2. **The optimization manifest** (operational specification):
+   *how* the codegen should compile the algorithm (cost-model
+   knobs, cache mode, scan strategy, demand-filter spec, etc.).
+
+The split exists because the two have different owners and lifecycles:
+
+- Algorithm: stable, owned by whoever wrote the workload, rarely
+  changes.
+- Optimizations: tuning artefacts, owned by whoever cares about
+  perf, change with measurements / hardware / data scale.
+
+Without the split, every bench harness and every call site
+re-declares the same option bundle. With it, an algorithm is
+declared once with its current best-known optimization profile;
+any caller that names the algorithm gets the optimizations for
+free.
+
+This document specifies the mechanism. It's deliberately small —
+the abstraction is a thin layer over the existing option-list
+plumbing the resolvers already consume.
+
+## Connection to existing patterns
+
+This generalises three patterns already present in the codebase:
+
+| existing | how the manifest subsumes it |
+|---|---|
+| `user:demand_filter_spec(Strategy, Opts)` | an entry in the manifest's option list |
+| `user:max_depth/1`, `user:dimension_n/1` | algorithm-level metadata in `algorithm/2` |
+| `statistics:declare_cache_hints/1` | `workload_locality/1` entry in the manifest |
+
+The manifest is **additive**, not replacing existing mechanisms.
+The codegen still reads `user:demand_filter_spec/2`, etc., for
+backwards compatibility. New workloads use the manifest; old
+workloads keep working unchanged.
+
+## API
+
+Two user-level predicates. Both declared in workload files via
+`:- ...` directives (the standard Prolog idiom for facts about
+the workload).
+
+### `algorithm(+Name, +AlgorithmOpts) is det.`
+
+Declares an algorithm. `Name` is an atom identifying the
+algorithm (e.g. `effective_distance`, `transitive_closure`,
+`shortest_path`). `AlgorithmOpts` is an option list describing
+the algorithm's structural metadata:
+
+- `kernel(P/N)` — the recursive kernel predicate.
+- `seeds(P/N)` — the seed predicate (input).
+- `roots(P/N)` — the root predicate (target).
+- `max_depth(D)` — bound on recursion depth.
+- Other algorithm-specific keys (e.g. `dimension_n/1` for
+  effective_distance).
+
+Example:
+
+```prolog
+:- algorithm(effective_distance, [
+       kernel(category_ancestor/4),
+       seeds(article/1),
+       roots(root_category/1),
+       max_depth(10),
+       dimension_n(5)
+   ]).
+```
+
+At most one `algorithm/2` declaration per file. Multiple
+declarations with the same name in different files are an error;
+the codegen rejects compilation rather than silently picking one.
+
+### `algorithm_optimization(+Name, +OptList) is det.`
+
+Declares optimizations for an algorithm. `Name` matches an
+`algorithm/2` declaration. `OptList` contains any options that
+downstream resolvers consume.
+
+Example:
+
+```prolog
+:- algorithm_optimization(effective_distance, [
+       cache_strategy(auto),
+       working_set_fraction(0.001),
+       expected_query_count(10),
+       lmdb_cache_mode(auto),
+       workload_locality(unknown),
+       scan_strategy(auto),
+       tree_cost_function(flux, [iterations(1), parent_decay(0.5), child_decay(0.3)]),
+       tree_retention(snapshot_only),
+       warm_budget_nodes_fraction(0.1),
+       demand_filter_spec(hop_limit, [max_hops(10)])
+   ]).
+```
+
+Multiple `algorithm_optimization/2` facts for the same algorithm
+are allowed — their option lists concatenate with `option/3`
+first-match semantics. This lets workload authors split
+optimizations by concern:
+
+```prolog
+:- algorithm_optimization(effective_distance, [
+       cache_strategy(auto), working_set_fraction(0.001), ...
+   ]).
+:- algorithm_optimization(effective_distance, [
+       scan_strategy(auto), tree_cost_function(flux, [...]), ...
+   ]).
+```
+
+If a key appears in multiple facts, the first occurrence wins
+(matches `option/3` semantics). To override, place the overriding
+fact earlier or pass the option directly to the codegen entry
+point.
+
+## Merge semantics
+
+When the codegen enters `write_wam_haskell_project/3` (or
+`compile_wam_runtime_to_haskell/3`), the option pipeline is:
+
+```
+1. Caller-provided options: Options0
+2. Manifest options:        ManifestOpts
+                              = concat of all algorithm_optimization(Name, _)
+                                facts for the declared algorithm
+3. Merged options:          Options' = merge(Options0, ManifestOpts)
+   - For each key in either:
+       if Options0 has key  → take from Options0
+       else                 → take from ManifestOpts
+4. Resolvers run on Options' in their existing order.
+```
+
+**Caller wins on conflict.** This is the override path: a bench
+that wants to experiment with `tree_retention(live)` passes that
+option directly, and it beats the manifest's `snapshot_only`.
+
+**Manifest is best-known-defaults, not policy.** The manifest is
+a recommendation derived from measurement and cost-model logic.
+Callers should feel free to override when they have a reason.
+
+## Composition with resolvers
+
+The manifest sits *before* the existing resolvers in the option
+pipeline. By the time `resolve_auto_cache_strategy/2`,
+`resolve_auto_lmdb_cache_mode/2`, etc., run, they see a merged
+option list with manifest defaults filled in.
+
+```
+Options0 (caller)
+  → merge with ManifestOpts
+  → resolve_auto_use_lmdb
+  → resolve_auto_cache_strategy
+  → resolve_auto_lmdb_cache_mode
+  → resolve_auto_scan_strategy   (new in P4 of the scan-strategy plan)
+  → codegen
+```
+
+The resolvers themselves don't change. The manifest just provides
+their inputs.
+
+## Where the manifest is read
+
+A small Prolog helper, conceptually:
+
+```prolog
+%% load_algorithm_manifest(+Options0, -OptionsWithManifest) is det.
+%
+%  Reads user:algorithm/2 and user:algorithm_optimization/2 (if
+%  declared) and merges the optimization options into the codegen's
+%  option list. Caller-provided options always win on conflict.
+load_algorithm_manifest(Options0, OptionsWithManifest) :-
+    (   user:algorithm(Name, _AlgOpts)
+    ->  findall(Opts, user:algorithm_optimization(Name, Opts), AllOpts),
+        flatten(AllOpts, ManifestOpts),
+        merge_options(Options0, ManifestOpts, OptionsWithManifest)
+    ;   OptionsWithManifest = Options0
+    ).
+```
+
+`merge_options/3` is implemented to give caller options precedence
+on duplicate keys (this is *not* the SWI-Prolog library
+`merge_options/3` — that one's argument order convention is
+different; we'd implement our own or pick a clear name).
+
+This helper runs once at the top of
+`write_wam_haskell_project/3` (and `compile_wam_runtime_to_haskell/3`
+for the codegen-test path), before any resolver.
+
+## Edge cases
+
+### No algorithm declared
+
+The codegen continues to work without a manifest — `Options0`
+flows straight through. Existing workloads that don't use
+manifests are unaffected.
+
+### Algorithm declared but no optimization manifest
+
+Same as caller-provided options only. The `algorithm/2` declaration
+itself is informational (the codegen reads its `kernel/1`,
+`seeds/1`, etc. — same way it currently reads `user:max_depth/1`),
+but no optimization options are merged.
+
+### Multiple algorithms declared in one workload
+
+Currently an error. A workload that genuinely computes two
+algorithms in one binary is rare; the design defers the
+multi-algorithm case to a future extension (probably named
+manifests selected at call time). For now: one algorithm per
+workload.
+
+### Optimization for an undeclared algorithm
+
+`algorithm_optimization(foo, _)` without a matching `algorithm(foo,
+_)` is a warning. The optimization options are ignored; the
+codegen logs that the manifest was orphaned.
+
+### Manifest option that no resolver understands
+
+Unknown keys are passed through. If a future resolver adds support,
+existing manifests pick it up without changes. If a resolver
+removes support for a key, the unknown entry is silently ignored
+(or warned, depending on strict-mode setting).
+
+### Conflicting `algorithm_optimization/2` facts
+
+Multiple facts concat with first-match semantics. If two facts
+both set `cache_strategy/1` to different values, the first one
+encountered wins. Workload authors who need explicit ordering
+should consolidate into a single fact.
+
+## Example: full workload file
+
+```prolog
+:- module(workload_effective_distance, []).
+
+%% Algorithm declaration
+:- algorithm(effective_distance, [
+       kernel(category_ancestor/4),
+       seeds(article/1),
+       roots(root_category/1),
+       max_depth(10),
+       dimension_n(5)
+   ]).
+
+%% Optimization manifest, split by concern
+:- algorithm_optimization(effective_distance, [
+       %% Cache cost-model
+       cache_strategy(auto),
+       working_set_fraction(0.001),
+       expected_query_count(10),
+       mem_available_bytes(8_000_000_000)  % overrides /proc/meminfo
+   ]).
+:- algorithm_optimization(effective_distance, [
+       %% Cache tier
+       lmdb_cache_mode(auto),
+       workload_locality(unknown),
+       cache_tier_floor_bytes(8_000_000)
+   ]).
+:- algorithm_optimization(effective_distance, [
+       %% Scan strategy
+       scan_strategy(auto),
+       tree_cost_function(flux, [
+           iterations(1),
+           parent_decay(0.5),
+           child_decay(0.3),
+           flux_merge(sum)
+       ]),
+       tree_retention(snapshot_only),
+       warm_budget_nodes_fraction(0.1)
+   ]).
+:- algorithm_optimization(effective_distance, [
+       %% Demand filter
+       demand_filter_spec(hop_limit, [max_hops(10)])
+   ]).
+
+%% Kernel definition (unchanged)
+category_ancestor(Cat, Root, Hops, Visited) :-
+    category_parent(Cat, Mid),
+    \+ member(Mid, Visited),
+    category_ancestor(Mid, Root, H, [Mid|Visited]),
+    Hops is H + 1.
+```
+
+A bench harness or end-user code that wants to run this workload
+now just names the algorithm:
+
+```prolog
+:- use_module(workload_effective_distance).
+:- write_wam_haskell_project(
+       [category_ancestor/4, ...],
+       [],  % no caller-level options — manifest provides everything
+       'output/effective_distance_bench'
+   ).
+```
+
+To experiment with a different optimization:
+
+```prolog
+:- write_wam_haskell_project(
+       [category_ancestor/4, ...],
+       [tree_retention(live)],  % overrides manifest's snapshot_only
+       'output/effective_distance_bench_live'
+   ).
+```
+
+## What this isn't
+
+- **Not a DSL for algorithms.** `algorithm/2` is a metadata
+  declaration, not a language for *defining* algorithms. The
+  kernel still lives in normal Prolog clauses.
+- **Not a runtime mechanism.** Manifests are read at codegen time
+  only. They don't affect runtime behaviour directly; their effect
+  is mediated through the options they feed to the resolvers.
+- **Not a replacement for explicit options.** Callers retain full
+  control via the caller-wins merge rule.
+
+## See also
+
+- `docs/design/CACHE_COST_MODEL_PHILOSOPHY.md` — the cost-model
+  resolvers that consume manifest entries.
+- `docs/design/SCAN_STRATEGY_SPECIFICATION.md` — the scan-strategy
+  options that go in a manifest.
+- `docs/design/SCAN_STRATEGY_IMPLEMENTATION_PLAN.md` — Phase 0
+  introduces the manifest abstraction; later phases assume it.
--- a/docs/design/ALGORITHM_MANIFEST_SPECIFICATION.md
+++ b/docs/design/ALGORITHM_MANIFEST_SPECIFICATION.md
@@ -1,5 +1,15 @@
 # Algorithm Optimization Manifest — Specification
 
+## Naming note
+
+The user-level directives are `decl_algorithm/2` and
+`decl_algorithm_optimization/2`. The `decl_` prefix signals that
+these are *declarative directives* (facts that describe metadata
+about the workload), not callable predicates. The prefix also
+avoids any collision with library predicates that might define a
+plain `algorithm/2` — defensive even if no current standard
+library uses that name.
+
 ## Terminology note
 
 In this document and the scan-strategy docs that reference it,
@@ -57,7 +67,7 @@ This generalises three patterns already present in the codebase:
 | existing | how the manifest subsumes it |
 |---|---|
 | `user:demand_filter_spec(Strategy, Opts)` | an entry in the manifest's option list |
-| `user:max_depth/1`, `user:dimension_n/1` | algorithm-level metadata in `algorithm/2` |
+| `user:max_depth/1`, `user:dimension_n/1` | algorithm-level metadata in `decl_algorithm/2` |
 | `statistics:declare_cache_hints/1` | `workload_locality/1` entry in the manifest |
 
 The manifest is **additive**, not replacing existing mechanisms.
@@ -71,7 +81,7 @@ Two user-level predicates. Both declared in workload files via
 `:- ...` directives (the standard Prolog idiom for facts about
 the workload).
 
-### `algorithm(+Name, +AlgorithmOpts) is det.`
+### `decl_algorithm(+Name, +AlgorithmOpts) is det.`
 
 Declares an algorithm. `Name` is an atom identifying the
 algorithm (e.g. `effective_distance`, `transitive_closure`,
@@ -88,7 +98,7 @@ the algorithm's structural metadata:
 Example:
 
 ```prolog
-:- algorithm(effective_distance, [
+:- decl_algorithm(effective_distance, [
        kernel(category_ancestor/4),
        seeds(article/1),
        roots(root_category/1),
@@ -97,20 +107,20 @@ Example:
    ]).
 ```
 
-At most one `algorithm/2` declaration per file. Multiple
+At most one `decl_algorithm/2` declaration per file. Multiple
 declarations with the same name in different files are an error;
 the codegen rejects compilation rather than silently picking one.
 
-### `algorithm_optimization(+Name, +OptList) is det.`
+### `decl_algorithm_optimization(+Name, +OptList) is det.`
 
 Declares optimizations for an algorithm. `Name` matches an
-`algorithm/2` declaration. `OptList` contains any options that
+`decl_algorithm/2` declaration. `OptList` contains any options that
 downstream resolvers consume.
 
 Example:
 
 ```prolog
-:- algorithm_optimization(effective_distance, [
+:- decl_algorithm_optimization(effective_distance, [
        cache_strategy(auto),
        working_set_fraction(0.001),
        expected_query_count(10),
@@ -119,21 +129,30 @@ Example:
        scan_strategy(auto),
        tree_cost_function(flux, [iterations(1), parent_decay(0.5), child_decay(0.3)]),
        tree_retention(snapshot_only),
+       %% warm_budget_nodes_fraction(F) is a manifest-friendly alias
+       %% that scan_strategy's resolver converts to an absolute
+       %% warm_budget_nodes(N) before codegen. See the option list in
+       %% SCAN_STRATEGY_SPECIFICATION.md for both forms.
        warm_budget_nodes_fraction(0.1),
        demand_filter_spec(hop_limit, [max_hops(10)])
    ]).
 ```
 
-Multiple `algorithm_optimization/2` facts for the same algorithm
-are allowed — their option lists concatenate with `option/3`
-first-match semantics. This lets workload authors split
+Multiple `decl_algorithm_optimization/2` facts for the same algorithm
+are allowed — their option lists are concatenated in declaration
+order using `append/2` (shallow concat — see the `load_algorithm_manifest/2`
+sketch below for the trap with `flatten/2`). The manifest itself
+does no de-duplication at concat time. When the merged list is
+later read via SWI-Prolog's `option/3` accessor, the first
+occurrence of any key wins — that's `option/3`'s documented
+behaviour for repeated keys. This lets workload authors split
 optimizations by concern:
 
 ```prolog
-:- algorithm_optimization(effective_distance, [
+:- decl_algorithm_optimization(effective_distance, [
        cache_strategy(auto), working_set_fraction(0.001), ...
    ]).
-:- algorithm_optimization(effective_distance, [
+:- decl_algorithm_optimization(effective_distance, [
        scan_strategy(auto), tree_cost_function(flux, [...]), ...
    ]).
 ```
@@ -151,7 +170,7 @@ When the codegen enters `write_wam_haskell_project/3` (or
 ```
 1. Caller-provided options: Options0
 2. Manifest options:        ManifestOpts
-                              = concat of all algorithm_optimization(Name, _)
+                              = concat of all decl_algorithm_optimization(Name, _)
                                 facts for the declared algorithm
 3. Merged options:          Options' = merge(Options0, ManifestOpts)
    - For each key in either:
@@ -195,26 +214,46 @@ A small Prolog helper, conceptually:
 ```prolog
 %% load_algorithm_manifest(+Options0, -OptionsWithManifest) is det.
 %
-%  Reads user:algorithm/2 and user:algorithm_optimization/2 (if
+%  Reads user:decl_algorithm/2 and user:decl_algorithm_optimization/2 (if
 %  declared) and merges the optimization options into the codegen's
 %  option list. Caller-provided options always win on conflict.
 load_algorithm_manifest(Options0, OptionsWithManifest) :-
-    (   user:algorithm(Name, _AlgOpts)
-    ->  findall(Opts, user:algorithm_optimization(Name, Opts), AllOpts),
-        flatten(AllOpts, ManifestOpts),
-        merge_options(Options0, ManifestOpts, OptionsWithManifest)
+    (   user:decl_algorithm(Name, _AlgOpts)
+    ->  findall(Opts, user:decl_algorithm_optimization(Name, Opts), AllOpts),
+        %% IMPORTANT: shallow concat, not deep flatten. flatten/2 would
+        %% destructure option terms whose values are themselves lists —
+        %% e.g. `tree_cost_function(flux, [iterations(1), parent_decay(0.5)])`
+        %% would have its inner parameter list pulled into the outer
+        %% option list and corrupt the structure.
+        append(AllOpts, ManifestOpts),
+        merge_caller_and_manifest_options(Options0, ManifestOpts,
+                                          OptionsWithManifest)
     ;   OptionsWithManifest = Options0
     ).
 ```
 
-`merge_options/3` is implemented to give caller options precedence
-on duplicate keys (this is *not* the SWI-Prolog library
-`merge_options/3` — that one's argument order convention is
-different; we'd implement our own or pick a clear name).
+The merge helper is `merge_caller_and_manifest_options/3` — same
+name used in `SCAN_STRATEGY_IMPLEMENTATION_PLAN.md` P0. We
+deliberately avoid the bare name `merge_options/3` because
+SWI-Prolog's `library(option)` exports a different
+`merge_options/3` with reversed argument-order conventions;
+sharing the name would invite confusion at use sites.
 
-This helper runs once at the top of
-`write_wam_haskell_project/3` (and `compile_wam_runtime_to_haskell/3`
-for the codegen-test path), before any resolver.
+This helper runs once at the head of the codegen pipeline. There
+are two public entry points:
+
+- `write_wam_haskell_project/3` — the end-to-end path used by
+  bench harnesses and users.
+- `compile_wam_runtime_to_haskell/3` — the inner path used by
+  codegen tests that don't want to write a project to disk.
+
+`write_wam_haskell_project/3` calls `compile_wam_runtime_to_haskell/3`
+internally, so if the manifest load lives only at the inner entry
+point, both paths benefit. The current implementation puts it
+explicitly at both for symmetry with the existing cost-model
+resolver wiring (`resolve_auto_use_lmdb/2` already lives at both
+sites); P0's deliverables include making `load_algorithm_manifest/2`
+idempotent so the duplicate call in the outer path is harmless.
 
 ## Edge cases
 
@@ -226,7 +265,7 @@ manifests are unaffected.
 
 ### Algorithm declared but no optimization manifest
 
-Same as caller-provided options only. The `algorithm/2` declaration
+Same as caller-provided options only. The `decl_algorithm/2` declaration
 itself is informational (the codegen reads its `kernel/1`,
 `seeds/1`, etc. — same way it currently reads `user:max_depth/1`),
 but no optimization options are merged.
@@ -252,7 +291,7 @@ existing manifests pick it up without changes. If a resolver
 removes support for a key, the unknown entry is silently ignored
 (or warned, depending on strict-mode setting).
 
-### Conflicting `algorithm_optimization/2` facts
+### Conflicting `decl_algorithm_optimization/2` facts
 
 Multiple facts concat with first-match semantics. If two facts
 both set `cache_strategy/1` to different values, the first one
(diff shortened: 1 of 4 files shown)


docs/design/ALGORITHM_MANIFEST_SPECIFICATION.md
```diff
@@ -0,0 +1,355 @@
...
+`shortest_path`). `AlgorithmOpts` is an option list describing
+the algorithm's structural metadata:
+
+- `kernel(P/N)` — the recursive kernel predicate.
+- `seeds(P/N)` — the seed predicate (input).
+- `roots(P/N)` — the root predicate (target).
+- `max_depth(D)` — bound on recursion depth.
+- Other algorithm-specific keys (e.g. `dimension_n/1` for
+  effective_distance).
+
```
s243a, line 97: **`algorithm/2` arity confusion**: The spec says "at most one `algorithm/2` declaration per file" but allows multiple `algorithm_optimization/2` facts. Consider clarifying whether the file-scoped constraint is intentional or if it should be module/workload-scoped. Large workloads might span multiple files but logically describe a single algorithm.

Also, the `algorithm/2` predicate name collides with SWI-Prolog's built-in `algorithm/2` from `aggregate`. Consider a namespaced variant like `wam_algorithm/2` or wrapping it as `:- decl_algorithm(Name, Opts)` to make the directive nature clearer.

docs/design/ALGORITHM_MANIFEST_SPECIFICATION.md
```diff
@@ -0,0 +1,355 @@
...
+## Merge semantics
+
+When the codegen enters `write_wam_haskell_project/3` (or
+`compile_wam_runtime_to_haskell/3`), the option pipeline is:
+
+```
+1. Caller-provided options: Options0
+2. Manifest options:        ManifestOpts
+                              = concat of all algorithm_optimization(Name, _)
```
s243a: **`merge_options/3` naming conflict is documented but unresolved**: The spec explicitly notes this predicate name differs from SWI-Prolog's library `merge_options/3` and says "we'd implement our own or pick a clear name". Since this is a spec doc, it should commit to a name rather than leaving it open. Suggestion: `manifest_merge_options/3` or `merge_caller_wins/3` — the latter makes the semantics self-documenting.

The implementation plan (P0 deliverables) uses `merge_caller_and_manifest_options/3` which is clearer but verbose. Either way, reconcile the name across both docs before implementation.

docs/design/ALGORITHM_MANIFEST_SPECIFICATION.md
```diff
@@ -0,0 +1,355 @@
...
+codegen logs that the manifest was orphaned.
+
+### Manifest option that no resolver understands
+
+Unknown keys are passed through. If a future resolver adds support,
+existing manifests pick it up without changes. If a resolver
+removes support for a key, the unknown entry is silently ignored
+(or warned, depending on strict-mode setting).
+
+### Conflicting `algorithm_optimization/2` facts
```
s243a: **The full workload example splits optimizations across 4 separate `algorithm_optimization/2` facts by concern (cache cost-model, cache tier, scan strategy, demand filter) — this is a great pattern.** One minor issue: the memory-override comment uses `8_000_000_000` (Go/Haskell numeric literal style) which is not valid in SWI-Prolog. In SWI-Prolog, use `8000000000` or define a helper constant. Worth correcting in the example to avoid confusion.

docs/design/SCAN_STRATEGY_SPECIFICATION.md
```diff
@@ -0,0 +1,442 @@
...
+The cache is **memoised on miss**. A repeated query for the same
+node hits even when the cost function didn't predict it. This is
+the standard caching contract; the warm snapshot just provides
+the *initial state* of the cache, not its frozen contents.
+
+If a snapshot entry collides with a miss-memoised insert, the new
+insert wins the bucket. That's tolerable: re-cursoring the lost
+snapshot entry on a later query is a single seek, symmetric to any
+other miss. We're not trying to permanently protect the snapshot
+from the workload's access pattern; the cache is a fixed-capacity
```
s243a, line 92: **Flux formula asymmetry is well-motivated, but the formula has an indexing ambiguity**: `hops_from_endpoint(p)` is referenced for a *parent* of `n`, but which endpoint? If the graph is bidirectional and `n` can be both an ancestor and descendant of the root, the "nearest endpoint" is ambiguous when there are multiple roots.

Suggestion: clarify whether `hops_from_endpoint` is computed from the *seed* set, the *roots* set, or the nearest of either. A note like "endpoint = nearest seed node via BFS distance" would make this unambiguous for implementors.

docs/design/SCAN_STRATEGY_SPECIFICATION.md
```diff
@@ -0,0 +1,442 @@
...
+
+### Flux
+
+For each candidate node `n`, flux is the sum of two single-direction
+contributions:
+
+```
+flux(n) = parent_flux(n) + child_flux(n)
+
+parent_flux(n) = Σ_{p ∈ parents(n)}  (parent_decay / |children(p)|) ^ hops_from_endpoint(p)
```
s243a, line 139: **`WarmHeap` type sketch uses `Data.Heap.MinPrioHeap` but the eviction section describes evicting the _lowest_-scored node when at capacity — a max-heap behaviour from the cache's perspective.**

For cache warming, we want to *keep* the highest-scored nodes, so we evict the minimum. A `MinPrioHeap` (min at top) supports O(1) peek-min and O(log N) evict-min, which is correct here. However the spec should make this explicit: "min-heap keyed by cost; on insertion when at capacity, evict the minimum (lowest-cost) entry." Currently a reader might misread it as needing a max-heap.

docs/design/SCAN_STRATEGY_SPECIFICATION.md
```diff
@@ -0,0 +1,442 @@
...
+path; workload-authors who need converged scoring opt up.
+
+## Stage-1 → Stage-2 crossover
+
+The runtime tracks frontier size. When the frontier — i.e. *number
+of unvisited candidate nodes whose scores indicate worthwhile
+materialisation* — exceeds a threshold, switch from cursor seeks
+to a scan pass.
+
+```
```
s243a, line 216: **`K_cross_stage2` formula borrows from the cost model but the inputs diverge slightly**: The cost model's `K_cross` is computed at codegen time from static estimates. Here `W_remaining` is a *runtime* quantity (bytes of edge data not yet materialised). This is a key distinction that should be called out explicitly — the stage-2 crossover is a *runtime check* against a *statically derived threshold*.

Recommend adding a sentence clarifying: "The threshold `K_cross_stage2` is derived at codegen time using the same formula as the cost model. The frontier size `|Frontier|` is measured at runtime and compared against the precomputed threshold." This prevents implementors from trying to recompute the threshold dynamically.

docs/design/SCAN_STRATEGY_SPECIFICATION.md
```diff
@@ -0,0 +1,442 @@
...
+Two distinct policies, for two distinct structures:
+
+### Tree eviction (warm phase)
+
+When the heap is at capacity:
+
+- Compute candidate score.
+- Peek at the heap's lowest-scored element.
+- If candidate score > lowest, evict lowest, insert candidate.
+- Else discard candidate without inserting.
```
s243a, line 268: **`cache_capacity_buckets` default of `2 × warm_budget_nodes` is underdefined for collision probability.** A fixed-bucket hashtable with overwrite-on-collision has expected collision rate ≈ `N/B` where N = entries, B = buckets. At `B = 2N`, that's ~50% collision rate at snapshot time, meaning ~half the snapshot entries immediately displace each other during the initial population pass.

Consider either:
1. Increasing the default multiplier (e.g. `4×` gives ~25% collision density at snapshot time), or
2. Explicitly documenting the expected collision rate at the default and why it's acceptable (e.g., "we expect high miss-memoisation churn to overwrite snapshot entries anyway").

This is a correctness-adjacent concern: the snapshot's value comes from pre-populating the cache, but if half the entries collide during population, the warm benefit is halved.

docs/design/SCAN_STRATEGY_SPECIFICATION.md
```diff
@@ -0,0 +1,442 @@
...
+
+Implemented as a follow-up consumer in the implementation plan
+(P5); the snapshot just produces the view.
+
+### `live`
+
+`WarmHeap` and `WarmIndex` persist past snapshot. Algorithms can
+update the tree during steady-state via:
+
+```haskell
```
s243a, line 373: **Thread safety for `live` mode is deferred to P7+, but the `snapshot_only` mode (P5) also has a subtle concurrency concern**: `RankedView` is described as `[(NodeId, Cost)]` — an immutable list, so reads are safe. However, `partitionByRank` is called from within `mapConcurrently` (per the implementation plan). If the ranked view is produced by one thread and consumed by multiple, the handoff point needs to be documented.

Since it's immutable, this is safe in Haskell's GC model, but the spec should state this explicitly to close the loop: "`RankedView` is immutable post-snapshot; concurrent reads from multiple capabilities are safe without synchronisation."

docs/design/SCAN_STRATEGY_PHILOSOPHY.md
```diff
@@ -0,0 +1,259 @@
...
+
+Treating these as one structure forces a tradeoff: O(1)
+hashtable lookup competes with O(log N) heap-ordered eviction.
+Separating them lets each phase use the right shape.
+
+## Architectural alternatives we considered
+
+### 1. Tree-as-cache (single structure)
+
+Keep one data structure across all phases — a hybrid hashtable +
```
s243a, line 77: **Excellent framing of the core motivation** — the shift from "cursor vs in_memory" to "cursor vs cursor-then-scan" cleanly captures why Phase 2c's static resolver is insufficient at workloads with unknown frontier growth.

Minor suggestion: the phrase "absorbing the remaining workload in one scan" could be clarified as "absorbing the *frontier*'s edge data in one sequential scan pass" since the workload itself isn't absorbed — it's the pending edge-loading that triggers the crossover.

docs/design/SCAN_STRATEGY_PHILOSOPHY.md
```diff
@@ -0,0 +1,259 @@
...
+function works best. Flux makes sense for graph-shaped workloads
+where branching-factor decay tracks relevance. Hop distance is
+cheap and works for symmetric reachability problems. Semantic
+similarity needs embeddings but may dominate when those exist.
+Hardcoding forecloses A/B comparison; pluggable lets the workload
+author choose and lets the implementation evolve.
+
+## Separating the algorithm from its optimizations
+
+A meta-concern that surfaced during the design discussion: the
```
s243a, line 134: **Alternative 2 (Tree + L2 spillover) dismissal is reasonable but the reasoning conflates two separate cases**: "If predictions are poor, fix the cost function" is valid for *systematic* mis-prediction (the cost function is wrong for this workload type). But miss-memoisation in steady-state already handles *random* cache misses — so the real question is whether a *second structural miss path* adds value beyond the existing overwrite-on-collision cache.

The current design's answer (miss memoises back into the same hashtable) is actually a form of implicit spillover. Worth acknowledging this connection: "The overwrite-on-collision cache with miss memoisation is effectively a single-level spillover — misses become future hits if the same node is queried again."

docs/design/SCAN_STRATEGY_IMPLEMENTATION_PLAN.md
```diff
@@ -0,0 +1,457 @@
...
+    caller-wins merge semantics (not the SWI library
+    `merge_options/3`, whose convention is different).
+  - Validation: warn on orphan optimization, error on duplicate
+    `algorithm/2` declarations.
+- `src/unifyweaver/targets/wam_haskell_target.pl` — call
+  `load_algorithm_manifest/2` at the top of
+  `write_wam_haskell_project/3` and
+  `compile_wam_runtime_to_haskell/3`, before any resolver.
+- `tests/core/test_algorithm_manifest.pl` — unit tests:
+  - manifest absent → Options0 unchanged
```
s243a, line 53: **P0 success criteria references "167+ existing WAM-Haskell tests"** — this is a good concrete gate. Consider also adding a negative test: *a workload with `algorithm/2` pointing to a non-existent predicate should fail at codegen time with a clear error*, not silently produce incorrect code. The current edge-case section covers "optimization for undeclared algorithm" but not "algorithm declares a kernel predicate that doesn't exist in the module."

docs/design/SCAN_STRATEGY_IMPLEMENTATION_PLAN.md
```diff
@@ -0,0 +1,457 @@
...
+  emits the same Haskell code as the equivalent caller-options
+  invocation.
+
+### Estimated effort
+
+~1 session.
+
+---
+
+## Phase 1 — Pluggable cost-function slot
```
s243a, line 88: **P1 scope note: "The Flux implementation matches today's stub *exactly*"** — this is good discipline, but the stub path is `error "Flux strategy not implemented (Phase 2.5)"`. If P1 lands `hop_distance` as a working alternative and routes Flux through the new abstraction, it would be easy for a test to accidentally exercise the new `CostFn` dispatch path and hit the panic.

Suggestion: add a test that *explicitly* verifies the Flux-panic still fires through the new abstraction layer (not just the old direct call), so refactoring the dispatch doesn't silently swallow the error.

docs/design/SCAN_STRATEGY_IMPLEMENTATION_PLAN.md
```diff
@@ -0,0 +1,457 @@
...
+### Deliverables
+
+No code changes. Tooling and measurement only.
+
+- Re-run streaming pipeline to produce
+  `data/benchmark/simplewiki_cats/lmdb_proj/lmdb` and
+  `data/benchmark/enwiki_cats/lmdb_proj/lmdb`.
+- Run `convert_lmdb_to_phase1_layout.py` to produce
+  `lmdb_proj_resident` variants.
+- Generate matrix-bench projects at 1k, simplewiki, enwiki
```
s243a, line 157: **P3 memory estimate is helpful and largely correct, but missing one term**: The estimate accounts for `WarmHeap`, `WarmIndex`, `Frontier`, and `Visited` but not the edge data loaded during warm. For each of the `warm_budget_nodes` nodes, we load its edges into the heap entry. At simplewiki with average 5 children/node and say 16 bytes per edge reference, that's an additional `30k × 5 × 16 = 2.4 MB` — not large, but for enwiki at 990k nodes it becomes `~80 MB` additional. Worth adding `|WarmHeap| × avg_edges × bytes_per_edge` to the formula.

docs/design/SCAN_STRATEGY_IMPLEMENTATION_PLAN.md
```diff
@@ -0,0 +1,457 @@
...
+
+Haskell side (`templates/targets/haskell_wam/`):
+
+- `scan_strategy.hs.mustache` — new template with:
+  - `WarmHeap`, `WarmIndex`, `Frontier`, `Visited` types
+  - `runWarmBuild :: CostFn -> WarmConfig -> IO (WarmHeap, WarmIndex)`
+  - `snapshotCache :: WarmHeap -> IntMap NodeId [EdgeTarget]`
+  - `snapshotRanked :: WarmHeap -> [(NodeId, Cost)]`
+  - Stage-1 cursor loop + stage-2 scan trigger
+- `main.hs.mustache` — wire scan-strategy path conditionally on
```
s243a, line 216: **P4 resolver decision: "picks cost-function based on availability of embeddings (semantic if available, hop_distance otherwise) and graph properties (flux if branching distribution exists from P2; hop_distance fallback)"** — this creates a two-level heuristic that isn't specified in `SCAN_STRATEGY_SPECIFICATION.md`'s option list. The spec says `scan_strategy(auto)` picks options, but doesn't define the auto-selection logic for `tree_cost_function`.

Suggestion: add a subsection to `SCAN_STRATEGY_SPECIFICATION.md` (or note it here) describing the auto-selection priority: `semantic > flux (if branching data) > hop_distance`. This makes the resolver behaviour testable against the spec.

s243a (commented):
## Review: docs/scan-strategy-design (4 docs)

This is a well-structured, thorough design package. The separation into Philosophy / Specification / Implementation Plan / Manifest Specification is the right decomposition — the Philosophy doc reads cleanly as a standalone motivation, and the Spec is precise enough to implement from. The algorithm-manifest abstraction is a genuinely good catch: it cleanly separates the "what" from the "how to compile" concerns, and the caller-wins merge semantics are exactly right for a defaults-with-escape-hatch system.

**Strengths:**
- The three-phase warm/snapshot/steady-state architecture is well-motivated and the tradeoffs vs alternatives are clearly documented
- Flux's asymmetric parent/child decay for Wikipedia category graphs is the right call given the branching factor data from Phase M/L measurements
- The implementation plan has concrete, testable success criteria per phase; P0–P4 as core + P5–P7 as optional is sensible scoping
- The `live` mode deferral is the right call — the door is kept open without over-building
- The cross-cutting section (determinism, memory budgets, telemetry, out-of-scope) is excellent — prevents scope creep in implementation PRs

**Issues requiring attention before implementation (inline comments):**
1. `algorithm/2` name may collide with SWI-Prolog aggregation library — consider namespacing
2. `merge_options/3` name is left unresolved between the two docs — pick one name
3. `8_000_000_000` numeric literal in Prolog example is not valid SWI-Prolog syntax
4. Flux formula's `hops_from_endpoint` needs clarification for multi-root graphs
5. `cache_capacity_buckets` default of `2×` gives ~50% collision at snapshot time — worth revisiting the multiplier or documenting the tradeoff
6. `K_cross_stage2` static vs runtime distinction should be made explicit
7. P4's `scan_strategy(auto)` cost-function auto-selection logic is not captured in the Spec doc

**Minor suggestions** also inline (philosophy framing, memory estimate completeness, thread-safety callout for `snapshot_only`).

No blocking issues for merging the docs themselves — this is a doc-only PR. The inline items should be addressed either in a follow-up or before P0 implementation begins.

docs/design/ALGORITHM_MANIFEST_SPECIFICATION.md
```diff
@@ -0,0 +1,355 @@
...
+
+At most one `algorithm/2` declaration per file. Multiple
+declarations with the same name in different files are an error;
+the codegen rejects compilation rather than silently picking one.
+
+### `algorithm_optimization(+Name, +OptList) is det.`
+
+Declares optimizations for an algorithm. `Name` matches an
+`algorithm/2` declaration. `OptList` contains any options that
+downstream resolvers consume.
```
s243a, line 118: **Option name inconsistency: `warm_budget_nodes_fraction` vs `warm_budget_nodes`**

In `algorithm_optimization/2` example here (line ~108), the option is `warm_budget_nodes_fraction(0.1)` — a fraction of `fact_count`. But in `SCAN_STRATEGY_SPECIFICATION.md`'s option table, the option is `warm_budget_nodes(N)` with default `fact_count / 10` — an *absolute node count*.

These are two different semantics under nearly the same name. Either:
1. The spec should add `warm_budget_nodes_fraction(F)` as a distinct option that the resolver converts to `warm_budget_nodes(N)` before use, OR
2. The manifest example should use `warm_budget_nodes(N)` directly.

Currently a reader of the manifest spec will write `warm_budget_nodes_fraction(0.1)`, a reader of the strategy spec will write `warm_budget_nodes(30000)`, and the two will silently not compose unless the resolver handles both forms. This needs reconciling before P0.

docs/design/ALGORITHM_MANIFEST_SPECIFICATION.md
```diff
@@ -0,0 +1,355 @@
...
+```
+
+If a key appears in multiple facts, the first occurrence wins
+(matches `option/3` semantics). To override, place the overriding
+fact earlier or pass the option directly to the codegen entry
+point.
+
+## Merge semantics
+
+When the codegen enters `write_wam_haskell_project/3` (or
```
s243a, line 167: **`flatten/2` on nested option lists is fragile for nested option terms**

The `load_algorithm_manifest/2` Prolog snippet uses:
```prolog
findall(Opts, user:algorithm_optimization(Name, Opts), AllOpts),
flatten(AllOpts, ManifestOpts),
```

`flatten/2` is a deep flatten — it will destructure *all* nested lists, including option values that are themselves lists (e.g. `tree_cost_function(flux, [iterations(1), parent_decay(0.5)])` would have its inner parameter list `[iterations(1), parent_decay(0.5)]` flattened into the top-level option list, corrupting the structure).

The correct approach is a shallow concat of the outer list only:
```prolog
findall(Opts, user:algorithm_optimization(Name, Opts), AllOpts),
append(AllOpts, ManifestOpts),  % shallow concat, not deep flatten
```
This is a correctness bug in the spec's pseudocode that would carry into P0 if copied directly.

docs/design/ALGORITHM_MANIFEST_SPECIFICATION.md
```diff
@@ -0,0 +1,355 @@
...
+## Example: full workload file
+
+```prolog
+:- module(workload_effective_distance, []).
+
+%% Algorithm declaration
+:- algorithm(effective_distance, [
+       kernel(category_ancestor/4),
+       seeds(article/1),
+       roots(root_category/1),
```
s243a, line 310: **`category_ancestor/4` kernel has a correctness issue: accumulator-style recursion with no base case**

The kernel example:
```prolog
category_ancestor(Cat, Root, Hops, Visited) :-
    category_parent(Cat, Mid),
    \+ member(Mid, Visited),
    category_ancestor(Mid, Root, H, [Mid|Visited]),
    Hops is H + 1.
```

This has **no base case** — it will loop infinitely on any input because there is no clause that succeeds when `Cat = Root`. A typical base case would be:
```prolog
category_ancestor(Root, Root, 0, _).
```
Without it, the predicate always fails (assuming `category_parent` eventually has no solution), which means `seeds(article/1)` and `roots(root_category/1)` would produce zero results in the bench harness. This is an example in a spec doc, so implementors will copy it — worth fixing.

docs/design/SCAN_STRATEGY_IMPLEMENTATION_PLAN.md
```diff
@@ -0,0 +1,457 @@
...
+  but unused by any consumer until P5.
+
+### Success criteria
+
+- Unit tests pass for warm-budget computation, threshold derivation,
+  cost-function dispatch.
+- End-to-end smoke build of a `scan_strategy(auto)`-enabled project
+  on the 1k fixture.
+- Functional run on simplewiki produces correct results (tuple
+  count matches `resident_cursor` baseline within tolerance).
```
s243a, line 241: **`tree_retention(retain)` in P5 deliverables doesn't match the spec's enum**

P5 deliverables say:
> `main.hs.mustache` — when `tree_retention(retain)` is set, replace the seed `parMap rdeepseq`...

But `SCAN_STRATEGY_SPECIFICATION.md` defines the retention enum as:
```prolog
tree_retention(discard).
tree_retention(snapshot_only).
tree_retention(live).
```

There is no `tree_retention(retain)` — the correct value for P5's use case is `tree_retention(snapshot_only)`. This is a naming inconsistency that would cause a silent no-op at runtime (the `retain` option wouldn't match any clause). Should be `snapshot_only` throughout P5.

docs/design/SCAN_STRATEGY_IMPLEMENTATION_PLAN.md
```diff
@@ -0,0 +1,457 @@
...
+- Measured wall-clock at simplewiki: within 10% of
+  `resident_cursor` at -N4 (no regression at fixture sizes where
+  warming should be approximately neutral).
+
```
s243a, line 245: **P5 resolver hook condition is underspecified: `scan_strategy(auto)` + `expected_query_count > 1` → `capability_local`**

This auto-opt-in to `tree_retention(snapshot_only)` + `spark_routing(capability_local)` when `expected_query_count > 1` and L1 is the cache mode has a hidden dependency: the resolver needs to know the *resolved* cache mode (output of `resolve_auto_lmdb_cache_mode/2`), not the *declared* cache mode. If a workload declares `lmdb_cache_mode(auto)`, the L1 decision won't be finalized until after that resolver runs.

The resolver chain order in the spec (`load_manifest → resolve_auto_cache_strategy → resolve_auto_lmdb_cache_mode → resolve_auto_scan_strategy`) correctly places `resolve_auto_scan_strategy` last — so it *does* see the resolved cache mode. But P5's description should explicitly say: "the hook reads the *resolved* `lmdb_cache_mode` value (post-`resolve_auto_lmdb_cache_mode`)" to make it clear this isn't a chicken-and-egg problem.

docs/design/SCAN_STRATEGY_PHILOSOPHY.md
```diff
@@ -0,0 +1,259 @@
...
+
+The retained flux-ranked view is precisely what MoE-style spark
+routing needs to break the per-HEC L1 cache-duplication problem.
+Memory note: "MoE-style spark routing would unlock L1; until then
+sharded is the right default" because parMap has no region
+affinity. With flux-sorted seeds:
+
+1. Partition seeds into N capability-local chunks by flux score
+   (high-flux chunk to capability 0, next to capability 1, etc.).
+2. Each thread's L1 cache accumulates non-overlapping coverage
```
s243a, line 207: **`No memoisation of misses` in "What we picked" contradicts the Spec doc**

The Philosophy doc states:
> No memoisation of misses (no spillover) for the first cut.

But `SCAN_STRATEGY_SPECIFICATION.md` (Phase R lookup logic) explicitly describes:
```
_   -> do   -- miss
  es <- cursor n
  insertBucket n es cache   -- memoise; may overwrite
```

The Spec *does* memoise misses back into the hashtable. The Philosophy doc says it doesn't. These directly contradict each other. The Spec appears to be the authoritative version (miss-memoisation is what makes the overwrite-on-collision cache learn the workload's actual access pattern), so the Philosophy doc should be corrected: remove the "no memoisation of misses" sentence, or change it to "misses are memoised back into the same hashtable (overwrite-on-collision)."

docs/design/SCAN_STRATEGY_PHILOSOPHY.md
```diff
@@ -0,0 +1,259 @@
...
+   per region is how we capture that.
+4. **Workloads where parallelisation matters** — even without the
+   cold regime, the retained flux ranking unlocks
+   region-affinity-aware spark routing.
+
+At today's fixture scale (simplewiki at 297k, enwiki at 9.9M),
+this work is groundwork. Like the cost model itself.
+
+## What this isn't doing
+
```
s243a, line 241: **"Warming a tree per region" is mentioned but has no corresponding option in the Spec**

The Philosophy doc says:
> Warming a tree per region is how we capture [repeated queries against a hot root subset].

But the Spec and Implementation Plan describe exactly *one* warm phase per workload run — a single `runWarmBuild` call seeded from the full `seeds/1` predicate. There is no per-region warm mechanism specified anywhere.

This is either:
1. A future capability worth noting as deferred (like `live` mode), or
2. An aspirational statement that overclaims what the design delivers.

Suggestion: Add a note like "Per-region warming is not in scope for P0–P5; the design supports one warm phase per workload run. Region-specific warming would require either multiple manifest declarations or a `live`-mode extension."

docs/design/SCAN_STRATEGY_SPECIFICATION.md
```diff
@@ -0,0 +1,442 @@
...
+
+The `Cache` is the same shape as the existing L1/L2 hashtables: a
+fixed-bucket array, where insert into a non-empty bucket overwrites
+the prior occupant. No LRU/LFU tracking, no metadata per entry,
+no eviction policy beyond "the latest insert wins the bucket". This
+keeps lookups branch-free and matches the existing code paths the
+implementation can reuse.
+
+Lookup logic:
+
```
s243a, line 72: **`snapshotCache` return type loses the Cost dimension**

```haskell
snapshotCache :: WarmHeap -> IntMap NodeId [EdgeTarget]
```

The `WarmHeap` holds `(NodeId, Edges)` keyed by `Cost`. The snapshot discards the cost and produces only edges. That's correct for the `Cache` — steady-state only needs edge lookup by node. But this means the `Cache` has no way to answer "what was this node's warm score?" without also keeping `RankedView`.

This is a design choice, not a bug — but it should be stated explicitly: **"Cost values are not retained in the Cache. If cost information is needed post-snapshot (e.g. for P5 routing), it is retained separately via `snapshotRanked`, not embedded in the Cache."** Without this note, an implementor might try to add cost metadata to Cache entries unnecessarily.

docs/design/SCAN_STRATEGY_SPECIFICATION.md
```diff
@@ -0,0 +1,442 @@
...
+tree_cost_function(flux,                [iterations(1), parent_decay(0.5), child_decay(0.3)]).
+tree_cost_function(hop_distance,        [max_hops(5)]).
+tree_cost_function(semantic_similarity, [dim(128), embedding_path('foo.bin')]).
+```
+
+The Haskell-side generated code consumes a `CostFn` abstraction:
+
+```haskell
+data CostFn = CostFn
+  { cfScore   :: !(NodeId -> Edges -> Cost)   -- pure scorer
```
s243a: **`CostFn` record has a `cfScore` field that overlaps with `cfRelax` semantically**

```haskell
data CostFn = CostFn
  { cfScore   :: !(NodeId -> Edges -> Cost)          -- pure scorer
  , cfRelax   :: !(NodeId -> [NodeId] -> Cost -> ScoreMap -> ScoreMap)  -- propagation step
  , cfInitial :: !(NodeId -> Cost)                   -- endpoint score
  }
```

`cfScore` takes `(NodeId, Edges)` and returns `Cost` — but during warm-build, the score of a node is computed incrementally via `cfRelax` (propagating from already-scored neighbours). When would you call `cfScore` directly vs `cfRelax`?

For `hop_distance`, `cfScore` could plausibly be `const maxHops` (initial score at endpoint), but then it's the same as `cfInitial`. For `flux`, `cfScore` would need to sum contributions from all scored neighbours, which requires the `ScoreMap` — making `cfScore :: NodeId -> Edges -> Cost` insufficient (no `ScoreMap` argument).

Suggestion: Either remove `cfScore` (it appears to be subsumed by `cfRelax` + `cfInitial`), or clarify its distinct role with a concrete example for both `hop_distance` and `flux`.

docs/design/SCAN_STRATEGY_SPECIFICATION.md
```diff
@@ -0,0 +1,442 @@
...
+
+### Semantic similarity
+
+```
+score(n) = dot_product(embedding(n), query_embedding) / (|embedding(n)| * |query_embedding|)
+```
+
+Requires an embeddings table indexed by `NodeId`. Out of scope for
+the first cuts; included in the strategy enum because the slot
+should be open to it.
```
s243a, line 183: **`iterations(auto)` is listed in the option table but explicitly deferred with no gate**

The iterations table says:
> `auto` — until score-change-per-iteration drops below a threshold (deferred — needs a convergence metric to be defined).

But the option *is* listed in the public API table without any `(not yet implemented)` marker. An implementor reading the Spec would need to handle this case. P6 in the implementation plan adds `convergence_threshold(F)` for adaptive iteration, but `iterations(auto)` doesn't appear as a P6 deliverable — P6 is about lifting the `stage2_max_scans` bound, not about per-iteration convergence.

Either:
1. Map `iterations(auto)` → `iterations(1)` with a warning in P3, and explicitly list making it functional as a P6 deliverable, OR
2. Remove it from the option table and add it as a separate "Future extensions" section.

Currently it's a dangling `auto` value with no implementation path.

docs/design/SCAN_STRATEGY_SPECIFICATION.md
```diff
@@ -0,0 +1,442 @@
...
+The new scan-strategy resolver:
+
+- Fires only when `scan_strategy(auto)` is in the (merged) options.
+- Reads `tree_cost_function/1`, `expected_query_count/1`,
+  `working_set_fraction/1`, `mem_available_bytes/1`, etc. — most of
+  which typically come from the algorithm manifest, with caller
+  overrides on top.
+- Outputs concrete `tree_cost_function/2`, `warm_budget_nodes/1`,
+  `stage2_scan_threshold/1`, `tree_retention/1`.
+
```
s243a, line 458: **`cache_strategy_verbose(true)` used here but `scan_strategy_verbose(true)` defined in the Implementation Plan**

The edge-case section says:
> Recommend exposing `cache_strategy_verbose(true)`-style stats (hit/miss counters)...

But the Implementation Plan's cross-cutting "Telemetry" section defines a *separate* `scan_strategy_verbose(true)` option. The spec's "Cost function disagrees with workload" edge case should reference `scan_strategy_verbose(true)`, not `cache_strategy_verbose(true)`. These are different option keys with different scopes — mixing them here will confuse implementors about which flag to toggle for scan-strategy diagnostics.

docs/design/SCAN_STRATEGY_IMPLEMENTATION_PLAN.md
```diff
@@ -0,0 +1,457 @@
...
+- `src/unifyweaver/core/cost_function.pl` — new module:
+  - `tree_cost_function(+Name, +Params)` term constructor
+  - `validate_cost_function/1` predicate
+  - Registry of supported names: `flux`, `hop_distance`,
+    `semantic_similarity`
+- `templates/targets/haskell_wam/cost_function.hs.mustache` — new
+  Haskell-side `CostFn` record + concrete implementations.
+- `src/unifyweaver/targets/wam_haskell_target.pl` — codegen for
+  the cost-function strategy slot; ensures the existing Flux path
```
s243a, line 108: **P2 success criterion tolerance is too tight given fixture variability**

Success criterion: "`resident_auto` runs within 5% of `resident_cursor` baseline at every scale."

Phase L measurements historically show multi-trial variance of 3–8% at -N4 for BFS workloads, purely from GC timing and LMDB page-cache state. A 5% gate will flap on otherwise-correct implementations due to environmental noise, especially on enwiki at -N1 where cold-start variance is higher.

Suggestion: use 10% tolerance (consistent with P3's success criteria which use 10%), or qualify with "median of 3 trials within 5% of median `resident_cursor` over 3 trials." The current absolute wording will cause false failures.

docs/design/SCAN_STRATEGY_IMPLEMENTATION_PLAN.md
```diff
@@ -0,0 +1,457 @@
...
+### Success criteria
+
+- `resident_auto` runs within 5% of `resident_cursor` baseline at
+  every scale (resolver agrees with empirical best).
+- Branching-factor distribution captured: median, P90, P99 for
+  parent and child legs separately.
+
+### Estimated effort
+
+~1 session for ingest + sweep + write-up. Wall-clock dominated by
```
s243a, line 180: **P3 Haskell template `runWarmBuild` returns `(WarmHeap, WarmIndex)` but snapshot needs `WarmHeap` only**

The deliverable lists:
```
runWarmBuild :: CostFn -> WarmConfig -> IO (WarmHeap, WarmIndex)
```

But `snapshotCache :: WarmHeap -> IntMap NodeId [EdgeTarget]` only takes `WarmHeap`. The `WarmIndex` is needed during warm (to check node membership) but has no role post-warm unless we're in `live` mode. Returning it from `runWarmBuild` forces the caller to carry a potentially large `IntMap` until GC happens.

Consider either:
1. Return only `WarmHeap` from `runWarmBuild` (the index is an internal implementation detail), or
2. Document explicitly why the caller needs `WarmIndex` post-warm (perhaps for `live`-mode consumers who need it before GC).

For `discard` and `snapshot_only` modes this is just unnecessary memory pressure.

docs/design/SCAN_STRATEGY_IMPLEMENTATION_PLAN.md
```diff
@@ -0,0 +1,457 @@
...
+May be unnecessary. Filed as optional. Likely de-prioritised in
+favour of cache-warming follow-ups if P3/P4 data shows 2 scans
+captures >95% of the tree.
+
+### Estimated effort
+
+~1 session if pursued.
+
+---
+
```
s243a, line 369: **"What we won't do" contradicts the Spec on miss memoisation**

This section lists as out of scope:
> Online cache growth (cache misses don't memoise).

But this directly contradicts `SCAN_STRATEGY_SPECIFICATION.md`, which defines miss memoisation as a core feature of Phase R steady-state (the lookup pseudo-code explicitly calls `insertBucket n es cache` on every miss).

This is the same contradiction flagged in `SCAN_STRATEGY_PHILOSOPHY.md` line 200 — it appears to be a copy-paste of the same erroneous statement. The phrase "cache misses don't memoise" needs to be removed from both docs. The out-of-scope statement should be something like:
> Online cache *growth beyond the fixed bucket count* (the bucket array is sized at codegen time; misses are memoised via overwrite-on-collision but the array does not resize).

docs/design/ALGORITHM_MANIFEST_SPECIFICATION.md
```diff
@@ -0,0 +1,355 @@
...
+Example:
+
+```prolog
+:- algorithm_optimization(effective_distance, [
+       cache_strategy(auto),
+       working_set_fraction(0.001),
+       expected_query_count(10),
+       lmdb_cache_mode(auto),
+       workload_locality(unknown),
+       scan_strategy(auto),
```
s243a, line 129: **`option/3` vs `merge_options/3`: the text cites two different merge predicates for the same semantics**

This paragraph says:
> their option lists concatenate with `option/3` first-match semantics

But `option/3` is SWI-Prolog's option-reading predicate (`option(Key(Value), OptionList, Default)`) — it's not a merge predicate. What's meant here is likely the semantics of `nb_getval`/`option/3` used by SWI's `options` library where the first occurrence of a key wins when reading.

The actual merge is done by `merge_options/3` (or the custom variant discussed elsewhere). The sentence should clarify: "their option lists are *concatenated* (via `append`), and the merged list is read with first-match semantics: the first occurrence of any key wins when resolving option values." Citing `option/3` as the merge mechanism is technically incorrect.

docs/design/ALGORITHM_MANIFEST_SPECIFICATION.md
```diff
@@ -0,0 +1,355 @@
...
+```
+
+**Caller wins on conflict.** This is the override path: a bench
+that wants to experiment with `tree_retention(live)` passes that
+option directly, and it beats the manifest's `snapshot_only`.
+
+**Manifest is best-known-defaults, not policy.** The manifest is
+a recommendation derived from measurement and cost-model logic.
+Callers should feel free to override when they have a reason.
+
```
s243a, line 189: **`load_algorithm_manifest/2` runs under `write_wam_haskell_project/3` but `compile_wam_runtime_to_haskell/3` is an alias or separate entry point?**

The spec says:
> This helper runs once at the top of `write_wam_haskell_project/3` (and `compile_wam_runtime_to_haskell/3` for the codegen-test path)

This implies two separate codegen entry points both need to call `load_algorithm_manifest/2`. If they share an internal implementation, the manifest load should happen in the shared call, not be duplicated at both entry points. If they don't share, this is two integration points to maintain.

The Implementation Plan (P0 deliverables) lists both call sites explicitly. Suggest adding a note clarifying the relationship: "Both entry points funnel through a common internal codegen pipeline; `load_algorithm_manifest/2` is called once at the pipeline's head, not once per entry point." Otherwise implementors may wire it redundantly.

s243a (commented):
## Second-pass review: docs/scan-strategy-design

This pass focused on cross-doc consistency, semantic correctness of pseudocode, and implementation traps not caught in the first review. The docs are in good shape overall — the issues below are mostly spec-level ambiguities and two outright contradictions that need fixing before P0.

### 🔴 Must fix before P0 (correctness issues)

1. **Miss-memoisation contradiction** (`PHILOSOPHY.md` line 200 + `IMPLEMENTATION_PLAN.md` line 344): Both docs say "cache misses don't memoise" but the Spec's Phase R lookup pseudocode explicitly calls `insertBucket n es cache` on every miss. The Spec is correct — the Philosophy and Implementation Plan have a stale copy of an earlier design decision. Remove "no memoisation of misses" from both places and replace with the accurate description (misses overwrite into the fixed-size hashtable).

2. **`option/3` cited as a merge predicate** (`ALGORITHM_MANIFEST_SPEC.md` line 119): `option/3` is SWI-Prolog's option-reading predicate, not a merge predicate. The sentence conflates the *reading* semantics (first-match wins) with the *merging* mechanism (append + read). Clarify: lists are concatenated with `append`, and first-match-wins applies when *reading* via `option/3`.

3. **`iterations(auto)` is in the public API with no implementation path** (`SCAN_STRATEGY_SPEC.md`): Listed in the option table, marked deferred, but not assigned to any phase. Either add it to P6's deliverables or move it to a "Future" section to avoid implementors needing to handle it as a stub.

### 🟡 Should fix before P0 (ambiguities that become bugs)

4. **`CostFn.cfScore` overlaps with `cfRelax` + `cfInitial`** (`SCAN_STRATEGY_SPEC.md` line 100): Unclear when `cfScore :: NodeId -> Edges -> Cost` is called vs `cfRelax`. For flux, computing a node's score requires the `ScoreMap` (neighbours' scores), which `cfScore` doesn't take. Either justify `cfScore`'s distinct role or remove it.

5. **`snapshotCache` silently drops Cost** (`SCAN_STRATEGY_SPEC.md` line 62): The snapshot discards cost values; only edges are kept in the Cache. This is correct but should be stated explicitly to prevent implementors from redundantly embedding cost in Cache entries.

6. **`verbose` option name mismatch** (`SCAN_STRATEGY_SPEC.md` line 395): Edge case section references `cache_strategy_verbose(true)` but the Implementation Plan defines `scan_strategy_verbose(true)` as the correct option for scan-strategy diagnostics.

7. **`runWarmBuild` returns `WarmIndex` unnecessarily for `discard`/`snapshot_only` modes** (`IMPLEMENTATION_PLAN.md` line 168): The index has no post-warm use in non-`live` modes. Either return only `WarmHeap`, or document why the caller needs the index.

### 🟢 Minor / documentation quality

8. **Per-region warming claim** (`PHILOSOPHY.md` line 232): The design only supports one warm phase per workload run; "warming a tree per region" is unspecified. Scope it as future work.

9. **P2 tolerance gate too tight** (`IMPLEMENTATION_PLAN.md` line 103): 5% wall-clock tolerance will flap with GC variance. Use 10% (consistent with P3) or specify median-of-3.

10. **Two codegen entry points** (`ALGORITHM_MANIFEST_SPEC.md` line 170): Clarify whether `write_wam_haskell_project/3` and `compile_wam_runtime_to_haskell/3` share an internal pipeline (and therefore `load_algorithm_manifest/2` is called once) or are independent wiring points.

PR #2130, opened 2026-05-15
feat(wam-elixir): call/N meta-call (PR #1 of 5)

s243a:
## Summary

  PR #1 from [`WAM_ELIXIR_GAPS_SPECIFICATION.md`](docs/design/WAM_ELIXIR_GAPS_SPECIFICATION.md) §3.
  Closes the `call/N` gap so user code can write higher-order patterns
  and so `catch/3` (PR #2) has the goal-term dispatch it needs.

  ### Audit findings (the gap before this commit)

  - `WamDispatcher.call/2` dispatches by predicate-string with builtin
    fallback, but had no understanding of `"call/N"` as a meta-call.
  - `:call` / `:execute` step arms only do label lookup, so
    `{:call, "call/N", N}` fell through to `:fail`.
  - `execute_builtin` had no `"call/N"` arm — lowered code emitting
    `WamDispatcher.call("call/N", state)` hit the `unknown_builtin`
    throw.
  - No `true/0` builtin, so `call(true)` would have crashed even if
    routing worked.

  ### Changes

  | Component | Purpose |
  |---|---|
  | `dispatch_call_meta/3` runtime helper | Reads A1 as base goal, gathers A2..A_TotalArity as extras, builds combined
  `"name/total_arity"` key, loads regs, routes through `WamDispatcher.call`. Mirrors C++ `dispatch_call_meta` in
  `wam_cpp_target.pl`. |
  | `execute_builtin "call/N"` arm | Catches lowered-mode dispatch from `WamDispatcher.call("call/N", state)` |
  | `:call` / `:execute` step arms | Catch `"call/N"` before label lookup (interpreter-mode coverage) |
  | `true/0` builtin arm | Required for `call(true)` |
  | `build_call_target/3`, `load_args_into_regs/3` | Internal helpers — handle bare-string atom, BEAM-atom, and `{:ref,
  addr} → {:str, "name/baseN"}` compound goal shapes |

  Reference: C++ commit `b5fc6041` *feat(wam-cpp): call/N meta-call*.

  ### Bench fixtures

  Deferred to PR #2 (catch/3 + throw/1) per spec §6.1's "may also
  pre-land" wording — no bench infrastructure changes were
  incidental to this PR.

  ## Test plan

  ### Unit tests (`tests/test_wam_elixir_target.pl`)

  - [x] `WAM-Elixir: dispatch_call_meta helper present in runtime`
  - [x] `WAM-Elixir: :call and :execute step arms catch "call/N" before label lookup`
  - [x] `WAM-Elixir: true/0 builtin arm exists`
  - [x] `WAM-Elixir: build_call_target + load_args_into_regs helpers present`

  ### E2E acceptance tests (`tests/test_wam_elixir_classic_programs.pl`)

  Full Prolog → WAM → Elixir → elixir-subprocess execution.

  - [x] `call_atom_true` — `?- call(true).` succeeds → `"true"`
  - [x] `call_atom_fail` — `?- call(fail).` fails → `"false"`
  - [x] `call_compound_goal` — `?- G = (X is 1+1), call(G), R = X.` with R=2 → `"true"`
  - [x] `call_with_extras` — `?- call(append, [a,b], [c,d], R).` with R=[a,b,c,d] → `"true"`

  ### Regression

  - [x] All existing `tests/test_wam_elixir_target.pl` tests pass.
  - [x] All existing `tests/test_wam_elixir_classic_programs.pl` tests pass (fibonacci, ackermann, pythagoras).

  ## Reviewer notes

  - The `:call` / `:execute` step arm changes serve interpreter mode;
    the `execute_builtin "call/N"` arm serves lowered mode. Both
    share `dispatch_call_meta/3`. Belt-and-suspenders so future
    callers from either mode get the same dispatch behaviour.
  - No changes to existing `\+/1` / `findall/3` paths, which use
    `WamDispatcher.call` directly (not via `"call/N"`). The new arm
    is additive.
  - `true/0` was added because the `_ -> throw({:unknown_builtin, ...})`
    hardening would otherwise crash on `call(true)` even after
    routing worked. Same justification used for the existing `fail/0`
    arm (see line 821 comment).

### DIFF
--- a/src/unifyweaver/targets/wam_elixir_target.pl
+++ b/src/unifyweaver/targets/wam_elixir_target.pl
@@ -232,9 +232,11 @@ wam_elixir_case(call,
         %{state | pc: target_pc, cp: state.pc + 1}
       # Meta-call call/N — caught BEFORE label lookup so the
       # "call/N" string is dispatched as the meta-call, not
-      # treated as a missing user predicate.
-      {:call, "call/" <> _, n} ->
-        dispatch_call_meta(state, n, state.pc + 1)
+      # treated as a missing user predicate. The `n` field IS
+      # the total arity (= 1 base goal + extras), matching the
+      # WAM compiler''s emission convention.
+      {:call, "call/" <> _, total_arity} ->
+        dispatch_call_meta(state, total_arity, state.pc + 1)
       # Fallback: unresolved string label (cross-module/orphan).
       {:call, p, _n} when is_binary(p) ->
         case Map.get(state.labels, p) do
@@ -248,9 +250,14 @@ wam_elixir_case(execute,
         %{state | pc: target_pc}
       # Tail-position meta-call call/N. Total arity comes from
       # the op-name suffix (Execute carries no arity field).
+      # Integer.parse/1 instead of String.to_integer/1 so a
+      # malformed opcode (corrupted instr stream) fails cleanly
+      # rather than raising ArgumentError.
       {:execute, "call/" <> arity_str} ->
-        total_arity = String.to_integer(arity_str)
-        dispatch_call_meta(state, total_arity, state.cp)
+        case Integer.parse(arity_str) do
+          {total_arity, ""} -> dispatch_call_meta(state, total_arity, state.cp)
+          _ -> :fail
+        end
       # Fallback: unresolved string label.
       {:execute, p} when is_binary(p) ->
         case Map.get(state.labels, p) do
@@ -1197,9 +1204,17 @@ compile_utility_helpers_to_elixir(Code) :-
   # Then load the combined args into A1..A_total, set cp to after_pc,
   # and route through WamDispatcher.call (which knows about user
   # predicates and falls through to execute_builtin for builtins).
-  @doc "Dispatch a call/N meta-call. After the called goal returns, control resumes at after_pc."
-  def dispatch_call_meta(state, total_arity, after_pc) do
+  #
+  # `defp`: only invoked from within WamRuntime (execute_builtin
+  # arm + the :call/:execute step arms via the same module). Not
+  # part of the runtime''s external API.
+  defp dispatch_call_meta(state, total_arity, after_pc) do
     goal = deref_var(state, get_reg(state, 1))
+    # Elixir descending-range trap: `for i <- 2..1` would raise
+    # in Elixir 1.16+ (without an explicit step) and would yield
+    # [2, 1] in older Elixir. The `total_arity > 1` guard
+    # short-circuits the 0-extras case so the comprehension only
+    # runs on a real ascending range.
     extras =
       if total_arity > 1 do
         for i <- 2..total_arity, do: deref_var(state, get_reg(state, i))
@@ -1212,12 +1227,31 @@ compile_utility_helpers_to_elixir(Code) :-
         call_state = %{prepared_state | cp: after_pc}
         try do
           case WamDispatcher.call(pred_arity, call_state) do
+            # Why override pc to after_pc here? WamDispatcher.call
+            # returns the dispatched goal''s post-state, whose
+            # pc is whatever the called goal''s last instruction
+            # left it at (typically state.pc + 1 for builtins, or
+            # the user-pred''s last-instr pc for lowered predicates).
+            # Neither matches the meta-call''s logical resumption
+            # point, which is after_pc (= state.pc + 1 for non-tail
+            # call, state.cp for tail execute). Forcing pc to
+            # after_pc makes the meta-call behave like a single
+            # logical instruction regardless of what the dispatched
+            # goal did internally. PRs #2-5 (catch/throw + ISO)
+            # depend on this invariant.
             {:ok, post_state} -> %{post_state | pc: after_pc}
             :fail -> :fail
           end
         catch
+          # :fail and {:fail, _} — internal control-flow throws used
+          # for backtrack propagation. Cargoed from the \\+/1 block
+          # (see negation_as_failure handling); same semantics.
           :fail -> :fail
           {:fail, _} -> :fail
+          # {:return, _} — same provenance as the catch arms in the
+          # negation block. Some lowered code paths use throw({:return,
+          # state}) to short-circuit out of a body. Re-pin pc the
+          # same way the success path does.
           {:return, post_state} -> %{post_state | pc: after_pc}
         end
       :fail -> :fail
@@ -1251,9 +1285,21 @@ compile_utility_helpers_to_elixir(Code) :-
       _ -> :fail
     end
   end
+  # Catch-all: unbound goal, malformed term, or otherwise unrecognised
+  # shape. Returns :fail (lax semantics), matching the rest of the
+  # current Elixir runtime. PR #3 (ISO errors plumbing) will replace
+  # this with `instantiation_error` for the unbound-goal case
+  # specifically; the malformed-shape cases stay :fail.
   defp build_call_target(_state, _goal, _extras), do: :fail
 
-  # Write a list of arg values into consecutive A-regs starting at start_reg.
+  # Write a list of arg values into consecutive A-regs starting at
+  # start_reg. Note: callers pass start_reg=1, which means the FIRST
+  # extra (or first base-arg, for compound goals) overwrites A1 — the
+  # register that originally held the base goal-term. That is correct:
+  # after dispatch_call_meta builds the combined name/total_arity,
+  # the dispatched predicate''s arg 1 is the FIRST argument of the
+  # combined call, not the base-goal term. The base-goal value is no
+  # longer needed once pred_arity is built.
   defp load_args_into_regs(state, start_reg, args) do
     args
     |> Enum.with_index(start_reg)
--- a/src/unifyweaver/targets/wam_elixir_target.pl
+++ b/src/unifyweaver/targets/wam_elixir_target.pl
@@ -230,6 +230,11 @@ wam_elixir_case(call,
 '      # Fast path: label pre-resolved at codegen to integer PC.
       {:call, target_pc, _n} when is_integer(target_pc) ->
         %{state | pc: target_pc, cp: state.pc + 1}
+      # Meta-call call/N — caught BEFORE label lookup so the
+      # "call/N" string is dispatched as the meta-call, not
+      # treated as a missing user predicate.
+      {:call, "call/" <> _, n} ->
+        dispatch_call_meta(state, n, state.pc + 1)
       # Fallback: unresolved string label (cross-module/orphan).
       {:call, p, _n} when is_binary(p) ->
         case Map.get(state.labels, p) do
@@ -241,6 +246,11 @@ wam_elixir_case(execute,
 '      # Fast path: label pre-resolved at codegen to integer PC.
       {:execute, target_pc} when is_integer(target_pc) ->
         %{state | pc: target_pc}
+      # Tail-position meta-call call/N. Total arity comes from
+      # the op-name suffix (Execute carries no arity field).
+      {:execute, "call/" <> arity_str} ->
+        total_arity = String.to_integer(arity_str)
+        dispatch_call_meta(state, total_arity, state.cp)
       # Fallback: unresolved string label.
       {:execute, p} when is_binary(p) ->
         case Map.get(state.labels, p) do
@@ -826,6 +836,11 @@ compile_utility_helpers_to_elixir(Code) :-
         # failure. (Pre-hardening, fail/0 worked accidentally because
         # the default arm returned :fail — same semantic, wrong reason.)
         :fail
+      {"true/0", 0} ->
+        # ISO `true` — succeeds unconditionally. Same default-arm
+        # hardening reasoning as fail/0 above.
+        new_pc = if is_integer(state.pc), do: state.pc + 1, else: state.pc
+        %{state | pc: new_pc}
       {"</2", 2} ->
         v1 = eval_arith(state, get_reg(state, 1))
         v2 = eval_arith(state, get_reg(state, 2))
@@ -1144,6 +1159,14 @@ compile_utility_helpers_to_elixir(Code) :-
             preserved_aggs ++ state.cut_point
           end
         %{state | choice_points: new_cps, pc: new_pc}
+      {"call/" <> _, n} ->
+        # call/N meta-call. Lowered-mode dispatch arrives here via
+        # WamDispatcher.call("call/N", state) -> execute_builtin/3
+        # falling through. Interpreter mode also reaches the same
+        # helper from the :call / :execute step arms (see
+        # wam_elixir_case(call, ...) and wam_elixir_case(execute, ...)).
+        # See WAM_ELIXIR_GAPS_SPECIFICATION.md PR #1.
+        dispatch_call_meta(state, n, state.cp)
       _ ->
         # Hardening: surface unimplemented builtins instead of silently
         # returning :fail. The pre-hardening default arm masked
@@ -1155,6 +1178,86 @@ compile_utility_helpers_to_elixir(Code) :-
         # be added rather than relying on the silent-:fail accident.
         throw({:unknown_builtin, op, arity})
     end
+  end
+
+  # ---- call/N meta-call dispatch -----------------------------------
+  #
+  # Mirrors the C++ dispatch_call_meta in wam_cpp_target.pl. The
+  # WAM compiler emits call(Goal[, X1..Xk]) usage as
+  #   {:execute, "call/N"}        -- when call/N is the body tail
+  #   {:call,    "call/N", N}     -- otherwise (non-tail)
+  # with the base goal in A1 and any extras in A2..AN. The :call
+  # and :execute step arms catch the "call/" prefix BEFORE doing
+  # label lookup and route here.
+  #
+  # Build the combined goal:
+  #   - A1 atom        (`true`, `fail`, `append`)  -> use as-is, arity = total - 1
+  #   - A1 compound    ({:ref, addr} -> {:str, "name/baseN"}) ->
+  #       gather base args from heap, append extras, total = baseN + (total - 1)
+  # Then load the combined args into A1..A_total, set cp to after_pc,
+  # and route through WamDispatcher.call (which knows about user
+  # predicates and falls through to execute_builtin for builtins).
+  @doc "Dispatch a call/N meta-call. After the called goal returns, control resumes at after_pc."
+  def dispatch_call_meta(state, total_arity, after_pc) do
+    goal = deref_var(state, get_reg(state, 1))
+    extras =
+      if total_arity > 1 do
+        for i <- 2..total_arity, do: deref_var(state, get_reg(state, i))
+      else
+        []
+      end
+
+    case build_call_target(state, goal, extras) do
+      {:ok, pred_arity, prepared_state} ->
+        call_state = %{prepared_state | cp: after_pc}
+        try do
+          case WamDispatcher.call(pred_arity, call_state) do
+            {:ok, post_state} -> %{post_state | pc: after_pc}
+            :fail -> :fail
+          end
+        catch
+          :fail -> :fail
+          {:fail, _} -> :fail
+          {:return, post_state} -> %{post_state | pc: after_pc}
+        end
+      :fail -> :fail
+    end
+  end
+
+  # Build the combined predicate-key + load A-regs from the goal+extras.
+  # Atomic goal: arity is just the count of extras.
+  # Compound {:ref, addr} -> {:str, "name/baseN"}: read base args
+  # from heap, concat extras, total = baseN + length(extras).
+  defp build_call_target(state, goal, extras) when is_binary(goal) do
+    arity = length(extras)
+    pred_arity = "#{goal}/#{arity}"
+    new_state = load_args_into_regs(state, 1, extras)
+    {:ok, pred_arity, new_state}
+  end
+  defp build_call_target(state, goal, extras) when is_atom(goal) and not is_nil(goal) do
+    build_call_target(state, Atom.to_string(goal), extras)
+  end
+  defp build_call_target(state, {:ref, addr}, extras) do
+    case Map.get(state.heap, addr) do
+      {:str, base_pred_arity} ->
+        base_arity = parse_functor_arity(base_pred_arity)
+        name = parse_functor_name(base_pred_arity)
+        base_args = heap_slice(state, addr + 1, base_arity)
+        combined = base_args ++ extras
+        total = base_arity + length(extras)
+        pred_arity = "#{name}/#{total}"
+        new_state = load_args_into_regs(state, 1, combined)
+        {:ok, pred_arity, new_state}
+      _ -> :fail
+    end
+  end
+  defp build_call_target(_state, _goal, _extras), do: :fail
+
+  # Write a list of arg values into consecutive A-regs starting at start_reg.
+  defp load_args_into_regs(state, start_reg, args) do
+    args
+    |> Enum.with_index(start_reg)
+    |> Enum.reduce(state, fn {arg, i}, s -> %{s | regs: Map.put(s.regs, i, arg)} end)
   end
 
   # Extract the functor-name part from a "name/arity" string.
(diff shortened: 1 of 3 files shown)


src/unifyweaver/targets/wam_elixir_target.pl
```diff
@@ -241,6 +246,11 @@
 '      # Fast path: label pre-resolved at codegen to integer PC.
       {:execute, target_pc} when is_integer(target_pc) ->
         %{state | pc: target_pc}
```
s243a, line 250: **Nit / robustness — `String.to_integer/1` can raise on a malformed opcode**

`String.to_integer(arity_str)` will throw an `ArgumentError` if `arity_str` is not a valid decimal string (e.g. if a future codegen bug emits `{:execute, "call/meta"}` or a trailing newline slips through). Consider guarding with `Integer.parse/1`:

```elixir
{:execute, "call/" <> arity_str} ->
  case Integer.parse(arity_str) do
    {total_arity, ""} -> dispatch_call_meta(state, total_arity, state.cp)
    _ -> :fail   # or throw({:unknown_builtin, "call/" <> arity_str, 0})
  end
```

The `:call` arm doesn't have this issue because the arity `n` comes directly from the decoded tuple field, not string parsing.

src/unifyweaver/targets/wam_elixir_target.pl
```diff
@@ -1157,6 +1180,86 @@
...
+  # label lookup and route here.
+  #
+  # Build the combined goal:
+  #   - A1 atom        (`true`, `fail`, `append`)  -> use as-is, arity = total - 1
+  #   - A1 compound    ({:ref, addr} -> {:str, "name/baseN"}) ->
+  #       gather base args from heap, append extras, total = baseN + (total - 1)
+  # Then load the combined args into A1..A_total, set cp to after_pc,
+  # and route through WamDispatcher.call (which knows about user
+  # predicates and falls through to execute_builtin for builtins).
+  @doc "Dispatch a call/N meta-call. After the called goal returns, control resumes at after_pc."
```
s243a: **`dispatch_call_meta/3` is `def` (public) — should it be `defp`?**

This helper is an internal runtime detail and nothing outside `wam_elixir_target.pl` (or the generated module) should call it directly. The existing parallel helpers (`build_call_target`, `load_args_into_regs`, `parse_functor_name`, etc.) are all `defp`. Keeping it public widens the API surface unnecessarily and could invite misuse from driver code.

If the public visibility is intentional (e.g. so the `WamDispatcher` or external test harness can call it directly), please add a comment explaining why.

src/unifyweaver/targets/wam_elixir_target.pl
```diff
@@ -1157,6 +1180,86 @@
...
+        for i <- 2..total_arity, do: deref_var(state, get_reg(state, i))
+      else
+        []
+      end
+
+    case build_call_target(state, goal, extras) do
+      {:ok, pred_arity, prepared_state} ->
+        call_state = %{prepared_state | cp: after_pc}
+        try do
+          case WamDispatcher.call(pred_arity, call_state) do
```
s243a, line 1229: **`{:return, post_state}` catch arm — semantics question**

The `catch {:return, post_state}` arm restores `pc` to `after_pc`, but `{:return, _}` is not a standard WAM control-flow signal documented elsewhere in the runtime (the canonical ones are `{:fail, state}` and the choice-point throw). Is this a legacy throw shape from an older version of the runtime, or does it originate from something like `cut` or `proceed`?

If it is dead code, removing it avoids confusion. If it is needed, a one-line comment explaining which instruction produces `{:return, post_state}` would help future readers.

src/unifyweaver/targets/wam_elixir_target.pl
```diff
@@ -1157,6 +1180,86 @@
...
+          end
+        catch
+          :fail -> :fail
+          {:fail, _} -> :fail
+          {:return, post_state} -> %{post_state | pc: after_pc}
+        end
+      :fail -> :fail
+    end
+  end
+
```
s243a, line 1260: **Missing goal shape: `{:unbound, _}` / free variable passed to `call/1`**

`build_call_target/3` handles binary strings, atoms, and `{:ref, addr}` compounds, but the catch-all `defp build_call_target(_state, _goal, _extras), do: :fail` will silently fail if an unbound variable (e.g. `{:unbound, {:heap_ref, n}}`) reaches it. ISO Prolog requires this to be an `instantiation_error`.

Consider adding:
```elixir
defp build_call_target(_state, {:unbound, _}, _extras) do
  throw({:error, :instantiation_error})
end
```
before the catch-all clause. This is consistent with how the rest of the runtime surfaces type errors (cf. `unknown_builtin` throw).

src/unifyweaver/targets/wam_elixir_target.pl
```diff
@@ -1144,6 +1159,14 @@
             preserved_aggs ++ state.cut_point
           end
         %{state | choice_points: new_cps, pc: new_pc}
+      {"call/" <> _, n} ->
+        # call/N meta-call. Lowered-mode dispatch arrives here via
```
s243a, line 1170: **`execute_builtin` `call/N` arm uses `state.cp` but tail position should use the *caller's* `cp`**

In the `:execute` step arm (line ~250) you correctly pass `state.cp` as `after_pc`, meaning "return to where the caller of `execute` was going to return". The same value reaches this `execute_builtin` arm via `dispatch_call_meta(state, n, state.cp)` which is correct.

However the `:call` step arm passes `state.pc + 1` as `after_pc` (non-tail), while this `execute_builtin` arm always passes `state.cp` regardless of how it was reached. In lowered mode, `WamDispatcher.call("call/N", state)` arrives here with `state.cp` already set to the correct return address by the lowered `call` instruction, so this is fine — but it would be worth a short comment noting the assumption so it is not accidentally broken when call-convention changes are made in future PRs:

```elixir
{"call/" <> _, n} ->
  # In lowered mode, the :call instruction has already set state.cp
  # to the correct return address before entering execute_builtin.
  dispatch_call_meta(state, n, state.cp)
```

tests/test_wam_elixir_target.pl
```diff
@@ -156,6 +156,54 @@
...
+        sub_string(ExecCode, _, _, _, '"call/" <> arity_str'),
+        sub_string(ExecCode, _, _, _,
+                   'dispatch_call_meta(state, total_arity, state.cp)')
+    ->  pass(Test)
+    ;   fail_test(Test, 'call/N dispatch missing from step arms')
+    ).
+
+test_true_zero_builtin :-
+    Test = 'WAM-Elixir: true/0 builtin arm exists',
+    (   compile_wam_helpers_to_elixir([], Code),
```
s243a, line 202: **Test coverage gap: `call/N` with a compound goal and extras is not unit-tested in `test_wam_elixir_target.pl`**

The four unit tests added here only verify *string presence* in the emitted code (emit-and-grep posture). The E2E tests in `test_wam_elixir_classic_programs.pl` cover `call_compound_goal` and `call_with_extras`, but those require a live Elixir install.

Consider adding a unit-level test for `build_call_target` with a `{:ref, addr}` input (the compound-goal heap-dereferencing path) to catch regressions without needing Elixir. Something like:

```prolog
test_call_n_compound_goal_dispatch :-
    Test = 'WAM-Elixir: build_call_target resolves {:ref, addr} compound goal shape',
    (   compile_wam_helpers_to_elixir([], Code),
        atom_string(Code, S),
        sub_string(S, _, _, _, 'defp build_call_target(state, {:ref, addr}, extras)')
    ->  pass(Test)
    ;   fail_test(Test, 'compound {:ref, addr} goal shape missing from build_call_target')
    ).
```

This is a low-cost addition that guards the most complex dispatch path.

s243a (commented):
## Review: `feat(wam-elixir): call/N meta-call` (PR #1 of 5)

### Summary

This PR correctly closes the `call/N` gap documented in `WAM_ELIXIR_GAPS_SPECIFICATION.md §3`. The three entry points (`:call` step arm, `:execute` step arm, `execute_builtin`) all funnel to a single `dispatch_call_meta/3` helper, which mirrors the C++ reference implementation cleanly. The `true/0` addition is a necessary companion fix. Tests are well-structured with four unit-level emit-and-grep checks plus four E2E acceptance tests.

### Findings

I've left inline comments on six items. In priority order:

1. **Medium — `String.to_integer/1` in the `:execute` arm can raise** (line ~248): use `Integer.parse/1` with a guard to avoid a crash on malformed opcodes.
2. **Medium — missing `instantiation_error` for an unbound variable goal** (line ~1226): `build_call_target/3`'s catch-all silently returns `:fail`; ISO requires an `instantiation_error` throw.
3. **Low — `dispatch_call_meta/3` is `def` (public)** (line ~1200): all sibling helpers are `defp`; widen the API surface only if there is a caller outside this module.
4. **Low — `{:return, post_state}` catch arm** (line ~1214): unclear which instruction produces this throw shape; either document or remove.
5. **Low — `execute_builtin` `call/N` arm `state.cp` assumption** (line ~1163): correct behaviour but the invariant deserves a comment for future maintainers.
6. **Suggestion — add a unit-level test for the `{:ref, addr}` compound-goal path** (test file line ~188): the most complex dispatch branch is only covered by E2E tests that require a live Elixir install.

### Positive notes

- The belt-and-suspenders approach (both interpreter and lowered modes covered) is the right architecture for a runtime that supports both execution paths.
- The block comment above `dispatch_call_meta/3` is excellent — it explains the data-flow clearly and cross-references the C++ analogue.
- No changes to existing `\+/1` / `findall/3` paths; the new arm is purely additive. ✅
- All four new unit tests and all four new E2E acceptance tests are correctly wired into `run_tests/0`. ✅

src/unifyweaver/targets/wam_elixir_target.pl
```diff
@@ -1157,6 +1180,86 @@
     end
   end
 
+  # ---- call/N meta-call dispatch -----------------------------------
+  #
+  # Mirrors the C++ dispatch_call_meta in wam_cpp_target.pl. The
```
s243a: **Off-by-one: `dispatch_call_meta` reads `total_arity` extras but A1 IS the goal, so extras should span `A2..A_total_arity`, i.e. `total_arity - 1` extra args**

The `total_arity` argument this function receives is the **full** arity of the emitted `call/N` opcode — meaning `N = 1 + k` where `k` is the number of extra arguments appended to the base goal. A1 holds the goal itself; A2..A_N hold the extras.

The current code:
```elixir
extras =
  if total_arity > 1 do
    for i <- 2..total_arity, do: deref_var(state, get_reg(state, i))
  else
    []
  end
```

`2..total_arity` generates `total_arity - 1` elements, which is correct — but only because Elixir ranges are inclusive on both ends and A1 is skipped. **This is correct as written**, but the comment block above says `"arity = total - 1"` for the atomic-goal case and then in `build_call_target/3` the pred_arity is computed as `"#{goal}/#{arity}"` where `arity = length(extras) = total_arity - 1`. So for `call(append, [a,b], [c,d], R)` emitted as `{:call, "call/4", 4}`, this would look up `"append/3"` — which is correct.

However, the **`execute_builtin` arm** receives `n` directly from the pattern `{"call/" <> _, n}`, but at that callsite the `n` is the **second element of the `{op, arity}` tuple** passed to `execute_builtin(state, op, arity)` — which is the builtin arity as registered, NOT the `total_arity` of the `call/N` opcode.

Check how `execute_builtin` is invoked from `builtin_call`: `{:builtin_call, op, arity} -> execute_builtin(state, op, arity)`. For a `call/4` the WAM compiler would emit `{:builtin_call, "call/4", 4}`, so `n = 4` and the extras span is `2..4` = 3 extras. That maps to `"append/3"` — still correct.

But: the `:execute` arm computes `total_arity = String.to_integer(arity_str)` from the opcode string `"call/4"`, giving `4`, which is also correct. **All three entry paths are consistent** — but this multi-path arity threading is fragile. A single table clarifying which `n`/`total_arity` value each entry point supplies would prevent future regressions.

src/unifyweaver/targets/wam_elixir_target.pl
```diff
@@ -230,6 +230,11 @@
 '      # Fast path: label pre-resolved at codegen to integer PC.
       {:call, target_pc, _n} when is_integer(target_pc) ->
         %{state | pc: target_pc, cp: state.pc + 1}
+      # Meta-call call/N — caught BEFORE label lookup so the
+      # "call/N" string is dispatched as the meta-call, not
```
s243a, line 234: **Arity mismatch: `:call` tuple carries `n` = *total* WAM arity (goal + extras), but `dispatch_call_meta` treats it as `total_arity` and reads A1..A_n**

In the WAM calling convention, `{:call, "call/N", N}` means "call the predicate at label `call/N` with N argument registers loaded". So `n` here equals the **total** number of argument registers used by the `call/N` instruction — i.e. A1 holds the base goal, and A2..An hold the extras. That is exactly what `dispatch_call_meta/3` expects, so the arithmetic is correct.

However the variable name `n` in the pattern match is easy to misread as "number of extras" (which would be `n - 1`). The `:execute` arm is unambiguous because `arity_str` comes from string-splitting the opcode name. Consider renaming to avoid future confusion:

```elixir
{:call, "call/" <> _, total_arity} ->
  dispatch_call_meta(state, total_arity, state.pc + 1)
```

This also makes it consistent with the naming used inside `dispatch_call_meta` itself (`total_arity`).

src/unifyweaver/targets/wam_elixir_target.pl
```diff
@@ -1157,6 +1180,86 @@
...
+  #
+  # Mirrors the C++ dispatch_call_meta in wam_cpp_target.pl. The
+  # WAM compiler emits call(Goal[, X1..Xk]) usage as
+  #   {:execute, "call/N"}        -- when call/N is the body tail
+  #   {:call,    "call/N", N}     -- otherwise (non-tail)
+  # with the base goal in A1 and any extras in A2..AN. The :call
+  # and :execute step arms catch the "call/" prefix BEFORE doing
+  # label lookup and route here.
+  #
+  # Build the combined goal:
```
s243a, line 1200: **`dispatch_call_meta/3`: `total_arity = 0` is a valid ISO case (`call/1` with just the goal) but the `extras` branch uses `2..total_arity` which **crashes** when `total_arity = 1`**

In Elixir, `for i <- 2..1` produces `[2, 1]` (a *descending* range), not an empty list. The guard `if total_arity > 1` correctly avoids the range, so the `extras = []` path is taken for `call/1` (total_arity = 1). This is safe.

But consider `total_arity = 0`: that would mean `call/0` was somehow emitted. The range `for i <- 2..0` would yield `[2, 1, 0]` — three spurious register reads — because the guard `total_arity > 1` is false, so it again takes `extras = []`. So `= 0` is also safe *by coincidence* of the guard.

This is correct today, but it is fragile reasoning. A defensive `total_arity >= 1` assertion or a comment explaining that `call/0` cannot be emitted by the WAM compiler would make the invariant explicit:

```elixir
# total_arity >= 1 guaranteed: WAM always emits at least the base-goal register.
extras =
  if total_arity > 1,
    do: for(i <- 2..total_arity, do: deref_var(state, get_reg(state, i))),
    else: []
```

src/unifyweaver/targets/wam_elixir_target.pl
s243a: **`dispatch_call_meta/3`: `after_pc` is written back with `%{post_state | pc: after_pc}` even when `WamDispatcher.call` returns `{:ok, post_state}` — this overwrites any `pc` advance the callee already made**

After `WamDispatcher.call` succeeds, `post_state.pc` should already be pointing at the correct continuation (the callee has run to completion and left `pc` at its return point). Overwriting it with `after_pc` a second time is redundant *if* the callee correctly restores `cp` and `pc` on success — but it would be **wrong** if the callee is itself a meta-call that has side-effects on `pc` that should be preserved.

In the `:return` catch arm the same overwrite happens. Please confirm that `WamDispatcher.call` is always guaranteed to return a `post_state` whose `pc` equals `after_pc` on success, or add a comment explaining the invariant. If that cannot be guaranteed, the correct pattern is:

```elixir
{:ok, post_state} -> post_state  # pc already set by callee
```

and rely on the `cp` set before the call to carry the return address.

src/unifyweaver/targets/wam_elixir_target.pl
```diff
@@ -1157,6 +1180,86 @@
...
+        call_state = %{prepared_state | cp: after_pc}
+        try do
+          case WamDispatcher.call(pred_arity, call_state) do
+            {:ok, post_state} -> %{post_state | pc: after_pc}
+            :fail -> :fail
+          end
+        catch
+          :fail -> :fail
+          {:fail, _} -> :fail
+          {:return, post_state} -> %{post_state | pc: after_pc}
```
s243a, line 1255: **`build_call_target` binary-goal arm: `load_args_into_regs` overwrites A1 — but A1 already holds the base goal and the extras start at A2**

For the atomic goal case (`call(append, [a,b], [c,d], R)`):
- A1 = `"append"` (base goal)
- A2 = `[a,b]`, A3 = `[c,d]`, A4 = R (extras, `length(extras) = 3`)
- `pred_arity = "append/3"`
- `load_args_into_regs(state, 1, extras)` writes extras into **A1, A2, A3**

This is correct for `append/3` because `append` expects its three arguments in A1..A3. The base-goal atom itself is no longer needed as a register value once the predicate key is built.

However the comment in the block comment above (`dispatch_call_meta`) says *"load the combined args into A1..A_total"* which would imply A1 gets the first extra (`[a,b]`) — that is what the code does. But for the **compound** goal case (`{:ref, addr}`), `combined = base_args ++ extras` is loaded from A1, meaning A1 gets the *first base arg*, not the original A1 compound reference. This is correct.

The subtle danger is: **what if `extras` is empty (call/1)?** Then `load_args_into_regs(state, 1, [])` is a no-op and A1 still holds the atom string `"true"` / `"fail"` / etc. Those goals expect 0 arguments in no registers, so A1's value is irrelevant. Still correct.

No code change required, but please add a one-line comment on the binary arm clarifying that `start_reg = 1` is intentional (extras displace A1 because the goal atom is consumed into the pred-key string):

```elixir
# Extras go into A1..A_arity; the atom itself is consumed into pred_arity.
new_state = load_args_into_regs(state, 1, extras)
```

src/unifyweaver/targets/wam_elixir_target.pl
```diff
@@ -1157,6 +1180,86 @@
...
+  # Build the combined predicate-key + load A-regs from the goal+extras.
+  # Atomic goal: arity is just the count of extras.
+  # Compound {:ref, addr} -> {:str, "name/baseN"}: read base args
+  # from heap, concat extras, total = baseN + length(extras).
+  defp build_call_target(state, goal, extras) when is_binary(goal) do
+    arity = length(extras)
+    pred_arity = "#{goal}/#{arity}"
+    new_state = load_args_into_regs(state, 1, extras)
+    {:ok, pred_arity, new_state}
+  end
```
s243a, line 1270: **`build_call_target` compound arm: `heap_slice(state, addr + 1, base_arity)` — is `heap_slice` bounds-checked?**

If a malformed or partially-initialised heap cell has `{:str, "foo/3"}` but the heap only contains 1 or 2 words after `addr`, `heap_slice` will silently return a short list (or crash, depending on its implementation). The resulting `combined` list would have fewer elements than `total` claims, and `load_args_into_regs` would load fewer registers than `pred_arity` expects — leading to a silent `get_reg` returning `nil` inside the callee.

Please confirm `heap_slice/3` raises or returns an error tuple on out-of-bounds access, or add a length check:

```elixir
base_args = heap_slice(state, addr + 1, base_arity)
if length(base_args) != base_arity, do: throw({:error, :heap_corrupt})
```

This is especially relevant when the compound goal was built by partially-evaluated code that may have left trailing unbound cells.

src/unifyweaver/targets/wam_elixir_target.pl
```diff
@@ -826,6 +836,11 @@
         # failure. (Pre-hardening, fail/0 worked accidentally because
         # the default arm returned :fail — same semantic, wrong reason.)
         :fail
```
s243a, line 845: **`true/0` builtin: `pc` increment is guarded by `is_integer(state.pc)` — inconsistent with every other builtin arm**

Every other builtin arm in this `execute_builtin` block unconditionally does `state.pc + 1` (or sets `pc` to a specific jump target). The `is_integer` guard here is unique and suggests `state.pc` might sometimes be a non-integer at this call site, which would be a pre-existing bug worth surfacing rather than silently swallowing.

If `state.pc` is always an integer when `execute_builtin` is reached (which it should be in a well-formed WAM state), remove the guard and make it consistent:

```elixir
{"true/0", 0} ->
  %{state | pc: state.pc + 1}
```

If it is *not* always an integer here, that is a runtime invariant violation that should be caught with a `raise` rather than silently preserved.

tests/test_wam_elixir_classic_programs.pl
```diff
@@ -66,6 +66,22 @@
...
+% --- call/N meta-call acceptance cases (PR #1) ---
+% Spec: docs/design/WAM_ELIXIR_GAPS_SPECIFICATION.md §3 PR #1.
+% Each predicate exercises one shape of meta-call dispatch:
+%   - elx_call_true/0       atom goal, 0 extras  (succeeds)
+%   - elx_call_fail/0       atom goal, 0 extras  (fails)
+%   - elx_call_compound/1   compound goal, 0 extras (heap-built goal)
+%   - elx_call_extras/1     atom goal, 3 extras   (partial-application)
+:- dynamic user:elx_call_true/0.
+:- dynamic user:elx_call_fail/0.
+:- dynamic user:elx_call_compound/1.
```
s243a, line 78: **`elx_call_compound/1` comment says "0 extras" but the goal `(X is 1+1)` is a compound term — the comment is misleading**

The header comment reads:
```
%   - elx_call_compound/1   compound goal, 0 extras (heap-built goal)
```

"0 extras" is technically correct (there is no partial application — only `call/1` is used), but calling it a "heap-built goal" alongside "0 extras" may confuse readers into thinking it tests the `{:ref, addr}` + `extras` dispatch path. It actually tests the `{:ref, addr}` + empty-extras compound path.

A clearer wording:
```prolog
%   - elx_call_compound/1   compound goal {:ref, addr}, no extra args (tests heap deref in build_call_target)
```

This is a documentation-only nit but the comments are the primary guide for what each acceptance test covers.

s243a (commented):
## Second Review Pass — `feat(wam-elixir): call/N meta-call` (PR #1 of 5)

This is an independent second pass focused on logic correctness, register-convention invariants, and edge cases not covered in the first review. Six new inline comments; none overlap with the first review's findings.

### New findings summary

| # | Priority | File | Topic |
|---|----------|------|-------|
| 1 | **Medium** | `wam_elixir_target.pl` line ~234 | Ambiguous variable name `n` in `:call` arm — rename to `total_arity` for clarity |
| 2 | **Medium** | `wam_elixir_target.pl` line ~1193 | Elixir descending-range trap: `2..total_arity` when `total_arity = 1`; correct today but needs a comment |
| 3 | **Medium** | `wam_elixir_target.pl` line ~1200 | Double `pc` overwrite after `WamDispatcher.call` succeeds — needs invariant confirmation or comment |
| 4 | Low | `wam_elixir_target.pl` line ~1221 | `load_args_into_regs` starts at A1 (displacing base-goal atom) — correct but needs a comment |
| 5 | Low | `wam_elixir_target.pl` line ~1236 | `heap_slice` bounds not verified — malformed heap could load wrong register count silently |
| 6 | Low | `wam_elixir_target.pl` line ~838 | `true/0` `is_integer(state.pc)` guard is inconsistent with every other builtin arm |
| 7 | Nit | `test_wam_elixir_classic_programs.pl` line ~78 | `elx_call_compound/1` comment says "0 extras" in a way that obscures what path it tests |

### Overall assessment

The core dispatch logic is sound and the C++ parity goal is achieved. The most actionable finding is #3 (potential double `pc` overwrite), which needs a comment at minimum to confirm the `WamDispatcher.call` return-value invariant. Findings #1 and #2 are naming/clarity issues that will help the next PRs in this series (catch/3, throw/1) land cleanly against a readable base.

PR #2612, opened 2026-05-30
Design note: metric-tree-likeness as a (graph, metric)-pair property

s243a:
## Summary

  Records a property we observed empirically while running the direction-weighted bidirectional kernel against the
  now-correctly-ingested simplewiki graph (PR #2568 + parser fix PR #2569). The new `correct`-mode ingest gave us
  294,773 walkable edges keyed by real page_ids — enough to run a meaningful convergence experiment.

  **The observation**: the weighted effective-distance metric `d_wPow` converges to four significant figures (5.044 →
  5.045) as we relax the budget from 1,118 upward-only paths to **2.7 million** mixed paths — a 2400× increase in path
   count that moves the metric **0.02%**. Per-seed (20 seeds at depth 4 from Physics), the worst-case drift is
  **0.007%**. The aggregate isn't hiding bimodal behavior; the convergence is genuinely per-pair.

  ## Why it happens (the geometric-series argument)

  Each path's weight is `w = (1/D)^N · (1/(b·D))^M` where:
  - `N` = parent hops, `M` = child hops
  - `D` = avg child fan-out (simplewiki: 7.34)
  - `b` = `E[d²_child] / E[d²_parent]` (simplewiki: 1353)

  So `b·D ≈ 9933`, meaning each additional child hop multiplies a path's weight by ~10⁻⁴. The number of M-child-hop
  paths grows ~15–100× per level within the budget, but the weight shrinks 10⁴× per level — a geometric series whose
  ratio is tiny. The total mixed-path contribution is bounded above by a convergent series whose first term already
  determines the answer to 4 significant figures.

  ## The operational claim

  > A (graph, metric) pair is **metric-tree-like** when searching the graph as if it were a tree (ignoring
  cross-edges) gives a metric value statistically close to the full-DAG value.

  The property **attaches to the pair** — neither alone suffices:
  - Same simplewiki graph under uniform weighting (prior experiments in PR #2502): metric drifted 90% (5.04 → 0.29).
  - Same calibrated power-mean metric on a symmetric DAG would presumably not be metric-tree-like (the falsification
  test, deferred).
  - The calibration of `b` and `D` is what *couples* graph to metric; the kernel already computes both during setup.

  ## What this is NOT

  Distinct from three existing concepts that sound similar:

  - **Structural tree-likeness (treewidth)**: simplewiki has enormous cycle space — every cross-cutting categorization
   is a cycle. Treewidth would be huge. Metric-tree-likeness measures something different.
  - **Gromov hyperbolicity**: a geometric property of distance, independent of any weighting.
  - **Bounded cycle space + small graph**: doesn't apply; the graph genuinely has lots of cross-edges.

  It's a third notion — a property of the (graph, metric) pair specifically.

  ## Algorithmic consequence

  A cheap convergence probe (run the kernel at `cc=∞` and `cc=5`, measure drift) is a **runtime certificate**: if
  drift < 0.1%, use tree-search safely. On simplewiki this is a measured **~2000× speedup** (11 ms vs ~43 s) with no
  observable accuracy loss to 4 significant digits.

  The framing is honest: it's a statistical claim about a specific (graph, metric, query-distribution) tuple, not a
  universal one. Some pair might fail per-pair convergence while passing aggregate — the probe can check both cheaply.

  ## What's in this PR

  A single design note at `docs/design/METRIC_TREE_LIKENESS.md` with:

  1. The empirical observation (full table of childCost × path count × metric values + per-seed table)
  2. The geometric-series mechanism with the exact weight formula
  3. The operational definition (aggregate vs per-pair versions)
  4. What this is NOT (treewidth, hyperbolicity, etc.)
  5. Evidence so far (simplewiki, with cross-reference to PR #2502 broken-ingest baseline)
  6. Open questions (sufficient/necessary conditions; falsification candidates)
  7. Algorithmic consequences (tree-search certificate + drift diagnostic)
  8. References to companion design docs

  ## What's NOT in this PR

  - **Not a theorem.** Explicitly flagged as such in §7. The name and framework are useful even without the proof.
  - **Not the falsification test.** Filed as a follow-up task: construct a symmetric DAG / diamond graph /
  bipartite-choke graph and verify the metric drifts more. If the dichotomy holds, we have a clean empirical
  distinction worth promoting into the cost-model framework.
  - **Not a code change.** Doc-only PR; no behavior modified.

  ## Test plan

  - [x] Markdown renders cleanly (verified locally).
  - [x] No code changes, so existing tests are unaffected.
  - [ ] Follow-up: synthetic-graph falsification test (filed).

### DIFF
--- a/docs/design/METRIC_TREE_LIKENESS.md
+++ b/docs/design/METRIC_TREE_LIKENESS.md
@@ -43,6 +43,29 @@ drift was 0.007% — every single pair converged.
 
 ## 2. Why this happens: the geometric-series argument
 
+### 2.0 Notation
+
+Several variant symbols denote the branching asymmetry at different
+stages of refinement. They are not interchangeable; the distinction
+matters for §4–§5.
+
+| Symbol | Definition | Where it's computed | Example value (simplewiki) |
+|---|---|---|---|
+| `D` | `E[d_child]` — average child fan-out | first moment of the child-degree distribution | 7.34 |
+| `b` (early) | `E[d²_child] / E[d²_parent]` — raw second-moment ratio | scan child/parent degree distributions globally | 1353 |
+| `b_eff` | `(E[d²_c]/E[d_c]) / (E[d²_p]/E[d_p])` — friendship-paradox-corrected branching asymmetry | same scan, with first-moment correction | 589 (global) / 9.59 (topical, §4.5) |
+| `BranchRatio` | `b_eff × routing_correction` — the composed scalar the kernel passes to the metric | adds the empirical routing factor | 226 (global) / 3.68 (topical, currently — see §5.4) |
+| `b'` | per-child-hop empirical path-count growth | measured by running the kernel at varying `cc` and reading path-count ratios | ~11 |
+
+`b` (early-formula) was used in §2 below before the first-moment
+correction; subsequent sections use `b_eff` exclusively. **The
+operative quantity for convergence is `b' < b_eff · D`, not the
+raw `b·D`.** §4.4 and §5.5 explain why the early `b·D ≈ 9933`
+number, while not wrong, is misleading once we account for
+inhomogeneity.
+
+### 2.1 The geometric-series argument
+
 Each path's weight in the metric is
 
 ```
@@ -50,11 +73,14 @@ w(path) = (1/D)^N · (1/(b·D))^M
 ```
 
 where N = parent hops, M = child hops, D = avg child fan-out, and
-b = E[d²_child] / E[d²_parent] (the calibrated branching asymmetry).
+b = E[d²_child] / E[d²_parent] (the early calibrated branching
+asymmetry — see §2.0; honest topical calibration uses `b_eff`
+instead, with a different numerical value).
 
-On simplewiki: D ≈ 7.34, b ≈ 1353. So a single child hop carries
-weight ≈ `1/(b·D) ≈ 10⁻⁴`, and each additional child hop multiplies
-by another factor of 10⁻⁴.
+On simplewiki under the original global calibration:
+D ≈ 7.34, b ≈ 1353. So a single child hop carries weight
+≈ `1/(b·D) ≈ 10⁻⁴`, and each additional child hop multiplies by
+another factor of 10⁻⁴.
 
 The contribution to the weighted sum from paths with M child hops
 is bounded by
@@ -69,6 +95,16 @@ growth by 2+ orders of magnitude. Result: a convergent geometric
 series whose terms shrink by ~10⁴ per level. The total contribution
 of mixed paths is asymptotically negligible compared to the
 upward-only baseline.
+
+> **Forward reference.** The `b·D ≈ 9933` figure here is the
+> *global* calibration. §4.4 shows this overestimates the
+> traversed branching by ~50× because the global moment scan
+> includes admin hubs the search never touches; §4.5 derives the
+> honest topical recalibration `b_eff ≈ 9.59` giving `b_eff·D ≈ 70`.
+> Both numbers satisfy the convergence inequality `b·D > b'` (with
+> empirical `b' ≈ 11` per §4.4), which is why the property holds
+> under either calibration — but §5.5 explains why the topical
+> number is the principled one for any quantitative use of `(b, D)`.
 
 ## 3. Proposed formal property: metric-tree-likeness
 
--- a/docs/design/METRIC_TREE_LIKENESS.md
+++ b/docs/design/METRIC_TREE_LIKENESS.md
@@ -683,7 +683,25 @@ For high-value queries, an extra 20-seed sample at cc=5 + cc=100
 costs ~1 second on simplewiki. The aggregate-vs-per-pair check
 gives us early warning if some pairs drift while others don't.
 
-### 6.4 Data preparation: build the LMDB from a topical root
+## 7. Data-prep consequences of the inhomogeneity finding
+
+§6 follows directly from the property: if a (graph, metric) pair
+is metric-tree-like, you can use tree-search. The items below are
+different — they are *production recommendations* that follow from
+§4.5's inhomogeneity finding, not from the property itself. They
+depend on the §4.5 evidence holding up (which it does on
+simplewiki) and on the §5.4 open question about the routing
+correction (which is *not yet settled*; see task #14).
+
+> **Status:** these recommendations are **tentative** until §5.4
+> is settled by task #14. The §4.5 evidence is empirical and
+> robust; the §5.4 claim that routing correction can be dropped
+> is supported by argument but not yet by a re-run of the drift
+> probe under topical calibration. Treat §7.1's "no
+> routing-correction band-aid" as conditional on task #14
+> confirming the drift criterion still holds without that factor.
+
+### 7.1 Data preparation: build the LMDB from a topical root
 
 The inhomogeneity result (§4.5) has a direct production
 consequence: **the LMDB should be built from a topical root, not
@@ -692,9 +710,11 @@ from the full category dump**.
 Recipe:
 
 1. Identify the topical root for the target wiki. On
-   simplewiki: `Category:Articles` (page_id 137597). On
-   enwiki: `Category:Main_topic_classifications` (page_id
-   7345184).
+   simplewiki: `Category:Articles` (page_id 137597, verified).
+   On enwiki: `Category:Main_topic_classifications` — the
+   page_id should be verified against the live wiki before use
+   (cited as 7345184 from earlier work, not re-verified in this
+   doc).
 2. During ingest, BFS from the topical root via the
    `cl_type = 'subcat'` edges; mark every reached page_id as
    in-scope.
@@ -706,25 +726,29 @@ Recipe:
 The result is an LMDB that contains the "topical core" only.
 Calibration on this LMDB will give `b_eff` matching the empirical
 per-hop branching out of the box (no topical-subgraph scoping
-needed at runtime, no routing-correction band-aid, no manual
-hub blacklist). The size reduction on simplewiki is ~75% (~80k
-in-scope nodes vs ~92k full), which is also cache-friendly.
+needed at runtime, no manual hub blacklist). The size reduction
+on simplewiki is ~75% (~80k in-scope nodes vs ~92k full), which
+is also cache-friendly. **Whether the routing-correction factor
+can be dropped from `BranchRatio` is task #14**; if it cannot,
+the same LMDB still works, the routing correction just stays
+applied at calibration time.
 
 Implementation: add `--filter-root <page_id>` to the
 `mysql_stream_lmdb` ingester (see the categorylinks 3-mode
 ingester for the natural extension point). The flag is optional;
 omitted, the full graph is ingested as today.
 
-### 6.5 Calibration recipe (summary)
+### 7.2 Calibration recipe (summary)
 
-Putting §4 and §6.4 together:
+Putting §4 and §7.1 together:
 
 1. **Build** the LMDB with `--filter-root <topical_root_id>`
-   (§6.4).
+   (§7.1).
 2. **Calibrate** at setup time: scan `category_child` and
    `category_parent` for degree moments, compute
-   `b_eff = (E[d²_c]/E[d_c]) / (E[d²_p]/E[d_p])`. No routing
-   correction (tentative — see §5.4).
+   `b_eff = (E[d²_c]/E[d_c]) / (E[d²_p]/E[d_p])`. Whether to
+   drop the routing correction factor is contingent on task #14
+   (see §5.4).
 3. **Certificate**: run the 20-pair drift probe at the production
    budget (§6.1). If `ε_agg < 0.1%`, deploy tree-search.
 4. **Per-query**: tree search using `(b_eff, D, parentCost)`.
@@ -735,7 +759,7 @@ entirely at step 1 (ingest), the calibration honesty lives at
 step 2, the certificate lives at step 3, and runtime stays
 fast.
 
-## 7. Status of this document
+## 8. Status of this document
 
 This is **not a theorem**. It's a name (`metric-tree-likeness`)
 attached to a phenomenon we've observed once, with a reasonable
@@ -759,10 +783,9 @@ a useful property worth adding to the cost-model framework:
 the fast tree-search path; pairs that fail get the full
 bidirectional search.
 
-## 8. References
+## 9. References
 
 - `WAM_FSHARP_CSR_KERNEL_INTEGRATION.md` — design of the
   bidirectional kernel itself.
 - `WIKIPEDIA_CATEGORYLINKS_INGEST_MODES.md` — the ingester whose
   correct-mode output made this measurement possible.
-- Simplewiki benchmark log: `/tmp/per_seed.log` (transient).
--- a/docs/design/METRIC_TREE_LIKENESS.md
+++ b/docs/design/METRIC_TREE_LIKENESS.md
@@ -78,6 +78,17 @@ upward-only baseline.
 > gives a metric value statistically close to the full-DAG
 > value.
 
+> **Homogeneity precondition (added on the basis of §4.5).** The
+> definition above assumes the graph is *statistically
+> homogeneous* — every region the query distribution can reach
+> has the same characteristic degree distribution and the same
+> calibrated `b_eff`. For inhomogeneous graphs the property must
+> be re-stated as: **there exists a decomposition into
+> homogeneous subgraphs such that each subgraph is metric-tree-
+> like under its own calibration**; the query's reachable set
+> determines which subgraph's calibration applies. See §4.5 for
+> the topical-core example on Wikipedia.
+
 This is the *practical* formulation. It tells you what to do —
 "just run tree search, the answer will be close enough" — rather
 than just describing a property of the graph.
@@ -189,6 +200,83 @@ drifts 0.02%. Two interpretations, both probably partly true:
 Either way, the convergence rate *changed with the graph*, which
 is what the property predicts.
 
+### 4.4 Calibration vs empirical: the inhomogeneity gap
+
+Running the full calibration probe on the correct-ingest simplewiki
+gives concrete numbers for each correction term:
+
+| Quantity | Value | Interpretation |
+|---|---|---|
+| `E[d²_c] / E[d²_p]` (second-moment ratio) | 1353 | Raw distributional asymmetry |
+| `× E[d_p] / E[d_c]` (first-moment correction) | 0.436 | Friendship-paradox normalisation |
+| `b_eff` (global, traversal-effective) | **589** | Calibrated asymmetry, global graph |
+| `× routing_correction` (avg_min_dist / avg_path_hops) | 0.384 | Hub-correlation pruning |
+| Final `BranchRatio` (global) | 226 | Composed calibration |
+
+And the *empirical* per-child-hop branching ratio (measured directly
+by running the kernel at multiple childCost values and reading the
+path-count growth):
+
+| childCost transition | extra child hops allowed | ratio of path counts |
+|---|---|---|
+| 100 → 10 | 0 → 1 | 1.05 (essentially none) |
+| 10 → 7  | 1 → 2 | **16.19** |
+| 7  → 5  | 2 → 3 | **7.58**  |
+
+Geometric mean across the meaningful transitions: **~11×** per added
+child-hop level. The global calibration says 226 (or 1353 without
+any corrections); the search actually experiences ~11.
+
+This is a **20–120× gap between calibration and reality** that no
+amount of correction term tuning fully closes — because it isn't a
+correction-factor problem. It's a *graph-decomposition* problem.
+
+### 4.5 Homogeneity evidence: the topical core
+
+The gap closes essentially completely when calibration is restricted
+to the **topical subgraph** — the descendants of a topical root like
+`Category:Articles` (page_id 137597 on simplewiki, analogous to
+`Category:Main topic classifications` on enwiki).
+
+| Calibration scope | b_eff | Admin hubs in scope | vs empirical (~11) |
+|---|---|---|---|
+| Global (whole simplewiki) | 589 | 20/20 | 53× too high |
+| `Category:Contents` (includes admin subroot) | 589 | 20/20 | 53× too high |
+| `Category:Articles` (topical core) | **9.59** | 3/20 (semi-topical only) | **within 13%** |
+| `Category:Physics` (topical subtree) | 9.51 | 3/20 (same three) | within 14% |
+
+Two things this shows:
+
+1. **The topical core is statistically homogeneous.** `Category:Articles`
+   and `Category:Physics` give essentially the same b_eff (9.59 vs
+   9.51, agreement to <1%). They reach different sets of nodes but
+   measure the *same statistical regime*. If the topical core were
+   itself inhomogeneous, these two scopes would disagree materially.
+   They don't.
+
+2. **The 20–120× calibration error was driven entirely by admin
+   hubs.** Once we BFS from a topical root, the maintenance
+   categories (`CatAutoTOC generates no TOC`, `Hidden categories`,
+   `Navseasoncats…`, etc.) are not reached and don't enter the
+   moment calculations. Three semi-topical categories remain
+   (`Births by year`, `Deaths by year`, `Songs by artist`), and
+   they're genuinely topical meta-indexes — their fan-out reflects
+   real categorical structure, not maintenance noise.
+
+The implication: **Wikipedia's category graph is inhomogeneous —
+two qualitatively different statistical regimes mixed together**.
+The topical regime (γ ≈ 2.4–2.5, characteristic fan-outs 1–100,
+tree-like organisation) and the administrative regime (list-shaped
+categories with fan-outs in the thousands, no internal hierarchy).
+Each regime can be metric-tree-like under its own calibration;
+together they are not metric-tree-like under any single calibration.
+
+The routing correction (the 0.384 factor in §4.4) was effectively
+*band-aiding* this inhomogeneity — compensating for hub categories
+that the search wasn't actually traversing. **With proper subgraph
+scoping the routing correction becomes unnecessary**: the topical
+b_eff = 9.59 already matches empirical without it.
+
 ## 5. Open questions
 
 ### 5.1 Sufficient conditions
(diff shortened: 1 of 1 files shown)


s243a (commented):
## Review: Design note — metric-tree-likeness as a (graph, metric)-pair property by perplexity Using Deep Research

Overall this is a well-crafted design note. The empirical grounding is strong, the geometric-series explanation is clear and honest, and the "property attaches to the pair" framing is exactly the right level of abstraction. I have a mix of structural suggestions, precision nits, and one broader open question about formalization readiness.

---

### ✅ What's working well

- **Honesty about status.** §7 is exemplary: explicitly not a theorem, one observation, explicitly lists falsification candidates. This is the right epistemics for a design note.
- **Operational definition first.** Leading with the actionable framing ("will tree-search give me an answer close enough?") before the formal one is a good choice for a practitioner document.
- **The b·D decomposition in §2** is satisfying. Having D ≈ 7.34 and b ≈ 1353 concrete lets a reader sanity-check the 10⁻⁴ per-hop claim immediately.
- **§4.3 (broken-ingest comparison)** is underrated — it's effectively a quasi-natural experiment: same graph topology (roughly), same metric, b went from 61 → 1353, drift went from ~1% → 0.02%. That's arguably the strongest evidence in the doc that b is the causal driver.

---

### 🔧 Structural / Content Suggestions

**1. §2 — weight formula needs a denominator clarification**

The formula `w(path) = (1/D)^N · (1/(b·D))^M` silently assumes the normalisation denominators are exactly `D` for parent hops and `b·D` for child hops. That's the calibrated model — but a reader who hasn't read the kernel design doc won't know where `b·D` came from as a *per-hop* penalty. Consider adding a one-sentence bridge: *"The bidirectional kernel assigns each parent hop a weight of 1/D and each child hop a weight of 1/(b·D), where b is computed from the ratio of second moments E[d²_child]/E[d²_parent] during setup (see WAM_FSHARP_CSR_KERNEL_INTEGRATION.md)."*

**2. §3 — the two versions of "statistically close" need thresholds**

Aggregate vs. per-pair is clearly distinguished, but the definition still says "statistically close" without pinning what ε means. Even a parenthetical like *(ε = 0.1% in our running example)* would make the operational definition actually operational. As written, a user can't evaluate whether their own graph+metric passes without knowing what the bar is. The convergence probe in §6.1 uses < 0.1% — that threshold should probably appear in §3 as the default.

**3. §4.2 — the per-seed table is truncated at 5 rows**

You ran 20 seeds; the worst-case seeds (211205 and 144727 at 0.007%) are in the table, which is good. But two of the five rows have drift 0.000–0.001%, which makes the sample look rosier than the full distribution. Consider either: (a) showing all 20 rows, (b) adding a histogram/percentile row, or (c) explicitly noting the median and 95th-percentile drift in the prose. The text already says "19 of 20 converged to within 0.003%" — that sentence does the work, but it would be even cleaner to have it next to the table rather than above it.

**4. §5.3 (falsification) — diamond graph description is slightly imprecise**

> *"Diamond graph with two distinct distance regimes between every leaf-pair, depending on whether you go up-then-down or the other way"*

This describes a scenario where the forward and reverse paths have *different costs*, but in the standard diamond (A→B, A→C, B→D, C→D), there are two paths A→D, not two distance regimes unless the edge weights differ. The intuition is right — you want a graph where non-tree cross-edges carry comparable weight to tree edges — but the example as stated would need asymmetric weights to actually defeat the metric. Either sharpen the example (*"assign B→D weight 1 and C→D weight 1, while the tree path A→B→D has weight (1/D)^2 — for small D the cross-path is competitive"*) or replace it with the symmetric DAG case in the same paragraph, which is the cleaner falsification.

**5. §6.3 — "costs ~1 second on simplewiki" needs context**

This is in the algorithmic-consequences section but the hardware / environment isn't specified. For reproducibility and cost-model use, it would help to note whether this is single-threaded F# on a laptop, the WSL environment, LMDB-backed, etc. Even a note that it scales with graph size would let a reader extrapolate to their own use case.

---

### 🔬 Precision / Nit-level

- **§1 table header**: "mixed (child-hop) paths" — consider "paths with ≥1 child hop" to avoid ambiguity; "mixed" could imply exactly one of each.
- **§2 bound**: `contribution(M) ≤ (number of M-child-hop paths) · (1/(b·D))^M` — this is correct but is technically an upper bound on a contribution to the path *count*, not to the weighted sum directly. The inequality holds for the weighted sum too (since each path weight ≤ (1/(b·D))^M for paths with exactly M child hops), but spelling that out in one line would make the argument self-contained.
- **§3.1 "not Gromov hyperbolicity"** — correct that they're different, but worth adding that *some* metric-tree-like graphs might *also* be Gromov hyperbolic (Wikipedia categories are empirically known to be approximately hyperbolic); the point is that metric-tree-likeness neither implies nor is implied by hyperbolicity.
- **§4.3**: `b ≈ 61` for the broken-ingest graph — this number isn't derived in this doc. A footnote or parenthetical citing where it was computed (presumably the §2 formula applied to the broken-ingest graph) would let a reader verify the story.

---

### ❓ Open question for the author

**Is the geometric-series argument tight, or just sufficient?**

The argument shows the contribution of M-hop child paths shrinks geometrically, which is sufficient to prove convergence. But is it also *tight* — i.e., could the property hold even when `b·D` is small (say ≈ 10), as long as the path-count growth is also slow? The empirical evidence (b ≈ 61, ~1% drift) suggests the answer is roughly "no — small b·D means bad convergence" — but it would sharpen §5.1 considerably to know whether the condition `b·D ≫ (path-count growth rate per child hop)` is really the operative criterion, rather than just `b·D large`.

---

### Minor markdown

- Line 48: `d_wPow` — consider wrapping in backticks consistently throughout (it's inline code for the metric name).
- Line 63: the `w(path) = ...` block could use a brief caption line before the variable definitions for readers scanning the document.
- The `**Date**: 2026-05-29` metadata at the top is useful but might belong in a front-matter block or a proper metadata table for consistency with other design docs in this repo.

---

### Verdict

**Approve with suggestions.** The core content is solid and honest about its status. The suggestions above are mostly about making the operational definitions more precise and the evidence presentation more complete — none of them block merge for a doc-only design note. The most important one is pinning the ε threshold in §3 so the definition is actually usable.

From the perplexity thread: https://www.perplexity.ai/search/599f78b3-8539-40f6-b5c4-c129568141d6

s243a (commented):
## Second Review: metric-tree-likeness design note (independent pass) by perplexity using GPT 4.5

*This is an independent second review. I've intentionally approached the document fresh, focusing on mathematical rigour, conceptual gaps, and practical completeness. Some overlap with the first review is unavoidable; where I overlap I've tried to add a different angle.*

---

### Overall assessment

The document earns its place. The naming of a property that was previously unnamed but operationally used is valuable on its own — even if the property is never elevated to a theorem, having a name lets future design docs say "this works because metric-tree-likeness" rather than re-deriving the geometric series each time. The status honesty is impeccable.

My main concerns are: (1) the definition in §3 is still informal where it needs to be precise, (2) the relationship between the *weight formula* and the *effective-distance metric* is never stated, leaving a gap in the argument, and (3) the algorithmic certificate in §6.1 is conflating two distinct things.

---

### 🔴 Substantive issues

**A. The link between the weight formula and `d_wPow` is implicit throughout**

The document shows the weight formula `w(path) = (1/D)^N · (1/(b·D))^M` and shows that `d_wPow` barely moves. But it never states how `d_wPow` is *computed from* path weights. The reader must infer that `d_wPow` is something like a power-mean over path lengths weighted by `w(path)`, but nowhere is the aggregation function written down.

This matters because the geometric-series argument in §2 bounds the *total weight* contributed by mixed paths, which only translates to "metric barely moves" if the metric is sufficiently Lipschitz in the path-weight distribution. For a power-mean that's plausible, but it's an assumption. State the aggregation formula (even just `d_wPow = (Σ_paths w(path) · len(path)^n)^(1/n)` or whatever it actually is) at the top of §2, then the argument closes.

**B. The definition in §3 uses "statistically close" in two different senses**

The aggregate version says the *means* match. The per-pair version says *every* pair matches. But "statistically close" in the aggregate sense usually means something like |E[d_tree] - E[d_DAG]| / E[d_DAG] < ε — a relative error on the expectation. For per-pair, it means max_i |d_tree(i) - d_DAG(i)| / d_DAG(i) < ε. These are different objects and the same ε might not apply to both. The evidence in §4 shows both hold with ε ≈ 0.01% for simplewiki, but a reader applying this to a different graph needs to know which ε to check and against what norm.

Suggestion: split the definition block into two explicit sub-definitions with their own ε notation — say ε_agg and ε_pair — and state that the empirical simplewiki values are ε_agg ≈ 0.02%, ε_pair ≈ 0.007%.

**C. §6.1 conflates the *certificate* with the *decision rule***

The convergence probe (run at cc=∞ and cc=5, measure drift < 0.1%) is described as a "runtime certificate". But a certificate in the algorithmic sense is a witness that the property holds *for this graph*, from which you can then deploy tree-search for *all* future queries on this graph. The document conflates this with a per-query runtime check. These are very different things:

- **Certificate** (one-time): Run the probe once on the graph at setup time. If it passes, permanently switch to tree-search. Cost: paid once.
- **Per-query check**: Run the probe for each query. This doesn't make sense at 11 ms vs 43 s — if the probe itself costs comparable time, there's no win.

The intended meaning is clearly the certificate interpretation ("for this (graph, metric, query-distribution) tuple"), but the phrasing "a cheap convergence probe" reads as per-query. §6.3 then says "for high-value queries, an extra 20-seed sample... costs ~1 second" which further blurs this. Clarify: the certificate is a *setup-time* check, and §6.3 is describing a *diagnostic mode* for auditing, not a per-query runtime overhead.

---

### 🟡 Gaps and missed angles

**D. The role of `budget` is underspecified**

The table in §1 uses budget=15 implicitly (stated in §4 but not §1). The convergence argument depends on the path-count growth rate, which depends on budget. For budget → ∞, even a small `b·D` might not be enough — you eventually enumerate enough child paths to shift the metric. The document should state explicitly what budget=15 implies for the geometric series (roughly: paths with M > 2 child hops are negligible, and the series terminates quickly within budget). Otherwise a reader might apply the certificate on a graph with budget=50 and get a false pass.

**E. §5.1 (sufficient conditions) conflates degree distribution with the calibration parameters**

The three hypotheses (power-law degree, dominant parent rule, sparse cross-edges) are stated as graph-structural properties, but the operative quantity is `b·D = E[d²_child]/E[d²_parent] · E[d_child]`. A graph could have a power-law degree distribution without having a large `b` — if parent and child fan-outs happen to be similar in second moment. The sufficient condition should be stated directly in terms of `b·D` (or equivalently, the ratio of second moments), with the power-law hypothesis as a *reason* that Wikipedia categories happen to have large `b·D` rather than as the condition itself.

**F. No discussion of query-distribution dependence**

The PR summary correctly notes "a statistical claim about a specific (graph, metric, query-distribution) tuple" — the query distribution is named as a parameter. But the document body never operationalizes it. The 20 seeds at depth 4 from Physics are a *specific* query distribution; a user querying from depth 1 (top-level categories) might see very different drift if the tree structure is sparser near the root. The open questions section would benefit from a question: *"Does the certificate hold uniformly over all query distributions, or only for the tested one?"* This isn't a blocker — it's an honest admission that would make the open-questions section more complete.

**G. §4.3 draws a causal conclusion from a confounded comparison**

The broken-ingest vs. correct-ingest comparison is compelling but the two graphs differ in more than just `b`: they differ in edge count (25k vs 294k), namespace composition, and likely graph diameter. The 90% drift in the broken graph (5.04 → 0.29) is so large it probably also reflects structural damage (incorrect cross-namespace edges creating false cross-paths) rather than just smaller `b`. The document partially acknowledges this ("Two interpretations, both probably partly true") but doesn't follow through. It would be stronger to say: *"We cannot cleanly isolate the effect of b from the graph structure change; the synthetic falsification test (§5.3) is what would isolate it."*

---

### 🟢 Minor / prose

- **§2, last paragraph**: "a convergent geometric series whose terms shrink by ~10⁴ per level" — this should be "shrink by a factor of ~10⁴ per level" (or "shrink 10⁴× per level"); "shrink by 10⁴" is ambiguous.
- **§3.2 title**: "the 'child shortcuts are statistically rare' property" — this framing emphasises the *path count* being rare, but the document has already established that the paths exist in abundance (2.7M at cc=3). The rarity is in *weight*, not count. Consider: "child shortcuts are statistically negligible" or "child shortcuts carry negligible weight".
- **§6.2**: "re-check b and D" — this implies miscalibration is the only other explanation for > 1% drift. A third explanation is that the query distribution has shifted to a region where per-pair metric-tree-likeness fails. Worth listing.
- **§7**: "The property attaches to the pair `(graph, metric)`" — this is the third time this statement appears verbatim. The first two instances (header and §3.2) already did the work. Consider replacing this with a sentence about what that means for the *cost model*: "A cost model that carries `(b, D)` already has what it needs to predict whether tree-search is safe."

---

### Verdict

**Approve with suggestions.** Issues A and C are the most actionable before merge: A is a one-sentence fix (state the aggregation formula), C is a paragraph clarification (certificate vs. per-query). B–G are good-to-have for long-term usefulness of the doc but don't block a design note of this status. The document is honest, well-structured, and introduces a genuinely useful concept.

From the perplexity thread: https://www.perplexity.ai/search/599f78b3-8539-40f6-b5c4-c129568141d6

s243a (commented):
## Re-review (second pass, updated branch)

The update is a substantial improvement. Sections 4.4, 4.5, 5.4, 5.5, 5.6, 6.4, and 6.5 are all new and collectively address the most important gaps from the previous reviews. The topical-core calibration result (b_eff = 9.59 vs empirical ~11) is the empirical highlight of the entire document now, and the §5.5 convergence-robustness analysis is genuinely insightful. Below I track which prior issues are resolved, what's partially addressed, and what is new.

---

### ✅ Previous issues: resolved

**Issue A (aggregation formula missing)** — Still not explicitly stated, but §5.6 now introduces the restatement `w ∝ 1/#paths(N,M)` and the ε-metric-tree-like criterion in §5.6.1. This is a *better* answer than just writing out the power-mean formula, because it explains *why* the weights have that form. I'm withdrawing this concern.

**Issue C (certificate vs. per-query conflation)** — §6.5 now clearly lays out the pipeline: step 3 is "one-time setup" certificate, §6.3 is explicitly "for high-value queries" diagnostic. The distinction is now recoverable from the document even if §6.1–6.3 still don't explicitly say "setup-time" vs. "per-query". Good enough for a design note.

**Issue F (query-distribution dependence)** — §4.5 adds the topical-core scope explicitly, and the homogeneity precondition in §3 names the dependency. The new §5.6.2 closes this carefully. Resolved.

**Issue G (§4.3 confounded comparison)** — Not rewritten, but §4.4–4.5 now provide a clean mechanism (admin vs. topical regimes) that explains the confounding. The reader can now understand why b jumped from 61 to 1353 without needing a pure controlled experiment. Partially resolved (see below).

---

### 🟡 Previous issues: partially addressed

**Issue B (ε_agg vs ε_pair thresholds)** — §6.5 step 3 uses `ε_agg < 0.1%` explicitly. But the definition in §3 still says "statistically close" without the threshold. The two-sub-definition suggestion (ε_agg and ε_pair separately) wasn't taken up. This remains an inconsistency: the definition says one thing, the certificate says another without cross-reference. Minor, but a future reader formalising this will have to reconcile them.

**Issue D (budget underspecified)** — §5.5 now discusses budget=15 implicitly through the path-growth table, and §4.4's path-count transition table makes the budget dependence visible. But the §1 table header still doesn't state budget, and the certificate in §6.1 doesn't say "at the production budget", only §6.5 step 3 does. The connection from budget to the geometric-series argument isn't made explicit.

**Issue E (sufficient conditions stated in graph terms, not b·D terms)** — §5.1 is unchanged. The new §5.6.1 operational restatement is exactly the right framing (b' is the operative criterion), but §5.1 still lists the three graph-structural hypotheses without promoting b·D to the first-class position. The two sections don't cross-reference each other, so a reader of §5.1 won't find the operationalisation.

---

### 🔴 New issues introduced by the update

**N1. §4.4 table: `b_eff` vs `b` notation is inconsistent**

The §4.4 calibration table shows "Raw distributional asymmetry" = 1353 (labelled `E[d²_c] / E[d²_p]`) and then `b_eff` = 589 after the first-moment correction. But §2 defines `b = E[d²_child] / E[d²_parent]` — no first-moment correction. So the `b` in the weight formula (§2) is 1353, while `b_eff` in §4.4 is 589. These are two different quantities. The document now uses both without clearly distinguishing them. Specifically:

- §2 uses `b` (raw second-moment ratio = 1353)
- §4.4 uses `b_eff` (second-moment ratio × first-moment correction = 589)
- §4.5 uses `b_eff` (topical = 9.59) in a third sense (which appears to include *all* corrections)
- §5.4 uses `b_eff(topical)` as the recommended calibration

These are three distinct quantities under the same symbol. The document needs a notation section or at minimum a clear statement in §4.4 of which `b` enters the weight formula and which enters the certificate.

**N2. §4.5 topical calibration: the `b_eff = 9.59` derivation is implicit**

The table in §4.5 shows `b_eff = 9.59` for `Category:Articles` but doesn't show *how* it was computed — the §4.4 correction chain (raw → first-moment → routing) gives 1353 → 589 → 226. Does the topical `b_eff = 9.59` go through the same chain, or is it a directly measured quantity? Given the note that the routing correction may be redundant under topical scoping (§5.4), the reader needs to know whether 9.59 is before or after the routing correction. The §4.5 note says admin hubs are excluded from the moment calculation, but doesn't say whether the 0.436 first-moment correction is also re-applied or what the raw second-moment ratio is for the topical subgraph.

**N3. §5.5 convergence table: the `path_growth / (b·D)` column uses inconsistent b**

The §5.5 table rows:
- "Topical, with routing": b·D = 27 → implies b = 27/7.34 ≈ 3.68 (= 9.59 × 0.384)
- "Topical, no routing": b·D = 70 → implies b = 70/7.34 ≈ 9.54 ≈ 9.59 ✓
- "Global, with routing": b·D = 226 → implies b = 226/7.34 ≈ 30.8 ≠ 589

The "Global, with routing" row appears to use `BranchRatio = 226` (the composed calibration from §4.4) divided by... nothing? The global b_eff is 589, and 589 × 7.34 ≈ 4323, not 226. The 226 is already `BranchRatio = b_eff × routing_correction × D` (or something like it), not `b·D`. The column header says `b·D` but the values are mixing `BranchRatio` (a composed scalar) with `b·D` (the two-parameter product). This needs clarification or the table will confuse anyone trying to verify the numbers.

**N4. §5.6.1 uses `b'` without defining it relative to `b` and `b_eff`**

The section introduces `b'` ("empirically calibrated") and says it "equals `b · D` in our notation when properly calibrated". But `b` is a dimensionless ratio and `D` is an average degree, so `b · D` has the units of an average degree, not a dimensionless per-hop shrinkage factor. The notation conflation here (using `b'` for what looks like the per-child-hop path-count growth rate) needs a one-line clarification. Looking at §5.6, `#paths(N,M) ≈ D^N · (b·D)^M` means the per-child-hop growth is `(b·D)`, which is a mixed-units quantity. If `b'` = `b·D` then `b'` ≈ 9933 globally or ≈ 70 topically-uncorrected, not ≈ 9.59. There's a dimensional inconsistency somewhere in this chain.

**N5. §6.4 operational recipe introduces a hard dependency without tracking it**

The recommendation in §6.4 ("build LMDB from a topical root") is presented as a direct production consequence, with a specific recipe including `--filter-root <page_id>`. But this recommendation is predicated on §5.4's **tentative** conclusion that the routing correction is redundant under topical scoping — a conclusion explicitly marked as needing verification (task #14). The §6.4 recipe should carry a similar caveat ("tentative, pending task #14 verification"), otherwise a reader treating §6 as implementation-ready will skip §5.4 and miss the open question. The recipe is also more specific than the evidence warrants: it names exact page_ids (137597, 7345184) for a recommendation that hasn't been verified on enwiki.

**N6. §5.5.1 is analytically strong but structurally orphaned**

Section 5.5.1 ("Where the slack actually lives") is some of the best analysis in the document — the deconstruction that `b` is constrained by calibration, `D` is the actual slack, and routing correction silently consumes that slack without naming itself as a slack consumer is sharp and useful. But it's buried in §5.5 under "Open questions", immediately followed by §5.6 which restates much of the same point from a different angle. The two sections reinforce each other but a reader scanning the document may not connect them. Consider a cross-reference: add a sentence at the end of §5.5.1: *"§5.6 restates this from the path-count-normaliser perspective, which makes the connection between calibration honesty and tree-likeness certification explicit."*

---

### Minor / prose

- §3 homogeneity precondition block: this is new and valuable, but it reads as a full paragraph inserted into a definition block, making §3 structurally awkward. The precondition should either be a named sub-definition (§3.0 or a bullet under the main definition) or moved to §3.2 which is the natural home for "what metric-tree-likeness actually requires".
- §4.4 table: "Friendship-paradox normalisation" as a column label is accurate but jargon-heavy. A parenthetical would help: *Friendship-paradox normalisation (accounts for high-degree nodes being over-sampled in neighbours)*.
- §5.5 second table: the "Boundary" row says `path_growth / (b·D) = 1.00 (no convergence)` but convergence fails strictly at ratio = 1 only if path_growth = b·D exactly; for ratio just below 1 the series diverges, for ratio just above 1 it converges. The table should say ratio = 1.00 is the *boundary* (not a threshold for "no convergence"), or the row should be removed to avoid implying the property is boolean at exactly this value.
- §6.2: as noted in the prior review, drift > 1% could also indicate a query-distribution shift — this is still not listed as a third explanation. Now that §4.5 establishes regimes, this is even more relevant: if a new query targets a seed in the admin regime, the per-pair drift will be high without any calibration error. Worth one bullet.

---

### Verdict

**Approve with suggestions.** The new sections add real substance and the topical-core result is strong. The notation inconsistencies (N1–N4) are the most pressing issues before this doc becomes a reliable reference: a reader trying to reproduce the §5.5 convergence table will hit the b/b_eff/BranchRatio ambiguity immediately. N5 is a scoping/caveat issue that's easy to fix with one sentence. N6 is a cross-reference suggestion. None of these block merge for a design note flagged as not-yet-a-theorem, but fixing N1–N3 would substantially increase the document's longevity as a reference.

s243a (commented):
## Third review — fresh read, new contributor perspective

This review approaches the document as someone reading it for the first time without the benefit of the PR discussion thread — the way it will be read in six months by a new contributor to the cost-model framework. I've focused on three questions: (1) can a reader reconstruct the argument from start to finish without ambiguity, (2) does the document cohere across its sections, and (3) where does it over-promise relative to its stated status as a design note?

---

### The document's narrative arc has a structural inversion

The document opens with an operational claim in the header, grounds it with the geometric-series argument in §2, proposes a definition in §3, then presents evidence in §4 — but §4.4 and §4.5 *substantially revise* the premise of §2. The geometric-series argument in §2 uses `b ≈ 1353` and `b·D ≈ 9933`, but §4.5 reveals that the honest calibration is `b_eff ≈ 9.59` and `b·D ≈ 70`. These are two different worlds: in §2 the mechanism produces 10⁻⁴ per child hop; in §4.5 the honest world is closer to 1/70 ≈ 1.4% per child hop.

The document knows this — §5.5 explicitly discusses the robustness band that makes both values compatible with the empirical observation. But §2 is never updated to reflect the §4.5 finding. A reader who reads §2 and stops will carry away `b·D ≈ 9933` as the operative mechanism. A reader who makes it to §4.5 will find that the mechanism actually operates at `b·D ≈ 70`. The design note reads like it was written in two sessions without a reconciliation pass.

**Suggestion**: Add a forward reference at the end of §2 — *"Note: §4.4–4.5 show that honest topical calibration gives b·D ≈ 70, not 9933; the geometric-series argument still holds, but the safety margin is ~70×, not ~9933×. See §5.5 for the convergence-robustness analysis."* This single sentence bridges the two views without rewriting §2.

---

### §3 defines the property but §5.6.1 redefines it better — and neither links to the other

§3 defines metric-tree-likeness operationally as "statistically close to the full-DAG value". §5.6.1 introduces **ε-metric-tree-likeness** as a formal criterion with an explicit summation inequality. These are the same property stated at different levels of precision. But a reader of §3 doesn't know §5.6.1 exists, and a reader of §5.6.1 may not notice it's meant to formalise §3.

Furthermore, §5.6.1 introduces the symbol `b'` ("empirically calibrated") and says it "equals `b · D` in our notation when properly calibrated" — but the §3 homogeneity precondition uses `b_eff`, §2 uses `b`, §4.4 uses both `b_eff` and `BranchRatio`, and §5.4 uses `b_eff(topical)`. There are now at minimum four symbols (`b`, `b_eff`, `BranchRatio`, `b'`) that all refer to variants of the same underlying quantity, and a reader has to reverse-engineer which applies where.

**Suggestion**: Add a small notation table — even just a three-row table in §2 or §3 — mapping each symbol to its definition and the section where it's computed. This is cheap to write and would dramatically reduce the cognitive load for any future reader.

---

### The homogeneity precondition in §3 is load-bearing but placed as an afterthought

The homogeneity precondition block in §3 is labelled "added on the basis of §4.5" — meaning it was inserted after the rest of §3 was written. Its placement (between the operational definition and the explanation of the practical formulation) interrupts the flow and signals to a careful reader that the definition wasn't settled before the evidence was gathered. That's fine epistemically — this is a design note — but it means the *definition* of the property formally requires a condition (homogeneity of the reachable subgraph) that isn't mentioned anywhere in §1 or §2.

The §1 observation (20 seeds from Physics) happens to land in the homogeneous topical core, so the observation is valid. But there's no statement in §1 or §2 that establishes this. A reader could correctly ask: "How do we know the 20 seeds at depth 4 from Physics were all in the topical core and not partly in the admin regime?" The answer is probably in §4.5 (3/20 seeds had semi-topical parents), but it isn't said in §1 where the observation is introduced.

**Suggestion**: Add one sentence to §1: *"All 20 seeds were drawn from depth 4 under `Category:Physics`, which §4.5 shows is within the topical core (b_eff ≈ 9.51); the homogeneity precondition of §3 is satisfied for this query distribution."*

---

### §6.4 is a production recommendation dressed as an algorithmic consequence

Sections §6.1–§6.3 are algorithmic consequences that follow from the property as defined. §6.4 is different: it's a production engineering recommendation ("build the LMDB differently") that follows from the §4.5 finding about inhomogeneity — not from metric-tree-likeness per se. A graph being metric-tree-like doesn't by itself imply you should filter the ingest; you should filter the ingest because the full graph is *inhomogeneous* and calibration on it is misleading.

This is a meaningful distinction because §6.4 could be correct even if metric-tree-likeness turned out to be false, and metric-tree-likeness could hold on the filtered graph without §6.4 being necessary (if calibration were somehow corrected at runtime instead). The two recommendations have different logical dependencies.

Moreover, §6.4 includes specific page_ids (137597, 7345184) presented as fact for enwiki, but §5.4 explicitly marks the topical-scoping recommendation as *tentative, pending task #14*. The enwiki page_id (7345184) appears to have been included without a corresponding verification note.

**Suggestion**: Rename §6 to "Algorithmic and operational consequences", move §6.4 to a new §7 "Data preparation recommendation", and add the task #14 pending caveat inline. This makes the logical dependency clearer: §6 consequences follow from the property; §7 recommendation follows from the inhomogeneity finding.

---

### §5.5's most important insight is buried four levels deep

The insight in §5.5.1 — that `D` is doing double duty as both a per-hop weight and as the convergence safety margin, making it principled rather than a free parameter — is arguably the conceptual centrepiece of the updated document. It explains *why* the certificate is robust (D is independently determined), *why* the routing correction is problematic (it consumes D's slack without naming itself), and *why* calibration on the topical core is the right default (b then matches path_growth, leaving D as the sole slack). This is a load-bearing insight for anyone designing future cost-model integrations.

But it's at §5.5.1, inside §5.5, inside §5 (Open questions), four levels from the top. A reader scanning the document for actionable architecture guidance will almost certainly miss it.

**Suggestion**: Pull the core sentence of §5.5.1 into a callout or bold paragraph near the top of §6 (before the algorithmic consequences): *"D is the convergence safety margin. Once b is calibrated honestly (b ≈ path_growth on the topical subgraph), the tree-search certificate reduces to checking b·D > path_growth — which is exactly 1/D from the root. D is the single parameter that determines how much slack the metric has, and it comes for free from the graph structure."*

---

### Minor issues

- **§4.4 table note**: The row `× E[d_p] / E[d_c]` = 0.436 is labelled "Friendship-paradox normalisation" but this ratio is the *inverse* of the classical friendship paradox correction (which would make high-degree nodes more prominent, not less). Either the label or the formula direction needs clarification.
- **§5.3 diamond graph**: As noted in prior reviews, the falsification argument requires asymmetric weights. This still isn't resolved in the text. Consider replacing the paragraph with: *"A diamond graph (A→B, A→C, B→D, C→D) with uniform edge weights defeats the metric only if b·D < path_growth — i.e., if the cross-path B→D carries comparable weight to the tree path A→B→D. For this to happen with our weight formula, D must be small (≤ 2), which a diamond literally is. This makes the diamond graph a natural minimal falsification target."*
- **§8 References**: The benchmark log is `/tmp/per_seed.log` — a transient file. Either preserve this as a committed artifact or replace the reference with the command to reproduce it. Design notes that reference transient files become self-inconsistent within days.
- **§4.3**: `b ≈ 61` still has no derivation or cross-reference. This was flagged in both prior reviews and remains unresolved.

---

### Verdict

**Approve with suggestions.** The document is substantially richer than its original form and the core ideas are sound. The most impactful single fix is the forward reference at the end of §2 bridging `b·D ≈ 9933` (the §2 calculation) to `b·D ≈ 70` (the honest topical calibration from §4.5) — without this, the document's argument is split across two incompatible framings and a casual reader will carry away the wrong number. The notation table is the second most impactful fix. Neither requires changing any of the substantive findings.

s243a (commented):
## Fourth review — post-update pass + answers to Claude's three deferred questions

This is a thorough re-read of the 853-line version. The update addresses nearly all outstanding issues from the three prior reviews. I begin with direct answers to the three questions Claude flagged, then track remaining issues.

---

### Answers to Claude's three deferred questions

**Q1: §3 ε threshold — is the cross-reference to §6.1 in the definition prose enough, or do you still want a threshold in the definition itself?**

The cross-reference is enough. §3's prose now correctly distinguishes aggregate vs. per-pair ε as quantities with different practical thresholds, and §6.1 is the right home for the concrete < 0.1% value. Pinning it into the definition would conflate the property's definition with the certificate's calibration — those belong in different places. The current structure is correct. **No change needed on this point.**

**Q2: §4.3 `b ≈ 61` — can the doc cite it as "from PR #2502 benchmark output" or does it need to be re-derived?**

Citing it as "from PR #2502 benchmark output" is sufficient *provided* that citation is added inline. Right now §4.3 uses `b ≈ 61` as a bare number with no provenance at all. A parenthetical like *(b ≈ 61; derived from PR #2502 benchmark output using the §2.0 formula applied to the broken-ingest LMDB)* is all that's needed — the reader doesn't need the LMDB, just a traceable claim. The number supports the narrative (b was lower under broken ingest → worse convergence); the narrative holds either way. **Add the provenance parenthetical; no re-derivation required.**

**Q3: Diamond graph — delete it and keep symmetric-DAG as the cleaner falsification, or specify the weight condition?**

**Specify the weight condition — and it's actually simpler than the earlier reviews suggested.** The diamond graph *does* defeat the metric under the natural weight assignment for a small graph: with D = 2 (each node has 2 children) and b ≈ 1 (symmetric), `b·D = 2`, which is below most realistic `path_growth` values. The convergence condition `b·D > path_growth` fails immediately for any graph where path-growth-per-hop exceeds 2. That means the diamond graph is a fine falsification example — it just needs one sentence stating the weights: *"For a diamond graph (A→B, A→C, B→D, C→D) with uniform weights, D = 2 and b ≈ 1, giving b·D = 2. With path-count growth ≈ 2 per additional child hop, the convergence margin is negligible and the metric drifts materially as mixed paths are admitted."* Don't delete it — it's the minimal falsification example, and pointing to D = 2 is a concrete, verifiable prediction. **Keep the diamond graph; add the two-sentence weight specification.**

---

### ✅ Issues resolved in this update

- **Notation table (§2.0)**: All four symbols (`b`, `b_eff`, `BranchRatio`, `b'`) are now defined in one place with example values. This was the single highest-impact fix and it's done cleanly. The note that `b·D ≈ 9933` is misleading relative to honest topical calibration is exactly right.
- **Forward reference at end of §2.1**: The callout box bridging global `b·D ≈ 9933` to topical `b_eff·D ≈ 70` closes the narrative split that was the biggest structural issue in the previous version. 
- **§5.5.1 insight promoted to §3.1**: Moving the "D does double duty" insight to §3.1 makes it visible to any reader who reads the property definition. Excellent placement.
- **§6/§7 separation**: §7 is now correctly framed as "data-prep consequences of the inhomogeneity finding" with explicit tentative status, separated from the algorithmic consequences of the property itself in §6. The status block at the top of §7 is exactly right.
- **§6.1 budget note**: The new paragraph explicitly states the certificate must be obtained at the production budget B and explains why. This closes Issue D from the second review cleanly.
- **§5.5.1 → §5.6 cross-reference**: The added sentence at the end of §5.5.1 pointing to §5.6 is present and correct.
- **§5.1 reframing**: The sufficient conditions are now correctly framed as "conditions under which `b·D > b'` is expected to hold" rather than as the property itself. The convergence inequality is now first-class.
- **Transient log reference**: `/tmp/per_seed.log` has been removed from §9 references.
- **§1 homogeneity note**: The new paragraph at the end of §1 explains that the Physics-rooted measurement already satisfied the homogeneity precondition without the authors noticing it.
- **§7.1 enwiki page_id verification note**: "Both verified against the live Wikipedia API on 2026-05-31" is added. Good.

---

### 🟡 Small remaining issues

**R1. §2.0 notation table: `b'` definition is circular**

The table says `b'` is the "per-child-hop empirical path-count growth" and its example value is "~11". But two rows up, `BranchRatio` is defined as `b_eff × routing_correction` with example 226 (global) / 3.68 (topical). The table note then says "the operative quantity for convergence is `b' < b_eff · D`" — which is fine. But a reader who wants to *compute* `b'` has only the description "measured by running the kernel at varying cc and reading path-count ratios". That's correct but it's the only symbol in the table with no algebraic definition. Consider adding a one-line algebraic description: `b' ≈ (#paths at cc=k) / (#paths at cc=k+1)` (geometric mean over transitions), or more precisely, the slope of `log #paths` vs `M`. This doesn't need to be a precise formula — just something that distinguishes `b'` from the structural quantities above it.

**R2. §3 definition — "statistically close" still has no ε in the definition block**

This is the Q1 partial-resolution item. The cross-reference to §6.1 is handled in the *prose* after the definition block, but the definition block itself (the blockquote) still ends with "statistically close to the full-DAG value" without any pointer. A reader who reads only the blockquote — which is the most quotable/reusable fragment of the doc — has no pointer to the certificate. Adding one sentence inside the blockquote: *"(The operational threshold is ε_agg < 0.1% for the aggregate version; see §6.1 for the certificate protocol.)"* would close this without disturbing the definition structure you've correctly kept.

**R3. §4.3 `b ≈ 61` — still no provenance (same as Q2 above)**

The update didn't add the provenance parenthetical yet. Just needs: *(b ≈ 61; from PR #2502 benchmark output, computed using the §2.0 formula on the broken-ingest LMDB)*.

**R4. §5.3 diamond graph — still no weight condition (same as Q3 above)**

The update didn't add the two-sentence weight specification. Per Q3 answer above: add the D=2, b≈1, `b·D = 2` note.

**R5. §4.4 friendship-paradox label direction**

The row `× E[d_p] / E[d_c]` = 0.436 is labelled "Friendship-paradox normalisation" but this factor is *less than 1* — meaning it *reduces* b_eff relative to the raw second-moment ratio. The classical friendship paradox makes high-degree nodes *more* prominent in neighbour sampling (the "your friends have more friends than you" effect), which typically inflates estimates, not deflates them. Either the label is backwards or the factor direction needs a clarifying note. Suggestion: rename to "Degree-bias correction (E[d_p]/E[d_c] < 1: parents have lower mean degree than children)" so the direction is self-explanatory.

---

### Verdict

**Approve — ready to merge.** The four updates in this pass have turned a structurally split document into a coherent, well-layered design note. The remaining five items (R1–R5) are all minor: two are the Q2/Q3 one-liners Claude explicitly deferred, one is a clarification inside a table label, one is a single sentence in the definition blockquote, and one is an algebraic note for `b'`. None of them affect the correctness or usability of the document. They are worth doing in a follow-up commit but do not need to block merge.

s243a (commented):
## Fifth review — mathematical self-consistency pass

Prior reviews have covered structure, narrative coherence, notation, and new-contributor readability. This pass focuses exclusively on whether the numbers and formulas in the document check out against each other. Working through the arithmetic independently.

---

### §2.0 notation table — numbers cross-check

**`b_eff` global (589):**
The table gives `E[d²_c] / E[d²_p]` = 1353, then `× E[d_p] / E[d_c]` = 0.436. So `b_eff = 1353 × 0.436 ≈ 589.9`. ✅ Checks out.

**`BranchRatio` global (226):**
The table gives `b_eff = 589 × routing_correction = 0.384`. So `589 × 0.384 ≈ 226.2`. ✅ Checks out.

**`b_eff` topical (9.59) → `BranchRatio` topical (3.68):**
`9.59 × 0.384 ≈ 3.68`. ✅ Checks out — the routing correction is the *same* 0.384 factor applied to both global and topical b_eff. This is notable: the routing correction was measured on the Physics-rooted subgraph (a topical traversal), so it makes sense it's the same value.

**`b' ≈ 11` vs. `b_eff(topical) = 9.59`:**
The table says the convergence condition is `b' < b_eff · D`. With `D = 7.34` and `b_eff(topical) = 9.59`: `b_eff · D = 70.4`. And `b' ≈ 11`. So `11 < 70.4` ✅ — convergence holds comfortably. No inconsistency.

---

### §5.5 convergence table — checking the `b·D` column

This is where Review #3's N3 issue flagged a potential inconsistency. Let me work through each row.

**Row: "Topical, no routing" → b·D = 70**
`b_eff(topical) × D = 9.59 × 7.34 ≈ 70.4`. ✅
`path_growth / (b·D) = 11 / 70.4 ≈ 0.156`. Table says 0.157. ✅

**Row: "Topical, with routing" → b·D = 27**
If we apply routing: `BranchRatio(topical) × D = 3.68 × 7.34 ≈ 27.0`. ✅
`path_growth / (b·D) = 11 / 27 ≈ 0.407`. Table says 0.407. ✅

**Row: "Global, with routing" → b·D = 226**
Here the table uses `BranchRatio(global) = 226` directly — *not* `b_eff(global) × D = 589 × 7.34 ≈ 4323`. This means the "b·D" column is actually mixing two different quantities: for the topical rows it's `b_eff × D`, but for the global row it's `BranchRatio` (which already has routing baked in) treated as if it were `b·D`.

`path_growth / 226 = 11 / 226 ≈ 0.049`. Table says 0.049. ✅ The arithmetic is correct *given what's been plugged in*, but what's been plugged in is `BranchRatio` not `b·D`. The column header "b·D" is therefore wrong for the global row — it should say `BranchRatio` (= `b_eff × routing × D` × ... actually, wait.

Let me re-examine: Is `BranchRatio = b_eff × routing_correction` or `BranchRatio = b_eff × routing_correction × D`?

From §4.4: `b_eff = 589`, `× routing_correction = 0.384`, `BranchRatio = 226`. So `226 = 589 × 0.384`, meaning `BranchRatio = b_eff × routing` — **without** multiplying by D. Then for the convergence table, `BranchRatio × D = 226 × 7.34 ≈ 1659`, not 226.

But the table uses 226 directly in the `b·D` column and the arithmetic gives the right path_growth ratio. This means either (a) `BranchRatio` is already `b_eff × routing × D` (and the §4.4 derivation is wrong), or (b) `D` is being absorbed somewhere else.

**This is a genuine internal inconsistency.** The §4.4 table implies `BranchRatio = b_eff × routing = 589 × 0.384 = 226` (no D). But the §5.5 table uses 226 as `b·D` in the convergence ratio. For the arithmetic to be right in §5.5, either D = 1 (trivially), or `BranchRatio` already includes D, or the convergence formula doesn't actually use `b·D` but uses `b` alone (with D cancelling elsewhere).

**Suggested resolution:** Check whether the weight formula `w(path) = (1/D)^N · (1/(b·D))^M` produces a convergence criterion of `b·D > b'` or `b > b'/D`. If the criterion is `b > b'/D` then `b_eff(topical) = 9.59 > 11/7.34 ≈ 1.5` and the global BranchRatio = 226 > 1.5 also — both work but with different margins. The table should clarify which formulation is being used.

---

### §2.1 geometric-series bound — the bound's tightness

The bound `contribution(M) ≤ #paths(M) · (1/(b·D))^M` is correct. But the document says the number of M-child-hop paths grows "~15-100× per extra child hop". The §4.4 empirical table shows ratios of 16.19× and 7.58× (geometric mean ~11). So the "15-100×" range in §2.1 was presumably an earlier estimate now superseded by the §4.4 measurement of ~11×. The bound in §2.1 should be updated to use the measured b' ≈ 11 rather than the pre-measurement "~15-100×" estimate. As written, a reader of §2.1 will use a different path-growth estimate than a reader of §4.4, which weakens the argument's precision when the two sections are read together.

---

### §3.1 convergence inequality formulation

§3.1 states: "The convergence inequality `b·D > b'` (path-count growth `b'` empirically matches `b_eff` once topical calibration is honest)".

This parenthetical claim — that `b' ≈ b_eff` when calibration is honest — deserves a cross-check. Empirically: `b' ≈ 11` (§4.4), `b_eff(topical) = 9.59` (§4.5). The ratio is `11 / 9.59 ≈ 1.15` — roughly 15% off. That's "approximately equal" but not tight. The text should say "within ~15%" or "within one measurement uncertainty" rather than "matches", which implies near-equality. The 15% gap is small enough that convergence holds well (the ratio `b'/b_eff·D ≈ 11/70 ≈ 0.16`), but the claim that `b' ≈ b_eff` when calibration is honest is only true to order-of-magnitude precision on this dataset.

---

### §4.1 and §4.2 — the 0.02% figure vs. §1's 0.02%

§1 says "aggregate metric moved 0.02%"; §4.1 says "avg d_wPow drift = 0.02%"; §4.2 says worst-case per-seed was 0.007%. These are consistent (aggregate ≈ worst-case individual on a 20-seed sample) and internally coherent. ✅

---

### §5.5 boundary row arithmetic

The "Boundary" row says `b·D = 11`, per-child weight = `1/11 = 9.1%`, and `path_growth / (b·D) = 1.00`. The claim is that at the boundary the series has ratio exactly 1.00 — neither convergent nor divergent.

But this is only exactly 1.00 if path_growth = b·D = 11. The document defines path_growth as the empirical ~11×. So the "boundary" row is: b·D is *set equal to* path_growth as a notional exercise — not a real calibration scenario. This is fine as a theoretical boundary, but it should be stated as "the theoretical convergence boundary: b·D := path_growth" to make clear this row isn't a real operational scenario. As currently written it reads as if b·D = 11 is a real calibration state we might be at.

---

### Summary of mathematical issues

| Issue | Severity | Fix |
|-------|----------|-----|
| §5.5 table: global row uses `BranchRatio` (226) as `b·D`, but derivation implies `BranchRatio` excludes D | **Moderate** — genuine arithmetic ambiguity, resolve by clarifying whether D is in `BranchRatio` | Clarify `BranchRatio` definition: does it include D or not? |
| §2.1: path growth estimate "~15-100×" superseded by §4.4's measured ~11× | **Minor** | Update §2.1 to "~11× (see §4.4)" |
| §3.1: `b' ≈ b_eff` claims equality; actual gap is ~15% | **Minor** | "within ~15%" or "approximately matches" |
| §5.5 boundary row: b·D = 11 reads as a real calibration scenario | **Minor** | Add "(theoretical boundary: b·D := path_growth)" |

All other numbers I checked are internally consistent. The §5.5 global-row inconsistency (whether BranchRatio includes D) is the one genuinely unresolved arithmetic question that a reader reproducing the table would hit.

---

### Verdict

**Approve — ready to merge.** The document's internal arithmetic is mostly self-consistent. The one genuine ambiguity (does `BranchRatio` include `D` or not?) is resolvable with a one-line clarification in §4.4 or §5.5. The other three items are precision improvements to language that's already directionally correct. None block a design note of this stated status.

PR #2664, opened 2026-06-01
docs: add tree-likeness index theoretical-foundations companion

s243a:
## Summary

  Adds `TREE_LIKENESS_INDEX_THEORY.md`, a formal companion to the design note. Separates the formal content
  (definitions, theorems, conjectures, citations) from the empirical / engineering content (which stays in the design
  note).

  **Stacked on `docs/tree-likeness-index-rename`** — please merge that PR first. After that, this branch can be rebased
  onto `main` and the diff will be just the new theory file.

  ## Structure

  | § | Content |
  |---|---|
  | 0 | Setup and notation — self-contained, LaTeX-formatted definitions for the weighted metric, path-count function,
  and TLI |
  | 1 | Established results with citations: Chung-Lu, Watts-Strogatz, Cohen-Havlin, Clauset-Shalizi-Newman, Feld, tree
  pair distance |
  | 2 | Three real proofs: Lemma 2.1 (path-count factorisation), Prop 2.2 (weights normalise), Theorem 2.3
  (geometric-series TLI bound) |
  | 3 | Four conjectures: main convergence conjecture (Theorem 2.3 is partial proof), topical homogeneity, b·D > b' is
  operative, symmetric-DAG falsification |
  | 4 | Empirical status — cross-references design note §4 |
  | 5 | Five open theoretical questions including a spectral-expansion conjecture |
  | 6 | References |

  ## Key framing decisions

  - **Directional asymmetry is the load-bearing property**, not acyclicity. Wikipedia is approximately DAG, but what
  matters for TLI is multi-parent (undirected-cycle) connectivity and the parent/child branching asymmetry it creates.
  - **Theorem 2.3 (the convergence bound)** establishes `TLI ≤ r/(1-r)` where `r = b'/(b·D)`, *conditional on* exact
  path-count factorisation (Lemma 2.1). On simplewiki this bounds TLI by ~18.6%; empirically TLI is ~0.02%, so the bound
  is loose by ~1000×. §5.1 lists this as an open question.
  - **Self-contained §0** but conservative §2 — we only prove what follows from definitions plus mild homogeneity.
  Bigger claims are conjectures with empirical support from the design note.

  ## Test plan

  - [ ] LaTeX renders correctly in GitHub markdown preview (MathJax)
  - [ ] Cross-references to design note resolve
  - [ ] No circular dependencies between theorem statements
  - [ ] Conjectures are clearly labelled as such

  ## Follow-up

  - Task #10 (synthetic non-tree-like graph) would empirically test Conjecture 3.4
  - Task #14 (routing correction redundancy) would settle the §5.4 question in the design note and update Conjecture
  3.1's calibration story
  - Task #15 (random-pair BFS) measures the geometric regime to test Conjecture 3.2 directly
  - The spectral-expansion question in §5.5 is genuinely open — would welcome reviewer perspectives on whether there's a
  known result connecting `b'/(b·D)` to the spectral gap

### DIFF
--- a/docs/design/TREE_LIKENESS_INDEX_THEORY.md
+++ b/docs/design/TREE_LIKENESS_INDEX_THEORY.md
@@ -134,21 +134,33 @@ $M$ ranges should yield consistent $b'$ estimates under
 homogeneity; the design note's §4.4 measurements at $cc \in
 \{100, 10, 5, 3\}$ are such an overlapping family.
 
-**Equivalent operational definition.** Equivalently,
+**Estimation in practice.** At fixed $B$, the maximum admissible
+$M$ is finite, so $b'$ cannot be extracted from a single
+$(B, cc)$ measurement. Instead, we run the kernel at a sequence
+of decreasing $cc$ values that admit successively more child
+hops. For two adjacent values $cc_1 > cc_2$ with
+$M_{\max}(B, cc_1) = m_1$ and $M_{\max}(B, cc_2) = m_2 > m_1$,
+under homogeneity
 $$
-b' = \lim_{cc \to 0^+}
-\left(\frac{\text{total } \#\text{paths at } (B, cc)}
-           {\text{total } \#\text{paths at } (B, cc = \infty)}\right)^{1/M_{\max}(B, cc)}
+\frac{\text{total } \#\text{paths at } (B, cc_2)}
+     {\text{total } \#\text{paths at } (B, cc_1)}
+\;\approx\;
+\frac{\sum_{M=0}^{m_2} (b')^M}{\sum_{M=0}^{m_1} (b')^M}
 $$
-where $M_{\max}(B, cc) = \lfloor B/cc \rfloor$. This makes the
-$cc \to 0^+$ regime explicit: $b'$ is the per-child-hop multiplier
-seen as we admit increasing numbers of child hops within fixed
-budget.
+so taking the geometric mean over transitions where
+$m_{k+1} - m_k = 1$ recovers $b'$ directly. For larger
+$M$ increments, the appropriate root corrects for the number
+of admitted child-hop levels.
 
-In practice (design note §4.4) we estimate $b'$ as the geometric
-mean of successive total-path-count ratios at $cc$ transitions
-$100 \to 10 \to 5 \to 3$, yielding $b' \approx 11$ on simplewiki
-topical core.
+In practice (design note §4.4), at $B = 15$ we use $cc \in
+\{100, 10, 5, 3\}$ with corresponding $M_{\max} \in \{0, 1, 3, 5\}$.
+The transition ratios are $1.05$, $122.7$, $18.7$; the
+per-child-hop growth recovers as $b' \approx \exp((\log 122.7)/2)
+\approx 11.1$ from the dominant transition (the
+$cc=100 \to 10$ ratio is essentially flat because that transition
+only admits $M = 1$ paths, which are few). The geometric mean
+across non-degenerate transitions gives $b' \approx 11$ on
+simplewiki topical core.
 
 ### 0.5 The tree-likeness index
 
@@ -163,6 +175,25 @@ $$
 i.e. the relative drift of the mean metric value as the child-step
 cost ranges from infinity (tree-search, $M = 0$ only) to
 $0^+$ (full DAG search, all $M$ admitted within budget).
+
+**Denominator choice.** We use
+$d_{\text{wPow}}(v; B, 0^+)$ — the full-DAG metric value — as the
+reference. The interpretation: tree-search is an *approximation*
+to the full-DAG value, and TLI is the relative approximation
+error of that approximation. Using $d_{\text{wPow}}(v; B, \infty)$
+as the denominator would give the inverse error; the absolute
+value in the numerator makes the index sign-agnostic
+regardless. (For the empirical drifts observed on simplewiki,
+the difference between the two normalisations is
+~0.02% × O(TLI) and immaterial.)
+
+**Absolute value.** The convergence direction is not guaranteed
+to be monotonic in $cc$ for all $(G, \mu, Q, B)$ — admitting
+more paths can either decrease or increase $d_{\text{wPow}}$
+depending on whether the new paths have above- or below-average
+$(h+1)^{-n}$ values. The absolute value handles both
+directions; empirically on simplewiki the drift is positive (the
+full-DAG value is slightly higher).
 
 In practice we estimate this with a finite contrast between two
 $cc$ values, typically $cc = 100$ (effectively tree-search at
@@ -197,10 +228,26 @@ homogeneous* if $\mathrm{supp}(Q)$ decomposes as
 $\bigsqcup_i V_i$ where each $(G|_{V_i}, Q|_{V_i}, B)$ is
 homogeneous. In this case the theorems below apply
 *piecewise* — each subgraph has its own $(D_i, b'_i,
-b_{\text{eff},i})$. This is the setting design note §4.5
-documents on Wikipedia: the global graph is piecewise
-homogeneous (topical regime + administrative regime) but not
-globally homogeneous.
+b_{\text{eff},i})$.
+
+**Important caveat.** The piecewise framing only applies when $Q$
+is restricted to a single component $V_i$ at calibration time
+(equivalently, the calibration sees only the subgraph
+$G|_{V_i}$). If $Q$ mixes nodes from multiple components — i.e.
+the query distribution genuinely spans different statistical
+regimes — Lemma 2.1's factorisation fails (different parts of
+$\mathrm{supp}(Q)$ produce different $(D, b')$ values), and
+neither Theorem 2.3 nor Corollary 2.3.1 applies. The
+recommended workflow is to *constrain* $Q$ to a single
+homogeneous component via the topical-root LMDB construction
+(design note §7.1), not to apply theorems piecewise to a query
+that crosses components.
+
+This is the setting design note §4.5 documents on Wikipedia: the
+global graph is piecewise homogeneous (topical regime +
+administrative regime) but not globally homogeneous, and the
+production recipe constrains $Q$ to the topical component
+specifically because cross-component queries break the theory.
 
 ## 1. Background: established results we build on
 
@@ -316,19 +363,40 @@ ultra-small-world consequences from this fit.
 distribution $P(k)$, the expected degree of a node reached by
 following a random edge is
 $$
-\frac{\mathbb{E}[k^2]}{\mathbb{E}[k]} \ge \mathbb{E}[k]
+\tilde{d} := \frac{\mathbb{E}[k^2]}{\mathbb{E}[k]} \ge \mathbb{E}[k]
 $$
-with equality iff degrees are constant.
+with equality iff degrees are constant. The quantity $\tilde{d}$
+is the *size-biased mean* or **Chung-Lu effective degree** (cf.
+§1.1).
 
 Source: S. L. Feld, "Why your friends have more friends than you
 do", *American Journal of Sociology* 96(6): 1464–1477 (1991).
 
-**Relevance to TLI.** The first-moment correction in our
-definition of $b_{\text{eff}}$ versus the raw second-moment ratio
-$b$ is precisely this size-biased correction applied separately
-to the child and parent edge populations. Design note §4.4
-labels this the "degree-bias correction"; this lemma is the
-underlying reason it has the form it does.
+**Relevance to TLI: $b_{\text{eff}}$ is the ratio of Chung-Lu
+effective degrees.** Applying the size-biased mean separately to
+the child and parent edge populations gives
+$$
+\tilde{d}_c = \frac{\mathbb{E}[d_c^2]}{\mathbb{E}[d_c]}, \qquad
+\tilde{d}_p = \frac{\mathbb{E}[d_p^2]}{\mathbb{E}[d_p]}
+$$
+and our calibrated branching asymmetry is exactly the ratio
+$$
+b_{\text{eff}}
+= \frac{\mathbb{E}[d_c^2]/\mathbb{E}[d_c]}{\mathbb{E}[d_p^2]/\mathbb{E}[d_p]}
+= \frac{\tilde{d}_c}{\tilde{d}_p}
+$$
+i.e. the Chung-Lu effective degree in the child direction divided
+by the Chung-Lu effective degree in the parent direction. **Low
+TLI thus has a clean interpretation: the child-direction
+Chung-Lu effective degree dominates the parent-direction one by
+a factor large enough that the convergence inequality
+$b_{\text{eff}} \cdot D > b'$ holds.**
+
+The raw second-moment ratio $b = \mathbb{E}[d_c^2]/\mathbb{E}[d_p^2]$
+omits the first-moment factors and is only approximately equal
+to $b_{\text{eff}}$ when $\mathbb{E}[d_c] \approx \mathbb{E}[d_p]$;
+on simplewiki the two differ by the factor 0.436 (design note
+§4.4), the same first-moment correction.
 
 ### 1.6 Tree pair distance
 
@@ -463,15 +531,26 @@ Multiplying:
 
 For the numerator,
 $$S_M \approx \sum_N r^M \cdot (N+M+1)^{-n} = r^M \cdot L_M$$
-where $L_M = \sum_{N \in [\text{admissible range}]} (N+M+1)^{-n}$.
-Since the length factor $(N+M+1)^{-n}$ decreases in both $N$
-and $M$, $L_M \le L_0$ for all $M \ge 0$, so
-$S_M / S_0 \le r^M$ and the geometric sum gives the bound.
+where $L_M := \sum_{N \in [d_{\min},\ B - M \cdot cc]} (N+M+1)^{-n}$.
+
+**Claim: $L_M \le L_0$ for all $M \ge 0$.** For each $N$ in the
+admissible range at level $M$ (which is a subset of the level-0
+range $[d_{\min}, B]$ truncated from the top by $M \cdot cc$),
+$(N+M+1)^{-n} \le (N+1)^{-n}$ pointwise since $x \mapsto x^{-n}$
+is decreasing. So $L_M$ is term-by-term dominated and has fewer
+or equal terms compared to $L_0$, giving $L_M \le L_0$. Hence
+$S_M / S_0 \le r^M$ and the geometric sum gives
+$\sum_{M \ge 1} S_M / S_0 \le r/(1-r)$.
 
 For the denominator,
-$$W_M \approx \sum_N r^M = r^M \cdot |\text{admissible } N \text{ range at level } M|$$
-The range size shrinks weakly in $M$ (the budget loses capacity
-$cc$ per child hop), so $W_M / W_0 \le r^M$ similarly. $\square$
+$$W_M \approx \sum_N r^M = r^M \cdot R_M$$
+where $R_M$ is the size of the admissible $N$-range at level $M$.
+
+**Claim: $R_M \le R_0$ for all $M \ge 0$.** The level-$M$ range
+$[d_{\min}, B - M \cdot cc]$ is a (possibly empty) subset of the
+level-0 range $[d_{\min}, B]$ obtained by truncating from the
+top, so $R_M \le R_0$. Hence $W_M / W_0 \le r^M$, hence
+$\sum_{M \ge 1} W_M / W_0 \le r/(1-r)$. $\square$
 
 **Caveats and tightness.**
 
@@ -578,26 +657,42 @@ test.
 This is the "decoupling geometry from metric" claim of design
 note §5.6.2 stated as a falsifiable conjecture.
 
-### 3.3 Sharpening of the bound: $\psi(r) \sim r^2 / (1 - r)$
+### 3.3 Sharpening of the bound (refinement of Conjecture 3.1)
+
+This conjecture *refines* Conjecture 3.1 by proposing a specific
+form for the function $\psi$ — it is not an independent claim
+but a guess at the leading-order asymptotic.
 
 **Conjecture 3.3.** The function $\psi$ in Conjecture 3.1
-satisfies $\psi(r) \le C \cdot r^2 / (1 - r)$ for some constant
-$C$ depending only on $n$ (the power-mean exponent) and
-$\mathrm{depth}_{\min}$ (the minimum distance from the support
-of $Q$ to the root).
+satisfies $\psi(r) \le C(n, B, d_{\min}) \cdot r^{\alpha} / (1 - r)$
+for some constant $C$ depending on $n$, the budget $B$, and the
+minimum distance $d_{\min}$ from the support of $Q$ to the
+root, with exponent $\alpha \ge 1$.
 
 **Motivation.** Theorem 2.3 bounds the *individual*
-contribution-sum drifts (numerator and denominator) by $r/(1-r)$.
-But $d_{\text{wPow}}$ is the ratio of these sums; if both drift
-by the same amount, the ratio drift is second-order in $r$. The
-empirical gap (theorem bound $r/(1-r) \approx 18.6\%$ vs
-empirical TLI ~$0.02\%$, a factor of ~1000) is consistent with
-this conjectured tighter scaling.
+contribution-sum drifts by $r/(1-r)$, but $d_{\text{wPow}}$ is
+the ratio of these sums and the drifts partially cancel. The
+question is *how much* they cancel — i.e. what $\alpha$ is.
 
-**Status.** Open. Even the constant $C$ is unknown; a careful
-calculation distinguishing the numerator and denominator
-contribution sums via their different length-factor dependences
-should resolve it.
+**Status.** Open in two directions:
+
+1. **First-order analysis at simplewiki parameters
+   ($B = 15, cc = 5, n = 2$) does NOT give clean $\alpha = 2$
+   scaling.** See §5.1 for the numerical check: $L_M / L_0$ and
+   $R_M / R_0$ differ by ~30–60% at level $M = 1$, so the $r^1$
+   coefficient does not cancel and the leading-order TLI
+   prediction is ~2.1%, not ~$r^2 \approx 0.025\%$.
+2. **The empirical TLI is still ~100× smaller than the first-
+   order prediction** (0.02% vs 2.1%). The source of this further
+   gap is unclear — candidates include budget-truncated $M$
+   ranges, higher-order Taylor terms with their own cancellation,
+   or measurement variance in the aggregate over only 20 seed
+   pairs.
+
+A definitive answer requires either a careful expansion to
+higher order in $r$, or empirical Monte Carlo at varying $r$ to
+fit $\alpha$ directly. The conjecture as stated is *consistent
+with* but not *derived from* the empirical observations.
 
 ### 3.4 Topical scoping is sufficient for homogeneity on Wikipedia
 
@@ -661,6 +756,17 @@ $b_{\text{eff}}$ to roughly $1/3$ of its honest value, which the
 robustness band absorbs but is not principled. See design note
 §5.4 for the discussion.
 
+**Directionality clarification.** A stronger statement than
+"unnecessary": $\rho < 1$ deflates $b_{\text{eff}}$, which
+*decreases* $\mathrm{BranchRatio}$, *increases* the convergence
+ratio $r = b'/(\mathrm{BranchRatio} \cdot D)$, and *loosens* the
+Theorem 2.3 bound. Dropping the routing correction therefore
+gives a *strictly tighter* theoretical bound on TLI, not just a
+simpler formula. The conjecture is more precisely "the routing
+correction was actively miscalibrating in the wrong direction,
+and dropping it gives a better-justified bound while leaving
+empirical TLI unchanged."
+
 **Status.** *Tentative.* The drift probe under the alternate
 composition has not been re-run on topical calibration. Task #14
 in the design note is to settle this empirically by measuring
@@ -670,9 +776,11 @@ TLI under both compositions and verifying both remain below the
 **Production consequence.** If confirmed, the data-prep recipe
 (design note §7.1) and calibration recipe (§7.2) become
 unconditionally simpler: drop the routing correction term, use
-$b_{\text{eff}}$ directly. If refuted, the routing correction
-stays but its theoretical role becomes "calibration-error
-band-aid" rather than "principled factor."
+$b_{\text{eff}}$ directly. The theoretical bound becomes tighter
+in the bargain. If refuted (i.e. dropping routing correction
+substantially increases TLI), the routing correction stays but
+its theoretical role becomes "calibration-error band-aid for an
+unmodelled effect" rather than "principled factor."
 
 ## 4. Empirical status
 
@@ -710,13 +818,21 @@ original design note pre-correction had this as $b_{\text{eff}}
 relationship between these calibrations is documented in design
 note §2.0's notation table.
 
-### 4.3 Conjecture 3.3 ($\psi \sim r^2/(1-r)$ tightening)
+### 4.3 Conjecture 3.3 ($\psi$ sharpening)
 
-Pending: a careful theoretical analysis distinguishing the
-numerator and denominator contribution sums via their different
-length-factor dependences. The empirical 1000× gap between the
-contribution-sum bound and the actual TLI is consistent with the
-conjectured $r^2$ scaling.
+**Mixed evidence.** First-order Taylor analysis at simplewiki
+parameters (§5.1) reduces the bound from the $r/(1-r) \approx
+18.6\%$ contribution-sum estimate to ~2.1% (after partial num/
+denom cancellation), but does *not* reach the empirical 0.02%.
+A clean $r^2/(1-r)$ scaling would predict ~0.025%, agreeing
+within 1.5× — but this is post-hoc fitting, not derivation.
+
+Until a careful expansion to higher order in $r$ resolves the
+exponent $\alpha$, Conjecture 3.3 is consistent with the data
+but not derived from theory. Pending: either an analytic
+expansion of $d_{\text{wPow}}(B, cc)$ to second order in $r$, or
+Monte Carlo evaluation at varying $r$ to fit $\alpha$
+empirically.
 
 ### 4.4 Conjecture 3.4 (topical homogeneity)
 
@@ -753,27 +869,56 @@ i.e. a bound on *individual* contribution-sum drifts. On
 simplewiki this gives $r/(1 - r) \approx 0.186$, while
 empirically TLI is ~0.02% — a factor of ~1000 gap.
 
-The gap arises because $d_{\text{wPow}} = (N/W)^{-1/n}$ is a
-*ratio*. To first order in $r$,
+Since $d_{\text{wPow}} = (N/W)^{-1/n}$ is a *ratio*, the first-
+order TLI estimate is
 $$
-\frac{\mathrm{TLI}}{1} \approx \frac{1}{n}
+\mathrm{TLI} \approx \frac{1}{n}
 \bigl|\Delta\log N - \Delta\log W\bigr|
 $$
-where $\Delta\log N \approx \sum r^k L_k / L_0$ and
-$\Delta\log W \approx \sum r^k R_k / R_0$ with $L_M$ the
-length-factor sum at level $M$ and $R_M$ the admissible-$N$
-range at level $M$. Both $\Delta\log N$ and $\Delta\log W$ are
-bounded by $r/(1-r)$ but their *difference* is much smaller —
-both are dominated by the same $r/(1-r)$ leading term, and the
-correction depends on the *ratio* $L_M/R_M$ vs $L_0/R_0$.
+where $\Delta\log N \approx \sum_{k \ge 1} r^k L_k / L_0$ and
+$\Delta\log W \approx \sum_{k \ge 1} r^k R_k / R_0$.
 
-**Open question (Conjecture 3.3).** Is $\psi(r) = O(r^2/(1-r))$
-the correct scaling, with the constant depending on $n$,
-$d_{\min}$, and the budget $B$? A careful Taylor expansion of
-the ratio $N/W$ in $r$ should resolve this. The empirical 1000×
-gap is consistent with $r^2$ scaling: $r^2/(1-r) \approx 0.029$
-for $r = 0.157$, only ~1.5× off from the empirical 0.02% (and
-closer once the constant $C$ is known).
+**Numerical check at simplewiki parameters does not support
+clean $r^2$ scaling.** For $B = 15$, $cc = 5$, $n = 2$,
+$d_{\min} = 4$:
+
+| Level $M$ | $L_M$ | $L_M / L_0$ | $R_M$ | $R_M / R_0$ | $L_M/L_0 - R_M/R_0$ |
+|---|---|---|---|---|---|
+| 0 | $\approx 0.149$ | 1.000 | 12 | 1.000 | 0 |
+| 1 | $\approx 0.140$ | 0.940 | 8 | 0.667 | 0.273 |
+| 2 | $\approx 0.122$ | 0.819 | 3 | 0.250 | 0.569 |
+
+The $r^1$ coefficient does *not* cancel — the difference at
+$M=1$ alone is $0.273 r$, giving a leading-order contribution of
+$\approx 0.5 \cdot 0.273 \cdot 0.157 \approx 0.021$, i.e. ~2.1%.
+This is much smaller than the contribution-sum bound (18.6%)
+but still ~100× larger than the empirical TLI of 0.02%.
+
+**Status: Conjecture 3.3's $\psi \sim r^2/(1-r)$ scaling is not
+yet supported.** A naive first-order Taylor expansion in $r$
+gives a ~2.1% prediction, not $r^2 \approx 0.025\%$. The
+remaining ~100× gap between first-order theory and empirical
+TLI likely has a different source — possibilities:
+
+- **Budget-constrained M range.** At $B = 15, cc = 5$ only
+  $M \in \{0, 1, 2, 3\}$ are admissible, and $M \in \{2, 3\}$
+  have very restricted $N$-ranges (3 values and 1 value
+  respectively). The effective sum is short and the higher-$M$
+  contributions are heavily truncated.
+- **Higher-order cancellation.** The Taylor expansion at $r=0$
+  may have larger higher-order terms that further reduce TLI;
+  computing the next term explicitly would resolve this.
+- **The simplewiki measurement itself.** TLI = 0.02% is the
+  *aggregate* over 20 seed pairs; per-pair drifts of 0.007%
+  worst-case (design note §4.2) are still 50× smaller than 2.1%.
+
+**Open question.** What is the actual scaling of $\psi(r)$ as a
+function of $r$ and the budget $B$? An honest answer requires
+either (i) a careful Taylor expansion accounting for the budget-
+truncated $L_M, R_M$ sums, or (ii) Monte Carlo evaluation at
+varying $r$ to fit the exponent empirically. Conjecture 3.3 as
+stated is *consistent with* the empirical gap but not derived
+from it.
 
 ### 5.2 Quantitative homogeneity ↔ calibration error
 
--- a/docs/design/TREE_LIKENESS_INDEX_THEORY.md
+++ b/docs/design/TREE_LIKENESS_INDEX_THEORY.md
@@ -1,0 +1,600 @@
+# Tree-likeness index: theoretical foundations
+
+**Companion to** [`TREE_LIKENESS_INDEX.md`](TREE_LIKENESS_INDEX.md), which
+documents the empirical observation, the calibration recipe, and the
+algorithmic consequences. This document attempts to formalise the
+theoretical content: what we can prove, what we conjecture, what is
+borrowed from existing literature.
+
+**Status**: Work in progress. Most central claims are conjectures
+rather than theorems. Established results from graph theory are
+cited; novel claims are clearly labelled. The design note's §4
+documents the empirical evidence supporting the conjectures here.
+
+**Scope**: We formalise the *tree-likeness index* `TLI(G, μ, Q, B)`
+as a statistical quantity computed from a directionally-weighted
+metric on a directed graph. The central conjecture (§3.1) gives an
+operational sufficient condition for low TLI in terms of two
+calibrated graph quantities `b_eff` and `D`. The non-trivial
+structural feature that makes the index interesting is *directional
+asymmetry* (parent-direction branching vs child-direction shortcut
+frequency), not acyclicity or any global structural label.
+
+## 0. Setup and notation
+
+### 0.1 The graph
+
+Let $G = (V, E)$ be a directed graph with a designated root
+$r \in V$. We do **not** require $G$ to be strictly acyclic: rare
+directed cycles are tolerated by visited-set pruning in traversal,
+and contribute negligibly to path counts when present below ~1%
+of vertices (which is the case for Wikipedia category graphs).
+
+What we *do* require is a parent/child decomposition: for each
+non-root $v$, define
+$$
+\mathrm{parents}(v) \subseteq V, \quad
+\mathrm{children}(v) \subseteq V
+$$
+as the sets of nodes connected to $v$ by an in-edge or out-edge
+respectively, in the categorisation interpretation. The
+distinction between "parent direction" and "child direction" is
+load-bearing — it is what gives the metric below its directional
+character.
+
+The relevant structural feature is **multi-parent connectivity**:
+typically $|\mathrm{parents}(v)| > 1$ for most non-root nodes,
+creating abundant undirected cycles (alternate ancestor paths
+through different parent chains). These undirected cycles are
+the *subject* of the index, not a perturbation to it.
+
+### 0.2 Paths
+
+A *path* from $v$ to $r$ in $G$ is a sequence
+$v = v_0 \to v_1 \to \dots \to v_k = r$ where each step is either a
+**parent hop** ($v_{i+1} \in \mathrm{parents}(v_i)$) or a
+**child hop** ($v_{i+1} \in \mathrm{children}(v_i)$). We
+decompose:
+
+- $N(p)$ = number of parent hops in path $p$
+- $M(p)$ = number of child hops in path $p$
+- $h(p) = N(p) + M(p)$ = total hop count
+
+We assume the traversal uses a visited-set so paths are simple
+(no node repeated).
+
+### 0.3 The directionally-weighted metric
+
+Following the design note's §2.0 notation, let:
+
+- $D := \mathbb{E}[d_c]$, the average child fan-out
+- $b := \mathbb{E}[d_c^2] / \mathbb{E}[d_p^2]$, the raw
+  second-moment ratio (the "early-formula" calibration)
+- $b_{\text{eff}} := \frac{\mathbb{E}[d_c^2]/\mathbb{E}[d_c]}{\mathbb{E}[d_p^2]/\mathbb{E}[d_p]}$, the
+  friendship-paradox-corrected branching asymmetry
+- $b'$ := the empirical per-child-hop path-count growth (defined
+  in §0.4 below)
+
+The path weight is
+$$
+w(p) = D^{-N(p)} \cdot (b \cdot D)^{-M(p)}
+$$
+and the **directionally-weighted power-mean metric** with exponent
+$n$ is
+$$
+d_{\text{wPow}}(v; B, cc) = \left(
+\frac{\sum_{p \in \mathcal{P}(v; B, cc)} w(p) \cdot (h(p)+1)^{-n}}
+     {\sum_{p \in \mathcal{P}(v; B, cc)} w(p)}
+\right)^{-1/n}
+$$
+where $\mathcal{P}(v; B, cc)$ is the set of paths from $v$ to $r$
+with cost at most $B$ when each parent hop costs $1$ and each
+child hop costs $cc$. We typically use $n = 2$ in experiments.
+
+### 0.4 Path-count growth
+
+For a node $v$ reachable to $r$ within budget $B$, define
+$$
+\#\text{paths}(v; N, M; B) :=
+\bigl|\{p : v \to r,\ p \text{ has } N \text{ parent hops}
+              \text{ and } M \text{ child hops, cost} \le B\}\bigr|
+$$
+The **empirical per-child-hop path-count growth** is
+$$
+b'(G, Q, B) := \lim_{M \to \infty}
+\frac{\mathbb{E}_{v \sim Q}\bigl[\#\text{paths}(v; \cdot, M+1; B)\bigr]}
+     {\mathbb{E}_{v \sim Q}\bigl[\#\text{paths}(v; \cdot, M; B)\bigr]}
+$$
+where the dot in $\#\text{paths}(v; \cdot, M; B)$ denotes summation
+over $N$, and the limit is taken over $M$ values for which both
+quantities are non-zero. In practice we estimate $b'$ from a
+finite range of $M$ by running the bidirectional kernel at
+varying $cc$ and reading the ratio of total path counts (design
+note §4.4).
+
+### 0.5 The tree-likeness index
+
+The **tree-likeness index** of the tuple $(G, \mu, Q, B)$ where
+$\mu$ is the directionally-weighted metric is
+$$
+\mathrm{TLI}(G, \mu, Q, B) :=
+\frac{\bigl|\mathbb{E}_{v \sim Q}\bigl[d_{\text{wPow}}(v; B, \infty)\bigr] -
+            \mathbb{E}_{v \sim Q}\bigl[d_{\text{wPow}}(v; B, 0^+)\bigr]\bigr|}
+     {\mathbb{E}_{v \sim Q}\bigl[d_{\text{wPow}}(v; B, 0^+)\bigr]}
+$$
+i.e. the relative drift of the mean metric value as the child-step
+cost ranges from infinity (tree-search, $M = 0$ only) to
+$0^+$ (full DAG search, all $M$ admitted within budget).
+
+In practice we estimate this with a finite contrast between two
+$cc$ values, typically $cc = 100$ (effectively tree-search at
+$B = 15$) and $cc = 5$ (admits up to 3 child hops); see design
+note §4.1.
+
+## 1. Background: established results we build on
+
+### 1.1 Random-graph distances (Erdős–Rényi / Chung–Lu)
+
+**Theorem (Chung & Lu, 2002).** For a random graph $G(n, p)$ with
+expected average degree $\langle k \rangle = pn$ satisfying
+$np \to \infty$, the average shortest-path length between random
+node pairs satisfies
+$$
+L \sim \frac{\log n}{\log \langle k \rangle}
+$$
+as $n \to \infty$.
+
+Source: F. Chung and L. Lu, "The average distances in random
+graphs with given expected degrees", *PNAS* 99(25): 15879–15882
+(2002).
+
+This is the baseline for "what branching alone achieves" in a
+homogeneous graph — relevant to §5.3 of the design note's
+geometric-regime comparison.
+
+### 1.2 The small-world property (Watts–Strogatz)
+
+**Definition (Watts & Strogatz, 1998).** A graph is *small-world*
+if it simultaneously satisfies:
+
+1. $L = \Theta(\log n)$ (logarithmic average shortest-path length)
+2. $C \gg C_{\text{random}}$ (clustering coefficient significantly
+   above that of a random graph of the same size and density)
+
+Source: D. J. Watts and S. H. Strogatz, "Collective dynamics of
+'small-world' networks", *Nature* 393: 440–442 (1998).
+
+### 1.3 Ultra-small-world for scale-free graphs (Cohen–Havlin)
+
+**Theorem (Cohen & Havlin, 2003).** For a scale-free network with
+degree distribution $P(k) \propto k^{-\gamma}$:
+
+- If $\gamma > 3$: $L \sim \log n / \log \langle k \rangle$
+- If $\gamma = 3$: $L \sim \log n / \log \log n$
+- If $2 < \gamma < 3$: $L \sim \log \log n$ (ultra-small-world)
+- If $\gamma \le 2$: $L = O(1)$ (degenerate; degree diverges
+  with $n$)
+
+The result requires *statistical homogeneity* of the graph: the
+degree distribution holds uniformly across $V$, with no
+sub-populations having qualitatively different scaling.
+
+Source: R. Cohen and S. Havlin, "Scale-free networks are
+ultrasmall", *Physical Review Letters* 90(5): 058701 (2003).
+
+**Relevance to TLI.** The Cohen–Havlin theorem governs the
+*geometric* regime of the graph (how distances scale with $n$).
+It does not directly bound TLI, but it constrains the family of
+graphs on which TLI can be low under any single global
+calibration (see §5.6.2 of the design note).
+
+### 1.4 Power-law fitting methodology (Clauset–Shalizi–Newman)
+
+**Method (Clauset, Shalizi & Newman, 2009).** Given empirical data
+$\{d_1, \dots, d_n\}$ believed to follow a power-law tail
+$P(k) \propto k^{-\gamma}$ for $k \ge d_{\min}$:
+
+1. For each candidate $d_{\min}$, compute the continuous MLE
+$$
+\hat\gamma(d_{\min}) = 1 + n_{\ge d_{\min}} \biggm/
+\sum_{d_i \ge d_{\min}} \log(d_i / d_{\min})
+$$
+2. Compute the Kolmogorov-Smirnov distance between the empirical
+   tail CDF and the fitted power-law tail CDF
+3. Select the $d_{\min}$ that minimises KS distance
+
+Source: A. Clauset, C. R. Shalizi, and M. E. J. Newman,
+"Power-law distributions in empirical data", *SIAM Review*
+51(4): 661–703 (2009).
+
+**Application here.** Design note §4.4 reports
+$\gamma = 2.41 \pm 0.05$ and $d_{\min} = 29$ with KS distance
+$0.048$ on the simplewiki child-degree distribution (global). The
+fit is statistically valid, but §5.6.2 explains why the
+homogeneity precondition is what blocks invoking Cohen-Havlin
+ultra-small-world consequences from this fit.
+
+### 1.5 Friendship paradox / size-biased degree (Feld)
+
+**Lemma (Feld, 1991).** For a graph with finite degree
+distribution $P(k)$, the expected degree of a node reached by
+following a random edge is
+$$
+\frac{\mathbb{E}[k^2]}{\mathbb{E}[k]} \ge \mathbb{E}[k]
+$$
+with equality iff degrees are constant.
+
+Source: S. L. Feld, "Why your friends have more friends than you
+do", *American Journal of Sociology* 96(6): 1464–1477 (1991).
+
+**Relevance to TLI.** The first-moment correction in our
+definition of $b_{\text{eff}}$ versus the raw second-moment ratio
+$b$ is precisely this size-biased correction applied separately
+to the child and parent edge populations. Design note §4.4
+labels this the "degree-bias correction"; this lemma is the
+underlying reason it has the form it does.
+
+### 1.6 Tree pair distance
+
+**Lemma (folklore).** In a balanced $D$-ary tree of depth $L$ with
+$n = (D^{L+1} - 1)/(D - 1)$ nodes:
+
+- Depth-to-root: $L = \log_D(n(D - 1) + 1) \approx \log_D n$
+- Average pair distance: for any two distinct nodes $u, v$,
+$$
+\mathbb{E}_{u,v}[d(u,v)] \le 2L \approx 2 \log_D n
+$$
+with equality approached as $D \to \infty$ (large branching makes
+LCAs shallow).
+
+**Proof sketch.** The distance between $u$ and $v$ is
+$\mathrm{depth}(u) + \mathrm{depth}(v) - 2 \cdot \mathrm{depth}(\mathrm{LCA}(u,v))$.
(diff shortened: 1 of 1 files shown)


s243a (commented):
## Review — TREE_LIKENESS_INDEX_THEORY.md

This is a well-structured theory companion. The separation of "what we can prove" (§2) from "what we conjecture" (§3) is exactly the right epistemic discipline, and the §0 notation is self-contained and precise. I've reviewed the document on five dimensions: (1) mathematical correctness of the proofs, (2) correctness of the cited external results, (3) logical consistency of the conjecture structure, (4) gaps between the proofs and the conjectures they are claimed to support, and (5) open questions quality.

---

## §0 Setup and notation

### Correctness

The definitions are clean. One precision issue:

**`b` vs `b_eff` in the weight formula.** The document defines the weight formula as

```
w(p) = D^{-N(p)} · (b · D)^{-M(p)}
```

where `b = E[d²_c] / E[d²_p]` (the raw second-moment ratio). But the design note §2.0 notation table explicitly states that the *early* `b` (raw second-moment ratio = 1353) was used in §2 *before* first-moment correction, and that "subsequent sections use `b_eff` exclusively." The weight formula in the design note is the one the kernel actually uses — which is calibrated with `b_eff`, not `b`.

As a result, Proposition 2.2 ("weights normalise path counts") requires `b · D = b'`, but the empirical calibration satisfies `b_eff · D ≈ b'`, not `b · D ≈ b'`. The calibration of the actual kernel uses `b_eff`; the weight formula in §0.3 uses `b`. These come apart. The definitions should use `b_eff` throughout (with a note that `b` is the un-corrected precursor), or the weight formula should explicitly say "where `b` here is the appropriately calibrated branching asymmetry (either `b` or `b_eff` depending on §2.0 of the design note)".

**`d_wPow` denominator.** The metric definition in §0.3 is:

$$d_{\text{wPow}} = \left( \frac{\sum w(p) \cdot (h(p)+1)^{-n}}{\sum w(p)} \right)^{-1/n}$$

This is the power-mean of `(h+1)` with weights `w(p)`, but the power-mean exponent is `−n` applied to the ratio, so the result is the weighted **harmonic** (if n=1) or a negative-exponent power-mean of `(h+1)`. More specifically: if we let `x = (h+1)` then this gives `(E_w[x^{-n}])^{-1/n}` which is not the standard power mean of `x` at exponent `n`; it is the power mean at exponent `−n`. This is correct for a metric that favours *shorter* paths (smaller h → larger `(h+1)^{-n}` → more weight in the numerator), but the document should state this explicitly: "this is the weighted power mean of `(h+1)` at exponent `−n`, equivalently the inverse `n`th root of the weighted mean of `(h+1)^{-n}`". A reader checking whether the normalisation makes sense will pause at the double inversion.

**TLI definition: `cc → 0⁺` vs `cc → ∞`.** The TLI formula uses `d_wPow(v; B, ∞)` (tree-search, no child hops) in the numerator and `d_wPow(v; B, 0⁺)` (full DAG, all child hops admitted) as the denominator normalisation. But in practice the estimate uses `cc = 100` as a proxy for ∞ and `cc = 5` as a proxy for `0⁺`. The document should state explicitly: (a) that `cc = 100` is a proxy and why it's sufficient at `B = 15` (one sentence from design note §4.1 would do), and (b) whether `d_wPow(v; B, 0⁺)` means the limit `cc → 0`, which would admit paths with arbitrarily many child hops within budget — and whether that limit even converges. At budget `B`, a path with `M` child hops each costing `cc` has child-hop cost `M · cc`, so as `cc → 0`, *all* paths within budget `B` are admitted regardless of M. The TLI is the relative drift from this admitting-all-paths baseline to the tree-search baseline; using `cc = 5` as the baseline is only an approximation, and the limit value needs a comment.

---

## §1 Cited external results

### Chung–Lu (§1.1)

The theorem as stated is for `G(n,p)` (Erdős–Rényi), but the citation is Chung & Lu (2002) which is a *generalisation* to graphs with given expected degree sequences, not just `G(n,p)`. The correct theorem statement for the Chung–Lu model is:

> For a random graph with degree sequence $(d_1, \ldots, d_n)$ with $\langle k \rangle = \sum d_i / n$ and $\max d_i^2 < \sum d_i$ (the condition that prevents a single node from connecting to all others), the average shortest-path length satisfies $L \sim \log n / \log \langle k \rangle$.

The `G(n,p)` version is a special case. Since the relevance here is to power-law degree sequences (non-uniform), the more general statement should be used. The current formulation is technically correct but weaker than the actual result cited.

### Cohen–Havlin (§1.3)

The result is correctly stated, including the critical precondition of statistical homogeneity. The relevance note is also correct. One addition: it is worth noting that the ultra-small-world result (`2 < γ < 3`) is asymptotic in `n` — for finite graphs the crossover to `L ~ log log n` behaviour may not be visible at Wikipedia's scale (simplewiki has ~80k topical nodes; enwiki ~3M). This affects §5.3's open question about whether `γ_topical > 3` would put the topical subgraph clearly in the small-world regime.

### Feld friendship paradox (§1.5)

The lemma is correctly stated. But the document's claim that "the first-moment correction in `b_eff` is precisely this size-biased correction" deserves a brief derivation. The raw ratio `b = E[d²_c] / E[d²_p]` is the ratio of *second moments*. The friendship-paradox correction gives `b_eff = (E[d²_c]/E[d_c]) / (E[d²_p]/E[d_p])`. The design note labels this a "degree-bias correction", but the explanation of *why* dividing by `E[d]` is the right correction is exactly Feld's lemma: the expected degree of a node reached via a random edge is `E[k²]/E[k]`. The document should add one sentence connecting Feld's statement to the `b_eff` formula: "The denominator `E[d_c]` in `b_eff` corrects for the fact that high-degree nodes are over-represented in neighbour samples by exactly this factor (Feld 1991)."

### Tree pair distance (§1.6)

The proof sketch contains a subtle error. The claim is that for uniform-random `u, v` in the leaves, the expected LCA depth is `O(1/(D-1))`, vanishing for large `D`. But this is only correct if `u` and `v` are drawn uniformly from all leaves — in which case the probability that the LCA is at depth `≥ ℓ` is indeed `D^{-ℓ}` asymptotically, but this gives expected LCA depth `∑_{ℓ≥0} D^{-ℓ} = D/(D-1)`, which is `O(1)` for any fixed `D`, not `O(1/(D-1))`. For `D → ∞`, the expected LCA depth is `D/(D-1) → 1`, so the pair distance approaches `2L - 2`, not `2L`. The claim that it's close to twice the depth `L` is still correct (the `−2` term is small relative to `L = log_D n`), but the proof sketch needs the correction: the expected LCA depth is `D/(D-1)`, not `O(1/(D-1))`.

---

## §2 The three proofs

### Lemma 2.1 (path-count factorisation)

This is stated as a lemma but is essentially an *assumption* — the homogeneity condition in the lemma statement *is* the factorisation; the "proof" is a one-line application of linearity of expectation. This is fine as long as it's clear the substantive content is the assumption, not the conclusion. The document does note this in the remark, but the lemma heading might better read "Lemma 2.1 (conditional factorisation)" to signal that the proof is trivial given the assumption, and the real question is whether the assumption holds.

### Proposition 2.2 (weights normalise path counts)

The proof is correct given Lemma 2.1 and the calibration `b · D = b'`. However, the proof silently assumes that the calibration sets `b · D = b'` exactly. In practice, `b_eff ≈ b'` with ~15% error. The proposition should note: "Exact calibration `b · D = b'` is assumed here; the practical calibration uses `b_eff` in place of `b`, and the approximation holds up to the ~15% agreement between `b_eff` and `b'` documented in the design note."

### Theorem 2.3 (geometric series bound)

The proof is broadly correct but has a gap at the key step. The proof says:

> The contribution at level `M` summed over `N` is proportional to `D^N · (b')^M · D^{-N} · (b·D)^{-M} = r^M` times the length factor.

The `D^N` cancellation is correct. But the "times the length factor" elides an important step: the length factor `(h+1)^{-n}` depends on *both* `N` and `M` (since `h = N + M`). To bound the sum over `M`, you need to bound `∑_N (N+M+1)^{-n}` for each `M`. For fixed `M`, this sum over `N` converges (since `n ≥ 1` and the terms decay as `N^{-n}`), but the bound you get depends on `M` through the starting point of the sum. Specifically:

$$\sum_{N \geq 0} (N + M + 1)^{-n} \leq \sum_{N \geq 0} (N + 1)^{-n} =: C_n$$

which is a constant independent of `M` (the Riemann zeta-like constant `ζ(n)` for `n ≥ 2`). So the geometric-series bound does go through, but the correct statement of the theorem should include the constant `C_n` in the bound:

$$\mathrm{TLI} \le \frac{r}{1-r} \cdot C_n$$

where `C_n = ∑_{k≥1} k^{-n}` (= ζ(n) for `n ≥ 2`). The current statement says TLI ≤ `r/(1-r)` with the implicit assumption that the `C_n` factor is absorbed or equals 1. For `n = 2`, `C_2 = π²/6 ≈ 1.645`, which is not 1 and would make the bound ~64% looser. This should be stated explicitly.

Additionally, the bound is on the *numerator contribution* of child paths, not directly on TLI. The document says the contribution is bounded by `r/(1-r) · d_wPow(M=0)` and calls this the TLI bound. But TLI is the *relative drift* — the numerator difference divided by the denominator. The denominator is `d_wPow(M=0)` (under tree-search), so the relative drift is indeed bounded by `r/(1-r)` (times `C_n`). This final step should be made explicit in the proof.

---

## §3 Conjectures

### Conjecture 3.1 (main convergence conjecture)

Well-stated. The use of an abstract `ψ` function rather than committing to `ψ(r) = r/(1-r)` is the right choice — it acknowledges that Theorem 2.3 gives one specific `ψ` (under strong assumptions) but the conjecture is about any bounding function.

One improvement: the conjecture as stated has condition (b) using `b_eff · D > b'`, but Theorem 2.3 uses `b · D > b'` (with the raw `b`). The transition from `b` to `b_eff` in the conjecture should be explicitly motivated: "We use `b_eff` rather than `b` in the conjecture because the honest calibration uses the first-moment-corrected quantity (§0.3)." This is implicit from the design note context but not stated here.

### Conjecture 3.3 (b·D > b' is operative)

This is the strongest conjecture — it claims necessity, not just sufficiency. The partial support section only provides sufficiency evidence (all three robustness-band calibrations exceed b' and all yield low TLI). The *necessity* direction requires a case where `b·D < b'` and TLI is high. The design note's broken-ingest comparison (b ≈ 61, drift ~1%) is the closest evidence but confounds b with graph structure changes. Conjecture 3.4's synthetic-graph experiment (task #10) is what would actually test necessity.

Suggest adding a sentence to the partial support: "The necessity direction (that failing `b_eff · D > b'` necessarily produces high TLI) has no clean empirical support yet; task #10 is the designed test."

### Conjecture 3.4 (symmetric DAG falsification)

The conjecture claims TLI is bounded *below* by a function growing with B for symmetric DAGs. This is plausible but needs one clarification: the claim uses "for *any* directionally-weighted metric of the form `w(p) = c₁^{N(p)} c₂^{M(p)}`". But for a symmetric DAG, `b ≈ 1`, so `c₁ ≈ 1/D` and `c₂ ≈ 1/D`. In this case the weight formula gives equal weight to parent and child hops, and path-count growth equals `D` per hop regardless of direction — meaning `b' ≈ D`, and `b_eff · D ≈ D` as well. So `r = b'/(b·D) ≈ 1`. The geometric series in Theorem 2.3 diverges at `r = 1`, consistent with high TLI. This is actually a direct consequence of Theorem 2.3 (plus Conjecture 3.1's reverse direction). The conjecture could note this cleaner derivation path.

---

## §5 Open questions

### §5.1 Tightness of the convergence bound

The 1000× gap between the Theorem 2.3 bound (18.6%) and empirical TLI (0.02%) is correctly identified and the two reasons given (longer child paths → additional length suppression; power-mean Lipschitz constant) are right. The conjecture `ψ(r) ~ r^{n+1}` would give `ψ(0.157) ≈ 0.157³ ≈ 0.004` for `n = 2` — a factor of 200 tighter, closer to empirical but still not matching. A potentially sharper hypothesis: the effective `r` is not 0.157 (based on path-count growth alone) but `r_eff = r · ((L_child)/(L_parent))^n` where `L_child / L_parent > 1` is the length ratio between child-mixed paths and parent-only paths. If child paths are on average `k` times longer than parent-only paths, the effective penalty per child hop is `r · k^n` rather than just `r`, giving `r_eff = r · k^n` with `k > 1`. For `n = 2` and `k ≈ 3`, `r_eff ≈ 0.157 × 9 ≈ 1.4` — overcorrects. But this line of reasoning is what connects the tightness question to the length-factor structure, and it's worth stating explicitly as a potential resolution path.

### §5.5 Spectral expansion connection

This is the most interesting open question in §5. One observation to add: `b' ≈ D` in a random regular graph (each node has the same degree `D`, so adding one child hop multiplies paths by `D`). For this graph, `b_eff = 1` (symmetric), `b_eff · D = D`, and `r = b'/( b_eff · D) = D/D = 1`. The spectral gap of a `D`-regular graph is `λ₁ - λ₂ = D - λ₂`, and for a Ramanujan expander this is `D - 2√(D-1)`. As `D` grows, the spectral gap grows — but `r = 1` regardless of `D`. So a large spectral gap does *not* guarantee low TLI; large `b_eff` (directional asymmetry) does. This suggests `r` and the spectral gap are measuring different things, and the relationship in §5.5's open question is likely not a simple equivalence. Worth adding this observation to sharpen the question: *"Note that `D`-regular graphs (large spectral gap, fast expansion) have `r = 1` (no convergence), suggesting the spectral gap and `r` are not equivalent. The question is whether there exists a spectral characterisation that does capture the directional asymmetry `b_eff > 1`"*.

---

## Minor issues

- **§0.5 TLI definition**: the absolute value in the numerator `|E[...∞] - E[...0⁺]|` is redundant if `d_wPow` is monotone in `cc` (more child hops admitted should weakly decrease or increase the metric monotonically). Adding a note on whether TLI is always non-negative would make the absolute value either justified or removable.
- **§2.3 Corollary 2.3.1**: labelled "informal version of the convergence conjecture" but it is actually a direct consequence of Theorem 2.3 (a formal corollary), not an informal version of Conjecture 3.1. Rename to "Corollary 2.3.1 (partial support for Conjecture 3.1)".
- **§4.3 empirical status for Conjecture 3.3**: the robustness-band calibrations have `b_eff · D` values ranging from 27 to 1659, but the statement says "27 to 1659" — the 1659 figure does not appear in the design note's §5.5 table (which tops out at 226). The 1659 would come from `b_eff(global) × D = 589 × 7.34 ≈ 4323`... or `BranchRatio × D = 226 × 7.34 ≈ 1659`. This is the same `BranchRatio` vs `b·D` ambiguity flagged in the fifth review of the design note. Whichever value is correct, the source should be stated.
- **§6 references**: all five external references are correctly cited. The Chung-Lu paper is PNAS 99(25): 15879–15882 (2002) — confirm page numbers (the doi is 10.1073/pnas.252631999).

---

## Summary table

| Issue | Section | Severity |
|-------|---------|----------|
| `b` vs `b_eff` in weight formula — should use `b_eff` consistently | §0.3 | **Moderate** |
| `d_wPow` definition: double inversion not explained | §0.3 | Minor |
| `cc → 0⁺` convergence not verified | §0.5 | Minor |
| Chung–Lu statement is G(n,p) special case, not the full result cited | §1.1 | Minor |
| Cohen–Havlin: finite-n caveat missing | §1.3 | Minor |
| Feld lemma: connection to `b_eff` formula not derived | §1.5 | Minor |
| Tree pair distance proof: expected LCA depth should be D/(D-1), not O(1/(D-1)) | §1.6 | **Moderate** |
| Lemma 2.1 heading should say "conditional" | §2.1 | Minor |
| Prop 2.2: ~15% calibration gap not acknowledged | §2.2 | Minor |
| Theorem 2.3: C_n = ζ(n) factor missing from bound; also the final TLI step implicit | §2.3 | **Moderate** |
| Conjecture 3.1: `b` vs `b_eff` transition not motivated | §3.1 | Minor |
| Conjecture 3.3: necessity direction has no empirical support; should say so | §3.3 | Minor |
| Conjecture 3.4: direct derivation from Theorem 2.3 not noted | §3.4 | Minor |
| §4.3: `b_eff · D` range "27 to 1659" — 1659 not traceable to design note table | §4.3 | Minor |
| §5.1: length-ratio resolution path not sketched | §5.1 | Minor |
| §5.5: regular-graph counterexample to spectral-gap equivalence not noted | §5.5 | Minor |

---

## Verdict

**Approve with suggestions.** The document achieves its goal: it clearly separates proved results from conjectures, provides correct citations for the external background, and the three proofs are broadly sound. The three moderate issues (§0.3 b/b_eff, §1.6 LCA depth, §2.3 missing C_n factor) should be addressed before merge since they affect the correctness of stated results. The minor issues are improvements to precision and completeness that could be done in a follow-up. The open questions in §5 are well-chosen and the spectral expansion question is genuinely interesting — the regular-graph observation in §5.5 would sharpen it considerably.

s243a (commented):
## Second review — TREE_LIKENESS_INDEX_THEORY.md (fresh pass)

This review approaches the document independently, with particular attention to: (1) whether the conjecture structure is logically well-ordered, (2) whether the scope boundaries between this doc and the design note are clean, (3) the adequacy of the §0 definitions as a self-contained foundation, and (4) whether the open questions are well-posed as research questions.

---

## Overall framing

The document makes a commendable decision to keep the "what we can prove" section (§2) small and honest. Three results is the right number: more would overreach, fewer would leave the theory companion without any formal content. The conjecture labelling is also disciplined — exactly four conjectures, each with partial support clearly delimited.

One structural observation before the details: **the document currently conflates two distinct audiences**. §1 (background results with citations) is aimed at a reader who doesn't know graph theory. §2–§3 assume a reader comfortable with asymptotic notation, measure-theoretic expectations, and geometric series. §5 presupposes familiarity with spectral graph theory. These are not the same reader. A short "how to read this document" note after the status block would help: *"§0–§1 are self-contained; §2–§3 require only undergraduate-level analysis; §5 presupposes familiarity with spectral graph theory."*

---

## §0 Definitions — precision issues

### The path budget and childCost interact in ways the metric definition obscures

The metric `d_wPow(v; B, cc)` is defined over `P(v; B, cc)` — paths with cost ≤ B where each parent hop costs 1 and each child hop costs `cc`. But the weight formula `w(p) = D^{-N(p)} · (b·D)^{-M(p)}` does not depend on `cc`. This means the *set* of paths that enter the sum changes with `cc`, but the *weights* assigned to any given path do not. This is the correct interpretation (and the right design — `cc` is a search parameter, not a metric parameter), but it's easy to misread the definition as implying `w(p)` varies with `cc`. The definition should add: *"Note: `cc` controls which paths are admitted to the sum; the weight `w(p)` of any admitted path does not depend on `cc`."*

### `b'` definition has a limiting issue

The formal definition of `b'(G, Q, B)` in §0.4 takes a limit as `M → ∞`. But at finite budget `B` with child hops costing `cc > 0`, there is a maximum `M` of `⌊B/cc⌋`. As `cc → 0`, this maximum grows, but for any fixed `cc` the limit in `M` cannot be taken. The definition implicitly requires `cc → 0` as part of the `M → ∞` limit, but this isn't stated. More precisely: the definition should be the limit as `cc → 0` of the ratio at the maximum achievable `M(cc) = ⌊B/cc⌋`, or alternatively it should be stated as a growth rate of the path-count function, defined separately from the `cc` parameterisation. As written, the limit definition is technically ill-posed for any fixed `cc > 0`.

### The homogeneity assumption in Lemma 2.1 should appear in §0 as a named condition

Lemma 2.1 introduces "statistical homogeneity" as a condition inline, and Conjecture 3.1 references it again. But §0 (Setup and notation) never defines it. For a document that aims to be self-contained in §0, statistical homogeneity should be a named definition in §0 — something like:

> **Definition 0.6 (Statistical homogeneity).** A graph `G` is *statistically homogeneous* with respect to root `r` and query distribution `Q` if there exist constants `D` and `b'` such that `#paths(v; N, M; B) ≈ D^N · (b')^M` uniformly over the support of `Q`, up to multiplicative `(1+o(1))`.

This would make Lemma 2.1 and Conjecture 3.1 immediately checkable against a named definition rather than requiring the reader to reconstruct it from the Lemma statement.

---

## §1 Background citations — scope questions

### §1.1 Chung–Lu: the connection to TLI is understated

The relevance note says the Chung–Lu result is "relevant to §5.3 of the design note's geometric-regime comparison." But the actual relevance to *this* document is more direct: Chung–Lu gives the baseline path-count growth in a random graph with `⟨k⟩` average degree as `⟨k⟩` paths per hop (since random graphs have roughly Poisson neighbourhood growth). This is precisely the setting where `b' ≈ D` and `b_eff ≈ 1`, meaning `r ≈ 1` and TLI is near its maximum. So the Chung–Lu model is the canonical *high-TLI* case, sitting at the boundary of Conjecture 3.4. This connection should be made explicit.

### §1.2 Watts–Strogatz: the clustering coefficient is not used

The Watts–Strogatz definition requires high clustering (`C ≫ C_random`). But clustering coefficient never appears anywhere in §2–§5. The relevant Watts–Strogatz property for TLI is only the `L = Θ(log n)` part. Including the clustering condition without using it introduces unused notation. Either use it (e.g., argue that high clustering implies many short cycles which implies `b' > b_eff · D`?) or narrow the citation to just the logarithmic-distance property.

### §1.6 Tree pair distance: this result is used backwards

The lemma establishes that in a tree, pair distance ≈ 2L. The relevance note says trees are the "structural limit where TLI = 0." That's correct. But the result as stated (about pair *distance*) is not what you need to prove TLI = 0 for trees — you need that in a tree there are no cross-edge paths at all (M = 0 for all paths), so `d_wPow(cc = ∞) = d_wPow(cc = 0+)` trivially. The tree pair distance lemma is about *how long* paths in a tree are, which is a different question from *how stable* the metric is under cross-edge admission. Consider either: (a) replacing this with a trivial TLI = 0 proof for trees (one sentence), or (b) keeping the pair distance result and noting its separate relevance to the §5.3 geometric-regime question.

---

## §2 Proofs — logical structure

### The proof chain has an unacknowledged circularity risk

The chain is: Lemma 2.1 (homogeneity → factorisation) → Proposition 2.2 (factorisation + calibration → weights normalise) → Theorem 2.3 (weights normalise + calibration → geometric bound). Each step depends on the previous. Fine. But Theorem 2.3 is then cited as "partial proof of Conjecture 3.1", which is the claim that `b_eff · D > b'` is sufficient for low TLI. The problem is that the chain *assumes* `b · D = b'` (exact calibration, Proposition 2.2) and then proves that under this assumption, convergence holds. But the Conjecture 3.1 condition is `b_eff · D > b'` (strict inequality, approximate calibration). So Theorem 2.3 proves the conjecture only in the degenerate limit where `b_eff · D = b'` exactly — precisely the case where the convergence ratio `r = 1` and the series *doesn't* converge. The Theorem actually proves the conjecture only for `b · D > b'` (strict), which is the non-degenerate case, and the proof goes through with `r = b'/(b·D) < 1`. This is fine but the relationship between "Proposition 2.2 assumes exact calibration" and "Theorem 2.3 uses strict inequality" needs one clarifying sentence: *"Proposition 2.2 is proved in the exact-calibration limit `b·D = b'`; Theorem 2.3 uses `b·D > b'` (strict), which means the weights slightly *over-normalise* path counts, giving a residual convergent series rather than exact cancellation."*

### Theorem 2.3 bound is on the numerator contribution, not directly on TLI

The theorem states: *"the contribution to the numerator of `d_wPow` from paths with M ≥ 1 child hops is bounded by `r/(1-r) · d_wPow(M=0)`"*. But TLI is defined as the *relative change in the mean metric value*, not the relative contribution of child paths to the numerator. The step from "child paths contribute at most `r/(1-r)` of the parent-path numerator contribution" to "TLI ≤ `r/(1-r)`" requires the metric to be approximately linear in its numerator, which is not immediate for a power-mean. Specifically:

- Let `S_0 = ∑_{M=0} w · (h+1)^{-n}` (parent-only numerator sum)
- Let `S_1 = ∑_{M≥1} w · (h+1)^{-n}` (child-path numerator sum)
- Let `W_0`, `W_1` be the corresponding weight sums
- Then `d_wPow(cc=∞) = (S_0/W_0)^{-1/n}` and `d_wPow(cc=0+) = ((S_0+S_1)/(W_0+W_1))^{-1/n}`

The bound `S_1 ≤ r/(1-r) · S_0` doesn't directly give a bound on `|d_wPow(cc=∞) - d_wPow(cc=0+)| / d_wPow(cc=0+)` without also bounding `W_1/W_0`. Under Proposition 2.2, `W_1/W_0 ≈ r/(1-r)` as well (same counting argument), so both numerator and denominator shift by the same relative amount, and the first-order change in the ratio is second-order in `r`. This would actually give a *tighter* bound `TLI = O(r²/(1-r))` rather than `O(r/(1-r))`. If this sharper bound holds, it would partially explain the 1000× gap between Theorem 2.3 and the empirical TLI. This connection is worth a remark after the theorem.

---

## §3 Conjectures — ordering and completeness

### The conjecture ordering should follow logical dependency

Currently: 3.1 (main convergence) → 3.2 (topical homogeneity) → 3.3 (b·D > b' is operative) → 3.4 (symmetric DAG falsification). But logically:
- 3.2 is a *precondition* for 3.1 (you need homogeneity to invoke the main conjecture)
- 3.3 is a *strengthening* of 3.1 (adding necessity to 3.1's sufficiency)
- 3.4 is the *falsification* case (what happens when 3.3's condition fails)

A cleaner ordering would be: 3.1 (homogeneity, enabling condition) → 3.2 (main convergence, given homogeneity) → 3.3 (necessity, strengthening 3.2) → 3.4 (falsification, testing 3.3). This would make the logical dependency explicit and the progression natural.

### A fifth conjecture is missing: the routing correction is redundant under topical scoping

Design note §5.4 makes a tentative recommendation (pending task #14) that the routing correction is redundant once topical calibration is used. This is stated as an open question in the design note but it's also a conjecture with partial support — exactly the kind of claim that belongs in this theory document as a named conjecture. Without it, the theory document doesn't formally capture what the design note calls its "most consequential open question for production use."

---

## §5 Open questions — assessment

### §5.2 (homogeneity ↔ calibration error) is the most practically important question

This question — how much deviation from exact homogeneity degrades the TLI bound — is directly relevant to any deployment on a new graph. But it's listed second, after the more esoteric tightness question. Consider promoting it to §5.1 or flagging it as "most practically important."

### §5.4 (extension to undirected graphs) poses the question incorrectly

The question asks whether there's a generalisation of TLI to undirected graphs using "some other asymmetry (e.g. spectral, expansion-based) in place of directional." But TLI as defined measures drift as `cc` varies — a parameter that has no meaning in an undirected graph without a parent/child decomposition. The question isn't "can TLI be generalised" but rather "is there a *different* index for undirected graphs that captures a similar concept?" These are different research questions. The section would benefit from separating: (a) can TLI be extended to directed-but-symmetric graphs (where the parent/child decomposition exists but `b ≈ 1`)? and (b) is there an analogous index for undirected graphs using non-directional structure?

---

## Verdict

**Approve with suggestions.** The document is doing something genuinely valuable: externalising the formal structure that was implicit in the design note. The three proofs are sound in substance (with the caveats noted in the first review regarding the C_n factor in Theorem 2.3). The main structural improvements for this pass are:

1. Add Definition 0.6 (statistical homogeneity) to §0 so it's a named object rather than reconstructed from Lemma 2.1
2. Fix the `b'` definition's limiting issue (it's ill-posed at fixed `cc > 0`)
3. Reorder conjectures 3.1–3.4 to follow logical dependency
4. Consider adding Conjecture 3.5 (routing correction redundancy, pending task #14) to capture the design note's most consequential open production question
5. Clarify that the Theorem 2.3 bound may actually give TLI = O(r²/(1-r)) when both numerator and weight sums are bounded simultaneously — this would partially explain the 1000× empirical gap

s243a (commented):
## Third review — post-update pass

This is a thorough re-read of the 881-line version. The update is comprehensive and addresses the majority of issues from both prior reviews. I track what's resolved, what's partially addressed, and what is new.

---

### ✅ Fully resolved

- **`b` vs `b_eff` in weight formula (§0.3)**: `b_eff` is now the canonical quantity in the weight formula. `b` is demoted to "early-formula / historical context" with an explicit note. The substitution note ("all theorems stated in terms of `b_eff`") is exactly right.
- **Definition 0.6 (Statistical homogeneity)**: Now a named definition in §0 with three explicit conditions (H1, H2, H3), a piecewise-homogeneous extension, and a concrete example (Wikipedia topical/admin decomposition). The piecewise extension is a nice addition not requested by either review — it's the right level of generality.
- **`b'` definition limiting issue (§0.4)**: Fixed cleanly. The formal definition is now as a growth-rate constant under homogeneity, with an equivalent operational definition via `cc → 0⁺`. Both forms are given. The note that different `(B, cc)` pairs should yield consistent `b'` estimates under homogeneity is exactly the right calibration consistency check.
- **Chung–Lu theorem (§1.1)**: Now stated for the full given-expected-degree model (not just `G(n,p)`), with `ℓ̃ = E[w²]/E[w]` (size-biased degree) as the operative quantity. The connection to `b_eff`'s size-biased correction is also now explicit.
- **Watts–Strogatz clustering coefficient (§1.2)**: The relevance note now explicitly says "only the distance criterion is operative" and the clustering coefficient is included for terminology completeness only. This is the right framing.
- **Tree pair distance proof (§1.6)**: The LCA depth calculation is now correct: `E[LCA depth] = D/(D-1)`, and the pair distance is stated as `2L - 2D/(D-1)`. The note that TLI = 0 for trees follows trivially (no cross-edge paths) is also correctly added and separated from the pair-distance lemma's purpose.
- **Conjecture ordering**: The ordering now follows logical dependency with an explicit preamble explaining the structure. Conjectures 3.1 through 3.6 are ordered as: main convergence → necessity → tightening → Wikipedia instance → falsification target → routing-correction production question.
- **Conjecture 3.6 (routing-correction redundancy)**: Added. The statement, motivation, status (tentative), and production consequence are all present. The "status: tentative, pending task #14" caveat is explicit.
- **Theorem 2.3 — both numerator and denominator bounded**: The theorem now explicitly bounds both `S_M/S_0` and `W_M/W_0` by `r/(1-r)`, and the caveat section correctly notes that TLI is a ratio, so the drifts partially cancel and the actual TLI bound is smaller.
- **Corollary 2.3.1 relabelling**: Now correctly labelled as "partial proof of Conjecture 3.1" rather than "informal version".
- **`b'` ≈ `b_eff` claim precision**: The 15% gap is acknowledged in the Proposition 2.2 caveat.
- **§4.2 note on the "1659" figure**: The derivation `b_eff · routing · D = 226 · D ≈ 1659` is now spelled out explicitly, resolving the ambiguity flagged in the fifth review of the design note.
- **§5.5 regular-graph counterexample**: Added to §5.5 as the "sharpening observation", correctly showing that a `D`-regular graph has `r = 1` despite large spectral gap. The conclusion (rules out equivalence with spectral gap; leaves open directional spectral quantity) is exactly right.
- **Conjecture 3.3 (`ψ ∼ r²/(1-r)`)**: Added as a named conjecture, motivated by the ratio-drift argument. §5.1's tightness analysis supports it with a first-order estimate.

---

### 🟡 Partially addressed

**P1. §0.5 TLI definition — absolute value and `cc → 0⁺` convergence**

The definition still uses `|E[d_wPow(B, ∞)] - E[d_wPow(B, 0⁺)]|` in the numerator and `E[d_wPow(B, 0⁺)]` in the denominator. The `cc → 0⁺` form is now defined via §0.4's operational definition, which resolves the earlier limiting-issue concern. However: (a) the absolute value is still present but its justification (whether `d_wPow` is monotone in `cc`) is not stated, and (b) the denominator normalises by `d_wPow(B, 0⁺)` — the all-paths-admitted baseline — rather than `d_wPow(B, ∞)` — the tree-search baseline. This choice matters: if TLI is intended to measure "how much does metric change when we switch from tree to full-DAG", either normalisation works, but they give different numerical values. The design note uses the 0⁺ denominator; the theory doc should state why (or make it a convention).

**P2. §2.3 Theorem 2.3 proof — length factor bound**

The proof says `S_M ≈ r^M · L_M` where `L_M = Σ_N (N+M+1)^{-n}`, then bounds `S_M/S_0 ≤ r^M` by claiming `L_M ≤ L_0`. This is correct: `(N+M+1)^{-n} ≤ (N+1)^{-n}` for all `M ≥ 0`, so `L_M ≤ L_0`. But the proof as written says "since the length factor decreases in both N and M, L_M ≤ L_0 for all M ≥ 0" without the key step: `Σ_N (N+M+1)^{-n} ≤ Σ_N (N+1)^{-n}` by term-by-term comparison. This is a one-line addition but the proof currently reads as asserting it without showing it. Similarly for `W_M ≤ W_0`: the admissible-N-range shrinks with M (budget constraint), which is correct, but the direction isn't immediately obvious (a smaller range means fewer terms in the sum, which could decrease or increase the sum depending on which terms are dropped). Specifically, at level M, child hops consume `M · cc` budget, leaving `B - M · cc` for parent hops, so the max N is `⌊(B - M·cc)/1⌋`. As M increases, the N range shrinks, and since all N terms are positive, `W_M ≤ W_0`. This direction is correct and the proof sketch is right, but needs this explicit step.

**P3. Feld lemma relevance note (§1.5)**

The relevance note says the `b_eff` correction "is precisely this size-biased correction applied separately to the child and parent edge populations" but still doesn't show the algebra. The first review asked for a one-sentence derivation connecting Feld's `E[k²]/E[k]` to the `b_eff = (E[d²_c]/E[d_c]) / (E[d²_p]/E[d_p])` formula. This still isn't present. This is a minor gap — just one sentence: "Feld's lemma gives `E[k²]/E[k]` as the expected degree of a random edge-neighbour; applying this separately to child edges gives `E[d²_c]/E[d_c]` (size-biased child degree) and to parent edges gives `E[d²_p]/E[d_p]` (size-biased parent degree); `b_eff` is their ratio."

---

### 🔴 One new issue

**N1. §5.1 — first-order TLI estimate is not derived correctly**

The new §5.1 introduces a first-order estimate:

```
TLI ≈ (1/n) · |Δlog N - Δlog W|
```

where `N` and `W` are the full numerator and denominator sums. The derivation path is: `d_wPow = (N/W)^{-1/n}`, so `log d_wPow = -(1/n) log(N/W)`. The drift in `log d_wPow` as `cc` varies is `-(1/n)(Δlog N - Δlog W)`. For TLI defined as the *relative* drift in `d_wPow` (not `log d_wPow`), the first-order expansion gives:

```
TLI ≈ |Δ(d_wPow)| / d_wPow ≈ |Δ log d_wPow| = (1/n)|Δlog N - Δlog W|
```

This is correct as a first-order approximation. But the document then goes on to say: "both Δlog N and Δlog W are bounded by r/(1-r) but their *difference* is much smaller — both are dominated by the same r/(1-r) leading term, and the correction depends on the *ratio* L_M/R_M vs L_0/R_0."

The issue: `Δlog N ≈ Σ r^k L_k/L_0` and `Δlog W ≈ Σ r^k R_k/R_0` where `L_M` is the length-factor sum and `R_M` is the admissible-N range. For these to nearly cancel (and give a factor of r² in the TLI), you would need `L_k/L_0 ≈ R_k/R_0` for all `k`. But `L_k = Σ_N (N+k+1)^{-n}` decreases as `k` grows, while `R_k = B - k·cc` also decreases — so they both shrink, but *at different rates*. The claim that they nearly cancel to give O(r²) is not shown; it's asserted. The §5.1 analysis is suggestive but stops short of the calculation that would actually verify the r² scaling. This is labelled as an open question, so asserting it as a conjecture is fine — but the numerical check `r²/(1-r) ≈ 0.029 for r = 0.157, which is ~1.5× off from empirical 0.02%` is presented as supporting evidence. The factor of 1.5× is not particularly tight for a first-order estimate; the claim "closer once the constant C is known" is true but the constant C is precisely what the uncompleted analysis would determine. This is worth flagging: the numerical coincidence is consistent with the conjecture but does not strongly confirm r² scaling over, say, r^1.5 or r^2.5.

**Suggestion**: Add a sentence: "The 1.5× agreement is suggestive but not conclusive — confirming the r² scaling requires either the full Taylor expansion of N/W in r, or a Monte Carlo test on synthetic graphs at varying r values."

---

### Minor items

- **§3.2 partial support section**: The robustness-band robustness evidence (27 to 1659) now has the "1659" figure explained in §4.2, but the §3.2 prose still references "1659" without a note. Since §4.2 explains it, a cross-reference would help: "(see §4.2 for derivation of the 1659 figure)".
- **§3.5 (symmetric DAG) vs Conjecture 3.2 relationship**: Conjecture 3.5 is the falsification target for Conjecture 3.2's *necessity* direction. The preamble to §3 says "3.5 is the falsification target" which is correct, but the connection should also be noted in 3.2's partial-support section: "Conjecture 3.5 is the designed test of the necessity direction — if a symmetric DAG has `b_eff · D ≈ b'` and high TLI, that confirms necessity."
- **§5.4 (undirected graphs)**: Both reviews asked for a split between (a) directed-but-symmetric graphs and (b) truly undirected graphs. The update didn't add this split. Both cases reduce to `b_eff ≈ 1` via different paths, but they're structurally different (directed-symmetric has a decomposition; undirected doesn't). Still minor, still worth one sentence.
- **§6 references**: Chung-Lu PNAS 99(25): 15879–15882 — this is correct. The DOI is 10.1073/pnas.252631999 if you want to add it for completeness.

---

### Verdict

**Approve — ready to merge.** The update resolves virtually all substantive issues from the first two reviews. The document is now in good shape as a theory companion: the definitions are precise, the proofs are honest about their assumptions, the conjecture structure is logically ordered, and the open questions are well-posed. The three partially-addressed items (P1–P3) and one new minor issue (N1's 1.5× caveat) are all low-priority improvements that can be done in a follow-up. The only item that genuinely strengthens the document before merge would be the one-sentence Feld algebra note (P3) — everything else is fine as-is.

s243a (commented):
## Independent second review of updated document (881-line version)

This review reads the updated document fresh, without treating prior review threads as a checklist. Focus areas: (1) internal logical consistency end-to-end, (2) whether the six-conjecture structure is tight, (3) whether any new content introduced in the update creates new problems, and (4) readability as a standalone companion.

---

## Opening framing — one structural concern

The "How to read this document" gap is still present. The document now has a well-written **Status** and **Scope** block at the top, but the scope block is aimed at someone who already knows what `b_eff`, `D`, and `TLI` mean. A first-time reader arriving from the companion link in `TREE_LIKENESS_INDEX.md` needs a single sentence orienting them to the reading order: §0–§1 are self-contained background; §2 requires comfort with geometric series; §3–§4 can be read for results without following the proofs; §5 is for specialists. Without this, the document reads as if §5.5's spectral-gap question has the same intended audience as §0.1's graph definition. One sentence would fix it.

---

## §0 Definitions — new content review

### Definition 0.6 piecewise-homogeneity extension

The piecewise-homogeneous extension is a genuine improvement. However, the definition says:

> *"In this case the theorems below apply piecewise — each subgraph has its own (D_i, b'_i, b_eff,i)."*

This is asserted but not verified. Theorem 2.3 applies to a single homogeneous tuple `(G, Q, B)`. For a piecewise-homogeneous graph, the metric `d_wPow` aggregates over all nodes in `supp(Q)`, which may span multiple subgraphs `V_i`. The theorem does *not* automatically apply piecewise unless the query distribution `Q` is also decomposed as `Q|_{V_i}` — which requires `Q` to be supported entirely within one `V_i`, or the TLI to be computed per-component and then averaged. If `Q` mixes nodes from multiple components, the path-count factorisation `D^N · (b')^M` fails (different nodes have different growth constants), and neither Lemma 2.1 nor Theorem 2.3 applies.

The document should add a clarification: *"The theorems apply piecewise only when `Q` is restricted to a single component `V_i`, or when TLI is computed per-component and the reported value is a Q-weighted average of per-component TLIs."*

### §0.4 equivalent operational definition of b'

The new operational definition:

$$b' = \lim_{cc \to 0^+} \left(\frac{\text{total #paths at }(B, cc)}{\text{total #paths at }(B, cc=\infty)}\right)^{1/M_{\max}(B,cc)}$$

has a subtle issue: the exponent `1/M_max(B, cc) = cc/B` diverges as `cc → 0` (the exponent goes to zero, not infinity). So the expression `(ratio)^{cc/B}` → `1` for any finite ratio as `cc → 0`. This is not the intended limit. The correct form should involve the ratio of successive path counts at decreasing `cc` values, not the absolute ratio relative to `cc = ∞`. The geometric-mean estimator described in the practice note (design note §4.4) is actually computing something different: the geometric mean of *successive* path-count ratios as `cc` decreases, which is `(total_paths(cc=3) / total_paths(cc=100))^{1/k}` for `k` steps. The operational definition as written doesn't match this estimator. Either the formula needs to be corrected, or the operational definition should be stated in terms of successive ratios rather than the absolute ratio.

---

## §1 Background — new content review

### §1.1 Chung–Lu relevance note

The updated relevance note says: *"For graphs with heavy-tailed degree distribution (like Wikipedia categories), d̃ >> ⟨k⟩ because the second moment is dominated by hubs. The Chung-Lu theorem then predicts substantially shorter distances than the constant-degree estimate would suggest. This size-biased correction is the same one entering our b_eff definition (§1.5)."*

This is correct and adds genuine value. One additional connection worth making: the Chung–Lu theorem's `d̃ = E[w²]/E[w]` is precisely the denominator term in `b_eff = (E[d²_c]/E[d_c]) / (E[d²_p]/E[d_p])` applied to child edges. So `b_eff`'s numerator is `d̃_child` (the Chung–Lu effective degree for child hops) and its denominator is `d̃_parent`. The TLI being low is, in this framing, the statement that child-direction effective degree dominates parent-direction effective degree. This phrasing is a clean one-sentence connection that no prior review drew out.

### §1.2 Watts–Strogatz note

The updated note says: *"A random small-world (Erdős-Rényi or Chung-Lu) has C → 0 but still has the distance scaling."*

This is slightly imprecise: Erdős–Rényi has `C → 0` as `n → ∞` for sparse graphs (`p = c/n`), but the Chung–Lu model also has `C → 0` for power-law graphs in the same regime. More importantly: Erdős–Rényi is usually *not* called a "small-world" because its clustering is negligible — Watts–Strogatz small-worlds specifically have *high* clustering. The sentence could be read as calling Erdős–Rényi a small-world, which would confuse readers familiar with the Watts–Strogatz definition. Better phrasing: *"The key property for TLI regime classification is `L = Θ(log n)`, which holds for both Watts–Strogatz small-worlds (high clustering) and Chung–Lu random graphs (low clustering). We use 'small-world distance scaling' to refer to this `Θ(log n)` property throughout §5.3, regardless of clustering."*

---

## §2 Proofs — updated theorem 2.3

### The two-sided bound (new)

The updated Theorem 2.3 now bounds both `Σ S_M / S_0 ≤ r/(1-r)` and `Σ W_M / W_0 ≤ r/(1-r)`. This is the right structure. The proof of the W bound says:

> *"The range size shrinks weakly in M (the budget loses capacity cc per child hop), so W_M / W_0 ≤ r^M similarly."*

The W_M bound requires two steps: (1) W_M ≈ r^M · |admissible N range at M|, and (2) |admissible N range at M| ≤ |admissible N range at 0|. Step (2) is the budget argument (range = B - M·cc), which is correct. But step (1) requires the path-count factorisation from Lemma 2.1 applied to the weight sum, not just the length-weighted sum. Specifically, `W_M = Σ_p:M(p)=M w(p)`. By Lemma 2.1, the number of paths with exactly M child hops and N parent hops is ≈ D^N · (b')^M. Their weight is `D^{-N} · (b_eff·D)^{-M}`. So `W_M ≈ Σ_N D^N (b')^M · D^{-N} (b_eff·D)^{-M} = Σ_N r^M = r^M · |N range|`. This derivation is correct but entirely omitted from the proof — the proof jumps directly to the conclusion. Adding this two-line derivation would make the W bound self-contained.

### Caveat 1 in Theorem 2.3

The updated caveat says: *"A first-order estimate: TLI ≲ |S_drift - W_drift|/n, typically a fraction of r/(1-r) rather than the full value."*

The formula `|S_drift - W_drift|/n` is not derived here and its form is not obvious. From the §5.1 discussion, the correct first-order estimate is `TLI ≈ (1/n)|Δlog N - Δlog W|`. This is the same expression, but with `S_drift` meaning `Δlog N` and `W_drift` meaning `Δlog W` — except that `S_drift` as defined in the theorem is the *absolute* contribution-sum drift (≤ r/(1-r)), not the *log* drift. If `S_drift ≈ r/(1-r)` and `W_drift ≈ r/(1-r)`, then `|S_drift - W_drift|` could be anywhere from 0 to `2r/(1-r)`, and dividing by `n` doesn't give anything sharper than `2r/(n(1-r))`. The correct statement should reference the *log* drifts: `Δlog N ≈ Σ r^k L_k/L_0` and `Δlog W ≈ Σ r^k R_k/R_0`, whose difference is second-order in `r` because the leading `r` terms nearly cancel. The caveat in the theorem should either just say "see §5.1 for the tightening argument" without the intermediate formula, or state the formula correctly. As written, `|S_drift - W_drift|/n` with the linear-drift definitions of S and W would not give the O(r²) claim.

---

## §3 Conjectures — new conjecture quality

### Conjecture 3.3 (`ψ ~ r²/(1-r)`)

Added as a named conjecture. The motivation paragraph is correct in spirit but Conjecture 3.3 is essentially Conjecture 3.1 with a specific `ψ` function proposed. The relationship should be made explicit: *"Conjecture 3.3 is a sharpening of Conjecture 3.1: it asserts that the function `ψ` of 3.1 satisfies `ψ(r) ≤ C·r²/(1-r)`, where Theorem 2.3 currently only establishes `ψ(r) ≤ r/(1-r)` (the contribution-sum bound, not a direct TLI bound)."* Without this, a reader may wonder why 3.3 is a separate conjecture rather than a remark on 3.1.

### Conjecture 3.6 (routing-correction redundancy)

Well written. The production consequence section is the most practically useful passage in the entire document for a developer reading the companion. One note: the conjecture says `BranchRatio = b_eff · ρ` should be replaced by `BranchRatio = b_eff`. But if ρ < 1 (the routing correction deflates b_eff), then dropping ρ increases `BranchRatio`, which decreases `r = b'/(BranchRatio · D)`. A smaller `r` means a tighter convergence bound, so if the conjecture is correct, dropping the routing correction *improves* the theoretical TLI bound rather than degrading it. This is worth stating explicitly because it makes the conjecture non-obvious: it's not "routing correction is helping but is unnecessary" — it's "routing correction is actively *hurting* the theoretical calibration, and dropping it gives a stricter (better-justified) bound."

---

## §5 Open questions — new content review

### §5.1 First-order TLI estimate

The expression:

> `TLI ≈ (1/n)|Δlog N - Δlog W|`

where `Δlog N ≈ Σ r^k L_k/L_0` and `Δlog W ≈ Σ r^k R_k/R_0`.

The claim that `Δlog N` and `Δlog W` "nearly cancel" at leading order in `r` deserves a single explicit statement of *why*. At `M = 1` (first correction), `L_1/L_0 = Σ_N (N+2)^{-n} / Σ_N (N+1)^{-n}` (a ratio slightly below 1) and `R_1/R_0 = (B-cc)/B` (also slightly below 1 for small `cc`). The claim is that `L_1/L_0 ≈ R_1/R_0`, so the `r¹` terms cancel. Is this true? For `B = 15`, `cc = 5`, `n = 2`: `R_1/R_0 = 10/15 = 0.667`; `L_1/L_0 = Σ_{N=0}^{10} (N+2)^{-2} / Σ_{N=0}^{15} (N+1)^{-2} ≈ 1.485 / 1.580 ≈ 0.940`. These are *not* close to each other (0.667 vs 0.940), so the `r¹` terms do *not* cancel at leading order for this budget. The first-order TLI estimate would be `(1/2)|r(0.940 - 0.667)| = (1/2) · 0.157 · 0.273 ≈ 0.021`, which is ~2.1% — much closer to the 18.6% theorem bound than to the empirical 0.02%. The `r²` scaling conjecture would then need a second-order cancellation, which also doesn't obviously occur.

This is a substantive concern: the §5.1 analysis asserts the `r²` scaling as if the first-order cancellation is guaranteed by the ratio structure, but a rough numerical check suggests the first-order correction is not small (0.667 vs 0.940 are meaningfully different). The 0.02% empirical TLI may instead reflect a structure not captured by the geometric-series bound at all — for example, the budget constraint at `B = 15` means that at `cc = 5`, only paths with M ≤ 3 child hops are admitted, and those paths have a specific length structure that the bound doesn't use. The §5.1 open question should note this: *"A direct numerical evaluation of the first-order terms at the simplewiki parameters (B=15, cc=5, n=2) does not immediately yield cancellation; the source of the 1000× gap may require a more detailed accounting of the length structure at each M level, not just the geometric-series envelope."*

### §5.5 spectral expansion connection — new counterexample

The D-regular graph counterexample is correct and well-placed. The conclusion drawn ("rules out spectral-gap equivalence; leaves open directional spectral quantity") is appropriate. One additional observation: for the D-regular directed graph in the counterexample, the *singular value decomposition* of the child-adjacency matrix versus the parent-adjacency matrix would show equal singular values (by symmetry), so the directional singular-value approach also fails for this graph. This suggests the relevant spectral quantity is something like the *ratio* of the largest singular values of the child vs parent adjacency matrices — which would be `b_eff` itself in disguise. Worth one sentence.

---

## Minor items

- **§3.2 partial support**: The robustness-band range "27 to 1659" still appears in §3.2 without a cross-reference to the §4.2 explanation. Add "(see §4.2 for derivation of the 1659 figure)".
- **§3 preamble**: The ordering description says "3.5 falsification target" but Conjecture 3.5 is the symmetric-DAG conjecture and 3.6 is the routing conjecture. The preamble should list all six conjecture summaries, not stop at 3.5's description.
- **§4 status section**: §4.3 (Conjecture 3.3, the tightening) is listed as "pending" without any estimated approach. Given that §5.1 now has a partial first-order analysis, §4.3 could note "§5.1 provides a first-order estimate consistent with the r² scaling; the constant C requires the full Taylor expansion or Monte Carlo tests at varying r."  
- **Reference list (§6)**: All five citations are correctly formatted. The Chung–Lu DOI (10.1073/pnas.252631999) would be a useful addition for the PNAS citation.

---

## Verdict

**Approve with two suggestions before merge:**

1. **Fix the §0.4 operational b' formula** — the limit `(ratio)^{cc/B} → 1` as `cc → 0` is not the intended definition. Either replace with a successive-ratio formulation or remove the formula and keep only the prose description plus the practice note.

2. **Add a caveat to §5.1** — the first-order cancellation claim (`r²` scaling from partial cancellation of `Δlog N` and `Δlog W`) is not verified numerically for the simplewiki parameters and appears not to hold at leading order. The 1000× empirical gap likely has a different explanation (budget-constrained M range, not just the geometric-series ratio structure). Stating this uncertainty honestly is consistent with the document's epistemics.

All other items are minor polish for a follow-up commit.

s243a (commented):
## Fourth review — post-second-update pass (1026-line version)

This pass focuses on the new material: the revised §0.4 estimation section, the revised §0.5 with denominator and absolute-value notes, the §0.6 piecewise-caveat, the §1.5 Feld algebra, the §2.3 proof completions, the revised Conjecture 3.3, and the new §5.1 numerical table.

---

## ✅ Fully resolved since last pass

- **§0.4 operational `b'` definition**: The broken `(ratio)^{cc/B} → 1` formula is gone. Replaced with a well-posed successive-ratio estimator: ratio of total path counts at adjacent `(B, cc)` values, with explicit `M_max` counts `{0, 1, 3, 5}` at `cc ∈ {100, 10, 5, 3}`. The derivation of `b' ≈ 11.1` from `exp((log 122.7)/2)` is correct and gives a concrete example.
- **§0.5 denominator choice and absolute value**: Both now have explicit justifications. Denominator uses the full-DAG value as reference ("tree-search is an approximation to full-DAG, TLI is the relative error"). Absolute value is justified by the fact that drift direction is not guaranteed monotonic (new paths can increase or decrease `d_wPow` depending on their length). Empirical note ("drift is positive on simplewiki") is a useful grounding.
- **§0.6 piecewise-homogeneity caveat**: Added. The caveat is strong and correct: "if Q mixes nodes from multiple components, Lemma 2.1's factorisation fails and neither Theorem 2.3 nor Corollary 2.3.1 applies." The recommended workflow (constrain Q to a single component via topical-root LMDB) is the right resolution.
- **§1.5 Feld lemma algebra**: The `b_eff = d̃_c / d̃_p` derivation is now fully explicit. Better still, the note adds a clean semantic interpretation: "Low TLI = child-direction Chung-Lu effective degree dominates parent-direction one by a factor large enough that `b_eff · D > b'`."
- **§2.3 proof completeness (L_M ≤ L_0 and R_M ≤ R_0)**: Both claims now have explicit term-by-term arguments. The `L_M ≤ L_0` proof uses pointwise domination `(N+M+1)^{-n} ≤ (N+1)^{-n}` plus the fact that the range at level M is a subset of the range at level 0. The `R_M ≤ R_0` proof uses the budget constraint explicitly (`[d_min, B - M·cc] ⊆ [d_min, B]`). Both are correct.
- **Conjecture 3.3 relationship to 3.1**: Now explicitly states "This conjecture *refines* Conjecture 3.1 by proposing a specific form for the function `ψ`."
- **§5.1 numerical table**: The `L_M / L_0` and `R_M / R_0` comparison at M = 0, 1, 2 is a genuine contribution. The conclusion (first-order prediction ~2.1%, still 100× off from 0.02%) is honest and correct. The three candidate explanations for the remaining gap (budget truncation, higher-order cancellation, measurement variance) are all plausible.
- **Conjecture 3.6 directionality**: The new "Directionality clarification" paragraph is exactly right and adds real value: dropping `ρ < 1` *increases* BranchRatio, *decreases* r, and *tightens* the bound. The rephrasing ("routing correction was actively miscalibrating in the wrong direction") is honest and precise.
- **§3 preamble fix**: The preamble now lists all six conjectures with correct summaries.

---

## 🟡 Residual issues (all minor)

**R1. §0.4 estimation note: the `b' ≈ 11.1` derivation is from a single transition**

The text says: `b' ≈ exp((log 122.7)/2) ≈ 11.1` from "the dominant transition." The 122.7 ratio is from `cc = 10 → 5`, which admits M ∈ {0,1,2,3} at cc=5 vs M ∈ {0,1} at cc=10. The ratio `Σ_{M=0}^{3} (b')^M / Σ_{M=0}^{1} (b')^M` ≈ `(1 + b' + (b')^2 + (b')^3) / (1 + b')` for b' ≈ 11 gives approximately `(1 + 11 + 121 + 1331) / (1 + 11) = 1464 / 12 = 122`. So the ratio is correctly attributed to admitting 2 additional child-hop levels (M=2 and M=3). But the document says `exp((log 122.7)/2)` as if extracting a square root — the exponent 2 is the number of *additional* M levels (M=2 and M=3 beyond M=1). This is correct but the derivation should say: "admitting M=2 and M=3 beyond the M≤1 baseline contributes two new child-hop levels, so `b' ≈ (122.7)^{1/2} ≈ 11.1`" rather than just citing `exp((log 122.7)/2)`. The current phrasing is slightly opaque.

**R2. §1.2 Watts–Strogatz relevance note: Erdős–Rényi terminology**

The updated note says: "A *random* small-world (Erdős-Rényi or Chung-Lu) has C → 0 but still has the distance scaling." As flagged in the last pass, calling Erdős–Rényi a "random small-world" is slightly confusing because Watts–Strogatz small-worlds specifically contrast with random graphs. Better: drop "Erdős–Rényi or" and say "A Chung-Lu random graph has C → 0 but still has L = Θ(log n) distance scaling." This is still unaddressed from the previous review.

**R3. §2.3 proof: weight algebra at the key step**

The proof writes: `D^{-N} (b_eff · D)^{-M} = D^{-N} (r · D / D)^{-M} · D^{-M}`. This simplification is algebraically wrong: `(b_eff · D)^{-M} = ((b'/r) · ??? )`. Let's check: `r = b'/(b_eff · D)`, so `b_eff · D = b'/r`. Therefore `(b_eff · D)^{-M} = (b'/r)^{-M} = (r/b')^M`. The correct simplification is: `w(p) · #paths(N,M) ≈ D^{-N} (b'/r)^{-M} · D^N (b')^M = (r/b')^M · (b')^M = r^M`. The intermediate step `(r · D/D)^{-M} · D^{-M}` seems to be an attempt to factor out D, but it equates to `r^{-M} · D^{-M}`, which is not the right direction. The final result `r^M` is correct, but the intermediate algebra as written is garbled. Replace with: "Each path at level (N,M) carries weight `w = D^{-N} · (b_eff · D)^{-M} = D^{-N} · (b'/r)^{-M}`. The number of such paths is `D^N · (b')^M` (Lemma 2.1). Their product is `(b'/r)^{-M} · (b')^M = r^M`."

**R4. §5.1 table: `L_0 ≈ 0.149` needs verification**

The table gives `L_0 ≈ 0.149` for `d_min = 4, B = 15, n = 2`. `L_0 = Σ_{N=4}^{15} (N+1)^{-2} = 5^{-2} + 6^{-2} + ... + 16^{-2}`. Let’s check: `Σ_{k=5}^{16} k^{-2} ≈ 0.04 + 0.0278 + 0.0204 + 0.0156 + 0.0123 + 0.01 + 0.0083 + 0.0069 + 0.0059 + 0.0051 + 0.0044 + 0.0039 ≈ 0.1806`. So `L_0 ≈ 0.181`, not 0.149. For `d_min = 4` and range `N ∈ [4, 15]` (12 terms as stated for R_0 = 12), the sum is `Σ_{N=4}^{15} (N+1)^{-2} = Σ_{k=5}^{16} k^{-2} ≈ 0.181`. The document gives 0.149 which would correspond to a different range, perhaps `N ∈ [4, 12]` (9 terms, `R_0 = 9`) or `d_min = 5`. This is a numerical error in the table that also makes `L_1 / L_0 = 0.140 / 0.149 = 0.940` suspicious (since the absolute values aren’t correct). The *ratio* `L_1/L_0` matters for the argument, so if both values are off by the same factor the ratio could still be approximately right — but the absolute figures should be corrected. The first-order prediction of ~2.1% depends on these values being accurate.

**R5. §5.1 first-order prediction arithmetic**

The document gives the leading-order TLI contribution as: `≈ 0.5 · 0.273 · 0.157 ≈ 0.021 = 2.1%`. The `0.5 = 1/n` for `n = 2` is correct. But the formula `(1/n)|r · (L_1/L_0 - R_1/R_0)|` gives the coefficient of `r^1` in the TLI expansion. This is `0.5 · 0.157 · 0.273 ≈ 0.0214`. For this to be the correct first-order TLI, the formula needs to be `TLI ≈ (1/n) |Σ_k r^k (L_k/L_0 - R_k/R_0)|`. The M=2 contribution would add `0.5 · r^2 · (0.819 - 0.250) = 0.5 · 0.0246 · 0.569 ≈ 0.007`. So the two-term estimate is ~2.8%, still dominated by the M=1 term. This is consistent with the document's claim that the first-order estimate is ~2.1% and the M=2 term is small. But the table should probably show the running total to make this clear.

---

## New observation: §5.1's gap explanation should mention the `cc` proxy effect

The three candidate explanations for the 100× remaining gap are: budget-constrained M range, higher-order cancellation, and measurement variance. There’s a fourth candidate not mentioned: **the empirical TLI uses `cc = 100` vs `cc = 5` as a proxy for `cc = ∞` vs `cc = 0⁺`**. At `cc = 100` with `B = 15`, paths with M = 1 child hop would cost 101, which exceeds B = 15, so `cc = 100` does act as `cc = ∞` (no child hops admitted). But at `cc = 5`, only M ≤ 3 is admitted. The theoretical TLI is the full `cc = 0⁺` limit (all M admitted), while the empirical measurement is the contrast at a *finite* `cc = 5`. The theoretical first-order prediction at `cc = 5` is the correct comparison (M ≤ 3 only), not the full `cc = 0⁺` limit. The document's `L_M / L_0` table at M = 0, 1, 2 already accounts for this (the M = 3 row is implicitly included since R_2 = 3 and R_3 would be 0 or 1). But making this explicit would sharpen the explanation: "the ~2.1% first-order estimate is already for the `cc = 5` approximation, not the theoretical `cc = 0⁺` limit; the gap between 2.1% and 0.02% cannot be explained by the finite-cc proxy."

---

## Verdict

**Approve — ready to merge.** The document is in excellent shape. All major structural and correctness issues from prior reviews are resolved. The §5.1 numerical table is a genuine contribution that makes the open question concrete and honest.

The only pre-merge recommendation is **R3** (the garbled weight algebra in the Theorem 2.3 proof) — the conclusion is correct but the intermediate step `(r · D/D)^{-M} · D^{-M}` is algebraically wrong and could confuse a reader checking the proof. Everything else (R1, R2, R4, R5, the cc-proxy observation) is follow-up polish.

s243a (commented):
## Independent second review of 1026-line version

Fresh read, no deference to prior review threads. Focus: (1) mathematical correctness of the new §5.1 table and first-order analysis, (2) whether the revised Conjecture 3.3 is correctly framed, (3) new content in §0.4–0.5, (4) end-to-end logical coherence of the full document.

---

## §0.4 — Estimation in practice (new section)

### The `b' ≈ 11.1` derivation is correct but the dominant transition claim needs unpacking

The text says: `b' ≈ exp((log 122.7)/2) ≈ 11.1` from "the dominant transition." The dominant transition is `cc = 10 → 5`, which moves from `M_max = 1` to `M_max = 3`. The ratio of total path counts is:

$$\frac{\sum_{M=0}^{3}(b')^M}{\sum_{M=0}^{1}(b')^M} = \frac{1 + b' + (b')^2 + (b')^3}{1 + b'} \approx \frac{(b')^3}{b'} = (b')^2 \text{ for large } b'$$

So `ratio ≈ (b')²`, giving `b' ≈ √122.7 ≈ 11.1`. This is correct. But the text says "taking the geometric mean over transitions where `m_{k+1} - m_k = 1` recovers `b'` directly" — yet the dominant transition has `m_{k+1} - m_k = 2` (not 1). The document handles this with "For larger M increments, the appropriate root corrects for the number of admitted child-hop levels", but then immediately uses the square root (which corresponds to 2 additional levels, not 1). The reconciliation is correct but the reader has to piece it together. A one-sentence explicit statement would close the gap: *"The cc = 10 → 5 transition admits two new child-hop levels (M = 2 and M = 3), so we take the square root of the ratio to recover the per-level growth `b' ≈ √122.7 ≈ 11.1`."*

### The `cc = 100 → 10` transition ratio of 1.05 is suspicious

The text reports the `cc = 100 → 10` transition ratio as 1.05. At cc = 100, budget B = 15: M_max = 0 (since 1 child hop costs 100 > 15). At cc = 10, M_max = 1. So this transition admits M = 1 for the first time. The ratio should be:

$$\frac{1 + b'}{1} = 1 + b' \approx 1 + 11 = 12$$

not 1.05. A ratio of 1.05 would imply that M = 1 paths contribute only 5% of the M = 0 path count, which contradicts `b' ≈ 11` (where M = 1 paths should contribute ~11× the M = 0 count weighted by `r = 0.157`, i.e. ~1.73 times net). Unless "total #paths" here is *unweighted* count of paths, not weighted by `w(p)` — in which case the ratio depends on the raw branching, not the convergence ratio. The text should clarify: are the transition ratios computed from *weighted* path counts (using `w(p)`) or *unweighted* path counts? If unweighted, the transition ratios behave differently from the `r^M` argument in Theorem 2.3, which is about *weighted* sums. If weighted, the 1.05 figure for the cc=100→10 transition is inconsistent with `b' ≈ 11`.

---

## §0.5 — TLI definition (new clarifying notes)

Both the denominator-choice note and the absolute-value note are well-written and resolve prior concerns. One small remaining issue:

### Monotonicity claim is stated without proof

The absolute-value note says: "admitting more paths can either decrease or increase `d_wPow` depending on whether the new paths have above- or below-average `(h+1)^{-n}` values." This is correct for `n > 0`. But the document earlier (§0.3) notes that `d_wPow` is a *negative* power mean — it is the inverse `n`th root of a weighted mean of `(h+1)^{-n}`. For this metric, adding paths with *longer* total hop count `h` (smaller `(h+1)^{-n}`) to the sum increases the weighted mean of `(h+1)^{-n}` (since the new terms pull the mean up from below), which then *decreases* `d_wPow` (since `d_wPow = (mean)^{-1/n}` and the mean increased). Wait — that's the opposite: adding longer paths decreases the metric. Adding shorter paths increases it. So child-hop paths, which are longer on average, should *decrease* `d_wPow`. But the empirical note says "drift is positive (the full-DAG value is slightly higher)", which would mean adding child-hop paths *increases* `d_wPow`. These two claims are in tension. Either (a) child-hop paths are shorter than parent-only paths on average (which seems unlikely given they mix parent and child steps), or (b) the weight `w(p)` effect dominates the length effect. The absolute-value justification should acknowledge this: it's not just that new paths might be above- or below-average, but that the weight-length interaction is non-trivial and the empirical direction (positive drift) is noted without a mechanistic explanation.

---

## §1.5 — Feld lemma (new algebra)

The new derivation is correct and the semantic interpretation ("Low TLI = child-direction Chung-Lu effective degree dominates parent-direction") is genuinely clarifying. One addition would strengthen it:

The document notes that `b = E[d²_c]/E[d²_p]` differs from `b_eff` by the factor `E[d_p]/E[d_c]` on simplewiki (= 0.436). This means `E[d_c] / E[d_p] = 0.436`, i.e. the average child degree is 43.6% of the average parent degree. This is a concrete fact about the simplewiki category graph (nodes have on average fewer children than parents) that directly motivates why the first-moment correction matters. Currently only the ratio 0.436 is mentioned; stating the direction explicitly ("nodes have fewer children than parents on average, so the size-biased correction deflates `b` to `b_eff`") would help readers understand why `b ≠ b_eff` matters.

---

## §2.3 — Theorem 2.3 proof (revised)

### The weight algebra step is still garbled (flagged in prior review, not yet fixed)

The proof writes:
```
D^{-N} (b_eff · D)^{-M} = D^{-N} (r · D / D)^{-M} · D^{-M}
```
The right-hand side expands to `D^{-N} · r^{-M} · D^{-M}`, which equals `D^{-N-M} · r^{-M}`. This is not the weight `w(p)`; it has `r` in the *denominator* when it should be in the numerator. The correct algebra is:
- `r = b' / (b_eff · D)`, so `b_eff · D = b' / r`
- `w(p) = D^{-N} · (b_eff · D)^{-M} = D^{-N} · (b'/r)^{-M} = D^{-N} · (r/b')^M`
- `w(p) · #paths(N,M) ≈ D^{-N} · (r/b')^M · D^N · (b')^M = r^M` ✓

The current intermediate step is wrong in both sign and grouping. This is the one pre-merge fix recommended by the prior review and is still present.

---

## §3.3 — Revised Conjecture 3.3

The revised conjecture is significantly improved: it now uses an exponent `α ≥ 1` rather than committing to `α = 2`, and the status section honestly states that the first-order analysis at simplewiki parameters does *not* support clean `r²` scaling. The conjecture is correctly framed as consistent-with but not derived-from the empirical observations.

One new concern: the conjecture says `ψ(r) ≤ C(n, B, d_min) · r^α / (1 - r)`. This form has `r^α / (1 - r)`, which for `α = 1` recovers the Theorem 2.3 contribution-sum bound. But Theorem 2.3 already establishes the bound *on contribution sums*, not on TLI directly. Conjecture 3.1 uses `ψ` as a bound on TLI. So Conjecture 3.3 is saying that the TLI-to-contribution-sum conversion factor is `O(r^{α-1})`. This is a cleaner way to state it: *"Conjecture 3.3 asserts that the ratio TLI / (r/(1-r)) is O(r^{α-1}), i.e. the TLI is smaller than the contribution-sum bound by a power of r."* This formulation makes the relationship to Theorem 2.3 and Conjecture 3.1 explicit in one sentence.

---

## §5.1 — Numerical table and first-order analysis

### The table values need verification (flagged in prior review)

For `d_min = 4`, `B = 15`, `n = 2`:
- Level 0: N ranges over [4, 15] (R_0 = 12 values). `L_0 = Σ_{N=4}^{15} (N+1)^{-2} = 5^{-2} + 6^{-2} + ... + 16^{-2}`.
  - `= 0.04 + 0.0278 + 0.0204 + 0.0156 + 0.0123 + 0.01 + 0.0083 + 0.0069 + 0.0059 + 0.0051 + 0.0044 + 0.0039 ≈ 0.181`
  - The table gives **L_0 ≈ 0.149** — discrepancy of ~21%.
- Level 1: At cc = 5, one child hop costs 5, leaving budget B - 5 = 10 for parent hops. N ranges over [4, 10] (R_1 = 7 values). `L_1 = Σ_{N=4}^{10} (N+1+1)^{-2} = Σ_{N=4}^{10} (N+2)^{-2} = 6^{-2} + 7^{-2} + ... + 12^{-2}`.
  - `= 0.0278 + 0.0204 + 0.0156 + 0.0123 + 0.01 + 0.0083 + 0.0069 ≈ 0.101`
  - The table gives **L_1 ≈ 0.140** — discrepancy of ~39%.
  - Also, R_1 = 7 (N ∈ [4,10]), but the table gives **R_1 = 8**. At cc = 5, one child hop costs 5, leaving 10 budget for parent hops. N ∈ [d_min, 10] = [4, 10] = 7 values, not 8.

These errors propagate into the key ratio calculations:
- Correct `L_1/L_0 ≈ 0.101/0.181 ≈ 0.558`, not 0.940
- Correct `R_1/R_0 = 7/12 ≈ 0.583`, not 0.667
- Correct difference `L_1/L_0 - R_1/R_0 ≈ 0.558 - 0.583 = -0.025`, not +0.273

This is a qualitatively different result: the two ratios are **nearly equal** (0.558 vs 0.583, difference of -0.025), which means the `r^1` coefficients nearly cancel after all! The first-order TLI would be `(1/2) · |−0.025| · 0.157 ≈ 0.002 = 0.2%`, not 2.1%. This is much closer to the empirical 0.02%, and the gap is now only 10× rather than 100×.

If this calculation is correct, it substantially changes the conclusion of §5.1:
- The near-cancellation **does** occur at leading order
- The `r²` scaling hypothesis may be recoverable from the data
- The three "candidate explanations" for the remaining gap are still relevant but explain a 10× gap, not 100×

This should be recalculated carefully before merge, as it affects the status of Conjecture 3.3 materially. The current table appears to have used different parameters than stated (perhaps `d_min = 0` or `cc = 1` rather than `cc = 5`).

### The absolute-value convention in the table

The table column `|L_M/L_0 - R_M/R_0|` should be unsigned differences for the bound, but the formula uses `Δlog N - Δlog W` which can be positive or negative. If `L_1/L_0 < R_1/R_0` (as in the corrected values), the difference is negative, meaning admitting child-hop paths *decreases* `d_wPow` (consistent with the observation that child paths are longer and the metric decreases when longer paths are added). This would also explain why the empirical drift is positive: the denominator normalisation uses the *full-DAG* value (lower), and `d_wPow(cc=∞) > d_wPow(cc=5)`, so TLI = |higher - lower| / lower > 0. The sign structure should be made explicit.

---

## §3.2 — Cross-reference to §4.2 for the "1659" figure

The "1659" figure appears in §3.2's partial-support section with no cross-reference to §4.2's derivation, despite §4.2 now containing the explicit derivation. A parenthetical "(see §4.2 for derivation)" is still missing. This was flagged in two prior reviews.

---

## Verdict

**Approve with one pre-merge fix and one pre-merge recomputation:**

1. **Fix the weight algebra in §2.3** (the garbled intermediate step `(r·D/D)^{-M}·D^{-M}` is algebraically incorrect). This is a correctness issue in a proof section.

2. **Recompute the §5.1 table** using the stated parameters (`d_min = 4`, `B = 15`, `cc = 5`, `n = 2`). The current values appear inconsistent with those parameters, and the corrected values likely show near-cancellation of the `r^1` terms (difference ≈ -0.025, not +0.273), which would materially strengthen the case for Conjecture 3.3's `α = 2` hypothesis and change the document's conclusion from "the r² scaling is not supported" to "the first-order analysis is consistent with near-cancellation and the r² hypothesis."

PR #2680, opened 2026-06-02
docs: LaTeX rendering fixes + Bayesian interpretation of routing correction

s243a:
## Summary

  Two distinct changes on top of the merged theory doc:

  ### Commit 1: LaTeX rendering fixes

  GitHub MathJax has two strict rules the merged doc violated:

  1. **Bare `#` produces "macro parameter character" error.** The doc used `\#\text{paths}(...)` in 8 instances plus a
  `\#\mathrm{children}` in §0.6. Even with the `\#` escape, GitHub's MathJax doesn't handle this correctly inside
  `\text{}`. Replaced with `\mathrm{paths}(...)` and `|\mathrm{children}(v)|`.

  2. **Whitespace between `$$` and formula content prevents rendering.** The doc had 25 display math blocks formatted
  `$$\n formula \n$$`, where the newlines counted as whitespace. Joined `$$` to the first formula line and the last
  formula line to the closing `$$`. Multi-line readability preserved.

  Both issues identified via direct reader feedback on the rendered doc.

  ### Commit 2: Bayesian interpretation + regime classification (new content)

  Two new subsections that came out of follow-up discussion:

  **§3.6.1 Bayesian interpretation of the routing correction.** Frames the metric weight `w(p)` as a prior probability
  that path `p` represents the "true" route. The base metric encodes a uniform prior over path shapes (path-count
  normalisation, §5.6). Adding `ρ < 1` shifts the prior toward "child paths informative" by factor `(1/ρ)^M`. The TLI
  measurement itself shows child paths contribute negligibly, so the prior shift contradicts the data — and dropping ρ
  both aligns prior with data and tightens the Theorem 2.3 bound. Restates Conjecture 3.6 in Bayesian form.

  **§5.6 Regime classification: when does ρ exceed D's buffer?** Answers when the buffer condition `ρ > r` holds across
  graph regimes:

  | Regime | γ | Buffer for ρ |
  |---|---|---|
  | Tree-like | > 3 | wide |
  | Wikipedia-like | 2.4–3 | moderate |
  | Ultra-small-world | 2 < γ < 3 | shrinks |
  | Degenerate | ≤ 2 | none |

  Open question: identify the critical γ* at which (a) r → 1 and (b) ρ > r becomes restrictive. Quantitative answer
  would sharpen Conjecture 3.6 into a regime statement.

  ## Test plan

  - [ ] LaTeX renders correctly in GitHub markdown preview (no "macro parameter character" errors)
  - [ ] Display math blocks render (no missing/broken formulas due to whitespace)
  - [ ] Cross-references between §3.6, §3.6.1, §5.6 resolve consistently
  - [ ] §3.6.1 directionality argument matches §3.6's "tightens the bound" claim

  ## Notes

  - §3.6.1 and §5.6 are formal elaborations of the same point from two angles (Bayesian inference / regime
  classification). Reviewers may notice some redundancy between them — intentional, since they target different reader
  entry points.
  - Task #14 (settle Conjecture 3.6 empirically) now has a refined design: report per-pair `d_wPow` values, not just
  aggregate TLI. If aggregate passes under both compositions but per-pair distributions differ, ρ was encoding something
  real — and we'd need to investigate further.

### DIFF
--- a/docs/design/TREE_LIKENESS_INDEX_THEORY.md
+++ b/docs/design/TREE_LIKENESS_INDEX_THEORY.md
@@ -32,10 +32,8 @@ of vertices (which is the case for Wikipedia category graphs).
 
 What we *do* require is a parent/child decomposition: for each
 non-root $v$, define
-$$
-\mathrm{parents}(v) \subseteq V, \quad
-\mathrm{children}(v) \subseteq V
-$$
+$$\mathrm{parents}(v) \subseteq V, \quad
+\mathrm{children}(v) \subseteq V$$
 as the sets of nodes connected to $v$ by an in-edge or out-edge
 respectively, in the categorisation interpretation. The
 distinction between "parent direction" and "child direction" is
@@ -82,17 +80,13 @@ Following the design note's §2.0 notation, let:
   context; $b_{\text{eff}}$ is the canonical asymmetry quantity.
 
 The path weight is
-$$
-w(p) = D^{-N(p)} \cdot (b_{\text{eff}} \cdot D)^{-M(p)}
-$$
+$$w(p) = D^{-N(p)} \cdot (b_{\text{eff}} \cdot D)^{-M(p)}$$
 and the **directionally-weighted power-mean metric** with exponent
 $n$ is
-$$
-d_{\text{wPow}}(v; B, cc) = \left(
+$$d_{\text{wPow}}(v; B, cc) = \left(
 \frac{\sum_{p \in \mathcal{P}(v; B, cc)} w(p) \cdot (h(p)+1)^{-n}}
      {\sum_{p \in \mathcal{P}(v; B, cc)} w(p)}
-\right)^{-1/n}
-$$
+\right)^{-1/n}$$
 where $\mathcal{P}(v; B, cc)$ is the set of paths from $v$ to $r$
 with cost at most $B$ when each parent hop costs $1$ and each
 child hop costs $cc$. We typically use $n = 2$ in experiments.
@@ -107,11 +101,9 @@ $b_{\text{eff}}$.
 ### 0.4 Path-count growth
 
 For a node $v$ reachable to $r$ within budget $B$, define
-$$
-\#\text{paths}(v; N, M; B) :=
+$$\mathrm{paths}(v; N, M; B) :=
 \bigl|\{p : v \to r,\ p \text{ has } N \text{ parent hops}
-              \text{ and } M \text{ child hops, cost} \le B\}\bigr|
-$$
+              \text{ and } M \text{ child hops, cost} \le B\}\bigr|$$
 At any fixed budget $B$ and child-step cost $cc > 0$, the maximum
 admissible $M$ is finite ($\lfloor B/cc \rfloor$), so a literal
 $M \to \infty$ limit does not exist. We instead define $b'$ as
@@ -119,11 +111,9 @@ the **asymptotic per-child-hop path-count growth rate** under
 the homogeneity assumption: there exists a constant $b' \ge 1$
 and a node-dependent constant $C(v) > 0$ such that, for $M$ in
 the range admissible at $(B, cc)$,
-$$
-\mathbb{E}_{v \sim Q}\bigl[\#\text{paths}(v; \cdot, M; B)\bigr]
-\approx C \cdot (b')^M
-$$
-where $\#\text{paths}(v; \cdot, M; B)$ denotes summation over $N$
+$$\mathbb{E}_{v \sim Q}\bigl[\mathrm{paths}(v; \cdot, M; B)\bigr]
+\approx C \cdot (b')^M$$
+where $\mathrm{paths}(v; \cdot, M; B)$ denotes summation over $N$
 and the approximation holds up to a multiplicative
 $(1 + o(1))$ factor uniformly in $M$.
 
@@ -141,12 +131,9 @@ of decreasing $cc$ values that admit successively more child
 hops. For two adjacent values $cc_1 > cc_2$ with
 $M_{\max}(B, cc_1) = m_1$ and $M_{\max}(B, cc_2) = m_2 > m_1$,
 under homogeneity
-$$
-\frac{\text{total } \#\text{paths at } (B, cc_2)}
-     {\text{total } \#\text{paths at } (B, cc_1)}
+$$\frac{\mathrm{paths_{total}}(B, cc_2)}{\mathrm{paths_{total}}(B, cc_1)}
 \;\approx\;
-\frac{\sum_{M=0}^{m_2} (b')^M}{\sum_{M=0}^{m_1} (b')^M}
-$$
+\frac{\sum_{M=0}^{m_2} (b')^M}{\sum_{M=0}^{m_1} (b')^M}$$
 so taking the geometric mean over transitions where
 $m_{k+1} - m_k = 1$ recovers $b'$ directly. For larger
 $M$ increments, the appropriate root corrects for the number
@@ -169,12 +156,10 @@ $b' \approx 11$ on simplewiki topical core.
 
 The **tree-likeness index** of the tuple $(G, \mu, Q, B)$ where
 $\mu$ is the directionally-weighted metric is
-$$
-\mathrm{TLI}(G, \mu, Q, B) :=
+$$\mathrm{TLI}(G, \mu, Q, B) :=
 \frac{\bigl|\mathbb{E}_{v \sim Q}\bigl[d_{\text{wPow}}(v; B, \infty)\bigr] -
             \mathbb{E}_{v \sim Q}\bigl[d_{\text{wPow}}(v; B, 0^+)\bigr]\bigr|}
-     {\mathbb{E}_{v \sim Q}\bigl[d_{\text{wPow}}(v; B, 0^+)\bigr]}
-$$
+     {\mathbb{E}_{v \sim Q}\bigl[d_{\text{wPow}}(v; B, 0^+)\bigr]}$$
 i.e. the relative drift of the mean metric value as the child-step
 cost ranges from infinity (tree-search, $M = 0$ only) to
 $0^+$ (full DAG search, all $M$ admitted within budget).
@@ -212,7 +197,7 @@ $(G, Q, B)$ is *statistically homogeneous* if there exist
 constants $D, b' \ge 1$ such that, uniformly over $v$ in the
 support of the marginal distribution $Q$:
 
-(H1) $\mathbb{E}\bigl[\#\mathrm{children}(v)\bigr] = D$
+(H1) $\mathbb{E}\bigl[|\mathrm{children}(v)|\bigr] = D$
 (within multiplicative factor $1 + o(1)$);
 
 (H2) The path-count growth in §0.4 has the same $b'$ at every
@@ -262,9 +247,7 @@ $\mathbf{w} = (w_1, \dots, w_n)$ — where edge $(i,j)$ is present
 with probability $w_i w_j / \sum_k w_k$ — and average degree
 $\langle k \rangle = \mathbb{E}[w] > 1$, the average shortest-path
 length between random node pairs satisfies
-$$
-L \sim \frac{\log n}{\log \tilde{d}}
-$$
+$$L \sim \frac{\log n}{\log \tilde{d}}$$
 as $n \to \infty$, where $\tilde{d} = \mathbb{E}[w^2] / \mathbb{E}[w]$
 is the *expected size-biased degree* (the average degree of a
 random endpoint of a random edge).
@@ -343,10 +326,8 @@ $\{d_1, \dots, d_n\}$ believed to follow a power-law tail
 $P(k) \propto k^{-\gamma}$ for $k \ge d_{\min}$:
 
 1. For each candidate $d_{\min}$, compute the continuous MLE
-$$
-\hat\gamma(d_{\min}) = 1 + n_{\ge d_{\min}} \biggm/
-\sum_{d_i \ge d_{\min}} \log(d_i / d_{\min})
-$$
+$$\hat\gamma(d_{\min}) = 1 + n_{\ge d_{\min}} \biggm/
+\sum_{d_i \ge d_{\min}} \log(d_i / d_{\min})$$
 2. Compute the Kolmogorov-Smirnov distance between the empirical
    tail CDF and the fitted power-law tail CDF
 3. Select the $d_{\min}$ that minimises KS distance
@@ -367,9 +348,7 @@ ultra-small-world consequences from this fit.
 **Lemma (Feld, 1991).** For a graph with finite degree
 distribution $P(k)$, the expected degree of a node reached by
 following a random edge is
-$$
-\tilde{d} := \frac{\mathbb{E}[k^2]}{\mathbb{E}[k]} \ge \mathbb{E}[k]
-$$
+$$\tilde{d} := \frac{\mathbb{E}[k^2]}{\mathbb{E}[k]} \ge \mathbb{E}[k]$$
 with equality iff degrees are constant. The quantity $\tilde{d}$
 is the *size-biased mean* or **Chung-Lu effective degree** (cf.
 §1.1).
@@ -380,16 +359,12 @@ do", *American Journal of Sociology* 96(6): 1464–1477 (1991).
 **Relevance to TLI: $b_{\text{eff}}$ is the ratio of Chung-Lu
 effective degrees.** Applying the size-biased mean separately to
 the child and parent edge populations gives
-$$
-\tilde{d}_c = \frac{\mathbb{E}[d_c^2]}{\mathbb{E}[d_c]}, \qquad
-\tilde{d}_p = \frac{\mathbb{E}[d_p^2]}{\mathbb{E}[d_p]}
-$$
+$$\tilde{d}_c = \frac{\mathbb{E}[d_c^2]}{\mathbb{E}[d_c]}, \qquad
+\tilde{d}_p = \frac{\mathbb{E}[d_p^2]}{\mathbb{E}[d_p]}$$
 and our calibrated branching asymmetry is exactly the ratio
-$$
-b_{\text{eff}}
+$$b_{\text{eff}}
 = \frac{\mathbb{E}[d_c^2]/\mathbb{E}[d_c]}{\mathbb{E}[d_p^2]/\mathbb{E}[d_p]}
-= \frac{\tilde{d}_c}{\tilde{d}_p}
-$$
+= \frac{\tilde{d}_c}{\tilde{d}_p}$$
 i.e. the Chung-Lu effective degree in the child direction divided
 by the Chung-Lu effective degree in the parent direction. **Low
 TLI thus has a clean interpretation: the child-direction
@@ -410,9 +385,7 @@ $n = (D^{L+1} - 1)/(D - 1)$ nodes:
 
 - Depth-to-root: $L = \log_D(n(D - 1) + 1) \approx \log_D n$
 - Average pair distance: for uniform-random distinct nodes $u, v$,
-$$
-\mathbb{E}_{u,v}[d(u,v)] \approx 2L - \frac{2D}{D-1} \approx 2 \log_D n
-$$
+$$\mathbb{E}_{u,v}[d(u,v)] \approx 2L - \frac{2D}{D-1} \approx 2 \log_D n$$
 with the constant offset $2D/(D-1)$ converging to $2$ as
 $D \to \infty$ (large branching makes LCAs shallow).
 
@@ -432,7 +405,7 @@ average pair distance", which we compare against
 small-world and ultra-small-world distance scalings. The lemma
 is *not* used to argue $\mathrm{TLI} = 0$ for trees — that
 follows trivially from the definition (a tree has no cross-edge
-paths, so $\#\text{paths}(v; N, M; B) = 0$ for all $M \ge 1$,
+paths, so $\mathrm{paths}(v; N, M; B) = 0$ for all $M \ge 1$,
 hence $d_{\text{wPow}}$ does not depend on $cc$ and TLI = 0
 identically).
 
@@ -447,15 +420,11 @@ conjecture (§3.1); they are building blocks supporting it.
 **Lemma 2.1.** Suppose $(G, Q, B)$ is statistically homogeneous in
 the sense of Definition 0.6. Then for any node $v$ in the support
 of $Q$,
-$$
-\#\text{paths}(v; N, M; B) \approx D^N \cdot (b')^M
-$$
+$$\mathrm{paths}(v; N, M; B) \approx D^N \cdot (b')^M$$
 as $N, M$ range over admissible hop counts within budget $B$,
 where the approximation holds up to a multiplicative
 $(1 + o(1))$ factor. Consequently,
-$$
-\mathbb{E}_{v \sim Q}\bigl[\#\text{paths}(v; N, M; B)\bigr] \approx D^N \cdot (b')^M
-$$
+$$\mathbb{E}_{v \sim Q}\bigl[\mathrm{paths}(v; N, M; B)\bigr] \approx D^N \cdot (b')^M$$
 with the same asymptotic factor.
 
 **Proof.** Condition (H2) of Definition 0.6 fixes the asymptotic
@@ -478,19 +447,15 @@ $D \approx 7.34$ and $b' \approx 11$.
 
 **Proposition 2.2.** Under Definition 0.6 homogeneity with
 calibration $b_{\text{eff}} \cdot D = b'$,
-$$
-w(p) \cdot \#\text{paths}\bigl(\cdot;\ N(p), M(p);\ B\bigr) \approx 1
-$$
+$$w(p) \cdot \mathrm{paths}\bigl(\cdot;\ N(p), M(p);\ B\bigr) \approx 1$$
 for all paths $p$, i.e. the weight formula is precisely a
 path-count normaliser.
 
 **Proof.** Direct substitution:
-$$
-w(p) = D^{-N(p)} (b_{\text{eff}} \cdot D)^{-M(p)}
-     = D^{-N(p)} (b')^{-M(p)}
-$$
+$$w(p) = D^{-N(p)} (b_{\text{eff}} \cdot D)^{-M(p)}
+     = D^{-N(p)} (b')^{-M(p)}$$
 By Lemma 2.1,
-$\#\text{paths}(N, M) \approx D^N (b')^M$, so the product is
+$\mathrm{paths}(N, M) \approx D^N (b')^M$, so the product is
 $1$ up to the $(1+o(1))$ factor. $\square$
 
 This is the formal statement of design note §5.6's "weights as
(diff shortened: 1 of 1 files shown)


s243a (commented):
## Review of PR #2680: LaTeX fixes + Bayesian interpretation

Two clearly separable commits. I review each in turn.

---

## Commit 1: LaTeX rendering fixes

### ✅ Both fix classes are correct

**`\#` → `\mathrm{paths}` replacements.** All eight `\#\text{paths}` instances and the `\#\mathrm{children}` instance in §0.6 are correctly replaced. The new `\mathrm{paths}(v; N, M; B)` notation is consistent throughout the document. The transition ratio formula gets `\mathrm{paths_{total}}(B, cc)` which is a reasonable shorthand.

**`$$\n formula \n$$` → `$$formula$$` whitespace joins.** All 25 display math blocks are correctly tightened. Multi-line formulas (e.g. the `d_wPow` definition, the `b_eff` equality chain, the Theorem 2.3 proof steps) keep their internal newlines but have `$$` joined to first and last lines. This is the correct GitHub MathJax fix.

**H1 fix in §0.6.** `\mathbb{E}[\#\mathrm{children}(v)]` → `\mathbb{E}[|\mathrm{children}(v)|]` is correct: the cardinality of the children set is `|\mathrm{children}(v)|`, not the count-symbol `#`.

### ⚠️ One minor inconsistency introduced by the `\mathrm{paths}` change

The Proposition 2.2 statement now reads:
```
w(p) \cdot \mathrm{paths}(\cdot;\ N(p), M(p);\ B) \approx 1
```
but the proof a few lines later has:
```
\mathrm{paths}(N, M) \approx D^N (b')^M
```
The first form uses the full argument list `(\cdot;\ N(p), M(p);\ B)` (dot for the node, explicit N, M, budget) while the proof uses the abbreviated `(N, M)`. This is a minor notational inconsistency introduced when the `\#` was stripped — the original had `\#\text{paths}(N, M)` in the proof, which was already abbreviated from the definition. The abbreviated form is fine, but either both should use the full signature or both should use the abbreviated form. Suggest changing the proof line to `\mathrm{paths}(\cdot;\ N, M;\ B)` for consistency with the statement.

---

## Commit 2: New content — §3.6.1 and §5.6

### §3.6.1 Bayesian interpretation of the routing correction

**The Bayesian framing is correct and adds genuine value.** The three-part structure (without ρ = uniform prior over path shapes; with ρ < 1 = prior shift toward child-informative; TLI data refutes that shift) is logically tight. The two consequences (maximum-entropy interpretation; convergence-ratio direction) are both correct and were implicit in the prior Conjecture 3.6 discussion but never stated in these terms.

**The `w(p) \propto 1/\mathrm{paths}(N, M)` claim deserves a citation.**

The "without ρ" section says:
> `w(p) \propto 1/\mathrm{paths}(N, M)` encodes a *uniform prior over path shapes*

This is exactly Proposition 2.2 (`w(p) \cdot \mathrm{paths}(N, M) \approx 1` under exact calibration). Since §3.6.1 is building Bayesian interpretation on this claim, it should cite the proposition: add "(cf. Proposition 2.2)" at the end of the claim. Without this, a reader could mistake the proportionality for a definition rather than a proven consequence of calibration.

**The "prior on branching" restatement has a sign issue.**

The text says:
> `\rho < 1` shifts the prior on the expected per-child-hop branching *down* from `b_{eff} \cdot D` to `b_{eff} \cdot \rho \cdot D`.

This is stated correctly — the effective BranchRatio decreases. The next sentence says: "Less branching expected means each branch carries more information." This is the correct Bayesian reading: if you expect fewer branches, finding a child path is more informative (higher likelihood ratio). The logic is right, but the semantic jump may be opaque to a reader not thinking in information-theoretic terms. Adding "(higher likelihood ratio per child-hop observed)" would close the gap.

**The `w(p) \propto 1/(\mathrm{paths}(N, M) \cdot \rho^M)` formula is subtly off.**

With routing correction `\rho`, the weight is:
```
w(p) = D^{-N} \cdot (b_{eff} \cdot \rho \cdot D)^{-M}
     = D^{-N} \cdot (b_{eff} \cdot D)^{-M} \cdot \rho^{-M}
```
So `w(p) = w_{no-\rho}(p) \cdot \rho^{-M}`. The normalised prior probability of path `p` is proportional to `w(p)`, which is `w_{no-\rho}(p) \cdot (1/\rho)^M`. Since `w_{no-\rho}(p) \propto 1/\mathrm{paths}(N, M)` (Proposition 2.2), we get `w_{with-\rho}(p) \propto (1/\rho)^M / \mathrm{paths}(N, M)`. The document writes `1/(\mathrm{paths}(N, M) \cdot \rho^M)` which equals `\rho^{-M} / \mathrm{paths}(N, M)` = `(1/\rho)^M / \mathrm{paths}(N, M)`. So the formula is correct; the only improvement would be to write it as `(1/\rho)^M / \mathrm{paths}(N, M)` to make the upweighting factor more visually obvious.

**The "Refined Conjecture 3.6 (Bayesian form)" is a restatement, not an independent conjecture.**

Conjecture 3.6 already contains the directionality argument (added in the previous PR). The Bayesian form is a *framing* of the same conjecture, not a new claim. The paragraph should say "Bayesian restatement of Conjecture 3.6:" rather than "Refined Conjecture 3.6 (Bayesian form)." Calling it a "refined" conjecture implies it adds new content to the falsifiable claim, which it doesn't — it adds interpretive framing.

**Task #14 design implication is excellent.** The per-pair `d_wPow` reporting suggestion ("aggregate TLI passes under both, per-pair distributions indistinguishable" vs "per-pair values differ materially") is a concrete and useful experimental design upgrade. This is the clearest actionable addition in the whole PR.

---

### §5.6 Regime classification

**The regime table is useful but the `b'` and `b_eff` scaling entries need clarification.**

For the "tree-like / tame" regime (`\gamma > 3`): the table says `b' \approx D` and `b_{eff} \approx D / K_p`. The quantity `K_p` is not defined anywhere in the document. It appears to mean the average parent degree or some parent-direction normalisation factor. Since `K_p` is introduced only in this table with no definition, a reader cannot evaluate the `r \approx K_p / D` entry. Either define `K_p` inline ("where `K_p = \mathbb{E}[d_p]` is the mean parent degree") or replace with the concrete simplewiki values.

**The "Wikipedia-like" row (`b' \sim 10 \cdot D`) is inconsistent with known calibration.**

The calibrated values from design note §4.5 are `b' \approx 11` and `D \approx 7.34`. So `b' / D \approx 1.5`, not `\sim 10`. The row should read `b' \sim 1.5 \cdot D` or just `b' \approx 11` for the Wikipedia-like regime. The `\sim 10 \cdot D` figure looks like a typo or unit error.

**The tree-like limit calculation is correct but the conclusion needs one more step.**

The text derives `r = 1/D` in the tree-like limit and says "the buffer condition becomes `\rho > 1/D` — easily satisfied." This is right for the tree-limit. But the table claims the tree-like regime has `r \approx K_p / D` (not `1/D`). These should be consistent: if `K_p = 1` (one parent per node, the pure tree case), then `K_p / D = 1/D`, recovering the tree-limit. If `K_p > 1` (multiple parents, as in multi-parent DAGs), `r > 1/D`, which tightens the buffer. The relationship between the tree-limit analysis and the table row should be made explicit.

**The hand-waving heuristic section is correctly labelled but the `b'` divergence argument needs care.**

The claim that in the ultra-small-world regime `b'` grows faster than `b_eff` because "path count is mediated by hub traversal (exponential in hub count), while `b_eff` is a size-biased moment ratio (polynomial in degree statistics)" is a reasonable heuristic, but:
- "Exponential in hub count" is imprecise; the path count growth depends on how many hubs are reachable within budget `B`, which is sub-exponential for fixed `B`.
- The claim that the race favours `b'` is plausible but not derived. For a fixed finite `B`, the hub-mediated path explosion is bounded by `B`.

This is fine as a heuristic — the section is appropriately labelled "hand-waving" — but the open question at the end ("What is `\gamma_*`?") would benefit from noting that the answer may depend on whether `B` is fixed or grows with `N`. If `B` is fixed (as in the kernel), the divergence is bounded differently than if `B \to \infty`.

**The cross-reference between §5.6 and Conjecture 3.6 / §3.6.1 is correct.** The "empirical Wikipedia case" paragraph correctly notes that global `\gamma \approx 2.41` is dangerous but the topical core is likely in the `\gamma > 3` safe regime. The Task #15 reference is appropriate.

---

## Verdict

**Approve with two suggestions before merge:**

1. **§5.6 table: define `K_p` or replace with concrete values, and fix `b' \sim 10 \cdot D` → `b' \sim 1.5 \cdot D` for the Wikipedia-like row.** The undefined `K_p` blocks a reader from evaluating the table, and the `10 \cdot D` figure contradicts the calibrated `b' \approx 11`, `D \approx 7.34` values from design note §4.5.

2. **§3.6.1: Relabel "Refined Conjecture 3.6 (Bayesian form)" to "Bayesian restatement of Conjecture 3.6".** The Bayesian framing is interpretive, not a new falsifiable claim. Calling it a "refined conjecture" overstates what it adds.

All other items (Proposition 2.2 cross-reference, `w(p)` formula rewrite, `K_p / D` consistency, `B`-dependence note) are follow-up polish.

s243a (commented):
## Independent second review of PR #2680

Fresh read of both commits. I'll be brief on the LaTeX fixes (mechanical, straightforward) and focus depth on the new content.

---

## Commit 1: LaTeX rendering fixes

### Correctness: all good

All replacements are correct. No issues with the whitespace joins or the `\mathrm{paths}` substitutions. One observation not in the prior review:

**The `\mathrm{paths_{total}}` macro in the transition-ratio formula is non-standard.** LaTeX subscripts inside `\mathrm{}` render differently across engines. GitHub MathJax renders `\mathrm{paths_{total}}` as `paths` with a subscript `total` in roman, which is fine visually but inconsistent with how subscripts are handled elsewhere in the doc (e.g. `d_{\text{wPow}}` uses `\text{}` for subscripts, not `\mathrm{}`). Suggest either `\mathrm{paths}_{\mathrm{total}}` or introduce a prose shorthand instead of the macro.

---

## Commit 2: New content

### §3.6.1 Bayesian interpretation

**Structural assessment.** The section does one thing well and one thing less well.

What it does well: the three-step argument (uniform prior → ρ shifts prior toward child-informative → TLI data refutes that shift) is clean. The Task #14 per-pair reporting suggestion is a concrete experimental upgrade. The maximum-entropy framing of Proposition 2.2 is a legitimate reinterpretation.

What it does less well: **the Bayesian language is applied asymmetrically.** The section argues that ρ < 1 encodes a prior that is "contradicted by the data" (TLI ≈ 0.02%). But this argument would equally apply to *any* deviation from exact calibration `b_eff · D = b'`. The design note acknowledges that empirical calibration gives `b_eff · D ≈ 70` while `b' ≈ 11` — a factor of ~6 discrepancy. Under the same Bayesian logic, the no-ρ metric also encodes a prior that is "off" from exact path-count normalisation (Proposition 2.2 requires exact calibration; the 15% caveat of Proposition 2.2 is the calibration error). The critique of ρ as encoding a "wrong prior" applies to the entire approximate calibration framework, not specifically to ρ. The section should acknowledge this: *"This critique applies to the full calibration, not ρ alone — ρ is one of several sources of prior mismatch. The specific case against ρ is that it pushes the prior in the wrong direction relative to data, whereas the b_eff approximation error pushes it in an uncontrolled direction."*

**The "prior on paths" restatement is the stronger argument; the "prior on branching" restatement adds confusion.** Both are stated as "equivalent restatements" but they have different implications:
- "Prior on branching" is about the effective branching ratio used in calibration.
- "Prior on paths" is about the per-path likelihood ratio.

These are equivalent mathematically but carry different intuitions. A reader may try to use both simultaneously and get confused about what the prior is defined *over* (branching rates? path probabilities?). Suggest keeping only the "prior on paths" framing (it's the more directly Bayesian one) and moving the "prior on branching" to a parenthetical or footnote.

**The "maximum-entropy" claim needs qualification.** The section says the no-ρ metric "satisfies Proposition 2.2: `w(p) · paths(N,M) ≈ 1`, the prior is the maximum-entropy choice given path-count structure." Maximum entropy *over what*? The claim is that the weight function which assigns equal prior probability to each path (i.e. uniform over all paths regardless of shape) is the maximum-entropy choice. This is true — it maximises entropy over the discrete distribution of paths. But it requires saying explicitly: "maximum entropy over the distribution over paths, conditioning on no information beyond path existence." Without this qualification, a reader might interpret "maximum entropy" as referring to the graph entropy or the degree distribution.

### §5.6 Regime classification

**The `K_p` problem (also flagged in prior review) is a blocker.** `K_p` appears in three cells of the table without definition. It's used to express `r ≈ K_p / D` in the tree-like regime. From context, `K_p` appears to be the mean parent degree `E[d_p]` — but in the pure tree limit described in the preceding paragraph, each node has exactly 1 parent, so `K_p = 1` and `r = 1/D`. If `K_p = E[d_p]` in general, the table is saying `r ≈ E[d_p] / D` in the tree-like regime, which would mean `r ≈ E[d_p] / E[d_c]`. This is *not* the convergence ratio formula `r = b' / (b_eff · D)` except in a very specific regime. The table needs this explained or `K_p` replaced.

**The `b' ∼ 10 · D` entry for the Wikipedia-like regime appears to be a factor-of-7 error.** Empirical: `b' ≈ 11`, `D ≈ 7.34`, so `b'/D ≈ 1.5`. The table claims `b' ∼ 10 · D ≈ 73`. This is flagged by the prior review as well. Either fix to `b' ∼ 1.5 · D` or to the absolute value `b' ≈ 11`.

**The regime table's `r` behaviour column is inconsistent with the body text.** The body derives `r = 1/D` in the tree limit. The table has `r ≈ K_p / D` for the tree-like regime. For K_p = 1 these agree. But the table also has `r ∼ 0.1–0.2` for the Wikipedia-like regime — the simplewiki value is `r ≈ 0.157`, which is in this range and consistent. The issue is that the table implies the Wikipedia-like row is a *different regime* from the tree-like row, when in fact simplewiki (with `γ ≈ 2.41` globally but `γ > 3` topically) straddles the boundary. The table should note whether the Wikipedia-like row describes the *global* graph or the *topical core* — these have very different `r` values (topical: 0.157; global: would be near 1 under any reasonable calibration given `b_eff = 589` globally vs `b' ≈ 11`).

**New observation not in prior review: the buffer condition `ρ > r` is stated without proof that this is the operative condition.**

The section frames everything around `ρ > r = b'/(b_eff · D)`. But Conjecture 3.6 says to *drop* ρ entirely, not to require `ρ > r`. The buffer condition `ρ > r` seems to be asking: "under what conditions does including ρ still keep `b_eff · ρ · D > b'`?" i.e. `ρ > b'/(b_eff · D) = r`. This is the condition under which the routing correction doesn't break convergence. This is a different question from whether ρ is theoretically justified. The section conflates two questions:
1. *When does ρ break convergence?* (Answer: when ρ ≤ r.)
2. *When should ρ be dropped?* (Answer: always, under homogeneity, per Conjecture 3.6.)

The section title "when does ρ exceed D's buffer?" suggests question 1, but the content also addresses question 2. These should be separated or the section retitled to "convergence robustness under ρ miscalibration" to clarify it's asking question 1 only.

---

## Verdict

**Approve with three pre-merge fixes:**

1. **§5.6 table: define `K_p` and fix `b' ∼ 10 · D` → `b' ∼ 1.5 · D`.** (Also flagged by prior review — must fix before merge.)

2. **§5.6: clarify whether the Wikipedia-like row describes the global graph or the topical core**, and add a note that the buffer condition (`ρ > r`) is a *convergence robustness* question distinct from the *should we drop ρ* question of Conjecture 3.6.

3. **§3.6.1: relabel "Refined Conjecture 3.6 (Bayesian form)" as "Bayesian restatement"** and add a sentence acknowledging that the maximum-entropy / prior-mismatch critique applies to the full approximate calibration, not ρ alone. (Also flagged by prior review.)

Everything else is polish.

s243a (commented):
## Re-review after second update

Tracking resolved items and new content. The update adds §3.5 preamble, revised §3.6.1, expanded §5.6 (retitled), and two new sections §5.7 and §5.8.

---

## ✅ Resolved

- **`\mathrm{paths_{total}}` macro**: Replaced with `T(B, cc)` with explicit definition. Clean fix.
- **Proposition 2.2 argument form**: Statement now uses `\mathrm{paths}(v;\ N(p), M(p);\ B)` and the proof uses `\mathrm{paths}(v;\ N, M;\ B)`. Consistent.
- **§3.6.1 section label**: Now "Bayesian restatement of Conjecture 3.6" with explicit upfront notice: "interpretive restatement, not an independent claim."
- **Asymmetric Bayesian critique**: The new §3.6.1 now explicitly addresses the symmetric observation: the no-ρ metric also has prior mismatch (~15%), and the distinction (ρ's mismatch is order-unity and directionally wrong; no-ρ mismatch is small and uncontrolled-but-bounded) is stated clearly. This is exactly the right framing.
- **`b' ∼ 10·D` error**: Fixed in the table. Now says `b' ≈ 1.5·D` (empirical) with explicit note `(b'/D ≈ 1.5, not the rough ~10 I had in an earlier draft)`.
- **`K_p` definition**: Now defined explicitly at the top of §5.6: `K_p := E[d_p]` is the mean parent in-degree.
- **§5.6 title and framing**: Retitled "Convergence robustness under ρ miscalibration" with explicit (i)/(ii) separation of the two questions. The prior version conflated them; this version separates them correctly.
- **Global vs. topical clarity in regime table**: Wikipedia topical and Wikipedia global are now separate rows with correct note that global is "n/a — Definition 0.6 violated."
- **Maximum-entropy qualifier**: §3.6.1 now says "maximum-entropy prior over path shapes" with explicit parenthetical: "among all priors with the constraint 'give each (N,M) stratum equal total mass after accounting for path counts,' the uniform-per-shape prior maximises entropy."

---

## New content review

### §3.5 preamble ("What TLI measures, precisely")

Well-placed. The clarification that TLI measures the contribution of *shorter-via-child paths* to the metric value, and that low TLI means either such paths don't exist or the weighting crushes them, is genuinely useful before the falsification conjecture. The simplewiki reference grounds it.

One small issue: the phrase "*either* such shorter-via-child paths don't exist, *or* the metric weighting crushes them to negligibility" suggests these are alternatives. But the simplewiki case has both: child paths do exist, but the weighting also crushes them. The "or" should be "and/or" — in practice we want to know which of the two mechanisms is dominant. A sentence noting this distinction would sharpen the preamble.

### §5.7 Weighting as expected-average over admissible paths

Strong addition. The three claims are all correct:

1. `w(p)` is the unique exponential weighting (up to constants) that cancels exponential path-count growth.
2. Under exact calibration, the metric is the expected `(h+1)^{-n}` over the distribution treating each path-shape class as equally likely.
3. Without convergence, increasing `B` keeps changing the metric value; with convergence, it saturates.

**Minor issue in the formal consequence paragraph.** The text says: "The double sum `\sum_{N,M} 1 · (h+1)^{-n} = \sum_{h ≥ h_min} (shape count at h) · (h+1)^{-n}` makes `d_wPow` the expected `(h+1)^{-n}` over the distribution that puts equal probability on each accessible (N,M) shape."

This is correct under exact calibration. But it should be stated that this uses Proposition 2.2's approximation — which requires `b_eff · D = b'`. Under the realistic 15% gap, the shape contributions are not exactly 1 but `≈ 1`. A parenthetical "(under exact calibration; ≈ holds up to the ~15% calibration gap per Proposition 2.2's caveat)" would make the approximation chain explicit.

**The "unique exponential weighting" claim is strong.** The text says `D^{-N}(b_eff·D)^{-M}` is "the unique exponential weighting (up to constants) that cancels the exponential growth." This is essentially true — any weighting of the form `α^N β^M` that makes `w(p)·#paths(N,M)` constant requires `α = D^{-1}` and `β = (b'·D)^{-1}` (from Lemma 2.1's path count estimate). But the qualifier "up to constants" is doing some work: a global scaling of all weights `w(p)` doesn't change `d_wPow` (since it's a ratio), so the claim should say "unique up to global scaling." The current phrasing could suggest uniqueness is stronger than it is.

### §5.8 Convergence on effectively-infinite graphs

Useful typology. The three categories (materially finite / effectively infinite / materially infinite) and the observation that "most theory targets the middle category" are well-framed.

**The `B^*` saturation claim needs a caveat.** The text says: "There exists a finite `B^*` such that `d_wPow(B) ≈ d_wPow(B^*)` for all `B ≥ B^*` within practical precision." This is a consequence of Theorem 2.3 (the geometric series bound means the M ≥ 1 contribution is bounded), but the saturation is in the *weighted contribution*, not the direct metric value. The claim that there exists a single `B^*` that works for all nodes in `supp(Q)` requires the convergence to be *uniform* over `Q`. Under homogeneity (Definition 0.6), this follows, but the text should say so: "Under homogeneity (Definition 0.6), there exists..." rather than asserting it unconditionally.

**The boolean-equivalent expression graph example is excellent.** It's a natural test case for §5.7's effectively-infinite graph motivation and §3.5's symmetric DAG point. The observation that boolean transformations are near-bidirectional (`b_eff ≈ 1`, convergence fails) makes it a clean falsification test for Conjecture 3.5. The deferral note is appropriate.

---

## Residual from prior passes

**One item still not resolved:** the §5.6 tree-like row still has `b_eff ≈ D / K_p` with `K_p` defined as `E[d_p]`. Under pure-tree conditions (1 parent per node), `K_p = 1` and `b_eff = D`. But the `b_eff` formula is `(E[d_c^2]/E[d_c]) / (E[d_p^2]/E[d_p])`, not `D / K_p`. For a pure tree with exactly `D` children per node and 1 parent: `E[d_c] = D`, `E[d_c^2] = D^2`, so `d̃_c = D^2/D = D`. Similarly `E[d_p] = 1`, `E[d_p^2] = 1`, `d̃_p = 1`. So `b_eff = D / 1 = D`. And `D / K_p = D / 1 = D`. These match. Good. But the formula `b_eff ≈ D / K_p` generalises incorrectly to non-tree cases: for a graph with `K_p` parents per node (all equal), `d̃_p = K_p^2/K_p = K_p`, and `d̃_c = D^2/D = D`, giving `b_eff = D / K_p`. This is correct for the *regular* case (all nodes have exactly `D` children and `K_p` parents). For the actual Wikipedia case with heavy-tailed degree distributions, the size-biased moments diverge from mean degrees, and `b_eff ≠ D / K_p`. The table row should note it's the *regular-graph approximation* of `b_eff`, not the general formula.

---

## Verdict

**Approve — ready to merge.** All pre-merge issues from prior reviews are resolved. The new sections (§5.7, §5.8) add genuine value: §5.7 gives the clearest explanation in the document of *why* the weight formula is what it is, and §5.8 grounds the theory in the practical case the kernel is designed for.

The three items above (§3.5 "and/or" phrasing, §5.7 exact-calibration qualifier, §5.8 uniformity caveat) are polish for a follow-up commit. The `b_eff ≈ D/K_p` item is the only one with a mild correctness flavour but is correctly scoped to the tree-like row.

s243a (commented):
## Independent second review — updated PR #2680

Fresh read of the full diff. The update adds §3.5 preamble, revised §3.6.1, expanded §5.6 (retitled), §5.7, and §5.8. I'll focus where I see things not covered or differently assessed.

---

## §3.5 preamble — "What TLI measures, precisely"

The preamble is well-placed and the core explanation is correct. One thing it omits that would strengthen it:

**The preamble conflates two different ways TLI can be low.** It says: "Low TLI means: *either* such shorter-via-child paths don't exist, *or* the metric weighting crushes them to negligibility."

These two cases have different implications for the rest of the theory:
- If shorter-via-child paths don't exist, TLI = 0 exactly (like the tree case, §1.6).
- If they exist but are crushed, TLI > 0 but small — this is the simplewiki case.

Conjecture 3.5 (symmetric DAG falsification) is relevant only to the second case: a symmetric DAG by construction *has* many child-path shortcuts, so the question is whether weighting can crush them. The preamble should note this distinction: "On simplewiki, child paths exist but are crushed (TLI ≈ 0.02% > 0). Conjecture 3.5 tests whether weighting can ever crush them in the symmetric case."

---

## §3.6.1 Bayesian restatement — qualified critique

The addition of the symmetric observation (no-ρ also has prior mismatch, ~15%) is good and directly addresses the prior review concern. The distinction between "order-unity and directionally wrong" vs "small and uncontrolled-but-bounded" is the right framing.

**One new concern: the 2.6× per-child-hop figure is not derived in the text.**

The section says ρ's mismatch is "order-unity (factor 2.6× per child hop on simplewiki)." Where does 2.6× come from? Presumably: `1/ρ = 1/0.384 ≈ 2.6`. That is correct — each child hop is upweighted by `(1/ρ)^1 ≈ 2.6` relative to the no-ρ metric. But this isn't stated. A reader who doesn't know `ρ = 0.384` from the design note cannot verify this figure. Add one sentence: "With `ρ = 0.384` (design note §4.5), each child-hop path is upweighted by factor `1/ρ ≈ 2.6` relative to the no-ρ metric."

**The "prior on paths" vs "prior on branching" consolidation is still present but confusing.**

The prior reviews flagged that having two "equivalent restatements" risks confusing readers about what the prior is defined over. This update keeps both but merges them into a single paragraph about ρ < 1 deflating expected branching and upweighting individual paths. This is better than the original but still reads as two separate claims. Specifically:
- "ρ deflates expected per-child-hop branching from `b_eff·D` to `b_eff·ρ·D`" — this is a statement about calibration.
- "each individual M-child-hop path is a priori more likely to be informative, by factor `(1/ρ)^M`" — this is the path-prior statement.

These are equivalent given Proposition 2.2 but require that proposition to connect them. The text says "read as a prior shift: ρ deflates..." but the mechanical connection is still implicit. One sentence bridging them would complete the argument.

---

## §5.6 Convergence robustness — regime table

**The `b_eff ≈ D/K_p` entry is the regular-graph approximation only.**

For the tree-like row, `b_eff ≈ D/K_p` (with `K_p = E[d_p]`) is derived from the regular-graph case where all degree second moments equal degree means squared. For heavy-tailed degree distributions, `E[d_c^2]/E[d_c] ≫ E[d_c]`, so `b_eff ≫ D/K_p`. The table entry should note this is a regular-graph lower bound on `b_eff`, not the general formula. This is flagged by the prior review as well — it is unresolved.

**The Wikipedia topical row is appropriately empirical but the γ entry "~3 (TBD, task #15)" deserves more context.**

γ ≈ 3 is the boundary between the Chung-Lu small-world and Cohen-Havlin ultra-small-world regimes. The row says "TBD, task #15" which is honest, but a reader unfamiliar with the significance of γ = 3 as a phase boundary won't understand why this is listed as "Wikipedia-like" and not "ultra-small-world." A brief note: "γ = 3 is the boundary below which Cohen-Havlin ultra-small-world behaviour emerges; the topical core is conjectured to sit at or above this boundary" would ground the table row.

**The ultra-small-world row's "race; can → 1" r-behaviour needs a concrete example.**

The row says r "can → 1" in the ultra-small-world regime without any indication of when it does vs. doesn't. For γ just below 3 (near-boundary), r should be close to but below 1 (close to failure but still converging). For γ near 2 (near degenerate), r → 1 strongly. A note like "r → 1 as γ → 2" would give the table a concrete anchor.

---

## §5.7 Weighting as expected-average

This is the strongest new addition. The three-part argument (path-count cancellation → equal-shape contribution → expected metric interpretation) is correct and clearly presented. The tree-case illustration of why weighting is *necessary* (unweighted sum diverges) is the best intuitive explanation of the weight formula in the entire document.

**One concern: §5.7's content partially duplicates §3.6.1.**

Both sections derive the interpretation of `w(p)` as encoding a prior/expectation over paths. §3.6.1 uses Bayesian language; §5.7 uses expectation language. They reach the same conclusion (maximum-entropy prior over shape classes = path-count normaliser). The duplication is intentional by design ("target different reader entry points"), as noted in the PR description. This is fine — but the two sections should explicitly cross-reference each other to avoid a reader thinking one is more authoritative than the other. Currently §3.6.1 references Proposition 2.2 but not §5.7, and §5.7 references Proposition 2.2 but not §3.6.1.

**The "unique exponential weighting" claim is slightly overstated.**

The text says the weighting is "unique (up to constants)". The qualifier should be "unique up to global scaling" (as noted in the prior review). Global scaling doesn't affect `d_wPow` since it's a ratio, so any `λ · D^{-N} (b_eff·D)^{-M}` gives the same metric. The current phrasing could mislead a reader into thinking `D` and `b_eff·D` are uniquely determined by the uniqueness argument; they are not — they're determined by the path-count factorisation in Lemma 2.1, which is separate from the uniqueness of the exponential form.

**The alternative-graph-regimes paragraph (sub-exponential path-count growth) adds a useful qualification.** No issues.

---

## §5.8 Convergence on effectively-infinite graphs

**The category table is well-conceived.** The three-way split (materially finite / effectively infinite / materially infinite) cleanly separates the mathematical ideal from the computational reality.

**The `B*` saturation claim needs the uniformity caveat.** The prior review flagged this: "there exists a finite `B*` such that..." requires uniform convergence over `supp(Q)`, which follows from homogeneity but should be stated. This is not yet resolved.

**The boolean expression graph example is genuinely illuminating.** Bidirectional transformations → b_eff ≈ 1 → convergence fails → clean Conjecture 3.5 test. One additional observation worth noting: in expression graphs, the "root" (the simplest or canonical form) may not be unique, and the parent/child decomposition (simplification vs. expansion) maps cleanly to the directed graph structure. This makes expression graphs not just a test case for Conjecture 3.5 but a potential *application domain* for the kernel once Conjecture 3.5 is resolved. Worth one sentence to plant this seed.

---

## Overall verdict

**Approve — ready to merge.** All prior pre-merge blockers are resolved. The new sections add real value. The items above are follow-up polish:

- §3.5: "and/or" phrasing + note that Conjecture 3.5 tests the crushing case specifically
- §3.6.1: derive the 2.6× figure explicitly; add one-sentence bridge between branching-deflation and path-upweighting interpretations
- §5.6 table: add γ = 3 boundary note; note that `b_eff ≈ D/K_p` is the regular-graph approximation; add "r → 1 as γ → 2" to ultra-small-world row
- §5.7: add cross-reference to §3.6.1; qualify "unique up to global scaling"
- §5.8: scope `B*` claim to homogeneous case; add one sentence on expression graphs as potential application domain

PR #2703, opened 2026-06-03
fix(wam-scala): LMDB cursor-scan reflection; verify LMDB-backed kernels

s243a:
## Summary

Validates the **capstone** integration — a graph kernel reading its edge relation directly from **LMDB** — which ties together the two flagship Scala features (`kernel_dispatch` + LMDB materialisation). Running it surfaced and fixed a latent reflection bug.

## Bug fixed

`LmdbFactSource.scanCursorAll` / `cursorSeekAndCollectDupSort` looked up the lmdbjava cursor accessor via `getMethod("`val`")` — the Scala keyword-escaping **backticks were embedded in the reflective name string**, so the lookup for the method actually named `val` threw `NoSuchMethodException` at runtime. The earlier LMDB tests only used ground-key lookups (`Dbi.get`) and **never the cursor scan**, so it never fired. A kernel reads its adjacency by enumerating the *whole* edge relation via `collectBinarySolutions → streamAll` (cursor scan), which hits this path. Fixed to `getMethod("val")`.

## Verified

A `transitive_closure2` kernel (`kernel_dispatch(true)`) whose `edge/2` relation is an `lmdb(...)` fact source reads its edges directly from LMDB and computes reachability correctly:

```
tc(a,b)=true  tc(a,e)=true  tc(c,e)=true  tc(e,a)=false  tc(a,x)=false
```

New gated `lmdb_backed_kernel` test in `tests/test_wam_scala_lmdb_runtime_smoke.pl` seeds an LMDB chain `a→b→c→d→e` and checks `tc` queries — and it covers the `streamAll` cursor-scan path the key-lookup tests miss. All 3 LMDB tests pass (lmdbjava 0.9.0 / OpenJDK 21). Non-LMDB projects still compile + run without lmdbjava on the classpath; generator structural suite (18) green.

## Why this matters

This is the intended **large-graph** path: native BFS/Dijkstra/A\* kernels running over edges materialised in LMDB (>100k facts), with no in-memory fact compilation. The codegen composes cleanly — `collectBinarySolutions` runs whatever backs the edge predicate, so kernels work over WAM facts, inline/CSV sources, *and* LMDB alike.

Docs gain an **LMDB-backed graph kernels** section; CHANGELOG updated.

https://claude.ai/code/session_013gLfigNMgCkxybbp7m6fXw

---
_Generated by [Claude Code](https://claude.ai/code/session_013gLfigNMgCkxybbp7m6fXw)_


s243a:
Updated during stale-PR triage (2026-06-10): merged current `main` into this branch (was 597 commits behind).

- Resolved the one conflict (`CHANGELOG.md` — both sides added entries under `### Added`; kept both).
- Verified the fix is still needed: `main` still has the broken `getMethod("` + backtick + `val` + backtick + `")` reflective lookups at `templates/targets/scala_wam/runtime.scala.mustache` lines 400 and 418.
- Re-ran the gated LMDB suite post-merge with Scala 2.13.16 / OpenJDK 21 / lmdbjava 0.9.0: **all 3 tests pass**, including the new `lmdb_backed_kernel` capstone test.

Still relevant and ready for review.

---

PR #2929, opened 2026-06-08
LLVM WAM: ==/2 + neq/2 deep deref correctness fix (M137)

s243a:
## Summary

**Correctness fix** for the `==/2` and `\==/2` recursion bug uncovered while writing the M136 `atom_split` tests. `[a, b] == [a, b]` was returning **false** for two literal lists.

## Root cause

`@value_equals` (in `templates/llvm_wam/value.ll.mustache`) only derefs the top-level operands — its comment notes *"callers do that"*. But its recursive arg loop loads `%Value` verbatim from the args array and passes it through to the recursive call **without dereferencing**.

When `put_structure` builds a nested compound — and the inner cons cell ends up stored as a `Ref` to a heap slot rather than directly inline — the recursive compare sees `Ref(payload=heap0)` on one side and `Ref(payload=heap1)` on the other. Both have tag 5, so `same_tag` branches to `shallow_payload` which compares heap addresses, finds them different, returns false. The compare bails even though the dereferenced values would be identical.

## Symptoms

| Expression | Before | Expected |
|---|---|---|
| `[a, b] == [a, b]` | false | true |
| `[a, b, c] == [a, b, c]` | false | true |
| `foo(bar(1), baz(qux)) == self` | false | true |
| `[a] == [a]` | true ✓ | true |
| `foo == foo` | true ✓ | true |
| Element-wise `A == a, B == b` | true ✓ | true |

The 1-level cases all worked because their args were inline atoms, not Refs to compound heap cells.

## Fix

Adds `@wam_strict_eq(%vm, %a, %b)` helper next to `@wam_collect_vars` / `@wam_is_ground`. Same structure as `@value_equals` **but**:

- Derefs both operands on every entry (including recursion).
- Passes raw args through to the recursive call, which derefs again.
- Otherwise identical control flow.

`builtin_eq` (`==/2`, op 7) and `builtin_neq` (`\==/2`, op 21) switched from `@value_equals` to `@wam_strict_eq`. Other callers (sort dedup, `is/2` result check, etc.) keep using `@value_equals` — their inputs are already dereferenced and don't traverse into compound args, so the shallow version is **correct AND faster** (no per-arg deref).

Prefix `seq.` for the helper's SSA values (no collisions).

## Tests

10 execution tests, all PASS:

| Test | What |
|---|---|
| `seq_lit_two` | `[a, b] == [a, b]` — the original regression |
| `seq_lit_three` | `[a, b, c] == [a, b, c]` |
| `seq_split_lit` | M136 `atom_split` result `== [a, b, c]` (was `=` workaround) |
| `seq_nested_compound` | `foo(bar(1), baz(qux))` self-equality |
| `seq_differ_at_depth` | Args differ at depth → `==` rejects |
| `sneq_lit_two` | `[a, b] \== [a, c]` succeeds |
| `sneq_equal_rejects` | `[a, b, c] \== [a, b, c]` rejects (equal terms) |
| `seq_singleton_still_works` | `[a] == [a]` — no shallow-fast-path regression |
| `seq_atom_eq_still_works` | `foo == foo` |
| `seq_int_eq_still_works` | `42 == 42` |

The last three confirm we haven't broken the shallow-fast paths that were already correct.

## Files

- `src/unifyweaver/targets/wam_llvm_target.pl` — `@wam_strict_eq` helper definition (~80 lines IR), `builtin_eq` and `builtin_neq` routed through it.
- `tests/core/test_wam_llvm_builtin_execution.pl` — 10 tests + `test_all` invocations.

## Test plan

- [x] Module loads (smoke) ✓
- [x] All 10 M137 tests PASS individually ✓
- [x] M136 atom_split tests still PASS with both `=` and `==` ✓
- [ ] Full suite regression check (esp. existing `==` tests)

https://claude.ai/code/session_012T8pkULrkmsMf91SSCvzim

---
_Generated by [Claude Code](https://claude.ai/code/session_012T8pkULrkmsMf91SSCvzim)_


s243a (commented):
Reviewed the diff against current `main` (with `main` merged into the branch locally). The fix is correct and well-scoped.

**Correctness analysis:**

- `@wam_strict_eq` is structurally identical to `@value_equals` (verified side-by-side) with exactly one change: `@wam_deref_value` applied to both operands on entry, including every recursive call — raw args are passed through and dereffed at the next level. This is the right shape for the bug: the legacy helper's recursive loop loaded heap `%Value`s verbatim, so `Ref`-stored compound args compared as heap addresses.
- The tag-3-only deep path preserves the legacy semantics for every other tag: atoms/integers payload-compare as before, distinct unbound vars stay unequal (`X == Y` false, `X == X` true via slot-address compare), and Float bit-pattern comparison is unchanged.
- Functor comparison by `i8*` pointer equality is sound — functor string globals are interned per module via `register_functor_string`, and this matches the legacy helper's convention.
- Switching `builtin_eq`/`builtin_neq` from `wam_get_reg_deref` to raw `wam_get_reg` is equivalent since the helper now derefs; no double-deref concern (`wam_deref_value` is idempotent on non-Ref values).
- Keeping the other `@value_equals` callers on the shallow helper is the right perf call, with one caveat below.

**One follow-up worth tracking (not a blocker):** the rationale that other callers' "inputs are already dereferenced" covers only the *top level*. A caller like sort dedup comparing compound elements whose **args** are heap-stored Refs would still hit the original bug class. That latent behavior is unchanged by this PR (strictly no regression), but an audit of `@value_equals` call sites with compound-element inputs would be a good M138 candidate.

**Verification in progress:** the PR's unchecked test-plan box (full-suite regression) is running now against this branch + merged `main` — `test_wam_llvm_builtin_execution`, all 840 exec tests with clang+llc. Currently ~300/840 with zero failures; final results will be posted on this PR when it completes, along with a push of the tested merge commit.

---

s243a:
Full-suite verification complete — **ready to merge**. ✅

- Merged current `main` into this branch (clean, no conflicts) and pushed, so the PR head (`ef71f89f`) is exactly the tested state.
- Ran the complete `test_wam_llvm_builtin_execution` suite against it: **843 PASS, 0 FAIL**, clean exit (clang 18.1.3 + llc 18, Ubuntu 24.04).
- The `--- M137 ==/2 deep deref correctness ---` section: all tests PASS, and the pre-existing `==`/`\==` tests elsewhere in the suite show no regressions — the shallow-fast-path cases (`[a] == [a]`, atom/int equality) and the sort/`is/2` callers that stayed on `@value_equals` are all green.

This checks the last box in the PR's test plan. Review feedback (separate comment) was positive with one non-blocking follow-up suggestion.

---

PR #2978, opened 2026-06-10
LLVM WAM: ISO is_list/1 cons-chain walk + project-root normalization (M141)

s243a:
## Summary

Two LLVM WAM engine fixes, both root-caused from the `is_list` follow-up flagged in #2976. Two commits.

## Fix 1: `is_list/1` never succeeded

`builtin_is_list_check` accepted only the legacy tag-4 List value, but the engine builds every actual list as `[|]/2` compounds with `[]` as an atom — so `is_list/1` failed on **every list the engine produces, including `is_list([])`**. The old comment even promised the `[]` check ("true if list (tag=4) or the empty list atom []"); it was never implemented.

Replaced with a proper ISO cons-chain walk: true at the `[]` atom (legacy tag-4 List still accepted), steps through `[|]/2` tails with full deref, false for atoms, partial lists (`[a|_]`), and unbound variables. The needed infrastructure (`@wam_empty_list_atom_id`, the interned `[|]` functor global) is pre-registered in every module by the M9 machinery, so the walk is self-contained.

## Fix 2: duplicate predicate compilation with unqualified root specs

The full-suite gate for fix 1 caught `test_sum_float` exiting 127 — the test binary was never produced because `llc` rejected the module: `redefinition of global '@node_switch_0'`. The `node/1` predicate was compiled into the merged module **twice**.

Root cause: `expand_wam_llvm_project_predicates` discovers call targets in qualified form (`user:node/1`), but the seen-set `memberchk` compared against the raw root list — and `run_pow_test` passes helper predicates **unqualified** (`[node/1]`), so the explicit root and the discovered dependency both compiled. Reproduced identically on the parent commit: pre-existing, masked by in-suite test ordering until M141's test additions shifted the schedule. Roots are now normalized to `Module:Pred/Arity` (and deduped) before seeding the expansion.

## Tests

6 tests in the new `--- M141 is_list/1 cons-chain walk ---` section: `[a,b,c]` / `[a]` / `[]` accept; `foo` / `[a|_]` (partial) / unbound var reject — per ISO. Fix 2 is covered by the existing `test_sum_float`/`test_sum_int` (unqualified-helper) tests, which now also pass **in isolation**, not just in suite order.

## Test plan

- [x] All 6 M141 tests PASS
- [x] `test_sum_float` passes in isolation (was exit 127)
- [x] Full `test_wam_llvm_builtin_execution` suite: **870 PASS, 0 FAIL** (clang 18.1.3 + llc 18)

https://claude.ai/code/session_01YNMFtZtCH9eMqTn6FToPmM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   deref, false for atoms, partial lists (`[a|_]`), and unbound
   variables. Six tests in the new
   `--- M141 is_list/1 cons-chain walk ---` section.
+- **WAM LLVM target (M141): duplicate predicate compilation with
+  unqualified root specs.** `expand_wam_llvm_project_predicates`
+  discovers call targets in qualified form (`user:node/1`), but the
+  seen-set dedup compared them against the raw root list — an
+  unqualified root (`node/1`) never matched, so the predicate was
+  compiled into the merged module twice and `llc` rejected the
+  duplicate `@<pred>_switch_N` globals (`redefinition of global`).
+  Roots are now normalized to `Module:Pred/Arity` before seeding the
+  expansion. Surfaced by `test_sum_float` failing in isolation;
+  previously masked by in-suite test ordering.
 - **WAM LLVM target (M140): remaining put-instruction bind-throughs**
   — completes the M139 audit. `put_constant`, `put_structure`, and
   `put_list` called `@wam_bind_through_if_unbound_ref(old, val)`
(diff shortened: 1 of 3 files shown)


s243a:
Full-suite verification complete — **ready to merge**. ✅

- Complete `test_wam_llvm_builtin_execution` run: **870 PASS, 0 FAIL**, clean exit (clang 18.1.3 + llc 18).
- The `--- M141 is_list/1 cons-chain walk ---` section: all 6 tests PASS.
- `test_sum_float` (`sum(1/N) -> 0.875`) passes both in-suite and in isolation — the duplicate-compilation fix holds in both orderings, closing the order-dependence that masked it.

That checks the last box in the test plan.

---

PR #3201, opened 2026-06-15
design(wam-rust): frame the cached object as a measure over path length (interface + endpoint care)

s243a:
## Summary

Folds your measure-theoretic framing into the design docs — and yes, your intuition is correct: the general object that supports "weight of an interval" **is a measure** `μ` over the path-length variable. This is the right abstraction because it gives clean **interface hooks** independent of representation, and it handles your delta-function caveat honestly.

## Specification §1 (rewritten)

- **§1 — the object is a measure** `μ`: a discrete histogram is an *atomic* measure (sum of point masses at integer lengths); a continuous approximation is an *absolutely-continuous* measure (density integrated); a single deterministic length is a *Dirac* point mass; mixtures are measures too.
- **§1a — the interface + the endpoint care.** Core read `interval_mass(a, b)` (a *sum* for atomic, an *integral* for continuous). The delta-on-an-endpoint ambiguity is pinned with a **right-continuous CDF** `F(x) = μ((-∞, x])`:
  ```
  interval_mass((a, b]) = F(b) − F(a)          (half-open, canonical)
  atom_mass(x)          = F(x) − F(x⁻)          (point mass AT x)
  interval_mass([a, b]) = F(b) − F(a) + atom_mass(a)
  ```
  So "centered / left-sided / right-sided" point masses are **explicit**, not ambiguous — the interface exposes both `interval_mass` (half-open default) and `atom_mass` so consumers never guess. Hooks: `interval_mass`, `atom_mass`, `total_mass`, `splice`, `truncate`.
- **§1b — backing representations as concrete measures**: discrete histogram / continuous-parametric / cumulative-CDF-basis / Dirac-truncated, with their `interval_mass` and splice costs.

Philosophy §4 reframed to the measure + interface-hooks + endpoint resolution.

## Why this matters

It makes the interface (the "hooks" you noted) representation-independent, and it resolves the genuinely awkward case (atoms on interval boundaries) with the standard right-continuous-CDF convention rather than leaving it implicit. The implementation can now expose a small measure trait and slot in histogram / parametric / CDF backings behind it.

Docs-only.

---

### DIFF
--- a/docs/design/WAM_RUST_BOUNDARY_DISTRIBUTION_SPECIFICATION.md
+++ b/docs/design/WAM_RUST_BOUNDARY_DISTRIBUTION_SPECIFICATION.md
@@ -4,43 +4,70 @@ Precise semantics of the boundary distribution optimization. See
 `WAM_RUST_BOUNDARY_DISTRIBUTION_PHILOSOPHY.md` (rationale) and
 `WAM_RUST_BOUNDARY_DISTRIBUTION_CACHE_PLAN.md` (phasing/status).
 
-## 1. The object: a path-length distribution *form*
+## 1. The object: a measure over path length
 
-The fundamental cached object is not "a histogram" specifically — it is a
-**path-length distribution form**: any representation from which we can compute
-the **mass between two lengths** (the PMF/CDF). The one operation everything else
-is built on is
+The cached object is, in general, a **measure** `μ` over the path-length variable
+`ℓ` (the single variable here being the seed→root path length). A measure assigns
+a non-negative **weight to a set** of lengths; everything the optimization needs is
+a *read* of `μ`. This is the abstraction the implementation should expose as an
+**interface** (the hooks), with the concrete representation behind it. It subsumes
+the cases under one idea:
+
+- a **discrete histogram** is an *atomic* measure — a sum of point masses at
+  integer lengths;
+- a **continuous approximation** is an *absolutely-continuous* measure — a density
+  integrated;
+- a **single deterministic length** is a *Dirac* point mass;
+- mixtures of the above are measures too.
+
+### 1a. The interface (one operation, and the endpoint care)
+
+The core read is the **mass of an interval**:
 
 ```
-range_mass(a, b) = (sum or integral) over [a, b] of the path-length distribution
+interval_mass(a, b) = μ of the length interval between a and b
 ```
 
-— a *sum* for a discrete form, an *integral* for a continuous one. From
-`range_mass` (equivalently, a cumulative `F(x) = range_mass(0, x)` and its
-differences) we get the PMF, the CDF, any bounded/constrained functional, and the
-truncations below. A concrete `H` (histogram) is the default backing, but the
-spec is written against the *form*, so a continuous or transformed backing drops
-in without changing the consumers.
+— a *sum* for an atomic (discrete) measure, an *integral* for a continuous one.
+Everything else (PMF, CDF, any bounded/constrained functional, truncations) is
+built from it.
 
-### 1a. Backing forms (representations)
+**Endpoint convention — the delta-function caveat.** A point mass sitting exactly
+on an endpoint forces a half-open-vs-closed choice. We pin it with a
+**right-continuous CDF** `F(x) = μ((-∞, x])` (so `F` *includes* the atom at `x`):
 
-| form | `range_mass` | splice (compose over a cut) | when |
-|------|--------------|------------------------------|------|
-| **discrete histogram** (PMF) `H[L]:u64` | partial sum | convolution — O(m²) direct, **O(m log m) FFT** | default; exact; small/medium support |
+```
+interval_mass((a, b]) = F(b) − F(a)              (half-open — the canonical form)
+atom_mass(x)          = F(x) − F(x⁻)             (the point mass AT x; 0 if no atom)
+interval_mass([a, b]) = F(b) − F(a) + atom_mass(a)
+```
+
+So a "centered / left-sided / right-sided" point mass is just where the atom is
+placed relative to the half-open intervals — in the discrete case the mass at
+integer `L` is `F(L) − F(L−1)`. The interface therefore exposes **both**
+`interval_mass` (half-open by default) **and** `atom_mass`, so consumers never
+guess the convention. `total_mass = F(∞)`; the splice (§3) and `truncate` are the
+other hooks.
+
+### 1b. Backing representations (concrete measures)
+
+| representation | `interval_mass` | splice (compose over a cut) | when |
+|----------------|-----------------|------------------------------|------|
+| **discrete histogram** (atomic) `H[L]:u64` | partial sum | convolution — O(m²) direct, **O(m log m) FFT** | default; exact; small/medium support |
 | **continuous / parametric** (density: normal, binomial, GMM…) | integral / closed-form CDF | parameter arithmetic (e.g. variances add) — often O(1) | **very large scale**, where the discrete grid is the cost; approximate |
-| **cumulative / transformed** (CDF table, or a pre-weighted basis `g_B`) | O(1) difference of cumulatives | depends; for a fixed functional the basis is a dot product (~1 ns) | hot `(functional, budget)`; O(1) range queries |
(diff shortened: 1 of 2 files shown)


docs/design/WAM_RUST_BOUNDARY_DISTRIBUTION_SPECIFICATION.md
```diff
@@ -4,43 +4,70 @@ Precise semantics of the boundary distribution optimization. See
...
+- a **single deterministic length** is a *Dirac* point mass;
+- mixtures of the above are measures too.
+
+### 1a. The interface (one operation, and the endpoint care)
+
+The core read is the **mass of an interval**:
+
+```
+interval_mass(a, b) = μ of the length interval between a and b
+```
```
s243a, line 34: **[minor] Signature inconsistency: `interval_mass(a, b)` vs `interval_mass((a,b])`**

This line introduces `interval_mass(a, b)` with two positional args and no open/closed marker. Lines 40–42 then write `interval_mass((a, b])` and `interval_mass([a, b])` — interval-notation with explicit endpoint markers — which reads as two different call signatures.

Since §1a's whole purpose is making the half-open/closed distinction explicit, the canonical signature should be stated once. **Suggested fix:** define `interval_mass(a, b)` here as the half-open `(a, b]` by convention (matching the "canonical form" label below), then present `interval_mass([a, b])` as a derived expression `interval_mass(a, b) + atom_mass(a)` rather than an alternate signature.

docs/design/WAM_RUST_BOUNDARY_DISTRIBUTION_SPECIFICATION.md
```diff
@@ -4,43 +4,70 @@ Precise semantics of the boundary distribution optimization. See
...
-differences) we get the PMF, the CDF, any bounded/constrained functional, and the
-truncations below. A concrete `H` (histogram) is the default backing, but the
-spec is written against the *form*, so a continuous or transformed backing drops
-in without changing the consumers.
+So a "centered / left-sided / right-sided" point mass is just where the atom is
+placed relative to the half-open intervals — in the discrete case the mass at
+integer `L` is `F(L) − F(L−1)`. The interface therefore exposes **both**
+`interval_mass` (half-open by default) **and** `atom_mass`, so consumers never
+guess the convention. `total_mass = F(∞)`; the splice (§3) and `truncate` are the
+other hooks.
```
s243a: **[minor] `truncate` listed as an interface hook but never defined**

"`truncate` are the other hooks" — but unlike `splice` (forward-referenced to §3), `truncate` has no signature or semantics anywhere in this PR.

**Suggested fix:** Add a one-liner, e.g.: `truncate(μ, lo, hi) = μ restricted to [lo, hi]` (mass outside the window set to 0). Clarify whether it renormalizes — this matters because the functionals in §2 ("integrate `w` only over `[lo, hi]`") rely on the *un*-normalized restriction.

**Bonus suggestion:** A short illustrative Rust trait sketch — even non-normative — would force two open decisions: (1) return type (`u64` for exact discrete, `f64` for continuous — the spec currently mixes both); (2) the `splice` associated-type problem (`Self × Self → Self` breaks when convolving different backings, e.g. histogram × parametric). Worth one code block in §1a.

docs/design/WAM_RUST_BOUNDARY_DISTRIBUTION_SPECIFICATION.md
```diff
@@ -4,43 +4,70 @@ Precise semantics of the boundary distribution optimization. See
...
-form convolves by parameter arithmetic and answers range queries from a closed-form
-CDF, avoiding the grid entirely — at the cost of being approximate (gated on a
-CDF/W1 error bound, per `DISTRIBUTIONAL_COMPRESSION_THEORY.md`).
+measure convolves by parameter arithmetic and answers interval reads from a
+closed-form CDF, avoiding the grid entirely — at the cost of being approximate
+(gated on a CDF/W1 error bound, per `DISTRIBUTIONAL_COMPRESSION_THEORY.md`).
 
-- **Boundary set** `Bset ⊆ V`: nodes near the root at which suffix forms are
+- **Boundary set** `Bset ⊆ V`: nodes near the root at which suffix measures are
   cached. The default backing is `boundary_suffix: FxMap<u32, Vec<u64>>` on
```
s243a: **[minor] `Bset` field `FxMap<u32, Vec<u64>>` cannot hold non-discrete backings**

The PR generalizes the cached object to a measure with multiple backings, but the concrete `WamState` field remains `Vec<u64>`. The text acknowledges an alternate backing "stores a CDF/basis/parametric measure instead" but the typed field can't hold that without a type change.

Acceptable for a spec that declares discrete-as-default, but this is the exact seam where "interface independent of representation" will first meet Rust's type system. **Suggested note:** call out that adopting non-discrete backings requires this field to become a sum type (e.g. `enum MeasureBacking { Histogram(Vec<u64>), Cdf(...), Parametric(...), Dirac(u32) }`) or `Box<dyn Measure>`. Most useful forward pointer for whoever implements §1a.

s243a (commented):
## Review Summary

This is a docs-only PR (+66/−33, two markdown files). The conceptual contribution is correct and valuable — the measure abstraction cleanly subsumes atomic/continuous/Dirac/mixture cases, and all endpoint math checks out (`interval_mass((a,b]) = F(b)−F(a)`, `atom_mass(x) = F(x)−F(x⁻)`, `interval_mass([a,b]) = F(b)−F(a)+atom_mass(a)`). No correctness, security, or performance concerns since no code ships.

**Required fix before merge:** The rename `range_mass → interval_mass` was applied only to §1/§1a. Four sites in §2/§5/§8 still say `range_mass` — including **the `range_mass(a,b)` result-mode name in the §5 table (line 139)**, which is a user-visible API identifier. A reader implementing §5 before reading §1a will expose `range_mass`; §1a says `interval_mass`. The spec contradicts itself on the API name — exactly the ambiguity it is trying to eliminate.

**Fix:** Either complete the rename across §2/§5/§8 (lines 81, 89, 139, 145), or add one sentence in §1a explicitly declaring `range_mass` a retained alias for `interval_mass`.

Other findings are in the inline comments below.

s243a:
@codex review

s243a (commented):
## Re-review — all issues resolved ✓

Verified commit `8b6604c` against every finding from the original multi-agent review (Claude Code ×2 + deep research). All required fixes are present and correct.

**Required fixes — all resolved:**

- **`range_mass → interval_mass` rename (blocking):** Completed across §2/§5/§6/§8. `grep -r range_mass docs/` returns empty — clean.
- **Canonical signature:** `interval_mass(a, b) := μ((a,b])` is now the single signature; the closed form is correctly presented as a derived expression `closed_interval_mass = interval_mass(a,b) + atom_mass(a)`.
- **`truncate` defined:** Line 52 defines `truncate(μ, lo, hi)` as the unnormalized restriction; line 118 confirms it suffices for the bounded-integral functional in §2. Correctly not renormalized.
- **`atom_mass(0) = 0` invariant stated:** Line 48 is explicit, and `total_mass = interval_mass(0, budget)` follows.

**High-value additions — all present:**

- **`trait Measure` sketch (lines 59–64):** `interval_mass`, `atom_mass`, `total_mass`, `truncate` — all returning `f64` (real mass, resolving the `u64`/`f64` ambiguity). `splice` is deliberately absent from the trait, with the associated-type seam called out.
- **`splice` as convolution of measures (line 129):** Representation-independent statement present; histogram convolution is its discrete realization.
- **`Vec<u64> → enum MeasureBacking` forward pointer (lines 95–100):** Exact enum variants named — `Histogram(Vec<u64>)`, `Cdf(Box<[f64]>)`, `Parametric(Params)`, `Dirac(i64)`.
- **measure/form bridge and domain `ℓ ∈ [0, budget]`:** Both present.

**Design note on `splice` (flagged for implementors):** The decision to keep `splice` off `trait Measure` (it cannot be `Self × Self → Self` across heterogeneous backings) is the right call and is clearly documented. This is the key constraint for the P2c-wiring implementation — `splice` is a separate combinator with its own output type, not a trait method.

**Recommendation: Approve and merge.** The spec is now internally consistent, mathematically rigorous, and provides the right forward pointers for implementation.

PR #3227, opened 2026-06-17
docs(wam-rust): theory note — graph functional semirings (functionals without the histogram)

s243a:
**Docs-only.** A self-contained theory note generalizing the (already-merged) boundary distribution cache from "cache the path-length histogram" to "propagate the **functionals** of that histogram directly, never forming it." Forward-looking — it scopes the next increments rather than describing shipped code. Posted as its own focused PR for review.

Adds `docs/design/WAM_RUST_GRAPH_FUNCTIONAL_SEMIRINGS.md` (+ a one-line cross-reference from the boundary spec). Four commits track how the idea was developed in discussion:

1. **The note itself** (`03257991`) — the core framework.
2. **Moment-jet precondition + graded reconstruction** (`a56188b5`).
3. **Odd-moment cancellation** (`8d86b807`).
4. **Convergence, not just acyclicity** (`fbb2d550`).

## What it argues

- **Composition law.** Concatenation convolves path-length histograms / unions add them, so a functional propagates on its own value exactly when it's a **semiring homomorphism**. Worked out for mass (counting `(+,×)`), the raw-moment jet `(M, m₁, …, m_K)` (the order-`K` binomial convolution `m_k(A⊛B) = Σ_j C(k,j) m_j(A) m_{k−j}(B)`), and the support interval `(min, max)` (min-plus × max-plus).
- **`PathSemiring` framework.** One algebraic-path-problem recurrence over a (product) semiring; n-tuples = product semirings; histogram / moment-jet / interval are all instances of the same generic solve.
- **Domain = the convergent ancestor direction.** The exact domain is a node's acyclic up-closure; the deeper governing property is *convergence to a unique sink* (the root), where the complete path set is canonical, finite, and meaningful — so relaxing the path-length budget is natural. The divergent descendant direction has no unique sink, is dominated by indirect paths, and needs a budget even in a pure DAG. Three-regime table included.
- **Moment-jet precondition (dual of the `g_B` caveat).** The binomial moment law is exact only for *untruncated* convolution — a length budget isn't representable in moment space — so the jet is budget-free-exact in the ancestor direction and breaks under the cyclic/descendant budget, where you fall back to the histogram.
- **Odd moments are fine.** Path lengths are non-negative, so raw moments propagate cancellation-free at any order; the one subtraction is the central read-out, which is benign (only ill-conditioned when the result is small, i.e. negligible) and gate-protected.
- **Certified bracket.** `d_eff` is a power mean of path lengths, so `d_eff ∈ [min, max]` exactly — a certified bound on the real metric from two scalars. (And the honest non-example: `WeightSum` is not multiplicative under convolution, so it has no scalar splice.)
- **Kernel-trick analogy.** The histogram is the implicit feature map; functionals are computed *through* it without materialization — with the honest limits (semiring homomorphism, not PSD; `WeightSum` = the non-factoring case).
- **CLT reconstruction.** The moment jet → discretised Normal at deep nodes, a CDF-gated rung complementary to the discretised-GMM; raw→central is exact algebra, the model enters only at moments→CDF, giving a graded family (Gaussian → Edgeworth/Gram–Charlier → Pearson → histogram).
- **Roadmap.** Generic `PathSemiring` payload on the ancestor space first (exact, validated bucket-for-bucket against the histogram), distance/shortest-path kernels + cyclic closure second.

---

### DIFF
--- a/docs/design/WAM_RUST_GRAPH_FUNCTIONAL_SEMIRINGS.md
+++ b/docs/design/WAM_RUST_GRAPH_FUNCTIONAL_SEMIRINGS.md
@@ -82,8 +82,10 @@ some small algebra `(⊗, ⊕)`. Working them out:
 > **acyclic ancestor space** (§4): there every path length is finite and bounded by the
 > DAG height, so no cap is ever needed and the jet is exact for free. A budget only
 > re-enters for *cyclic* graphs (to keep counts finite) — and there the moment shortcut
-> breaks and you fall back to the bucket-truncatable histogram. So the payloads
-> partition: **acyclic ⇒ moment jet exact; cyclic ⇒ histogram only.**
+> breaks and you fall back to the bucket-truncatable histogram. The governing property
+> is **direction of convergence**, not merely acyclicity (§4): the budget-free jet is at
+> home in the *convergent ancestor direction*, while the *divergent descendant
+> direction* wants a budget even in a pure DAG.
 - **support interval** `(min, max)`. Convolution adds the extremes, union takes the
   extremes: `min_{ac} = min_ab + min_bc`, `min_v = min_p (1 + min_p)`; symmetrically for
   `max` with `max`. → the **tropical** semirings min-plus and max-plus.
@@ -182,8 +184,37 @@ graph. This is the right and load-bearing domain:
   existing **budget + visited-guard** truncation. Exact on poset/taxonomic data;
   truncation-approximate on general graphs.
 
-This staging — **acyclic ancestor space first, cyclic closure second** — orders the
-implementation (§8).
+### Convergence, not just acyclicity: why unbounded length is meaningful here
+
+The deeper reason the ancestor direction tolerates an unbounded path length — the
+precondition the moment jet needs (§2) — is that it **converges to a unique sink, the
+root**. Every path funnels inward and *completes* at the root, so the path set is the
+canonical, finite, *total* set of the node's derivations ("all the ways `v` is-a … is-a
+root"). Completeness is well-defined and length is bounded by the DAG height for free, so
+relaxing the budget is natural, not a patch — and the complete statistics *are* the
+meaning.
+
+The **descendant direction is the opposite geometry**: paths *diverge* toward many leaves
+with **no unique sink**, so "all paths without a length cap" is not a canonical statistic
+— it is dominated by long, indirect routes and explodes with fan-out. There a budget is a
+**semantic filter** (which paths count as meaningful: the short, direct ones), and it is
+needed *even in a pure DAG*. So the precondition for the budget-free moment jet is
+properly the **convergent ancestor direction**, of which "acyclic up-closure" is the
+taxonomic instance — not acyclicity per se.
+
+| direction / structure | unique sink? | unbounded length is… | payload |
+|---|---|---|---|
+| ancestor → root (convergent) | yes (root) | finite **and** meaningful | **moment jet exact** (+ histogram) |
+| descendant (divergent, even acyclic) | no | finite but indirect-path-dominated | budget + histogram (jet breaks under the cap) |
+| cyclic (either way) | — | infinite | budget + histogram (finiteness necessity) |
+
+This is also why the moment-jet idea sits naturally on the existing machinery: the
+boundary cache already stores **suffixes toward the root** — the convergent direction — so
+budget-free moment propagation rides what exists, while descendant / general-direction
+search intrinsically carries the budget and stays on the histogram.
+
+This staging — **convergent ancestor space first, divergent/cyclic closure second** —
+orders the implementation (§8).
 
 ## 5. A certified bracket on `d_eff` from two scalars
 
--- a/docs/design/WAM_RUST_GRAPH_FUNCTIONAL_SEMIRINGS.md
+++ b/docs/design/WAM_RUST_GRAPH_FUNCTIONAL_SEMIRINGS.md
@@ -107,6 +107,23 @@ axiom, and that is precisely what makes the splice exact.
 > clean for both. `mean = m₁/M`, `var = m₂/M − mean²` are **nonlinear read-outs at the
 > end**, never carried through the recurrence.
 
+> **Odd moments and cancellation (why raw, again).** Because path lengths are
+> non-negative, every *raw* moment is a sum of non-negative terms `Lᵏ·H[L]`, and the
+> binomial convolution combines only non-negative quantities (`C(k,j) ≥ 0`, all
+> `m ≥ 0`) — so **propagation never cancels, at any order, odd or even.** Cancellation
+> appears only in the **central read-out** (`μ₃ = m₃′ − 3μ·m₂′ + 2μ³` subtracts
+> comparable terms), and only when the central moment is *small* — a near-symmetric
+> node, where `μ₃ ≈ 0` is a small difference of large raw moments. That is benign: you
+> only *need* `μ₃` accurate when it is **large** (strong skew), which is the
+> non-cancelling regime; when it is small the skew correction it feeds is negligible, so
+> the lost precision does not move the reconstruction. And the **CDF gate is the final
+> backstop** — a numerically degraded skew/kurtosis reconstruction misses the Kolmogorov
+> certificate and is rejected back to the Gaussian or the histogram, exactly as a *model*
+> mismatch would, so correctness never rests on the conditioning of the conversion. (A
+> shifted-origin moment conditions the read-out better but reintroduces signs and breaks
+> the non-negative, linear propagation, so it stays a read-out-time option, not a change
+> to what is carried.)
+
 ### The one that does *not* factor: `WeightSum`
 
 `WeightSum_N` is `⊕`-linear (it adds under union) but **not** `⊗`-multiplicative:
--- a/docs/design/WAM_RUST_GRAPH_FUNCTIONAL_SEMIRINGS.md
+++ b/docs/design/WAM_RUST_GRAPH_FUNCTIONAL_SEMIRINGS.md
@@ -61,15 +61,29 @@ some small algebra `(⊗, ⊕)`. Working them out:
 
 - **mass** `M = H(1)`. Convolution multiplies, union adds:
   `M_{a→c} = M_{a→b} · M_{b→c}`, `M_{v} = Σ_p M_p`. → the **counting semiring** `(+, ×)`.
-- **raw-moment jet** `(M, m₁, m₂)`. Under *union* every raw moment is linear
-  (`m_k(H₁+H₂) = m_k(H₁) + m_k(H₂)`). Under *convolution* the Leibniz/Vandermonde rule:
+- **raw-moment jet** `(M, m₁, …, m_K)`. Under *union* every raw moment is linear
+  (`m_k(H₁+H₂) = m_k(H₁) + m_k(H₂)`). Under *convolution* the **binomial convolution**
+  (because `(a+b)^k` expands binomially):
   ```
-  M_{ac}  = M_ab·M_bc
-  m₁_{ac} = m₁_ab·M_bc + M_ab·m₁_bc
-  m₂_{ac} = m₂_ab·M_bc + 2·m₁_ab·m₁_bc + M_ab·m₂_bc
+  m_k(A ⊛ B) = Σ_{j=0..k} C(k,j) · m_j(A) · m_{k-j}(B)
   ```
-  i.e. `⊗` is **upper-triangular Toeplitz (truncated power series / "jet")
-  multiplication** and `⊕` is componentwise addition. This is a commutative semiring.
+  — for `K = 2` this is `M_ac = M_ab·M_bc`, `m₁_ac = m₁_ab·M_bc + M_ab·m₁_bc`,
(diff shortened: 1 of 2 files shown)


docs/design/WAM_RUST_GRAPH_FUNCTIONAL_SEMIRINGS.md
```diff
@@ -0,0 +1,316 @@
...
+H_p)`.
+
+A functional `F` can therefore be propagated *on its own value* exactly when it respects
+both operations — i.e. `F` is a homomorphism from the histogram algebra `(⊛, +)` into
+some small algebra `(⊗, ⊕)`. Working them out:
+
+- **mass** `M = H(1)`. Convolution multiplies, union adds:
+  `M_{a→c} = M_{a→b} · M_{b→c}`, `M_{v} = Σ_p M_p`. → the **counting semiring** `(+, ×)`.
+- **raw-moment jet** `(M, m₁, …, m_K)`. Under *union* every raw moment is linear
+  (`m_k(H₁+H₂) = m_k(H₁) + m_k(H₂)`). Under *convolution* the **binomial convolution**
```
s243a, line 65: **[high-value, pre-§8.1] Cumulants vs raw moments — an unacknowledged design fork**

The note presents raw moments as the forced choice. They are a trade-off:

| Operation | Raw moments | Cumulants |
|---|---|---|
| Concatenation `⊗` (convolution of independent RVs) | O(K²) binomial | **O(K) componentwise addition** |
| Union `⊕` (mixture) | **O(K) linear** | Nonlinear (only κ₁ additive under mixing) |

Cumulants of independent summands add: `κ_k(A⊛B) = κ_k(A) + κ_k(B)` — O(K) vs the O(K²) binomial jet. But cumulants are not linear under union `⊕`. Neither representation is a semiring element under both operations — raw moments win on union-heavy (root-near, reconvergent) regions; cumulants win on concatenation-heavy (deep, thin) ancestor spines.

The conditioning argument later in §2.5 mitigates the `O(μ³)` cancellation in the central read-out. But carrying cumulants directly would avoid forming large raw moments `m₃ ~ μ³` at all — the problem §2.5 spends a paragraph on is exactly what cumulants prevent.

**Why this matters now:** §8.1 will freeze the `Elem` type into shipped code. Resolve the trade on paper first. The closest published cousin is [Tsujie & Ueda arXiv:2604.13819](https://arxiv.org/html/2604.13819v1) on t-deformed cumulant additivity in formal power series.

**Suggested addition:** A table row or subsection contrasting raw moments and cumulants, naming the ⊗/⊕ linearity trade, and noting the regime where each wins.

docs/design/WAM_RUST_GRAPH_FUNCTIONAL_SEMIRINGS.md
```diff
@@ -0,0 +1,316 @@
...
+f(root) = one;     f(v) = ⊕_{p ∈ parents(v)} ( edge ⊗ f(p) )
+```
+
+over a semiring `(S, ⊕, ⊗)`. An **n-tuple** of functionals is just the **product
+semiring** `S₁ × … × Sₖ`, propagated componentwise — so "arbitrary vectors of scalar
+difference equations on the graph" is exactly *pick a product semiring and run the one
+recurrence*. Exactness is automatic: `⊗` distributing over `⊕` is the defining semiring
+axiom, and that is precisely what makes the splice exact.
+
+> **Important state choice.** Store the **raw moments** `(M, m₁, m₂)`, not `(count,
```
s243a: **[minor] Promote this invariant to a normative rule — currently buried**

The single most important implementation rule in the document — _carry `(M, m₁, m₂)`, never `(count, mean, variance)`_ — is a blockquote that a skimming implementor will miss. They will write the natural-but-wrong mean/variance accumulator.

Suggested: elevate to a prominent `> **Rule:**` block before the binomial convolution law, not only as a buried-in-prose aside. The current wording is correct; it needs elevation, not rewriting.

s243a (commented):
## Multi-model review — PR #3227 (graph functional semirings)

**Reviewers:** Opus 4.8 · GPT-5.5 · Gemini deep research · prior joint review (Claude Code ×2)

**Verdict: ✅ Approve.** No blocking issues. All four new reviewers independently confirm the math is correct. The semiring framing is the right abstraction — GPT-5.5 specifically noted it consolidates rather than gold-plates by removing bespoke code paths. The [Abo Khamis et al. JACM 2024, arXiv:2105.14435](https://arxiv.org/abs/2105.14435) p-stable semiring results and [Sanmartín et al. ICML 2022](https://proceedings.mlr.press/v162/sanmarti-n22a/sanmarti-n22a.pdf) both independently ground the PR's claims about which functionals admit polynomial propagation vs require the full transitive closure.

**Most important before §8.1:** The cumulant vs raw moment design fork (see §2 inline comment) is the one decision that cannot easily be changed after `PathSemiring` is generic-ified.

**Required for the roadmap:** The `PathSemiring` trait cannot express the Kleene-star/closure contract §8.2 mandates — this should be flagged as deferred design work so implementors are not surprised (see §3 inline comment).

docs/design/WAM_RUST_GRAPH_FUNCTIONAL_SEMIRINGS.md
```diff
@@ -0,0 +1,316 @@
...
+`WeightSum_N` is `⊕`-linear (it adds under union) but **not** `⊗`-multiplicative:
+`Σ_{a+b=n} … (a+b)^{-N}` does not separate into `f(prefix)·g(suffix)` because `(a+b)^{-N}`
+is not a product of a-only and b-only terms. So there is no scalar concatenation law for
+the effective-distance weight. The existing `g_B` pre-weighted basis is the partial fix:
+fix the budget and pre-weight, `g_B[a] = Σ_b H_B[b]·(a+b)^{-N}`, giving a *vector indexed
+by prefix length* (not a scalar), valid only at that fixed `N` and budget. In kernel-trick
+terms (§6) this is a kernel that does **not** factor through a finite feature — you either
+carry the whole histogram, or accept the cheap **bracket** of §5 instead.
+
+## 3. The `PathSemiring` framework
```
s243a, line 192: **[roadmap-blocking at §8.2] `PathSemiring` trait cannot express the Kleene-star/closure contract**

The truncation hook is marked `// optional`. But the star/closure operation `a* = ⊕ aⁱ` has per-semiring convergence behavior:
- **min-plus (tropical):** converges freely (`a* = 0` for non-negative weights — Bellman-Ford).
- **counting semiring ℕ:** never converges on cyclic graphs (sum over all walks diverges).
- **moment-jet:** star converges iff mass component < 1 (WAM terminates w.p. 1, per [Droste & Kuich 2009](https://www.dmg.tuwien.ac.at/kuich/hbwa_chap1.pdf)).

The [Abo Khamis et al. JACM 2024](https://arxiv.org/abs/2105.14435) p-stable semiring result confirms: counting semiring is not p-stable on cyclic graphs; tropical is. An optional truncation hook cannot encode this per-payload.

**This is the gap most likely to force a mid-roadmap trait redesign** at §8.2.

Two additional gaps (GPT-5.5):
1. **No read-out/decode contract** — no typed surface for `Elem → query answer` projection (e.g. moment jet → mean/var → CLT reconstruction), which §8.1's bucket-for-bucket validation will need.
2. **Union/concat exactness asymmetry invisible** — the flat `add`/`mul` interface gives no place to encode that the budget breaks only `mul`, not `add`.

**Also:** the witness/dictionary-style design (marker type `S`, static methods on `Elem`) is orphan-safe and deliberate, but the note should state this explicitly — the natural "make it idiomatic via `num-traits` operator traits" path triggers orphan/overlap rules (tropical `Add = min` conflicts with numeric `Add` on the same scalar type). The product-semiring "free composition" story is also weaker than the prose implies: heterogeneous tuples like `(min, max, mass, m₁, m₂)` require hand-written composite impls, not generic product combinators.

**Recommended addition:** One paragraph in §3 or §8 flagging: (a) the trait is illustrative; (b) the star/closure contract is deferred design work — each payload's convergence must be characterized before §8.2; (c) each semiring instance will be a distinct newtype `Elem` by design.

docs/design/WAM_RUST_GRAPH_FUNCTIONAL_SEMIRINGS.md
```diff
@@ -0,0 +1,316 @@
...
+| interval `(min,max)` | min-plus × max-plus | 2 scalars | shortest + longest; brackets `d_eff` (§5) | both endpoints |
+| shortest scalar `min` | min-plus | 1 scalar | shortest distance (A* heuristic / landmark) | shortest only |
+
+## 4. The domain: a node's ancestor space
+
+The recurrence is an **up-propagation** — it walks `parents`, toward the root — so its
+support is exactly the **ancestor space (up-closure) of the query node**, not the whole
+graph. This is the right and load-bearing domain:
+
+- **Exactness needs only the reachable set to be acyclic**, and the reachable set *is*
```
s243a, line 237: **[minor, one sentence] §4 acyclicity argument is model-dependent**

Verified: the current codebase has no tabled/SLG machinery (zero `tabl`/`SLG` references in 51 Rust files) — the argument is correct for this codebase's execution model.

But the note's title says "WAM" generically. Under tabled/SLG resolution the relevant graph is the subgoal-dependency graph, whose SCCs are cyclic even over acyclic data. The [ProbLog (IJCAI 2007)](https://www.ijcai.org/Proceedings/07/Papers/396.pdf) and [YAP mode-directed tabling (PADL 2013)](https://www.dcc.fc.up.pt/~ricroc/homepage/publications/2013-PADL.pdf) systems handle this case via their own convergence mechanisms.

**Suggested fix:** Add one sentence to §4: _"Domain argument assumes a single bottom-up sweep over the data relation's parent graph (this target's model). Under tabled/SLG resolution the subgoal-dependency graph has SCCs cyclic over acyclic data — there the closed-semiring/truncation path (§4 bullet 3, §8.2) is the normal case, not the exception."_

docs/design/WAM_RUST_GRAPH_FUNCTIONAL_SEMIRINGS.md
```diff
@@ -0,0 +1,316 @@
...
+```
+min_L  ≤  d_eff  ≤  max_L      always.
+```
+
+So the tropical interval `(min, max)` is not just shortest/longest — it is an **exact,
+certified bracket on the real effective-distance metric**, for two integers, exact at
+the endpoints and sound everywhere between. It is the cheap surrogate for the
+`WeightSum` functional that §2 showed cannot be carried as a scalar.
+
+## 6. The analogy: the kernel trick
```
s243a: **[minor] Kernel-trick §6 should be demoted to a short aside**

All four independent reviewers this round agree: the analogy is correctly caveated and not wrong, but it occupies a full top-level section and 7-row table for what is operationally a mnemonic. A Rust developer implementing §8 never needs §6 — it is the least load-bearing section for the primary audience.

Recommend: keep the analogy (it is genuinely illuminating), but demote to a short aside within §1 or §2. The representer-theorem caveat (the analogy does not carry that theorem — no RKHS here) is correctly stated and should be preserved wherever it lands.

docs/design/WAM_RUST_GRAPH_FUNCTIONAL_SEMIRINGS.md
```diff
@@ -0,0 +1,316 @@
...
+  CLT-Gaussian for deep, well-mixed, unimodal nodes (3 scalars, no EM); GMM for shallow,
+  structured, multimodal nodes (expensive). If a node is multimodal the moment-Gaussian
+  misses the Kolmogorov gate and the chooser rejects it automatically — the correctness
+  certificate already guards the approximation, so it never silently fires where CLT
+  does not hold.
+- The win is exactly the implicitness of §1/§6: the moments propagate by their own
+  scalar recurrence, so the deep-node distribution estimate costs three accumulators and
+  never touches a histogram.
+
+## 8. Roadmap (increments)
```
s243a, line 367: **[non-blocking] §8 is missing a step 1.5 — convergence table belongs between the two increments**

Step 2 conflates two independent pieces of work:
1. **Per-payload star characterization:** min-plus has a well-defined star; counting/mass diverges without truncation; moment-jet star converges only if mass < 1. This must be resolved per payload *before* a cyclic query binding is correct.
2. **New query bindings:** `transitive_distance3`, `weighted_shortest_path3`, `astar_shortest_path4`.

Bundling them means the first cyclic query is also the first time the mass-payload's non-convergence is confronted — two unknowns at once. The [Bloom–Ésik–Kuich partial Conway framework arXiv:0712.2952](https://arxiv.org/pdf/0712.2952.pdf) handles semirings where star is only defined on an ideal — applicable here since the moment-jet star is only defined for mass < 1.

**Suggested revision:** 1 → **1.5** (characterize per-payload star/truncation on acyclic data, produce the convergence table) → 2 (cyclic query bindings).

Also: §8.1 wires the Gaussian CLT rung (safe). The higher-order Edgeworth/Gram-Charlier rungs (~line 301) should be explicitly deferred pending the clamp-or-Pearson decision. [Barton & Dennis (Biometrika 1952)](https://doi.org/10.1093/BIOMET/39.3-4.425) characterises the Gram-Charlier positive-definite region; [Jondeau & Rockinger (1999)](https://publications.banque-france.fr/sites/default/files/medias/documents/working-paper_56_1999.pdf) show QP enforcement of positivity. The Gaussian rung itself has no non-positivity risk.

PR #3228, opened 2026-06-17
feat(wam-rust): discretised-GMM approximation rung (escalation fallback)

s243a:
Completes the boundary-suffix approximation ladder (**Rungs 1–6**). This rung was developed alongside PR #3224 but landed after that PR was merged at the quantised-CDF commit, so it's broken out here on its own so it isn't lost.

Every prior parametric rung is a **binomial family**, whose per-component variance is pinned to its mean (`n·p·(1−p)`); that coupling means a binomial mixture cannot place a **narrow** mode in the *interior* of the support (a tight mode forces `p` toward an edge).

`HistRepr::DiscGmm{support,comps,total}` lifts the coupling — each component carries a free `(weight, mu, sigma)`, fitted by EM over the integer support (`fit_discretised_gmm`) and evaluated/renormalised by `discretised_gmm_pmf`. It costs the most params/mode of any rung, so the chooser selects it only when **every cheaper rung misses the CDF gate AND** it still undercuts the quantised-CDF table on bytes. Escalation only, never the default.

- `encode_repr`/`decode_repr` tag 5; `bytes()`/`pmf()`/`expand()`/`total()` arms.
- `representation_candidates` adds K=2,3 GMM fits, gated like the mixture.
- Test `disc_gmm_fits_narrow_interior_modes`: two `σ=1.5` modes at 20/40 — binomial, beta-binomial, and K≤3 binomial mixture all rejected; the GMM fits within `ε_K` and is chosen. **35 lib tests pass.**
- Docs: spec Rung 6 + §9b ladder-complete status; cache-plan rung DONE; how-to chooser order + representation map.

---

### DIFF
--- a/docs/design/WAM_RUST_BOUNDARY_DISTRIBUTION_CACHE_PLAN.md
+++ b/docs/design/WAM_RUST_BOUNDARY_DISTRIBUTION_CACHE_PLAN.md
@@ -412,8 +412,15 @@ Phasing:
     prefix-mass). `encode_repr`/`decode_repr` tag 4. Validated:
     `quant_cdf_is_the_no_fit_fallback` (five spikes — binomial/mixture rejected, the
     chooser falls back to QuantCdf, smaller than exact, within `ε_K`).
-  - **Fitted forms [next].** discretised-GMM as escalation only (the last rung;
-    bounded integer data make it lowest priority).
+  - **Discretised-GMM [DONE].** `HistRepr::DiscGmm{support,comps,total}` —
+    `fit_discretised_gmm` (EM over the integer support) fits free-`(weight,mu,sigma)`
+    components, so it places the narrow interior modes the mean-coupled binomial
+    families cannot. Most params/mode of any rung, so the chooser picks it only when
+    every cheaper rung misses the gate AND it still undercuts the quant-CDF table on
+    bytes. `encode_repr`/`decode_repr` tag 5. Validated:
+    `disc_gmm_fits_narrow_interior_modes` (two σ=1.5 modes at 20/40 — binomial, beta-
+    binomial, and K≤3 binomial mixture all rejected; the GMM fits and is chosen).
+    **The approximation ladder is now complete (Rungs 1–6).**
 
 ## 6. What to measure (re-derive, don't inherit)
 
--- a/docs/design/WAM_RUST_BOUNDARY_DISTRIBUTION_CACHE_PLAN.md
+++ b/docs/design/WAM_RUST_BOUNDARY_DISTRIBUTION_CACHE_PLAN.md
@@ -420,7 +420,12 @@ Phasing:
     bytes. `encode_repr`/`decode_repr` tag 5. Validated:
     `disc_gmm_fits_narrow_interior_modes` (two σ=1.5 modes at 20/40 — binomial, beta-
     binomial, and K≤3 binomial mixture all rejected; the GMM fits and is chosen).
-    **The approximation ladder is now complete (Rungs 1–6).**
+    **The fitted histogram-approximation ladder is now closed structurally (Rungs 1–6)**
+    — every *histogram-representation* form is in place. (A separate, cheaper moment-jet
+    / CLT *reconstruction* rung — carry `(M,m₁,m₂)`, no EM — remains unbuilt and is the
+    first increment in `WAM_RUST_GRAPH_FUNCTIONAL_SEMIRINGS.md` §7–§8; it is not part of
+    this fitted ladder.) "Closed" is structural, not a claim that a measured workload has
+    been shown to need Rung 6.
 
 ## 6. What to measure (re-derive, don't inherit)
 
--- a/docs/design/WAM_RUST_BOUNDARY_DISTRIBUTION_SPECIFICATION.md
+++ b/docs/design/WAM_RUST_BOUNDARY_DISTRIBUTION_SPECIFICATION.md
@@ -466,14 +466,27 @@ histograms (support ≤ `budget`+1) the work trigger never fires; this matters a
 
 ### 9b. Ladder status and future work
 
-All six rungs above are **implemented** (Rungs 1–6: tail-pruned exact, binomial
-[moment + CDF fits], beta-binomial, mixture-of-binomials, quantised-CDF, discretised-
-GMM). The error/CDF-gate machinery, the chooser (error- and storage-driven), and the
-persistence path are general — adding a representation is just another candidate in
+All six **fitted histogram-approximation rungs** are implemented (Rungs 1–6: tail-pruned
+exact, binomial [moment + CDF fits], beta-binomial, mixture-of-binomials, quantised-CDF,
+discretised-GMM) — i.e. the *histogram-representation* ladder is closed structurally.
+This is a structural closure, not an empirical claim that every rung has been observed
+necessary on a measured workload (the GMM test uses a synthetic ground truth; whether a
+real boundary node escalates past Rung 5 is a measurement question, §6). The error/CDF-
+gate machinery, the chooser (error- and storage-driven), and the persistence path are
+general — adding a representation is just another candidate in
 `representation_candidates` with a `bytes()` and a `pmf()`/`expand()`.
+
+A **separate, cheaper reconstruction rung is still unbuilt**: the moment-jet / CLT
+reconstruction described in `WAM_RUST_GRAPH_FUNCTIONAL_SEMIRINGS.md` §7 (carry
+`(M, m₁, m₂)`, reconstruct a discretised Normal). It is *not* part of this fitted ladder
+— it propagates without ever materialising the histogram — and per that note's roadmap
+it is the **first** (and cheapest, no-EM) reconstruction increment, even though the
+more-expensive GMM rung shipped first. So "ladder closed" here means the histogram-
+fitting forms; it does not mean the cheapest reconstruction path is done.
 
 Remaining, lower priority and added on demand:
 
+- **Moment-jet / CLT reconstruction rung** — see above and the functional-semiring note.
 - **heed-backend parity** for the `boundary_basis_repr` persistence and the spill
   sub-db (the `lmdb-zero` backend has both today).
 - **Higher-K / Bayesian model selection** for the mixture and GMM rungs (today `K =
--- a/docs/design/WAM_RUST_BOUNDARY_DISTRIBUTION_SPECIFICATION.md
+++ b/docs/design/WAM_RUST_BOUNDARY_DISTRIBUTION_SPECIFICATION.md
@@ -420,8 +420,9 @@ Implemented (Rust, `boundary_cache.rs`):
 - **Rung 4 — mixture of binomials** — `fit_binomial_mixture(h, k, iters)` (EM,
   shared `trials`, PMF-weighted) fits the **multimodal** nodes a single binomial
   rejects (a bottleneck / topic-mixture cone). `HistRepr::Mixture{trials,comps,total}`
-  joins the candidate set (`K = 2,3`). Discretised-GMM stays escalation-only —
-  bounded integer path-length data favour binomial families.
+  joins the candidate set (`K = 2,3`). Costlier-per-mode escalation (discretised-GMM)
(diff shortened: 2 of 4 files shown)


docs/design/WAM_RUST_BOUNDARY_DISTRIBUTION_CACHE_PLAN.md
```diff
@@ -412,8 +412,15 @@ Phasing:
...
-    bounded integer data make it lowest priority).
+  - **Discretised-GMM [DONE].** `HistRepr::DiscGmm{support,comps,total}` —
+    `fit_discretised_gmm` (EM over the integer support) fits free-`(weight,mu,sigma)`
+    components, so it places the narrow interior modes the mean-coupled binomial
+    families cannot. Most params/mode of any rung, so the chooser picks it only when
+    every cheaper rung misses the gate AND it still undercuts the quant-CDF table on
+    bytes. `encode_repr`/`decode_repr` tag 5. Validated:
+    `disc_gmm_fits_narrow_interior_modes` (two σ=1.5 modes at 20/40 — binomial, beta-
+    binomial, and K≤3 binomial mixture all rejected; the GMM fits and is chosen).
+    **The approximation ladder is now complete (Rungs 1–6).**
```
s243a, line 429: **[required — cross-PR wording]** "The approximation ladder is now complete (Rungs 1–6)" directly contradicts PR #3227 §7, which describes the CLT/moment-jet reconstruction rung as a cheaper, still-unbuilt complement to the GMM. Both PRs are in the same repo and will be read together.

Suggested rewording: *"All six fitted-representation rungs are now implemented. The CLT/moment-jet reconstruction rung (WAM_RUST_GRAPH_FUNCTIONAL_SEMIRINGS.md §7) is a complementary, cheaper future increment that does not require a stored histogram — it is not part of this ladder."*

Same fix applies to the matching sentence in SPEC §9b ("All six rungs above are implemented").

docs/design/WAM_RUST_BOUNDARY_DISTRIBUTION_SPECIFICATION.md
```diff
@@ -450,17 +464,21 @@ that certified bound for the affected nodes only). For typical small-budget boun
 histograms (support ≤ `budget`+1) the work trigger never fires; this matters at
 **large budget / deep paths**.
 
-### 9b. Remaining ladder rungs (future work)
+### 9b. Ladder status and future work
 
-The error/CDF-gate machinery, the chooser (error- and storage-driven), and the
+All six rungs above are **implemented** (Rungs 1–6: tail-pruned exact, binomial
+[moment + CDF fits], beta-binomial, mixture-of-binomials, quantised-CDF, discretised-
```
s243a: **[required — cross-PR wording]** Same issue as CACHE_PLAN: "All six rungs above are **implemented**" will be read as closing the roadmap, but #3227 §7 explicitly positions the CLT/moment-jet rung as cheaper and still-unbuilt. Please add a forward reference here too:

*"The CLT/moment-jet reconstruction rung (WAM_RUST_GRAPH_FUNCTIONAL_SEMIRINGS.md §7) is a complementary future increment — it reconstructs from moments without fitting a stored histogram and is not counted here."*

templates/targets/rust_wam/boundary_cache.rs.mustache
```diff
@@ -599,6 +696,14 @@ fn representation_candidates(h: &Hist, min_points: usize) -> Vec<(HistRepr, f64)
         let err = cdf_max_error_pmf(&exact_pmf, &mixture_pmf(trials, &comps));
         cands.push((HistRepr::Mixture { trials, comps, total }, err));
     }
+    for k in [2usize, 3] {
+        // escalation: free-variance Gaussian modes (narrow interior modes the
+        // binomial families cannot place); costlier than the mixture, so only wins
+        // in that niche.
+        let (support, comps) = fit_discretised_gmm(h, k, 80);
+        let err = cdf_max_error_pmf(&exact_pmf, &discretised_gmm_pmf(support, &comps));
+        cands.push((HistRepr::DiscGmm { support, comps, total }, err));
```
s243a, line 739: **[required — doc/terminology fix; code is correct]** `cdf_max_error_pmf` computes the deterministic sup-norm `max|F_fit(i) − F_exact(i)|` between the fitted PMF and the full empirical distribution — no sampling, no null hypothesis, no critical value. This is a **certified Kolmogorov-distance gate**, not a Kolmogorov–Smirnov statistical test.

The distinction matters: the feared "false pass" on discrete data (where the KS test is more conservative) **does not apply** — there is no distributional KS calibration happening at all. The gate is an exact hard bound against the empirical CDF.

Please replace any doc/PR-body references to "KS gate" or "KS test" with "sup-CDF tolerance" or "Kolmogorov distance gate". (The same rename should propagate to the related doc sections in SPEC/CACHE_PLAN.)

templates/targets/rust_wam/boundary_cache.rs.mustache
```diff
@@ -599,6 +696,14 @@ fn representation_candidates(h: &Hist, min_points: usize) -> Vec<(HistRepr, f64)
         let err = cdf_max_error_pmf(&exact_pmf, &mixture_pmf(trials, &comps));
         cands.push((HistRepr::Mixture { trials, comps, total }, err));
     }
+    for k in [2usize, 3] {
+        // escalation: free-variance Gaussian modes (narrow interior modes the
+        // binomial families cannot place); costlier than the mixture, so only wins
+        // in that niche.
+        let (support, comps) = fit_discretised_gmm(h, k, 80);
+        let err = cdf_max_error_pmf(&exact_pmf, &discretised_gmm_pmf(support, &comps));
```
s243a, line 738: **[non-blocking]** `fit_binomial_mixture` uses `iters = 60` (the existing call site) while this new call uses `iters = 80` with no comment explaining the difference. Neither has a named constant. Suggest extracting a shared `const EM_ITERS: usize = 80;` and updating both call sites, or at minimum adding a comment explaining why the GMM needs more iterations.

```rust
const EM_ITERS: usize = 80; // generous for the 1-D bounded case; add LL-delta exit if needed
```

templates/targets/rust_wam/boundary_cache.rs.mustache
```diff
@@ -568,6 +575,96 @@ pub fn fit_binomial_mixture(h: &Hist, k: usize, iters: usize) -> (usize, Vec<(f6
...
+                nmu2[j] += r * xi * xi;
+            }
+        }
+        let wsum: f64 = nw.iter().sum();
+        if wsum <= 0.0 { break; }
+        for j in 0..k {
+            if nw[j] <= 0.0 { continue; } // keep the stale component rather than NaN
+            let mu = nmu[j] / nw[j];
+            let var = (nmu2[j] / nw[j] - mu * mu).max(0.25); // floor sigma at 0.5 bin
+            comps[j] = (nw[j] / wsum, mu, var.sqrt());
```
s243a, line 696: **[non-blocking]** The EM σ-floor is applied here as `var.max(0.25)` (flooring σ ≥ 0.5 bin), but `gauss_density` independently floors σ at `1e-3`. These two floors disagree by ~3 orders of magnitude. In practice the EM floor dominates during fitting, but a *decoded* repr from persistence could carry σ between 1e-3 and 0.5 and be evaluated through the looser floor.

Recommend a single shared constant:
```rust
const SIGMA_FLOOR: f64 = 0.5; // 1 bin minimum; in gauss_density use sigma.max(SIGMA_FLOOR)
```
This makes the invariant explicit and covers both the fitting and evaluation paths.

templates/targets/rust_wam/boundary_cache.rs.mustache
```diff
@@ -568,6 +575,96 @@ pub fn fit_binomial_mixture(h: &Hist, k: usize, iters: usize) -> (usize, Vec<(f6
...
+                nw[j] += r;
+                nmu[j] += r * xi;
+                nmu2[j] += r * xi * xi;
+            }
+        }
+        let wsum: f64 = nw.iter().sum();
+        if wsum <= 0.0 { break; }
+        for j in 0..k {
+            if nw[j] <= 0.0 { continue; } // keep the stale component rather than NaN
+            let mu = nmu[j] / nw[j];
```
s243a, line 694: **[non-blocking]** When `nw[j] <= 0.0` the stale component is kept — correct defensive choice (avoids NaN). However, the surviving components are renormalised by `wsum` on line 263 while the dead component retains its old weight, so after a component death the persisted `comps` weights no longer sum to 1. `discretised_gmm_pmf` renormalises the PMF on output so the observable behaviour is correct, but a reader of the raw `comps` struct will see weights summing to < 1.

Suggest a one-line comment:
```rust
// component weights may not sum to 1 after a death; discretised_gmm_pmf renormalises
```

templates/targets/rust_wam/boundary_cache.rs.mustache
```diff
@@ -568,6 +575,96 @@ pub fn fit_binomial_mixture(h: &Hist, k: usize, iters: usize) -> (usize, Vec<(f6
     (trials, comps)
 }
 
+// --- Discretised Gaussian mixture (ladder rung 6): the escalation fallback ---
+//
```
s243a, line 597: **[non-blocking — splice invariant note]** Splice always expands a `DiscGmm` to a full histogram before convolution. This is intentional and correct: `discretise + renorm` is a nonlinear step, so analytically convolving two GMM parametrisations (means and variances add, but then you'd need to re-discretise and re-renorm) would not equal the convolution of the discretised PMFs and would silently break the certified-bound invariant.

A short comment here would prevent a future "optimisation" from reintroducing error:
```rust
// NB: DiscGmm is expanded to a Hist before splice; analytic GMM convolution is
// intentionally avoided — discretise+renorm is nonlinear, so it would break the
// exactness invariant (splice must be exact for all linear functionals at once).
```

templates/targets/rust_wam/boundary_cache.rs.mustache
```diff
@@ -921,6 +1051,38 @@ mod tests {
         assert!(cdf_max_error(&h, &repr.expand()) <= 0.01 + 1e-6);
     }
 
+    // Two NARROW interior modes (sigma well below any binomial's mean-coupled width
+    // at that location): the binomial families and K<=3 binomial mixture all miss the
+    // gate, but the free-variance discretised-GMM fits — and is chosen because it
+    // undercuts the quantised-CDF fallback on bytes at this support.
+    #[test]
+    fn disc_gmm_fits_narrow_interior_modes() {
+        let support = 60usize;
```
s243a, line 1094: **[non-blocking — test coverage gaps]** The new `disc_gmm_fits_narrow_interior_modes` test covers the primary happy path well (S=60, K=2, symmetric bimodal). Three paths remain untested:

1. **K=3 selection path** — there is no test where K=2 GMM misses the gate but K=3 passes. The K=3 loop iteration in `representation_candidates` is currently exercised only through roundtrip serialisation.
2. **Near-crossover support boundary** (S ≈ 26–38 for K=2/3) — the GMM only beats `QuantCdf` on bytes when S > 2+12K. A test at S=27 (K=2, just above crossover) and S=25 (just below, where QuantCdf should win) would exercise the byte-gating branch.
3. **Truncated-buffer decode** — the `if o + 24 > b.len() { break; }` branch in `decode_repr` is untested. A test passing a buffer truncated mid-component should return a partial (but non-crashing) decode.

s243a (commented):
## Multi-model code review — Rung 6 (discretised-GMM)

**Verdict: ready to merge after one documentation fix.**

The code is correct and safe. Every failure mode of the new rung (non-convergence, bad init, degenerate fit, underflow at depth-1000) is *safe-by-rejection*: a poor fit misses the certified sup-CDF tolerance and a cheaper rung is kept. Splice-by-expansion is the right choice and preserves the exactness invariant. Byte-cost gating is sound (DiscGmm < QuantCdf iff S > 2+12K; for K=2 that is S > 26, confirmed by the S=60 test).

**One required fix (doc only, code is correct):** the gate is not a KS *test* — see inline comment on `cdf_max_error_pmf`. Five non-blocking follow-ups are annotated inline.

PR #3256, opened 2026-06-17
WAM-Rust: cheap, non-circular hub-convergence metrics for caret-bridge selection

s243a:
## What & why

Caret-bridge selection needs a way to identify **convergence hubs** (the union-type
categories where the hierarchy funnels) so `caret_min_over_hubs` can pick the bridge giving
the minimum distance. The hard constraint that shaped this whole PR: the hub score **must not
call the upward distance/reachability work the bridges exist to amortize** — otherwise
identifying hubs costs exactly what hubs were meant to save (circular). Everything here reads
only local/structural information or a one-pass precompute; **no metric reads a distance**.

The design converged through four successive objections, each fixed by the next rung:

| rung | function(s) | what it answers | cost | objection it fixes |
|------|-------------|-----------------|------|--------------------|
| 1 | `convergence_fanin`, `hubs_by_fanin` | branchiness of the merge (in-degree) | free (already materialized as `children[B].len()`) | — |
| 2 | `descendant_weight`, `convergence_jump` | magnitude of the cone *step* | one O(V+E) pass | fan-in is count-only |
| 3 | `parent_reconvergence` | lineage overlap (ancestors lost going up ≪ branching factor) | local k-hop | down→**up** direction |
| 4 | `ancestor_minhash` + `sketch_jaccard` / `sketch_overlap_lift` | overlap at **any** height, vs **chance** | O(V·k) precompute, O(k) read | unknown crossover height + small-world baseline |

The quantized-LCA consumer `caret_min_over_hubs(u, v, hubs)` minimises `caret_through_bridge`
over a *cheaply pre-selected* hub set: with every node a hub it equals `caret_distance_lca`
exactly; with a sparser set it is that caret quantized up by the gap `2·d(LCA→nearest hub)`.

## Key ideas worth a reviewer's eye

- **Rung 2 over-counts diamonds on purpose** — a shared descendant is counted once per path.
  That over-count is the price of avoiding distinct-set reachability; it ranks hubs fine but
  is not a true cone count. (This is exactly the recursive-top-down over-counting concern: an
  additive pass cannot see *where* the fan-ins are that cause it — which is why rung 4 switches
  to a set sketch that dedups by construction.)
- **Rung 4(a) height-agnosticism**: a fixed k-hop probe only sees a crossover within k hops,
  but the crossover height is unknown per node. The KMV/MinHash ancestor sketch summarizes the
  *whole* lineage to root in fixed size, so overlap is read at any depth with no knob (error
  ∝ 1/√k, not a depth cutoff).
- **Rung 4(b) small-world correction**: in a small-world graph up-cones overlap with
  near-certainty for *every* pair, so raw overlap measures the world, not the hub. `lift =
  observed |A∩B| / (|A|·|B|/N)` corrects against the configuration-model null; `>1` = real
  funnel, `≈1` = background. The test shows **identical overlap** giving `lift 1.11` (N=5) vs
  `10.0` (N=45) — same overlap, opposite verdict.
- This is the §6 **kernel-trick** move applied to set overlap: the ancestor *set* is the
  never-materialized feature map, the sketch its inner-product handle.

## Tests (all exact — `k=16` ≫ the tiny test sets, so sketches keep all hashes)

- `convergence_fanin_and_min_over_hubs` — fan-in counts, hub selection, and the
  all-nodes-hub-set == `caret_distance_lca` / sparse-set ≥ exact quantization limits.
- `descendant_weight_and_dropoff` — additive weight with the visible diamond over-count,
  dropoff magnitudes, cycle ⇒ `None` (fan-in stays cycle-robust).
- `parent_reconvergence_diamond_vs_tree` — same branching factor, opposite verdict; the k-hop
  ladder reaching a shared great-grandparent only at `up_hops=2`.
- `ancestor_sketch_height_agnostic_and_lift` — deep crossover caught knob-free; the
  small-world lift contrast; lone shared root = exactly chance; cycle ⇒ `None`.

Full suite: **70 passed, 0 failed**. Theory written up as rungs 1–4 in
`docs/design/WAM_RUST_GRAPH_FUNCTIONAL_SEMIRINGS.md` §5c.

## Known limitations (fair game for review)

- Rung 4's configuration-model null `|A|·|B|/N` assumes independent ancestor membership, which
  a strongly hierarchical graph violates — so `lift` **ranks** hubs well but its absolute value
  is uncalibrated. A degree-aware null is the natural follow-up.
- Rung 2/4 precomputes need a DAG (a reverse/forward topo order); `None` on cycles. Fan-in
  (rung 1) is the cycle-robust fallback; cyclic graphs want SCC-condensation first.

---

### DIFF
--- a/docs/design/WAM_RUST_GRAPH_FUNCTIONAL_SEMIRINGS.md
+++ b/docs/design/WAM_RUST_GRAPH_FUNCTIONAL_SEMIRINGS.md
@@ -565,8 +565,25 @@ caching, the **designated-bridge** measure is cacheable.
 >   (a 2-hop neighbourhood, no walk to root); `up_hops → ∞` is the exact deficit. `up_hops` is
 >   the cost/sensitivity knob, and the depth bound makes it cycle-safe. (`parent_reconvergence`,
 >   returning the overlap fraction in `[0,1]`.)
+> - **Rung 4 — ancestor sketch + small-world lift (height-agnostic, baseline-corrected).** The
+>   fixed-`up_hops` probe of rung 3 has a fatal flaw: it only sees a crossover *within* `k`
+>   hops, but the **crossover height is unknown and varies per node** — too small misses deep
+>   hubs, too large walks to root (the reachability we refuse). Worse, in a **small-world**
+>   graph the up-cones cover most of the graph within a few hops, so *raw* overlap (at any `k`,
+>   even exact) approaches 1 for **every** pair — it measures "are we in a small world," not "is
+>   `B` a funnel." Two fixes, both O(k): (a) **height-agnosticism** — summarize each node's
+>   *whole* lineage to root once, as a fixed-size **KMV/MinHash ancestor sketch**
+>   `sig(B) = bottom-k( {B} ∪ ⋃_p sig(p) )`, one root→leaf pass; overlap (`sketch_jaccard`) is
+>   then read at *any* depth with no knob (error ∝ 1/√k, not a depth cutoff). (b) **baseline
+>   correction** — a real hub reconverges *more than chance*: against the configuration-model
+>   null `E|A∩B| ≈ |A|·|B|/N`, the signal is `lift = observed |A∩B| / E|A∩B|` (`sketch_overlap_
+>   lift`), `>1` an excess funnel, `≈1` just small-world background. The sketch yields `|A|`,
+>   `|B|` *and* `|A∩B|` from the same reads, so height-agnostic detection and the small-world
+>   correction share one precompute. This is the §6 kernel trick literally applied: the ancestor
+>   *set* is the never-materialized feature map, the sketch its inner-product handle. (`None` on
+>   a cycle — SCC-condense first, since a cycle's nodes share their entire up-cone.)
 > - **The min-over-hubs caret is then quantized-LCA.** With hubs *cheaply* pre-selected (by
->   fan-in / jump / reconvergence, **no distances**), `caret_min_over_hubs(u, v, hubs) = minᵦ
+>   fan-in / jump / reconvergence / lift, **no distances**), `caret_min_over_hubs(u, v, hubs) = minᵦ
 >   caret_through_bridge(u, v, B)` picks the hub giving the least distance. The only distance
 >   work runs over the *already-chosen small* hub set — bounded by hub count, not by ranking the
 >   whole graph — so selection stays free and only the final min-pick costs anything. With
--- a/docs/design/WAM_RUST_GRAPH_FUNCTIONAL_SEMIRINGS.md
+++ b/docs/design/WAM_RUST_GRAPH_FUNCTIONAL_SEMIRINGS.md
@@ -551,16 +551,30 @@ caching, the **designated-bridge** measure is cacheable.
 >   `1`, a hub merging several heavy subtrees jumps large. Needs a DAG (a reverse-topo order);
 >   `None` on a cycle — so rung 1 is the cyclic-graph fallback. (`descendant_weight`,
 >   `convergence_jump`.)
+> - **Rung 3 — parent reconvergence (the ancestor-side definition).** The two rungs above read
+>   the *descendant* cone (down from `B`); the sharpest hub definition reads the *ancestor* cone
+>   (up from `B`). The signature: ascending one level from `B` loses **far fewer distinct
+>   ancestors than the parent branching factor `b = |parents(B)|` predicts**, because the
+>   parents' upward cones *overlap* — the lineage re-merges above. The deficit (expected-from-`b`
+>   minus actual) *is* the convergence; `b` large with a big deficit is a hub, `b` large with
+>   none is just a fan. The exact deficit is reachability again — and note the additive
+>   ancestor-weight is no help, it *is* the disjoint/branching-factor expectation, blind to
+>   overlap, so a separate overlap-sensitive probe is required. The cheap one is **local**:
+>   overlapping parents first share **grandparents**, so probe each parent's bounded `up_hops`-
+>   deep up-cone and measure their overlap. `up_hops = 1` is "do the parents share grandparents"
+>   (a 2-hop neighbourhood, no walk to root); `up_hops → ∞` is the exact deficit. `up_hops` is
+>   the cost/sensitivity knob, and the depth bound makes it cycle-safe. (`parent_reconvergence`,
+>   returning the overlap fraction in `[0,1]`.)
 > - **The min-over-hubs caret is then quantized-LCA.** With hubs *cheaply* pre-selected (by
->   fan-in / jump, **no distances**), `caret_min_over_hubs(u, v, hubs) = minᵦ caret_through_
->   bridge(u, v, B)` picks the hub giving the least distance. The only distance work runs over
->   the *already-chosen small* hub set — bounded by hub count, not by ranking the whole graph —
->   so selection stays free and only the final min-pick costs anything. With **every** node a
->   hub it equals `caret_distance_lca` exactly (the unquantized shortest-path caret); with a
->   sparser hub set it is that caret **quantized up to the nearest hub level**, larger by the
->   gap `2·d(LCA→nearest hub)` of §5b. Tightness (low, dense hubs → small gap) trades against
->   reuse (high, sparse hubs → one field serves more pairs) — and *that* knob, unlike the cone
->   size, is chosen with arithmetic we already paid for.
+>   fan-in / jump / reconvergence, **no distances**), `caret_min_over_hubs(u, v, hubs) = minᵦ
+>   caret_through_bridge(u, v, B)` picks the hub giving the least distance. The only distance
+>   work runs over the *already-chosen small* hub set — bounded by hub count, not by ranking the
+>   whole graph — so selection stays free and only the final min-pick costs anything. With
+>   **every** node a hub it equals `caret_distance_lca` exactly (the unquantized shortest-path
+>   caret); with a sparser hub set it is that caret **quantized up to the nearest hub level**,
+>   larger by the gap `2·d(LCA→nearest hub)` of §5b. Tightness (low, dense hubs → small gap)
+>   trades against reuse (high, sparse hubs → one field serves more pairs) — and *that* knob,
+>   unlike the cone size, is chosen with arithmetic we already paid for.
 
 ## 6. Aside: the kernel-trick analogy
 
--- a/docs/design/WAM_RUST_GRAPH_FUNCTIONAL_SEMIRINGS.md
+++ b/docs/design/WAM_RUST_GRAPH_FUNCTIONAL_SEMIRINGS.md
@@ -564,7 +564,9 @@ caching, the **designated-bridge** measure is cacheable.
 >   deep up-cone and measure their overlap. `up_hops = 1` is "do the parents share grandparents"
 >   (a 2-hop neighbourhood, no walk to root); `up_hops → ∞` is the exact deficit. `up_hops` is
 >   the cost/sensitivity knob, and the depth bound makes it cycle-safe. (`parent_reconvergence`,
->   returning the overlap fraction in `[0,1]`.)
+>   returning the *duplicate-mass fraction* `overlap/total_mass = o/(Σsᵢ)` in `[0,1]` — **not**
+>   a Jaccard `o/(Σsᵢ−o)`; for two identical cones it returns `1/2`, intentionally, to avoid
+>   Jaccard's denominator instability near `o ≈ Σsᵢ`.)
 > - **Rung 4 — ancestor sketch + small-world lift (height-agnostic, baseline-corrected).** The
 >   fixed-`up_hops` probe of rung 3 has a fatal flaw: it only sees a crossover *within* `k`
 >   hops, but the **crossover height is unknown and varies per node** — too small misses deep
@@ -608,9 +610,13 @@ adds `2` per level (§5b), so it can never win the `min`. This gives an *exact, 
 algorithm that **does not climb to the root**: expand the joint up-BFS from `u` and `v` in
 lockstep by radius `r`, and stop once the best matched sum `≤ r+1` (any *unmatched* common
 ancestor has far-side depth `≥ r+1`, hence sum `≥ r+1`, so it cannot beat the best). The search
-radius is bounded by the boundary depth `≈ caret/2`, not the height to root — on a tall stem
-with a low fork it touches three nodes where the full-cone `caret_distance_lca` touches the
-whole stem. (`caret_distance_lca_boundary[_counted]`.) This is the honest framing of §5c: the
+radius is bounded by `max(d(u→LCA*), d(v→LCA*))` — in the **balanced** case `≈ caret/2`, but for
+an **asymmetric** pair (one node *is* the LCA) the near side speculatively climbs *above* the
+LCA up to `≈ caret` before the stop fires (it cannot know it is the LCA until the far side
+arrives). In every case it stays near the common-ancestor space rather than the height to root —
+on a tall stem with a low fork it touches a handful of nodes where the full-cone
+`caret_distance_lca` touches the whole stem — but the **worst-case** node-visit count is still
+`O(V+E)` for graphs with wide upward frontiers. (`caret_distance_lca_boundary[_counted]`.) This is the honest framing of §5c: the
 global hub set is an *approximate, reusable stand-in* for this boundary, justified only when
 **batching many pairs** amortizes its precompute; for a one-off pair, just search the boundary.
 
@@ -636,6 +642,18 @@ Wikipedia — so "geometric mean of the top-4 singular values, times parent coun
 guess. **Caveat:** this is an *ad-hoc proposal*; the truncation rank, parents-vs-child-centroids,
 and the count/diversity weighting are all unvalidated, and it presumes a meaningful embedding.
 Recorded as a future direction, not a recommendation.
+
+**Known limitation of the rung-4 lift null (deep DAGs).** The configuration-model null
+`E|A∩B| ≈ |A|·|B|/N` assumes *independent* ancestor membership, which a strongly hierarchical
+DAG violates: ancestor-set sizes grow as `Θ(branching^depth)`, so for deep nodes `|A|·|B|/N` can
+**exceed** the actual intersection, driving `lift < 1` (or undefined) even for genuine hubs. The
+null is therefore calibrated only for shallow/sparse hierarchies; for deep, high-branching DAGs
+the **absolute** lift values are unreliable, though the **ranking** of hubs against each other
+stays usable (the bias is roughly monotone in depth). The Gene Ontology semantic-similarity
+literature avoids this with an **information-content** null instead — `IC(t) = −log₂ P(node
+annotated under t)`, with similarity read from the IC of the LCA (Resnik 1995; Lin 1998) — a
+depth-aware baseline that does not inflate. For calibrated absolute scores, an IC-style null is
+the principled replacement; for bridge *selection* (a ranking), the current lift suffices.
 
 ## 6. Aside: the kernel-trick analogy
 
--- a/docs/design/WAM_RUST_GRAPH_FUNCTIONAL_SEMIRINGS.md
+++ b/docs/design/WAM_RUST_GRAPH_FUNCTIONAL_SEMIRINGS.md
@@ -521,6 +521,47 @@ caching, the **designated-bridge** measure is cacheable.
 > "common ancestor") is **robust** — pick the union hubs and go — while the nested-cut
 > factorisation (needs "dominator") is **fragile**, which is the deeper reason it is deferred.
 
+> **Defining "downward convergence" *cheaply* — and *non-circularly*.** Bridge selection (1)
+> asks us to score nodes by how much the hierarchy funnels through them. The honest constraint
+> is sharper than "make it fast": the score must **not call the very upward distance work the
+> bridges exist to amortize**. The natural definition — *descendant-cone size*, `|desc(B)|` —
+> needs reachability (a global per-node traversal); and ranking candidate bridges by
+> `caret_through_bridge` is the same circularity wearing a hat (it runs `up_distance_to` at
+> every candidate). Both are rejected: they spend exactly what hubs were meant to save. What is
+> left must be **structural and local**, readable *before* any query.
+>
+> - **The true signal is a cone-size *step*, not a large cone.** A hub is where the descendant
+>   cone drops off sharply as you descend through it: large at `B`, but each child just below
+>   carries only a slice. That dropoff *is* the right characterization — and its exact form
+>   (cone size at every node) is the reachability we are refusing. So we take its **cheap
+>   shadows**, in two rungs.
+> - **Rung 1 — local fan-in (free, cycle-robust).** `fanin[B] = #{v : B ∈ parents[v]}`, the
+>   in-degree in the child→parent graph. It is the *first derivative* of the dropoff: it counts
+>   *how many* cones merge at `B` in one hop without summing their sizes. One pass over the
+>   `parents` map we already hold — on the boundary-sweep path it is just `children[B].len()`,
+>   already materialized, so the score costs **nothing extra**. It reads no distance, so it
+>   cannot be circular, and it is well-defined even with cycles. Its blind spot: *branchiness
+>   only* — two giant subtrees merging is `fanin = 2` yet a huge step.
+>   (`convergence_fanin`, `hubs_by_fanin`.)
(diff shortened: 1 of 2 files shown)


templates/targets/rust_wam/boundary_cache.rs.mustache
```diff
@@ -624,6 +624,87 @@ pub fn caret_distance_budgeted(
...
+///
+/// `caret_distance_lca` builds the *entire* up-cone of `u` and of `v` and intersects — it
+/// always walks to root, even when the lineages merge one hop up. But the minimum
+/// `min_B (d(u→B)+d(v→B))` is achieved on the **merge frontier** (a common ancestor with a
+/// child still in a single lineage); every node above the frontier only adds `2` per level, so
+/// it can never win the `min`. So expand both BFSs in lockstep by radius `r` and stop once the
+/// best matched sum `≤ r+1`: any *unmatched* common ancestor has its far-side depth `≥ r+1`,
+/// hence sum `≥ r+1`, so it cannot beat the best. The search radius is bounded by the boundary
+/// depth (`≈ caret/2`), not the height to root — the search-space reduction. Returns
+/// `(caret, nodes_visited)`; the count is what the test uses to show the cone is never entered.
```
s243a: **[required — doc precision]** The claim "radius is bounded by the boundary depth (`≈ caret/2`)" is accurate only for the *balanced* case (both query nodes equidistant from their LCA). In the asymmetric case — e.g. querying `(50, 100)` on a chain `0←1←…←100` where the LCA is node 50 — the search expands from node 50 upward as the other side climbs from 100, running to radius ≈ `caret` (not `caret/2`), and visits the stem above the LCA.

The search is still much better than full-cone expansion in the common case, but the per-side bound is `max(d(u→LCA*), d(v→LCA*))`, not `≈ caret/2`. And worst-case complexity remains O(V+E) for wide-frontier graphs.

Suggested rewording:
> "The search radius is bounded by `max(d(u→LCA*), d(v→LCA*))` — in the balanced case `≈ caret/2`, in the worst case up to `caret`. Worst-case node-visit count is still O(V+E) for wide frontiers, but on narrow upward cones with a low fork this is dramatically less than full expansion."

Also add a test pairing `(50, 100)` on a tall chain to pin the asymmetric behaviour and document the visitation count.

templates/targets/rust_wam/boundary_cache.rs.mustache
```diff
@@ -624,6 +624,87 @@ pub fn caret_distance_budgeted(
...
+) -> (Option<u32>, usize) {
+    use std::collections::{HashMap, HashSet};
+    let mut du: HashMap<u32, u32> = HashMap::new();
+    let mut dv: HashMap<u32, u32> = HashMap::new();
+    du.insert(u, 0);
+    dv.insert(v, 0);
+    let mut fu: Vec<u32> = vec![u]; // u-side frontier at depth r
+    let mut fv: Vec<u32> = vec![v];
+    let mut best: Option<u32> = None;
+    let relax = |best: &mut Option<u32>, s: u32| {
```
s243a, line 658: **[non-blocking — comment precision]** The doc-comment says "e.g. `u == v`, *or one is the other*" but the r=0 bootstrap (`dv.get(&u)`, `du.get(&v)`) only fires for `u == v`. If `u` is a strict ancestor of `v`, neither lookup fires at r=0; the BFS correctly finds the match later as the `v`-side climbs to `u`. The value is right; only the comment oversells what the bootstrap catches.

Suggest narrowing: `// r = 0: u == v matches at sum 0.`

templates/targets/rust_wam/boundary_cache.rs.mustache
```diff
@@ -675,6 +756,334 @@ pub fn up_distance_to(
...
+        .collect();
+    let mut q: VecDeque<u32> = pending
+        .iter()
+        .filter(|&(_, &c)| c == 0)
+        .map(|(&n, _)| n)
+        .collect();
+    let mut w: HashMap<u32, u64> = HashMap::new();
+    while let Some(b) = q.pop_front() {
+        let sum: u64 = children
+            .get(&b)
```
s243a, line 925: **[required — safety]** The additive weight `w(B) = 1 + Σ w(c)` over-counts diamonds by path multiplicity. For a chain of `n` stacked diamonds the weight grows exponentially (≈ 2^(n/2)); around 64 stacked diamonds a `u64` overflows. In release builds this wraps silently and corrupts scores; in debug it panics.

The PR's own doc-comment acknowledges "the diamond over-count can grow fast" but provides no guard. For the intended type/category hierarchies (Wikipedia-scale, modest fan) `u64` is more than sufficient, but the invariant should be enforced.

Choose one:
```rust
// Option A: saturating (score saturates rather than wrapping/panicking)
let sum: u64 = children
    .get(&b)
    .map_or(0, |cs| cs.iter().map(|c| w[c]).fold(0u64, |a, x| a.saturating_add(*x)));
w.insert(b, 1u64.saturating_add(sum));
```
or document the precondition explicitly in the doc-comment:
> "Precondition: the maximum diamond-nesting depth must be < ~60 to avoid u64 overflow; for deeper adversarial DAGs use saturating arithmetic."

templates/targets/rust_wam/boundary_cache.rs.mustache
```diff
@@ -675,6 +756,334 @@ pub fn up_distance_to(
...
+/// fan. The exact deficit needs the parents' full ancestor sets (reachability — the global
+/// work we refuse), and the additive ancestor-weight is no help (it *is* the disjoint /
+/// branching-factor expectation, blind to overlap). But reconvergence shows up **locally**:
+/// overlapping parents first share **grandparents**. So we probe the bounded `up_hops`-deep
+/// up-cone of each parent and measure their overlap — `up_hops = 1` is "do the parents share
+/// grandparents," a 2-hop neighbourhood with no walk to root; `up_hops → ∞` is the exact
+/// deficit (the expensive limit). `up_hops` is the cost/sensitivity knob.
+///
+/// Returns the **overlap fraction** in `[0, 1]`: `(Σ|cone_p| − |⋃ cone_p|) / Σ|cone_p|`. `0`
+/// for disjoint lineages (tree-like, no convergence) or fewer than two parents (nothing to
```
s243a: **[required — doc fix: semantics of the overlap formula]** The formula `(Σ|cone_p| − |⋃ cone_p|) / Σ|cone_p|` is a **duplicate-mass fraction**, not a Jaccard index. For two parent cones of sizes `s1`, `s2` with intersection `o`:
- Implemented: `o / (s1 + s2)`
- Jaccard: `o / (s1 + s2 − o)`

For two *identical* cones the implemented measure returns `1/2`, not `1`. For `b` identical cones it approaches `1 − 1/b`. This is a deliberate and reasonable semantic choice (it measures "what fraction of per-parent observations were redundant"), but the current wording "overlap fraction in `[0,1]`" and "`up_hops → ∞` is the exact deficit" reads like a Jaccard-style quantity.

Please add to the doc-comment:
> "Note: the formula computes `overlap / total_mass = o/(Σsᵢ)`, the *duplicate-mass fraction*. This differs from the Jaccard index `o/(Σsᵢ − o)`; for identical cones it returns `1/2` (not `1`). The choice is intentional: it measures the fraction of per-parent ancestor observations that were duplicated, which ranks hubs without the Jaccard denominator instability near `o ≈ Σsᵢ`. Do not 'fix' this to Jaccard."

templates/targets/rust_wam/boundary_cache.rs.mustache
```diff
@@ -675,6 +756,334 @@ pub fn up_distance_to(
...
+        let mut seen: HashSet<u32> = HashSet::new();
+        let mut q: VecDeque<(u32, usize)> = VecDeque::new();
+        if let Some(gps) = parents.get(&p) {
+            for &g in gps {
+                if seen.insert(g) { q.push_back((g, 1)); }
+            }
+        }
+        while let Some((x, depth)) = q.pop_front() {
+            if depth >= up_hops { continue; }
+            if let Some(xps) = parents.get(&x) {
```
s243a, line 1018: **[required — bug: `up_hops=0` silently behaves like `up_hops=1`]** The BFS unconditionally inserts each parent's parents into `seen` in the outer setup loop (lines before the `while` loop), then only checks `if depth >= up_hops { continue; }` *inside* the BFS expansion. This means `up_hops=0` still visits one grandparent layer — identical to `up_hops=1`.

Fix:
```rust
pub fn parent_reconvergence(
    node: u32,
    parents: &std::collections::HashMap<u32, Vec<u32>>,
    up_hops: usize,
) -> f64 {
    if up_hops == 0 { return 0.0; } // zero hops: no ancestors visible, nothing to reconverge
    // ... rest of function unchanged
}
```
Also add a test: `assert_eq!(parent_reconvergence(3, &hub, 0), 0.0);`

templates/targets/rust_wam/boundary_cache.rs.mustache
```diff
@@ -675,6 +756,334 @@ pub fn up_distance_to(
...
+
+/// **Quantized-LCA caret**: minimise `caret_through_bridge` over a *pre-selected* hub set.
+///
+/// This is the "choose the hub that gives the minimum distance" primitive. Crucially the only
+/// distance work runs over the hubs `hubs_by_fanin` already chose — bounded by the hub count,
+/// not by the graph — so the expensive part is paid once, on a small set, not per candidate
+/// across the whole DAG. With **every** node a hub it equals `caret_distance_lca` exactly (the
+/// unquantized shortest-path caret); with a sparser hub set it returns an upper bound, the
+/// LCA caret *quantized to the nearest hub level*, larger by the gap `2·d(LCA→nearest hub)`.
+/// `None` if no hub is a common ancestor of both `u` and `v`.
```
s243a: **[required — doc qualification: quantized-LCA gap]** The claim "the LCA caret *quantized to the nearest hub level*, larger by the gap `2·d(LCA→nearest hub)`" is exactly true only when (a) there is a unique LCA and (b) the shortest paths from both query nodes to the chosen hub pass *through* that unique LCA. In a general DAG with multiple LCAs or divergent paths this need not hold.

Recommend replacing the unconditional gap statement with:
> "With a sparser hub set, `caret_min_over_hubs` returns an upper bound on `caret_distance_lca`. If a selected hub `h` lies above the optimal LCA `l` and both shortest paths `u→h` and `v→h` pass through `l`, the gap is exactly `2·d(l, h)`; in general the result is a valid upper bound but the gap expression may not simplify."

The test on lines 462–469 correctly demonstrates the gap for this specific tree; the doc should clarify it is a tree-like example, not a general DAG guarantee.

templates/targets/rust_wam/boundary_cache.rs.mustache
```diff
@@ -675,6 +756,334 @@ pub fn up_distance_to(
...
+/// to amortize; ranking candidate bridges by `caret_through_bridge` is the same circularity
+/// (it calls `up_distance_to` at every candidate). We refuse both. The **local in-degree**,
+/// `fanin[B] = #{v : B ∈ parents[v]}`, is the structural proxy: it never reads a distance, it
+/// is a property of the node (not of a `(u,v)` pair), and it falls out of one pass over the
+/// `parents` map we already hold — on the boundary-sweep path it is just `children[B].len()`,
+/// already materialized, so the score costs nothing extra. The trade is explicit: fan-in is
+/// the *local* convergence signal, not the true descendant funnel; we accept the proxy
+/// precisely because the exact funnel is the computation we are trying to avoid.
+pub fn convergence_fanin(
+    parents: &std::collections::HashMap<u32, Vec<u32>>,
```
s243a, line 783: **[required — doc: absent-means-zero semantics]** `convergence_fanin` only inserts nodes that are *somebody's parent*; pure leaves (in-degree 0 in the child→parent graph) are **absent from the returned map, not present with value 0**. The test on line 445 pins this: `fanin.get(&8) == None`.

This is internally consistent but is a latent foot-gun for consumers who `unwrap()` or match without a default. Please add to the doc-comment:
> "Nodes with zero fan-in (pure leaves in the parent-pointer graph) are **absent** from the returned map; callers should treat `None` as fan-in = 0."

templates/targets/rust_wam/boundary_cache.rs.mustache
```diff
@@ -2466,6 +2875,227 @@ mod tests {
...
+        let mut cyc: HashMap<u32, Vec<u32>> = HashMap::new();
+        cyc.insert(0, vec![1]); cyc.insert(1, vec![0]); cyc.insert(2, vec![0, 1]);
+        let _ = parent_reconvergence(2, &cyc, 10); // terminates; value not asserted
+    }
+
+    // Ancestor MinHash sketch: height-agnostic overlap + the small-world lift correction.
+    // k=16 >> every set here, so the sketches are EXACT (they keep all hashes) and the
+    // Jaccard/cardinality/intersection reads are exact -> deterministic assertions.
+    #[test]
+    fn ancestor_sketch_height_agnostic_and_lift() {
```
s243a, line 3098: **[required — missing test: saturated MinHash regime]** All sketch tests use `K=16` which far exceeds the tiny test sets (max 5 nodes), so the `s.len() < k` (exact) branch of `sketch_card` is always taken — the `(k-1)/h_k` estimator branch is **never exercised** by the test suite.

Add a test with `k` smaller than the ancestor set:
```rust
#[test]
fn ancestor_sketch_saturated_regime() {
    // Build a 20-node chain so ancestor sets of deep nodes have >4 members.
    let mut chain: HashMap<u32, Vec<u32>> = HashMap::new();
    for i in 1u32..=20 { chain.insert(i, vec![i - 1]); }
    const K: usize = 4; // force saturation: node 20 has 20-node ancestor set > K
    let sig = ancestor_minhash(&chain, K).unwrap();
    // sig[20] should have exactly K=4 entries (saturated)
    assert_eq!(sig[&20].len(), K);
    // sketch_card estimate should be in a reasonable range of the true 20
    let est = sketch_card_exposed(&sig[&20], K); // expose sketch_card for test, or inline
    assert!(est > 10.0 && est < 50.0, "KMV estimate {} out of range for n≈20, k={}", est, K);
}
```
(Either make `sketch_card` pub(crate) for testing or inline the formula.)

templates/targets/rust_wam/boundary_cache.rs.mustache
```diff
@@ -2466,6 +2875,227 @@ mod tests {
...
+        // Cyclic graph -> no topo order -> None (SCC-condense first).
+        let mut cyc: HashMap<u32, Vec<u32>> = HashMap::new();
+        cyc.insert(0, vec![1]); cyc.insert(1, vec![0]);
+        assert_eq!(ancestor_minhash(&cyc, K), None);
+    }
+
+    // The boundary-restricted caret equals the full caret everywhere, but stops at the mixing
+    // boundary instead of walking both cones to root.
+    #[test]
+    fn caret_boundary_equals_full_but_visits_fewer() {
```
s243a, line 3150: **[required — missing test: asymmetric boundary expansion]** Every boundary test uses pairs where both nodes are equidistant from their LCA (`d(u→LCA) = d(v→LCA) = 1`). The `≈ caret/2` claim and the stopping-condition correctness in the asymmetric case are untested.

Add:
```rust
#[test]
fn caret_boundary_asymmetric() {
    // Chain 0←1←2←3←4←5; query (1, 5). LCA=1, d(1→1)=0, d(5→1)=4, caret=4.
    // Boundary search must still find caret=4 correctly.
    let mut chain: HashMap<u32, Vec<u32>> = HashMap::new();
    for c in 1u32..=5 { chain.insert(c, vec![c - 1]); }
    assert_eq!(caret_distance_lca_boundary(1, 5, &chain), caret_distance_lca(1, 5, &chain));
    // Also verify the asymmetric stem-expansion: visited includes the path from 5 up to 1.
    let (caret, visited) = caret_distance_lca_boundary_counted(1, 5, &chain);
    assert_eq!(caret, Some(4));
    // The u-side (node 1) is the LCA itself; the v-side (node 5) must climb 4 hops.
    // visited should be at most {1,2,3,4,5} = 5 nodes, NOT the nodes below 1 (i.e. 0).
    assert!(visited <= 5, "should not visit below the LCA (node 0); visited={}", visited);
}
```

docs/design/WAM_RUST_GRAPH_FUNCTIONAL_SEMIRINGS.md
```diff
@@ -521,6 +521,122 @@ caching, the **designated-bridge** measure is cacheable.
...
+>   minus actual) *is* the convergence; `b` large with a big deficit is a hub, `b` large with
+>   none is just a fan. The exact deficit is reachability again — and note the additive
+>   ancestor-weight is no help, it *is* the disjoint/branching-factor expectation, blind to
+>   overlap, so a separate overlap-sensitive probe is required. The cheap one is **local**:
+>   overlapping parents first share **grandparents**, so probe each parent's bounded `up_hops`-
+>   deep up-cone and measure their overlap. `up_hops = 1` is "do the parents share grandparents"
+>   (a 2-hop neighbourhood, no walk to root); `up_hops → ∞` is the exact deficit. `up_hops` is
+>   the cost/sensitivity knob, and the depth bound makes it cycle-safe. (`parent_reconvergence`,
+>   returning the overlap fraction in `[0,1]`.)
+> - **Rung 4 — ancestor sketch + small-world lift (height-agnostic, baseline-corrected).** The
```
s243a, line 570: **[required — doc precision: overlap formula semantics]** The phrase "returning the overlap fraction in `[0,1]`" implies a Jaccard-like measure, but `parent_reconvergence` computes the duplicate-mass fraction `o/(s1+s2)`, not Jaccard `o/(s1+s2-o)`. For identical parent cones the score is `1/2`, not `1`.

Suggest adding a parenthetical:
> "(`parent_reconvergence`, returning the *duplicate-mass fraction* `overlap/total_mass` in `[0,1]` — not a Jaccard; for identical cones it returns `1/2`.)"

docs/design/WAM_RUST_GRAPH_FUNCTIONAL_SEMIRINGS.md
```diff
@@ -521,6 +521,122 @@ caching, the **designated-bridge** measure is cacheable.
...
+the **common-ancestor space** `CA(u,v) = anc(u) ∩ anc(v)`, which is upward-closed (once the two
+lineages mix, everything above is common). The minimum caret is achieved on its **lower
+boundary** — the lowest common ancestors, i.e. a node that is "mixed" (both lineages reach it)
+yet has **at least one child still in a single lineage**. Every node above the boundary only
+adds `2` per level (§5b), so it can never win the `min`. This gives an *exact, precompute-free*
+algorithm that **does not climb to the root**: expand the joint up-BFS from `u` and `v` in
+lockstep by radius `r`, and stop once the best matched sum `≤ r+1` (any *unmatched* common
+ancestor has far-side depth `≥ r+1`, hence sum `≥ r+1`, so it cannot beat the best). The search
+radius is bounded by the boundary depth `≈ caret/2`, not the height to root — on a tall stem
+with a low fork it touches three nodes where the full-cone `caret_distance_lca` touches the
```
s243a: **[required — doc precision: boundary-depth claim]** "The search radius is bounded by the boundary depth (`≈ caret/2`)" is accurate for balanced LCAs but misleading for asymmetric ones (one query node much closer to the LCA than the other). The per-side radius at termination is `max(d(u→LCA*), d(v→LCA*))`, which equals `caret/2` only when the depths are equal.

Suggest:
> "The search radius is bounded by `max(d(u→LCA*), d(v→LCA*))` — in the balanced case this is `caret/2`; for an asymmetric pair where one node *is* the LCA the radius approaches `caret`. In either case the search never climbs beyond the common-ancestor space, so it is strictly better than full-cone expansion on tall, narrow hierarchies."

Also note that worst-case node-visit count is still O(V+E) for graphs with wide upward frontiers (e.g. a bipartite-like layer immediately above both query nodes).

docs/design/WAM_RUST_GRAPH_FUNCTIONAL_SEMIRINGS.md
```diff
@@ -521,6 +521,122 @@ caching, the **designated-bridge** measure is cacheable.
...
+whole stem. (`caret_distance_lca_boundary[_counted]`.) This is the honest framing of §5c: the
+global hub set is an *approximate, reusable stand-in* for this boundary, justified only when
+**batching many pairs** amortizes its precompute; for a one-off pair, just search the boundary.
+
+**Global hub measure — deferred, and the following is a *conjecture*, not a result.** The
+global question ("rank all nodes as generic bridges") is harder, because — as the §5c rungs
+keep running into — *some* fan-in is near-universal (any multi-parent node reconverges
```
s243a, line 625: **[non-blocking — known-limitations expansion]** The §5d known-limitations section (in the PR description, and implicitly here) correctly flags that the configuration-model null `|A|·|B|/N` assumes independent ancestor membership. Deep research confirms this is a known failure mode with published precedent:

In deep DAG hierarchies ancestor set sizes grow as Θ(branching_factor^depth), so `|A|·|B|/N` can exceed the actual intersection for deep nodes — making lift undefined or negative. The Gene Ontology semantic similarity literature (Resnik 1995, Lin 1998) uses the information-content null instead:
> IC(t) = −log₂ P(random node annotated under t)

and measures similarity via IC of the LCA — a depth-aware null that does not suffer from this inflation.

The PR's lift measure is still useful for *ranking* hubs relative to each other (the ordering is consistent even if the absolute values are miscalibrated). But the known-limitations note should explicitly say: "the configuration-model null is calibrated for shallow/sparse hierarchies; for deep DAGs with high branching factor, expected overlap can exceed actual overlap, making absolute lift values unreliable — use IC-based alternatives for calibrated absolute scores."

s243a (commented):
## Multi-model code review — WAM-Rust hub-convergence metrics (Rungs 1–4)

**Reviewers:** Claude Opus 4.8 · GPT-5.5 · Gemini deep research (KMV/MinHash theory, DAG LCA literature, network-science overlap measures)

**Verdict: approve after required fixes. No blocking correctness issues.**

The mathematics is sound. The non-circularity invariant holds rigorously — no hub-scoring function touches `up_distance_to` or any distance primitive. The KMV cardinality formula `(k-1)/h_k` is the correct unbiased form (confirmed by Bar-Yossef et al. and Beyer et al.). The bottom-k Jaccard estimator is the standard Cohen & Kaplan (2007) one-pass form. The boundary-BFS stopping condition `bst ≤ r+1` is proved tight analytically and verified on 9,750 random-DAG pairs with zero mismatches.

Eight required items exist; all are documentation clarifications, one one-liner safety guard, one one-line early return, or test additions. One API gap (rung 2–4 selection helpers) can ship in a follow-up PR.

s243a:
## Review fixes — all 9 required items + the non-blocking ones addressed (commit `a74517f2`)

Thanks for the thorough Council-of-Models pass. Disposition:

**Code / safety**
- **R1** — `descendant_weight` now uses `saturating_add` (and `convergence_jump` `saturating_sub`); adversarial stacked diamonds pin at `u64::MAX` instead of wrapping/panicking. Growth (`~2^(n/2)`) and the saturation behaviour are documented.
- **R3** *(genuine bug)* — `parent_reconvergence(·, 0)` silently behaved like `up_hops=1`. Early-return `0.0`; test pins `0` at hops 0 vs `0.5` at hops 1.

**Doc precision**
- **R2** — documented as the *duplicate-mass fraction* `o/(s₁+s₂)`, **not** Jaccard; identical cones → `½`; explicit "do not fix to Jaccard." Same note added to the §5c rung-3 bullet.
- **R4** — `convergence_fanin` absent-means-zero documented (pure leaves omitted; treat `None` as 0).
- **R5** — quantized gap `2·d(LCA→hub)` qualified to the unique-LCA / paths-through-LCA (tree) case; general DAG is a valid upper bound only.
- **R6** — `≈ caret/2` corrected to `max(d(u→LCA*), d(v→LCA*))` (balanced `≈ caret/2`, asymmetric up to `≈ caret`, worst case `O(V+E)`), in both the doc-comment and §5d.
- **Deep-research finding** — §5d now records the rung-4 lift-null limitation on deep DAGs (`|A|·|B|/N` can exceed actual intersection) and the IC-based (Resnik 1995 / Lin 1998) alternative for calibrated absolute scores; ranking stays usable.

**Tests**
- **R7** — `caret_boundary_asymmetric`: `u` is the LCA, near side speculatively climbs `~caret` above it but not to root (visits 5 of 23 nodes). This corrected a wrong *assertion* of mine, not a code path — the behaviour is exactly the asymmetric `~caret` climb R6 describes.
- **R8** — `ancestor_sketch_saturated_regime` (`k=4 < |set|`) now exercises the `(k-1)/h_k` KMV estimator branch that was dead in the suite.

**API**
- **R9** — added `hubs_by_jump` (rung 2, `None` on cycles) and `hubs_by_lift` (rung 4, sketch-based) so rungs 2/4 feed `caret_min_over_hubs` like rung 1; both tested.

Also narrowed the `r=0` bootstrap comment in the boundary search (only `u==v` matches there). Full suite: **73 passed, 0 failed.**

Resolving the inline threads accordingly.

---

PR #3259, opened 2026-06-18
WAM-Rust: IC-based semantic similarity (Resnik/Lin) via a descendant sketch

s243a:
## What & why

#3256 documented (in §5d) that the rung-4 `sketch_overlap_lift` is **miscalibrated on deep
DAGs**: its configuration-model null `E|A∩B| ≈ |A|·|B|/N` assumes independent ancestor
membership, which a hierarchy violates — so for deep nodes the expected overlap can exceed the
actual, and absolute lift values go unreliable (the *ranking* survives, the *calibration* does
not). This PR implements the principled fix flagged there: the **information-content** null from
the Gene Ontology semantic-similarity literature (Resnik 1995, Lin 1998).

It falls out as the **downward mirror** of the rung-4 ancestor sketch — same machinery, pointed
at descendants instead of ancestors:

- **`descendant_minhash`** — one reverse-topological pass gives each node a fixed-`k` KMV sketch
  of its descendant cone. Being a *set*, it **dedups by construction**, so `sketch_card`
  estimates the **distinct** cone size `|desc(t)|` — exactly the quantity `descendant_weight`
  (rung 2) over-counts, and exactly what a calibrated frequency needs. `None` on cycles.
- **`information_content`** — `IC(t) = −log₂(|desc(t)| / N)`; high for specific nodes, `0` at the
  root. Uses *actual* cone frequencies → no deep-DAG inflation.
- **`resnik_similarity`** — `IC(MICA)`, the most informative common ancestor (max `IC` over
  common ancestors, which is a lowest one since `IC` is non-increasing upward).
- **`lin_similarity`** — `2·IC(MICA)/(IC(u)+IC(v)) ∈ [0,1]`.

Cost mirrors rung 4 exactly: `O(V·k)` precompute, `O(k)` reads, **no per-query reachability**.

## Scope (deliberately bounded)

This fixes the **calibration of the relatedness read-out** between two nodes. It does **not**
touch hub *selection* — *which* nodes make good generic bridges — which remains the open global
problem (§5d). Selection is a ranking; this is a read-out.

## Tests

`ic_resnik_lin_similarity` (`k=16`, so sketches are exact and the asserts are exact) on the
balanced tree `0; 1,2→0; 3,4→1; 5,6→2`:
- distinct cone sizes `7 / 3 / 1`; the three IC values (`0`, `log₂(7/3)`, `log₂7`);
- `Resnik(3,4) = IC(LCA=1)`, `Resnik(3,5) = 0` (only-root), self-similarity `= IC(self)`;
- `Lin(3,4)` value + `[0,1]`, only-root `→ 0`, identical `→ 1`;
- no-shared-ancestor `→ None`, cyclic graph `→ None`.

Full suite: **75 passed, 0 failed.**

## Docs

- §5d updated to mark the IC null **implemented** (with the descendant-sketch construction).
- **New §5e** — a from-scratch *primer* on IC similarity with worked numbers (rarity-as-
  information, why `−log₂`, Resnik, Lin, and why the dedup-ing sketch is required rather than the
  additive weight). Written to be learnable, not just decodable.

---

### DIFF
--- a/docs/design/WAM_RUST_GRAPH_FUNCTIONAL_SEMIRINGS.md
+++ b/docs/design/WAM_RUST_GRAPH_FUNCTIONAL_SEMIRINGS.md
@@ -660,6 +660,19 @@ annotated under t)`, with similarity read from the IC of the LCA (Resnik 1995; L
 depth-aware baseline that does not inflate. For calibrated absolute scores, an IC-style null is
 the principled replacement; for bridge *selection* (a ranking), the current lift suffices.
 
+**This IC null is now implemented** (`information_content`, `resnik_similarity`,
+`lin_similarity`). It rests on a **descendant sketch** — `descendant_minhash`, the exact
+downward mirror of the rung-4 `ancestor_minhash`: one reverse-topological pass gives each node a
+fixed-`k` KMV sketch of its descendant cone, which (being a *set*) **dedups by construction**, so
+`sketch_card` estimates the *distinct* cone size `|desc(t)|` that `descendant_weight` over-counts.
+Then `IC(t) = −log₂(|desc(t)|/N)`, `resnik = IC(MICA)` (the most informative common ancestor —
+max `IC` over the common ancestors, which is a lowest one since `IC` is non-increasing upward),
+and `lin = 2·IC(MICA)/(IC(u)+IC(v)) ∈ [0,1]`. The cost mirrors rung 4 exactly: `O(V·k)`
+precompute, `O(k)` reads, no per-query reachability. Unlike the configuration-model lift it uses
+*actual* cone frequencies, so it stays calibrated on deep DAGs — the principled absolute-score
+companion to the lift's ranking signal. (Hub *selection* from these scores is still the open
+global problem; this only fixes the calibration of the relatedness read-out.)
+
 ## 6. Aside: the kernel-trick analogy
 
 *(A mnemonic, not load-bearing — the mechanics above stand on their own; skip if you only
--- a/docs/design/WAM_RUST_GRAPH_FUNCTIONAL_SEMIRINGS.md
+++ b/docs/design/WAM_RUST_GRAPH_FUNCTIONAL_SEMIRINGS.md
@@ -666,12 +666,17 @@ downward mirror of the rung-4 `ancestor_minhash`: one reverse-topological pass g
 fixed-`k` KMV sketch of its descendant cone, which (being a *set*) **dedups by construction**, so
 `sketch_card` estimates the *distinct* cone size `|desc(t)|` that `descendant_weight` over-counts.
 Then `IC(t) = −log₂(|desc(t)|/N)`, `resnik = IC(MICA)` (the most informative common ancestor —
-max `IC` over the common ancestors, which is a lowest one since `IC` is non-increasing upward),
-and `lin = 2·IC(MICA)/(IC(u)+IC(v)) ∈ [0,1]`. The cost mirrors rung 4 exactly: `O(V·k)`
-precompute, `O(k)` reads, no per-query reachability. Unlike the configuration-model lift it uses
-*actual* cone frequencies, so it stays calibrated on deep DAGs — the principled absolute-score
-companion to the lift's ranking signal. (Hub *selection* from these scores is still the open
-global problem; this only fixes the calibration of the relatedness read-out.)
+max `IC` over the common ancestors, which for *exact* IC is a lowest one since `IC` is
+non-increasing upward; note the MICA can be non-unique, but we return the IC *value*, so ties are
+immaterial, and on *saturated* sketches the estimated-max node may not be the exact MICA),
+and `lin = 2·IC(MICA)/(IC(u)+IC(v)) ∈ [0,1]`, **undefined (→ `None`) when both nodes are the
+root** (`IC = 0`, a `0/0` ratio). The cost is split: the **sketch + IC** are the `O(V·k)`
+precompute / `O(k)` read; `resnik`/`lin` then add a **per-query `O(V+E)` ancestor BFS** to find
+the MICA (it is *not* an `O(k)` read — no DAG library does sub-`O(V+E)` MICA without all-pairs
+precompute). Unlike the configuration-model lift it uses *actual* cone frequencies, so it stays
+calibrated on deep DAGs — the principled absolute-score companion to the lift's ranking signal.
+(Hub *selection* from these scores is still the open global problem; this only fixes the
+calibration of the relatedness read-out.)
 
 ### 5e. A gentle primer on information-content similarity (for the reader learning this)
 
@@ -713,9 +718,14 @@ up — the maximum IC is at the bottom of the shared region, the merge frontier 
 
 **Lin similarity: normalize so "identical" scores 1.** Raw Resnik isn't on a fixed scale — a deep
 tree gives big IC numbers, a shallow one small. Lin divides by how specific the two items
-themselves are: `sim(u,v) = 2·IC(MICA) / (IC(u) + IC(v))`. If `u = v`, the MICA *is* `u`, so it is
-`2·IC(u)/2·IC(u) = 1`; if they share only the root, `IC(MICA)=0` so it is `0`. On the example,
-`Lin(3,4) = 2(1.22)/(2.81+2.81) = 0.43`. Now every pair sits in `[0,1]`, comparable across graphs.
+themselves are: `sim(u,v) = 2·IC(MICA) / (IC(u) + IC(v))`. If `u = v` (and `IC(u) > 0`) the MICA
+*is* `u`, so it is `2·IC(u)/2·IC(u) = 1`; if they share only the root, `IC(MICA)=0` so it is `0`.
+On the example, `Lin(3,4) = 2(1.22)/(2.81+2.81) = 0.43`. Now every pair sits in `[0,1]`, comparable
+across graphs. **One exception:** when both nodes are the root, `IC(u)=IC(v)=0`, the denominator is
+`0`, and Lin is undefined (`0/0`) — the implementation returns `None` there (while Resnik returns
+`0`). So the "identical → 1" identity holds for every node *except* the root. (`information_content`
+is an `O(k)` read, but calling `resnik`/`lin` is *not* `O(k)` — each runs a per-query `O(V+E)`
+upward BFS to find the MICA; cache the ancestor sets for repeated queries on the same graph.)
 
 **Why we needed the descendant *sketch* (and not the additive weight).** Every formula above needs
 `|desc(t)|`, the **distinct** count of nodes under `t`. Computing that exactly for all `t` is
@@ -723,9 +733,11 @@ reachability — the global blow-up we keep refusing. The cheap one-pass additiv
 (rung 2) is no good *here*: it counts a node reachable by two paths **twice**, so it inflates
 `|desc|`, distorts `p`, and would corrupt the IC. The fix is a **set**: `descendant_minhash` keeps
 a fixed-`k` sample of the cone, and a set automatically counts each member once — so its size
-estimate is the *distinct* `|desc|` we need. Same `O(V·k)` precompute / `O(k)` read as the rung-4
-ancestor sketch, just pointed downward. (And it is the §6 kernel-trick move once more: the cone is
-the big object we never materialize; the sketch is the small handle we read it through.)
+estimate is the *distinct* `|desc|` we need. Same `O(V·k)` precompute / `O(k)` read **for the
+sketch and `information_content`** as the rung-4 ancestor sketch, just pointed downward (the
+`resnik`/`lin` MICA search on top adds the per-query `O(V+E)` BFS noted above). (And it is the §6
+kernel-trick move once more: the cone is the big object we never materialize; the sketch is the
+small handle we read it through.)
 
 **Where this sits.** This gives a *calibrated relatedness read-out* between two nodes that does not
 inflate on deep graphs — the honest replacement for the rung-4 lift's absolute value. What it does
--- a/docs/design/WAM_RUST_GRAPH_FUNCTIONAL_SEMIRINGS.md
+++ b/docs/design/WAM_RUST_GRAPH_FUNCTIONAL_SEMIRINGS.md
@@ -673,6 +673,66 @@ precompute, `O(k)` reads, no per-query reachability. Unlike the configuration-mo
 companion to the lift's ranking signal. (Hub *selection* from these scores is still the open
 global problem; this only fixes the calibration of the relatedness read-out.)
 
+### 5e. A gentle primer on information-content similarity (for the reader learning this)
+
+*§5d is written for someone who already has the vocabulary. This subsection builds the same idea
+from scratch, with worked numbers — skip it if §5d read easily. Running example: the balanced
+tree the test uses — root `0`; `1,2 → 0`; `3,4 → 1`; `5,6 → 2`, so seven nodes total.*
+
+**One idea: rarity is information.** Imagine someone tells you "this article is filed under
+category `t`." How *informative* is that? If `t` is the root (every article is under it), you
+learned nothing — it was certain. If `t` is a tiny, specific leaf category, you learned a lot —
+that was surprising. So a node's information is its **rarity**: let `p(t) = |desc(t)| / N` be the
+fraction of all nodes that fall under `t` (its descendant cone over the total). The root has
+`p = 1`; a leaf has `p = 1/N`.
+
+**Why `−log₂`.** We want "information" to be `0` for the certain thing (`p=1`) and to *grow* as
+things get rarer (`p → 0`), and we want it to *add up* for independent facts. The function with
+those properties is `IC(t) = −log₂ p(t)` — the number of **bits of surprise**. Worked on the
+example (`N = 7`):
+
+| node | cone `desc(t)` | `\|desc\|` | `p = \|desc\|/7` | `IC = −log₂ p` |
+|------|----------------|-----------|------------------|----------------|
+| `0` (root) | all seven | 7 | 1.00 | **0.00** |
+| `1` (internal) | `{1,3,4}` | 3 | 0.43 | **1.22** |
+| `3` (leaf) | `{3}` | 1 | 0.14 | **2.81** |
+
+So depth/specificity shows up as higher IC, automatically — no hand-tuned "level" number, just
+the cone fraction.
+
+**Resnik similarity: how related are `u` and `v`? Look at the deepest category that holds both.**
+The common ancestors of `u` and `v` are the categories containing *both*. The **most informative**
+one — smallest cone, highest IC — is their *most specific shared category*, the `MICA`. Resnik
+says: `sim(u,v) = IC(MICA)`. Intuition: if the deepest thing that contains both *quantum
+electrodynamics* and *quantum chromodynamics* is the very specific *quantum field theory*, they
+are closely related; if the only thing containing both *QED* and *medieval poetry* is the root
+("everything"), they are unrelated (IC = 0). On the example: `Resnik(3,4)` — their deepest shared
+category is `1`, so `= 1.22`; `Resnik(3,5)` — they share only the root, so `= 0`. (Why the MICA is
+always a *lowest* common ancestor: cones only grow as you go up, so `IC` only *falls* as you go
+up — the maximum IC is at the bottom of the shared region, the merge frontier of §5d.)
+
+**Lin similarity: normalize so "identical" scores 1.** Raw Resnik isn't on a fixed scale — a deep
+tree gives big IC numbers, a shallow one small. Lin divides by how specific the two items
+themselves are: `sim(u,v) = 2·IC(MICA) / (IC(u) + IC(v))`. If `u = v`, the MICA *is* `u`, so it is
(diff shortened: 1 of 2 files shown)


docs/design/WAM_RUST_GRAPH_FUNCTIONAL_SEMIRINGS.md
```diff
@@ -660,6 +660,79 @@ annotated under t)`, with similarity read from the IC of the LCA (Resnik 1995; L
...
+**This IC null is now implemented** (`information_content`, `resnik_similarity`,
+`lin_similarity`). It rests on a **descendant sketch** — `descendant_minhash`, the exact
+downward mirror of the rung-4 `ancestor_minhash`: one reverse-topological pass gives each node a
+fixed-`k` KMV sketch of its descendant cone, which (being a *set*) **dedups by construction**, so
+`sketch_card` estimates the *distinct* cone size `|desc(t)|` that `descendant_weight` over-counts.
+Then `IC(t) = −log₂(|desc(t)|/N)`, `resnik = IC(MICA)` (the most informative common ancestor —
+max `IC` over the common ancestors, which is a lowest one since `IC` is non-increasing upward),
+and `lin = 2·IC(MICA)/(IC(u)+IC(v)) ∈ [0,1]`. The cost mirrors rung 4 exactly: `O(V·k)`
+precompute, `O(k)` reads, no per-query reachability. Unlike the configuration-model lift it uses
+*actual* cone frequencies, so it stays calibrated on deep DAGs — the principled absolute-score
```
s243a: **[required — doc defect: O(k) claim is false for Resnik/Lin]** The statement "O(k) reads, no per-query reachability" is accurate for `information_content` in isolation but **false** for `resnik_similarity` and `lin_similarity` as implemented.

`resnik_similarity` calls `reflexive_ancestors(u)` and `reflexive_ancestors(v)` — each is a full upward BFS allocating a fresh `HashSet` and `VecDeque`, costing O(|Anc(u)| + |Anc(v)|) ≈ O(V+E) worst case. The MICA search IS per-query reachability, directly contradicting the claim. `lin_similarity` inherits this cost.

This is confirmed by deep research: no published DAG semantic similarity library achieves sub-O(V+E) MICA queries without offline all-pairs precomputation (e.g., ancestor bitmask tables enabling O(n/64) bitwise intersection).

Required fix — choose one:
1. **Doc fix (simplest):** replace the `O(k)` sentence with: "`information_content` is an `O(k)` read after the `O(V·k)` precompute. `resnik_similarity` / `lin_similarity` additionally call a per-query O(V+E) BFS to find the MICA; cache ancestor sets or add an all-pairs precompute step for repeated queries."
2. **Code fix (optional future):** precompute ancestor sets once, or use ancestor bitmask tables for O(n/64) intersection.

docs/design/WAM_RUST_GRAPH_FUNCTIONAL_SEMIRINGS.md
```diff
@@ -660,6 +660,79 @@ annotated under t)`, with similarity read from the IC of the LCA (Resnik 1995; L
...
+are closely related; if the only thing containing both *QED* and *medieval poetry* is the root
+("everything"), they are unrelated (IC = 0). On the example: `Resnik(3,4)` — their deepest shared
+category is `1`, so `= 1.22`; `Resnik(3,5)` — they share only the root, so `= 0`. (Why the MICA is
+always a *lowest* common ancestor: cones only grow as you go up, so `IC` only *falls* as you go
+up — the maximum IC is at the bottom of the shared region, the merge frontier of §5d.)
+
+**Lin similarity: normalize so "identical" scores 1.** Raw Resnik isn't on a fixed scale — a deep
+tree gives big IC numbers, a shallow one small. Lin divides by how specific the two items
+themselves are: `sim(u,v) = 2·IC(MICA) / (IC(u) + IC(v))`. If `u = v`, the MICA *is* `u`, so it is
+`2·IC(u)/2·IC(u) = 1`; if they share only the root, `IC(MICA)=0` so it is `0`. On the example,
```
s243a: **[required — doc: Lin identity-at-root edge case]** The primer says "if `u = v`, the MICA is `u`, so it is `2·IC(u)/2·IC(u) = 1`" — but this is only true when `IC(u) > 0`. When `u = v = root`, `IC(root) = 0`, the denominator is zero, and `lin_similarity` returns `None` per the `denom <= 0.0` guard.

This creates an asymmetry: `resnik_similarity(root, root)` returns `Some(0.0)` but `lin_similarity(root, root)` returns `None`. The primer's identity claim silently fails for the root.

The standard resolution differs by library — WordNet::Similarity returns 0; mathematically, returning 1 for the identical-node case is defensible. Please:
1. Add to the primer: "Exception: when both nodes are the root (`IC = 0`), Lin is undefined (0/0); the function returns `None`."
2. Add a test: `assert_eq!(lin_similarity(0, 0, &t, &d, n, K), None);`
3. Optionally, special-case `if u == v && icu <= 0.0 { return Some(1.0); }` if identity-always-1 is the desired contract — document whichever choice is made.

templates/targets/rust_wam/boundary_cache.rs.mustache
```diff
@@ -1157,6 +1157,143 @@ pub fn sketch_overlap_lift(a: &[u64], b: &[u64], n_total: usize, k: usize) -> f6
...
+    parents: &std::collections::HashMap<u32, Vec<u32>>,
+    dsigs: &std::collections::HashMap<u32, Vec<u64>>,
+    n_total: usize,
+    k: usize,
+) -> Option<f64> {
+    let au = reflexive_ancestors(u, parents);
+    let av = reflexive_ancestors(v, parents);
+    au.intersection(&av)
+        .filter_map(|b| dsigs.get(b).map(|s| information_content(s, n_total, k)))
+        .fold(None, |acc, ic| Some(acc.map_or(ic, |a: f64| a.max(ic))))
```
s243a, line 1296: **[required — silent incorrect result: missing dsig entry skips MICA]** The `filter_map` on line 119 silently drops common ancestors whose sketch is absent from `dsigs`. If `dsigs` was computed on a different graph, or if the graph was modified after the sketch was built, a high-IC common ancestor's sketch can be missing. The function then returns the IC of a lower-IC common ancestor — a silently wrong answer — rather than `None` or a panic.

Worst case: the true MICA's sketch is missing; the root's sketch is present → `resnik_similarity` returns `Some(0.0)`, indistinguishable from "only shared ancestor is root."

Required fix: add a precondition check at the top of `resnik_similarity`:
```rust
// Precondition: dsigs must be the descendant_minhash output for the same `parents` graph.
// Missing entries for common ancestors cause silent IC underestimation; verify dsigs completeness
// or use debug_assert!s for correctness in debug builds.
debug_assert!(dsigs.contains_key(&u), "u not in dsigs — dsigs/parents graph mismatch");
debug_assert!(dsigs.contains_key(&v), "v not in dsigs — dsigs/parents graph mismatch");
```
Also document in the doc-comment: "`dsigs` must be the `descendant_minhash` output for the **same** `parents` graph with the same node set; missing entries cause silent IC underestimation, not a `None`." This requirement should mirror how `lin_similarity` already early-returns `None` via `dsigs.get(&u)?` — but `resnik_similarity` does not have the same protection for intermediate common ancestors.

templates/targets/rust_wam/boundary_cache.rs.mustache
```diff
@@ -1157,6 +1157,143 @@ pub fn sketch_overlap_lift(a: &[u64], b: &[u64], n_total: usize, k: usize) -> f6
...
+    dsigs: &std::collections::HashMap<u32, Vec<u64>>,
+    n_total: usize,
+    k: usize,
+) -> Option<f64> {
+    let mica_ic = resnik_similarity(u, v, parents, dsigs, n_total, k)?;
+    let icu = information_content(dsigs.get(&u)?, n_total, k);
+    let icv = information_content(dsigs.get(&v)?, n_total, k);
+    let denom = icu + icv;
+    if denom <= 0.0 { None } else { Some((2.0 * mica_ic / denom).min(1.0)) }
+}
```
s243a, line 1321: **[non-blocking — doc note: .min(1.0) rationale]** The `.min(1.0)` is theoretically unnecessary: since MICA is a common ancestor of both `u` and `v`, its descendant cone contains both nodes' cones, so `IC(MICA) ≤ min(IC(u), IC(v)) ≤ (IC(u)+IC(v))/2`, giving `2·IC(MICA)/(IC(u)+IC(v)) ≤ 1` exactly.

However, with KMV sketch estimation, sketch variance can make `|desc(MICA)|` appear *smaller* than `|desc(u)|` (even though MICA's exact cone ⊇ u's cone), inflating estimated `IC(MICA)` above `min(IC(u), IC(v))` and pushing the ratio above 1. The `.min(1.0)` is a correct KMV-error guard.

Suggest adding a comment:
```rust
// .min(1.0): exact Lin ≤ 1 by construction; guards KMV estimation variance that can
// make IC(MICA) appear > min(IC(u), IC(v)) when sketches are saturated.
```

templates/targets/rust_wam/boundary_cache.rs.mustache
```diff
@@ -3221,6 +3358,54 @@ mod tests {
         assert!((0.0..=1.0).contains(&j));
     }
 
+    // IC-based semantic similarity (Resnik/Lin) via the descendant sketch -- the calibrated,
+    // depth-aware alternative to the rung-4 lift. Balanced binary tree: 0 root; 1,2->0;
+    // 3,4->1; 5,6->2. N=7. Cones: |desc(0)|=7, |desc(1)|=|desc(2)|=3, leaves=1.
```
s243a, line 3389: **[required — missing tests: ancestor-of-other and diamond dedup]** The test covers sibling leaves, cross-subtree leaves, self-similarity (non-root), disconnected, and cycle. Two critical cases are missing:

**1. One node is an ancestor of the other.** Because `reflexive_ancestors` includes the node itself, `resnik_similarity(1, 3)` should return `IC(1)` (node 1 is the most specific common ancestor of itself and its descendant 3). Please add:
```rust
// ancestor-of-descendant: MICA is the ancestor node itself
assert!(close(resnik_similarity(1, 3, &t, &d, n, K).unwrap(),
    information_content(&d[&1], n, K)));
let lin13 = lin_similarity(1, 3, &t, &d, n, K).unwrap();
assert!(close(lin13, 2.0 * information_content(&d[&1], n, K)
    / (information_content(&d[&1], n, K) + information_content(&d[&3], n, K))));
```

**2. Diamond DAG — the dedup motivation.** The entire point of using a sketch over `descendant_weight` is deduplication. A diamond graph `3→[1,2], 1→0, 2→0` should show `sketch_card(&d[&0]) == 4` (distinct: {0,1,2,3}) while `descendant_weight` would return 5 (3 counted twice). Add:
```rust
let mut dia: HashMap<u32, Vec<u32>> = HashMap::new();
dia.insert(1, vec![0]); dia.insert(2, vec![0]); dia.insert(3, vec![1, 2]);
let dd = descendant_minhash(&dia, K).unwrap();
assert_eq!(sketch_card(&dd[&0], K), 4.0); // distinct: {0,1,2,3}, not 5
```

**3. root-root Lin is None:**
```rust
assert_eq!(lin_similarity(0, 0, &t, &d, n, K), None); // IC(root)=0 → denom=0
```

docs/design/WAM_RUST_GRAPH_FUNCTIONAL_SEMIRINGS.md
```diff
@@ -660,6 +660,79 @@ annotated under t)`, with similarity read from the IC of the LCA (Resnik 1995; L
...
+`Lin(3,4) = 2(1.22)/(2.81+2.81) = 0.43`. Now every pair sits in `[0,1]`, comparable across graphs.
+
+**Why we needed the descendant *sketch* (and not the additive weight).** Every formula above needs
+`|desc(t)|`, the **distinct** count of nodes under `t`. Computing that exactly for all `t` is
+reachability — the global blow-up we keep refusing. The cheap one-pass additive `descendant_weight`
+(rung 2) is no good *here*: it counts a node reachable by two paths **twice**, so it inflates
+`|desc|`, distorts `p`, and would corrupt the IC. The fix is a **set**: `descendant_minhash` keeps
+a fixed-`k` sample of the cone, and a set automatically counts each member once — so its size
+estimate is the *distinct* `|desc|` we need. Same `O(V·k)` precompute / `O(k)` read as the rung-4
+ancestor sketch, just pointed downward. (And it is the §6 kernel-trick move once more: the cone is
```
s243a: **[non-blocking — doc precision: O(k) read for IC only]** "Same `O(V·k)` precompute / `O(k)` read" is correct for `information_content` alone. But the similarity *functions* built on top of it (`resnik_similarity`, `lin_similarity`) add an O(V+E) per-query BFS. The reader of §5e may reasonably conclude that calling Resnik/Lin is O(k) — that conclusion would be wrong.

Suggest adding a parenthetical: "(Note: `information_content` itself is an `O(k)` read; `resnik_similarity`/`lin_similarity` additionally run a per-query O(V+E) ancestor BFS to locate the MICA — cache ancestor sets for repeated queries on the same graph.)"

templates/targets/rust_wam/boundary_cache.rs.mustache
```diff
@@ -1157,6 +1157,143 @@ pub fn sketch_overlap_lift(a: &[u64], b: &[u64], n_total: usize, k: usize) -> f6
...
+        return 0.0;
+    }
+    let card = sketch_card(dsig, k).max(1.0);
+    let p = (card / n_total as f64).min(1.0); // clamp sketch overestimate so IC ≥ 0
+    -p.log2()
+}
+
+/// Reflexive ancestor set of `node` (`node` itself plus every strict ancestor), by up-BFS.
+fn reflexive_ancestors(
+    node: u32,
```
s243a, line 1238: **[non-blocking — note: reflexive_ancestors is O(V+E) per call]** This helper is private but called twice per `resnik_similarity` / `lin_similarity` query. For sparse type hierarchies this is fine; for hot loops over many pairs, caller should cache ancestor sets. Suggest adding to the doc-comment:
```rust
/// O(|ancestors(node)| + edges above node) per call — allocates a fresh HashSet/VecDeque.
/// For repeated queries on the same node, cache the returned set rather than recomputing.
```
Also note: deep research confirms this is standard for per-query MICA — the alternative (all-pairs precompute via ancestor bitmask tables) enables O(n/64) intersection but requires O(n²/64) space, which is only worthwhile for repeated batch queries.

templates/targets/rust_wam/boundary_cache.rs.mustache
```diff
@@ -1157,6 +1157,143 @@ pub fn sketch_overlap_lift(a: &[u64], b: &[u64], n_total: usize, k: usize) -> f6
...
+}
+
+/// **Resnik similarity** — the calibrated, depth-aware alternative to the rung-4 lift: the
+/// information content of the **most informative common ancestor** (MICA), the common ancestor
+/// with the highest `IC` (= the smallest descendant cone). Since `IC` is non-increasing upward,
+/// the MICA is a *lowest* common ancestor; here we enumerate the common ancestors (reflexive,
+/// so one term may be the other's ancestor) and take the max `IC`. Uses *actual* cone
+/// frequencies, so — unlike the configuration-model lift — it does not inflate on deep DAGs.
+/// `None` if `u` and `v` share no ancestor. `dsigs` = `descendant_minhash`, `n_total` = nodes.
+pub fn resnik_similarity(
```
s243a, line 1282: **[non-blocking — doc note: dsigs/parents same-graph contract]** The doc-comment should state the contract explicitly, since violating it produces silent wrong answers (see the `filter_map` comment above):

Suggest adding to the `resnik_similarity` doc-comment:
> **Precondition:** `dsigs` must be the `descendant_minhash` output computed on the **same** `parents` graph — every node reachable from `u` or `v` upward must have a sketch entry. Missing entries cause silent IC underestimation (the MICA appears to have lower IC than it actually has). `u` and `v` must be nodes in the graph. If `dsigs` may be incomplete, call `descendant_minhash` freshly on the current `parents` map.

Same note applies to `lin_similarity` (though `lin_similarity` already short-circuits to `None` if `u` or `v` is missing — the issue is common ancestors of `u`/`v`, not `u`/`v` themselves).

docs/design/WAM_RUST_GRAPH_FUNCTIONAL_SEMIRINGS.md
```diff
@@ -660,6 +660,79 @@ annotated under t)`, with similarity read from the IC of the LCA (Resnik 1995; L
...
 the principled replacement; for bridge *selection* (a ranking), the current lift suffices.
 
+**This IC null is now implemented** (`information_content`, `resnik_similarity`,
+`lin_similarity`). It rests on a **descendant sketch** — `descendant_minhash`, the exact
+downward mirror of the rung-4 `ancestor_minhash`: one reverse-topological pass gives each node a
+fixed-`k` KMV sketch of its descendant cone, which (being a *set*) **dedups by construction**, so
+`sketch_card` estimates the *distinct* cone size `|desc(t)|` that `descendant_weight` over-counts.
+Then `IC(t) = −log₂(|desc(t)|/N)`, `resnik = IC(MICA)` (the most informative common ancestor —
+max `IC` over the common ancestors, which is a lowest one since `IC` is non-increasing upward),
+and `lin = 2·IC(MICA)/(IC(u)+IC(v)) ∈ [0,1]`. The cost mirrors rung 4 exactly: `O(V·k)`
```
s243a: **[non-blocking — MICA non-uniqueness and tie-breaking]** The claim "which is a lowest one since IC is non-increasing upward" is correct for *exact* IC: if `a` is a strict ancestor of `b`, `desc(a) ⊋ desc(b)`, so `|desc(a)| > |desc(b)|`, `IC(a) < IC(b)` — the max-IC common ancestor is indeed at the bottom of the common-ancestor region.

However, deep research flags two caveats to document:
1. **Tie-breaking:** In a symmetric DAG, two nodes can have identical descendant cone sizes and thus equal IC. The current `fold` picks whichever comes last in `au.intersection(&av)` iteration — non-deterministic across Rust HashMap runs. This may matter for reproducibility. Consider `a.max(ic)` in the fold being augmented with a tie-break rule (e.g., prefer larger node ID for determinism), or document that ties are resolved arbitrarily.
2. **Sketch monotonicity:** With saturated KMV sketches, estimated IC can be non-monotone (a parent's estimated cardinality can appear smaller than a child's due to hash variance). The implementation correctly takes max over ALL common ancestors, so no algorithmic error results — but the MICA returned may not be the true MICA when sketches are saturated. Document this in the doc-comment.

s243a (commented):
## Multi-model code review — PR #3259: IC-based semantic similarity (Resnik/Lin)

**Reviewers:** Claude Opus 4.8 · GPT-5.5 · Gemini deep research (Resnik 1995, Lin 1998, Seco et al. 2004, MICA literature, KMV error propagation)

**Verdict: approve after required fixes. Code is mathematically correct; two documentation defects and two test gaps need addressing before merge.**

The formulas are standard and well-validated: intrinsic IC via descendant-count (Seco et al. 2004 lineage) achieves 0.84 correlation with human benchmarks vs. Resnik's corpus-based 0.79. The Lin normalization is the canonical formula. The `descendant_minhash` dedup-by-construction correctly gives distinct cone sizes. The boundary guards (`.max(1.0)`, `.min(1.0)`) are sound.

The headline issue is a **documentation defect** that all three reviewers flagged independently: `resnik_similarity` and `lin_similarity` perform two O(V+E) BFS traversals per query — they are NOT O(k) reads. Four required items total.

s243a:
## Review fixes — all 4 required items + the non-blocking notes addressed (commit `eb14a4d6`)

Thanks for the multi-model pass — R1 was exactly the cost-claim issue I'd flagged as the likely gap. Disposition:

**Required**
- **R1** *(O(k) claim false for Resnik/Lin)* — documented that `information_content` is the `O(k)` read but `resnik_similarity`/`lin_similarity` add a **per-query `O(V+E)` ancestor BFS** to find the MICA, in the function doc-comments **and** §5d/§5e. Took the doc-fix route (option 1); the all-pairs/bitmask precompute is noted as the future code path.
- **R2** *(Lin root-root)* — fixed the §5e primer to state the `identical → 1` identity holds only for `IC > 0`; `Lin(root,root)` is `None` (`0/0`) while `Resnik(root,root)` is `Some(0)`. Tests added for both.
- **R3** *(silent wrong answer on incomplete `dsigs`)* — added `debug_assert!`s that `u,v ∈ dsigs` and a same-graph precondition to the doc-comment (a missing common-ancestor sketch is silently skipped → IC underestimate).
- **R4** *(missing tests)* — added the **diamond-DAG dedup** test (`sketch_card(dsig(root)) == 4` vs `descendant_weight == 5` — the whole motivation for a set sketch) and the **ancestor-of-other** test (`Resnik(1,3) == IC(1)`, explicit `Lin(1,3)`), plus `Lin(0,0) == None`.

**Non-blocking, folded in**
- `.min(1.0)` KMV-variance rationale comment on `lin_similarity`.
- `reflexive_ancestors` `O(V+E)`-per-call doc note.
- `dsigs`/`parents` same-graph contract on the doc-comment.
- MICA non-uniqueness (we return the IC **value**, so ties are immaterial) and the saturated-sketch non-monotonicity caveat (estimated-argmax may not be the exact MICA) documented.

On the deep-research pointer: **FaITH** (Pirró & Euzenat 2010), `IC(MICA)/(IC(u)+IC(v)−IC(MICA))`, which avoids the undefined-at-root pathology — noted as a clean future alternative; out of scope here.

Full suite: **75 passed, 0 failed.** Resolving the threads accordingly.

---

PR #3262, opened 2026-06-18
WAM-Rust: 3f — landmark-cached designated-bridge caret (O(1) per pair)

s243a:
## What & why

Closes roadmap **3f** and supplies the amortization that makes all the hub-selection work
actually pay off. Until now, `caret_min_over_hubs` recomputed `up_distance_to` (a BFS) per pair
per hub — so the "designated bridges are landmarks" reuse §5c describes wasn't realized. 3f
precomputes each bridge's distance field once, turning the designated-bridge caret into an O(1)
table lookup.

## Library (`boundary_cache`)

- **`bridge_distance_fields(bridges, parents)`** — one downward BFS per bridge over a **shared**
  children graph (built once), giving `d(·→B)` for every descendant of `B`. `O(E + Σ_B|desc(B)|)
  ≤ O(K·V)`. (`d(src→B)` over parent edges equals the reverse BFS depth `B→src` over child edges —
  reversing a shortest up-path gives an equal-length down-path.)
- **`caret_through_bridge_cached(u, v, field)`** — O(1) caret `field[u]+field[v]`; equals
  `caret_through_bridge` exactly, `None` on the same condition.
- **`caret_min_over_cached_bridges(u, v, fields)`** — the cached counterpart of
  `caret_min_over_hubs`: O(#bridges) lookups per pair instead of a BFS per bridge.

## Live (`WamState`)

`caret_landmarks` field + `build_caret_landmarks(bridges, edge_pred)` /
`category_caret_through_bridge` / `category_caret_min_over_landmarks`. Eager-edge only, `None`
until built.

## The pipeline this completes

**Pick** convergence hubs (the fan-in / jump / lift / reconvergence rungs) → **cache** their
distance fields once (this PR) → **answer** `caret_min_over_hubs` as O(#hubs) table lookups. The
cached field is the **min-plus distance functional** read off a compact precompute — the caret
answer never forms the pair's path-length distribution (the §8 "carry the functional, not the
distribution" theme).

## Tests

- `bridge_landmarks_cached_caret_matches_per_query` — field `== up_distance_to`; cached caret `==
  caret_through_bridge`; min `== caret_min_over_hubs`, across **all** pairs and bridges on a
  cross-listed tree.
- `live_bridge_landmarks_match_library` — parity through `WamState`; unbuilt `→ None`.

Full suite: **78 passed, 0 failed.** Roadmap 3f marked DONE in
`docs/design/WAM_RUST_GRAPH_FUNCTIONAL_SEMIRINGS.md`.

---

### DIFF
--- a/docs/design/WAM_RUST_GRAPH_FUNCTIONAL_SEMIRINGS.md
+++ b/docs/design/WAM_RUST_GRAPH_FUNCTIONAL_SEMIRINGS.md
@@ -596,7 +596,10 @@ caching, the **designated-bridge** measure is cacheable.
 >   `B` a funnel." Two fixes, both O(k): (a) **height-agnosticism** — summarize each node's
 >   *whole* lineage to root once, as a fixed-size **KMV/MinHash ancestor sketch**
 >   `sig(B) = bottom-k( {B} ∪ ⋃_p sig(p) )`, one root→leaf pass; overlap (`sketch_jaccard`) is
->   then read at *any* depth with no knob (error ∝ 1/√k, not a depth cutoff). (b) **baseline
+>   then read at *any* depth with no knob (error ∝ 1/√k, not a depth cutoff). (The sketch is the
+>   bottom-`k`/KMV lineage: MinHash for Jaccard, Broder 1997; the one-pass bottom-`k` Jaccard
+>   estimator, Cohen & Kaplan 2007; the `(k−1)/ĥ_k` distinct-count read, Bar-Yossef et al. 2002 /
+>   Beyer et al. 2007.) (b) **baseline
 >   correction** — a real hub reconverges *more than chance*: against the configuration-model
 >   null `E|A∩B| ≈ |A|·|B|/N`, the signal is `lift = observed |A∩B| / E|A∩B|` (`sketch_overlap_
 >   lift`), `>1` an excess funnel, `≈1` just small-world background. The sketch yields `|A|`,
@@ -709,7 +712,10 @@ category `t`." How *informative* is that? If `t` is the root (every article is u
 learned nothing — it was certain. If `t` is a tiny, specific leaf category, you learned a lot —
 that was surprising. So a node's information is its **rarity**: let `p(t) = |desc(t)| / N` be the
 fraction of all nodes that fall under `t` (its descendant cone over the total). The root has
-`p = 1`; a leaf has `p = 1/N`.
+`p = 1`; a leaf has `p = 1/N`. This descendant-fraction definition is **intrinsic** IC — it reads
+the rarity off the graph structure alone, needing no external corpus of annotation frequencies;
+Seco, Veale & Hayes (2004) introduced it and showed it matches corpus-based IC closely (≈ 0.84 vs
+0.79 correlation with human similarity benchmarks), which is why we use it here.
 
 **Why `−log₂`.** We want "information" to be `0` for the certain thing (`p=1`) and to *grow* as
 things get rarer (`p → 0`), and we want it to *add up* for independent facts. The function with
@@ -748,7 +754,7 @@ is an `O(k)` read, but calling `resnik`/`lin` is *not* `O(k)` — each runs a pe
 upward BFS to find the MICA; cache the ancestor sets for repeated queries on the same graph.)
 
 **FaITH similarity: the Jiang–Conrath-faithful sibling.** Lin isn't the only way to normalize
-Resnik. The *Jiang–Conrath distance* `JC(u,v) = IC(u) + IC(v) − 2·IC(MICA)` measures *how far
+Resnik. The *Jiang–Conrath distance* (Jiang & Conrath 1997) `JC(u,v) = IC(u) + IC(v) − 2·IC(MICA)` measures *how far
 apart* `u` and `v` are: it is the IC you'd have to "spend" climbing from each down to the MICA —
 `0` for identical nodes, large when they meet only high up. **FaITH** (Pirró & Euzenat 2010) turns
 that distance into a bounded similarity, `sim(u,v) = IC(MICA) / (IC(u) + IC(v) − IC(MICA))`, which
@@ -1022,3 +1028,53 @@ buy diminishing returns and is not carried).
 - `WAM_RUST_BOUNDARY_DISTRIBUTION_CACHE_PLAN.md` — phase status of shipped work.
 - This note — the algebraic generalization (product semirings, ancestor-space domain,
   the implicit-functional / kernel-trick framing) that the next increments build on.
+
+## 10. References
+
+Collected for understanding the theory and as a citation base for possible future write-up.
+Each is referenced inline at the section that uses it.
+
+**Information-content semantic similarity (§5d, §5e).**
+- Resnik, P. (1995). *Using Information Content to Evaluate Semantic Similarity in a Taxonomy.*
+  IJCAI-95. — `resnik_similarity = IC(MICA)`.
+- Lin, D. (1998). *An Information-Theoretic Definition of Similarity.* ICML 1998. —
+  `lin_similarity = 2·IC(MICA)/(IC(u)+IC(v))`.
+- Jiang, J. J., & Conrath, D. W. (1997). *Semantic Similarity Based on Corpus Statistics and
+  Lexical Taxonomy.* ROCLING X (also arXiv cmp-lg/9709008). — the JC distance `IC(u)+IC(v)−2·IC(MICA)`.
+- Seco, N., Veale, T., & Hayes, J. (2004). *An Intrinsic Information Content Metric for Semantic
+  Similarity in WordNet.* ECAI 2004. — the **intrinsic** (descendant-count) IC we use, `IC(t) =
+  −log₂(|desc(t)|/N)`, ≈0.84 vs 0.79 human-benchmark correlation against corpus IC.
+- Pirró, G., & Euzenat, J. (2010). *A Feature and Information Theoretic Framework for Semantic
+  Similarity and Relatedness.* ISWC 2010. — the **FaITH** measure `IC(MICA)/(IC(u)+IC(v)−IC(MICA))`.
+
+**MinHash / KMV sketches and distinct-count estimation (rung 4, §5d; descendant sketch, §5e).**
+- Broder, A. Z. (1997). *On the Resemblance and Containment of Documents.* SEQUENCES 1997. —
+  MinHash for Jaccard.
+- Bar-Yossef, Z., Jayram, T. S., Kumar, R., Sivakumar, D., & Trevisan, L. (2002). *Counting
+  Distinct Elements in a Data Stream.* RANDOM 2002. — k-minimum-values (KMV) distinct-count.
+- Beyer, K., Haas, P. J., Reinwald, B., Sismanis, Y., & Gemulla, R. (2007). *On Synopses for
+  Distinct-Value Estimation Under Multiset Operations.* SIGMOD 2007. — the bottom-`k` / KMV
+  `(k−1)/ĥ_k` cardinality estimator (`sketch_card`).
+- Cohen, E., & Kaplan, H. (2007). *Summarizing Data Using Bottom-k Sketches.* PODC 2007. — the
+  one-pass bottom-`k` Jaccard estimator (`sketch_jaccard`).
+
+**2-hop cover / hub labeling and landmark distance (§5c, roadmap 3f).**
+- Cohen, E., Halperin, E., Kaplan, H., & Zwick, U. (2002/2003). *Reachability and Distance
+  Queries via 2-Hop Labels.* SODA 2002 / SIAM J. Comput. 2003. — the 2-hop label framework that
+  `bridge_distance_fields` instantiates.
+- Abraham, I., Delling, D., Goldberg, A. V., & Werneck, R. F. (2012). *Hierarchical Hub Labelings
+  for Shortest Paths.* ESA 2012.
+- Akiba, T., Iwata, Y., & Yoshida, Y. (2013). *Fast Exact Shortest-Path Distance Queries on Large
+  Networks by Pruned Landmark Labeling.* SIGMOD 2013.
+- Tretyakov, K., Armas-Cervantes, A., García-Bañuelos, L., Vilo, J., & Dumas, M. (2011). *Fast
+  Fully Dynamic Landmark-based Estimation of Shortest Path Distances in Very Large Graphs.* CIKM
+  2011. — the `O(K·V)` space-tightness for O(1)-lookup landmarks.
+- Storandt, S. (2022). *Algorithms for Landmark Hub Labeling.* ISAAC 2022. — `min_B(d(u→B)+d(v→B))`
+  as a valid upper bound, exact iff an optimal meeting node is a landmark.
+
+**Distribution reconstruction (§7).**
+- Lindeberg, J. W. (1922). *Eine neue Herleitung des Exponentialgesetzes in der
+  Wahrscheinlichkeitsrechnung.* Math. Z. 15. — the CLT condition for the moment-jet → Gaussian rung.
+- Blinnikov, S., & Moessner, R. (1998). *Expansions for Nearly Gaussian Distributions.* A&A
+  Suppl. Ser. 130. — practical Gram–Charlier / Edgeworth series (the `MomentNormal` → `GramCharlier`
+  reconstruction rungs).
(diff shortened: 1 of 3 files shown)


templates/targets/rust_wam/boundary_cache.rs.mustache
```diff
@@ -880,6 +880,72 @@ pub fn caret_min_over_hubs(
...
+/// **Landmark distance fields for designated bridges** (roadmap 3f): for each bridge `B`,
+/// precompute `d(·→B)` — the up-distance from every descendant of `B` to `B` — as one downward
+/// BFS from `B` over the *children* graph (the reverse of `parents`). The children map is built
+/// **once** and shared across all bridges, so the whole precompute is `O(E + Σ_B |desc(B)|)`,
+/// at most `O(K·V)` for `K` bridges. With these fields cached, `caret_through_bridge` collapses
+/// from a per-query BFS to an **O(1)** table lookup (`caret_through_bridge_cached`) — the
+/// "designated bridges are landmarks" amortization (§5c): pay the field once, reuse it across
+/// every pair. (`d(src→B)` over parent edges equals the reverse BFS depth `B→src` over child
+/// edges, since reversing a shortest up-path gives an equal-length down-path.)
+pub fn bridge_distance_fields(
```
s243a, line 899: **[non-blocking — doc precision: BFS reversal wording]** "Reversing a shortest up-path gives an equal-length down-path" is correct but informal. The precise statement is: *the set of upward paths from `src` to `B` in the parent-edge graph bijects with the set of downward paths from `B` to `src` in the child-edge graph via element-wise edge reversal, preserving length.* Therefore their minima coincide — `d(src→B)` in the parent graph = `d(B→src)` in the child graph. This is the standard reversed-BFS correctness argument used in Contraction Hierarchies and bidirectional search; it holds for any directed graph (not just DAGs), and diamond structures introduce no subtlety since BFS correctly finds shortest paths regardless of how many routes exist.

Suggest tightening to: "(`d(src→B)` over parent edges equals the reverse BFS depth `B→src` over child edges: child edges are the exact reversal of parent edges, so upward and downward paths are in length-preserving bijection and their minima coincide.)"

templates/targets/rust_wam/boundary_cache.rs.mustache
```diff
@@ -880,6 +880,72 @@ pub fn caret_min_over_hubs(
         .min()
 }
 
+/// **Landmark distance fields for designated bridges** (roadmap 3f): for each bridge `B`,
+/// precompute `d(·→B)` — the up-distance from every descendant of `B` to `B` — as one downward
+/// BFS from `B` over the *children* graph (the reverse of `parents`). The children map is built
+/// **once** and shared across all bridges, so the whole precompute is `O(E + Σ_B |desc(B)|)`,
+/// at most `O(K·V)` for `K` bridges. With these fields cached, `caret_through_bridge` collapses
```
s243a: **[non-blocking — doc: complexity bound is approximate]** The stated `O(E + Σ_B|desc(B)|)` omits the intra-cone edge cost within each per-bridge BFS. The exact bound is `O(E)` (one-time children-map build) `+ Σ_B(|desc(B)| + edges_within_desc(B))`. For sparse DAGs this is essentially `O(E + Σ|desc(B)|)` as written; for dense DAGs with high fan-out, the edge term can dominate. The tighter — and still concise — bound is `O(E + K·(V+E)) = O(K·(V+E))` worst case.

Also note that memory is `Σ_B(|desc(B)|+1)` entries — up to `O(K·V)` — which is not currently documented anywhere. Deep research confirms O(K·V) space is tight for this approach (Akiba et al. 2013; Tretyakov et al. CIKM 2011).

Suggest updating the first doc-comment paragraph to: "`O(E + Σ_B(|desc(B)| + edges_within_desc(B))) ≤ O(K·(V+E))` time; `O(Σ_B|desc(B)|) ≤ O(K·V)` memory."

templates/targets/rust_wam/boundary_cache.rs.mustache
```diff
@@ -880,6 +880,72 @@ pub fn caret_min_over_hubs(
...
+/// `d(·→B)`: `field[u] + field[v]`. Equals `caret_through_bridge(u, v, B, parents)` exactly,
+/// and is `None` on the same condition (either node is not a descendant of `B`, so absent from
+/// the field).
+pub fn caret_through_bridge_cached(
+    u: u32,
+    v: u32,
+    field: &std::collections::HashMap<u32, u32>,
+) -> Option<u32> {
+    Some(field.get(&u)? + field.get(&v)?)
+}
```
s243a, line 946: **[required — minor: u32 overflow note on new public API]** `field.get(&u)? + field.get(&v)?` performs unchecked `u32` addition. Each addend is a BFS depth ≤ `V−1`, so the sum ≤ `2(V−1)`, which overflows `u32` only if `V > 2^31 ≈ 2.1B` — impossible for an in-memory `HashMap<u32,…>` type hierarchy in practice. This is also consistent with the existing `up_distance_to` arithmetic which uses the same unchecked `dx + 1`.

However, this is a **new public API** and the safety bound is non-local (a reader must trace through to `up_distance_to` to understand it). Add a one-line comment:
```rust
// Depths bounded by V−1 each; sum ≤ 2(V−1) < u32::MAX for any in-memory graph.
Some(field.get(&u)? + field.get(&v)?)
```

templates/targets/rust_wam/boundary_cache.rs.mustache
```diff
@@ -880,6 +880,72 @@ pub fn caret_min_over_hubs(
...
+/// designated bridges whose fields are precomputed (`bridge_distance_fields`). With the fields
+/// built for exactly the same bridge set, this equals `caret_min_over_hubs` — but at O(#bridges)
+/// table lookups per pair instead of a BFS per bridge.
+pub fn caret_min_over_cached_bridges(
+    u: u32,
+    v: u32,
+    fields: &std::collections::HashMap<u32, std::collections::HashMap<u32, u32>>,
+) -> Option<u32> {
+    fields.values().filter_map(|f| caret_through_bridge_cached(u, v, f)).min()
+}
```
s243a, line 965: **[non-blocking — doc: caller sync responsibility]** `caret_min_over_cached_bridges` minimizes over **all** fields present in `fields`, not over a passed-in bridge list. If the caller builds fields for bridges `{A, B, C}` and then later calls this with an extended `fields` that also includes `D`, the result is the minimum over all four — which may be correct or may not, depending on intent. Similarly, if bridges are added or removed from the hub set, `bridge_distance_fields` must be called again to regenerate.

Suggest adding to the doc-comment: "The caller is responsible for keeping `fields` in sync with the intended bridge set — this function minimizes over all entries in `fields`, not over an explicit bridge list. Rebuild via `bridge_distance_fields` if the graph or hub set changes."

templates/targets/rust_wam/boundary_cache.rs.mustache
```diff
@@ -880,6 +880,72 @@ pub fn caret_min_over_hubs(
...
+/// precompute `d(·→B)` — the up-distance from every descendant of `B` to `B` — as one downward
+/// BFS from `B` over the *children* graph (the reverse of `parents`). The children map is built
+/// **once** and shared across all bridges, so the whole precompute is `O(E + Σ_B |desc(B)|)`,
+/// at most `O(K·V)` for `K` bridges. With these fields cached, `caret_through_bridge` collapses
+/// from a per-query BFS to an **O(1)** table lookup (`caret_through_bridge_cached`) — the
+/// "designated bridges are landmarks" amortization (§5c): pay the field once, reuse it across
+/// every pair. (`d(src→B)` over parent edges equals the reverse BFS depth `B→src` over child
+/// edges, since reversing a shortest up-path gives an equal-length down-path.)
+pub fn bridge_distance_fields(
+    bridges: &[u32],
```
s243a, line 900: **[required — doc: duplicate bridges and pure-leaf bridge behaviour]** Two edge cases are undocumented:

1. **Duplicate bridges in the input slice:** `bridges: &[u32]` accepts duplicates. If `bridges = [1, 1, 2]`, the BFS for bridge `1` runs twice (each insert overwrites the previous — idempotent, but wasteful). Add either a `dedup` at the top of `bridge_distance_fields`, or a doc note: "deduplicate `bridges` before calling to avoid redundant BFS runs."

2. **Pure-leaf bridge (no children):** A bridge `B` with no children in the parent graph produces `d = {B: 0}`. `caret_through_bridge_cached(u, v, field)` returns `None` for any `u ≠ B` or `v ≠ B`. This is the correct equivalent of the per-query behaviour (a leaf bridge has no descendants to look up), but is not documented. Add: "A bridge with no descendants produces a field containing only `{B: 0}`; queries for pairs not equal to `B` correctly return `None`."

Neither is a logic bug — both are worth a test and/or doc note.

templates/targets/rust_wam/boundary_cache.rs.mustache
```diff
@@ -3137,6 +3203,42 @@ mod tests {
         assert_eq!(caret_through_bridge(4, 6, 1, &t), caret_distance_lca(4, 6, &t));
     }
 
+    // Roadmap 3f: the landmark-cached designated-bridge caret equals the per-query
+    // caret_through_bridge everywhere, at O(1) per pair after one precompute per bridge.
+    #[test]
+    fn bridge_landmarks_cached_caret_matches_per_query() {
+        // Same tree as above plus a cross-listing so some pairs route through several bridges.
```
s243a, line 3228: **[non-blocking — missing edge-case tests]** The existing tests give excellent all-pairs × all-bridges coverage of the cross-listed tree. Three small cases remain untested:

1. **No-common-bridge:** a pair (u, v) where no bridge is a common ancestor of both → `caret_min_over_cached_bridges` returns `None`. Currently this is only implicitly covered if no pair in the test has this property; add an explicit assertion.

2. **`build_caret_landmarks` returns false** for a non-existent `edge_pred` → add `assert!(!vm.build_caret_landmarks(&bridges, "no_such_pred"));`

3. **Pure-leaf bridge:** `bridge_distance_fields(&[leaf], &parents)` → field = `{leaf: 0}` → `caret_through_bridge_cached(leaf, leaf, &field) = Some(0)`, `caret_through_bridge_cached(other, leaf, &field) = None`.

templates/targets/rust_wam/state.rs.mustache
```diff
@@ -538,6 +538,12 @@ pub struct WamState {
     pub desc_sketch: std::collections::HashMap<u32, Vec<u64>>,
     pub desc_sketch_n: usize,
     pub desc_sketch_k: usize,
+    /// Designated-bridge landmark fields (roadmap 3f): bridge `B` -> `d(·→B)`, the up-distance
+    /// from every descendant of `B` to `B`. Built once by `build_caret_landmarks`; the
+    /// `category_caret_through_bridge` / `category_caret_min_over_landmarks` read-outs then
+    /// answer the designated-bridge caret in O(1) per pair instead of a per-query BFS. Empty =
+    /// landmark caret off (default). Eager-edge only.
+    pub caret_landmarks: std::collections::HashMap<u32, std::collections::HashMap<u32, u32>>,
     /// Int-native edge mode: when true, the lazy edge path treats the kernel's
```
s243a, line 547: **[non-blocking — serialization/Clone check]** If `WamState` derives `Clone` or `Serialize`/`Deserialize` (or if it's expected to in the future), the new `caret_landmarks: HashMap<u32, HashMap<u32, u32>>` field needs corresponding handling. `HashMap<u32, HashMap<u32, u32>>` is `Clone` and serde-serializable (with `serde` feature), so this is a no-op in practice — but if `WamState` has a custom `Clone` or manual serde impl, this field must be added there. Worth a quick grep for `impl Clone for WamState` or `#[derive(Clone)]` to confirm no custom impl is missing the new field.

docs/design/WAM_RUST_GRAPH_FUNCTIONAL_SEMIRINGS.md
```diff
@@ -971,10 +971,18 @@ buy diminishing returns and is not carried).
...
      interval, and compute multi-level budgeted carets between topics — the end-to-end
      composition on real (cyclic) data, where the 2a/2b cycle-correctness earns its keep.
-   - **[3f, buildable]** the **landmark-cached designated-bridge caret** (§5c): precompute
-     `d(·→B)` (one downward field) for each designated level `B`, so `caret_through_bridge`
-     is O(1) per pair — the boundary-cache reuse applied to the caret. `O(K·V)` for `K`
-     levels.
+   - **[3f DONE]** the **landmark-cached designated-bridge caret** (§5c): `bridge_distance_fields`
+     precomputes `d(·→B)` (one downward BFS per bridge over the shared children graph, `O(E +
+     Σ_B|desc(B)|) ≤ O(K·V)`), and `caret_through_bridge_cached` / `caret_min_over_cached_bridges`
+     then answer in **O(1)** / **O(#bridges)** per pair — the boundary-cache reuse applied to the
```
s243a, line 1003: **[non-blocking — context: caret-min is an upper bound, not always exact]** The updated 3f entry notes the min-plus connection accurately. One clarification worth adding for future readers: `caret_min_over_cached_bridges` / `caret_min_over_hubs` gives the **exact** LCA caret distance only if the true optimal common ancestor is one of the selected bridges. With a sparse bridge set this is an upper bound — the approximation gap is `2·d(optimal_LCA → nearest_bridge)`, as noted in §5c.

Deep research confirms this is the standard result in landmark distance estimation (Tretyakov et al. CIKM 2011; Storandt ISAAC 2022): `min_B(d(u→B) + d(v→B))` is a valid upper bound, exact iff the optimal meeting ancestor is a bridge. The current text implies this is exact by saying it "equals `caret_min_over_hubs`" — this is true (the cached function equals the per-query one), but neither is necessarily exact w.r.t. the true LCA caret unless bridges cover all optimal LCAs. A one-sentence clarifying note here (pointing back to §5c's existing approximation-gap discussion) would prevent misreading.

s243a (commented):
## Multi-model code review — PR #3262: landmark-cached designated-bridge caret O(1)

**Reviewers:** Claude Opus 4.8 · GPT-5.5 · Gemini deep research (hub labeling theory, BFS reversal, tropical semirings, landmark distance literature)

**Verdict: APPROVE. No blocking issues.**

The core mathematical claim is correct and verified. Opus 4.8 ran 2,000 random-DAG fuzzing trials plus hand-built diamond, direct-shortcut, multi-route asymmetry, and 3-cycle probes — zero mismatches. Deep research confirms BFS on the reversed graph is a textbook technique, the approach is a valid instantiation of the 2-hop cover / hub labeling framework (Cohen et al. 2002, Abraham et al. 2012), and the min-plus semiring framing is rigorous. Two required items (both documentation/test additions, no logic changes) and five non-blocking notes follow as inline comments.

s243a:
## Review fixes — both required items + all non-blocking notes addressed (commit `01f4b1e6`)

Thanks — APPROVE with only doc/test items, and the deep-research grounding was genuinely useful (folded into the theory doc, not just the inline comments). Disposition:

**Required**
- **R1** — `caret_through_bridge_cached` now carries the `u32`-overflow bound inline (depths `≤ V−1`, sum `< u32::MAX`; consistent with `up_distance_to`).
- **R2** — `bridge_distance_fields` now **de-duplicates** the `bridges` slice (a `HashSet` guard skips a repeated bridge's BFS), and the **pure-leaf-bridge** behaviour (`{B:0}` field → `None` for non-`(B,B)` pairs) is documented. Tests added: no-common-bridge `→ None` (with parity to `caret_min_over_hubs`), pure-leaf bridge, duplicate-bridge idempotence, and `build_caret_landmarks` `false` on an unknown predicate.

**Non-blocking, folded in**
- Complexity tightened to `O(E + Σ_B(|desc(B)| + edges_within)) ≤ O(K·(V+E))` time, `O(K·V)` memory (and the tightness noted).
- BFS-reversal wording made precise (length-preserving bijection ⇒ minima coincide).
- `caret_min_over_cached_bridges` caller-sync responsibility + upper-bound caveat documented.
- **Clone/serde:** `WamState` is `#[derive(Clone)]` with no custom or serde impl, so the new field is auto-handled — confirmed no manual handling needed.

**Theory captured in the docs** (per a follow-up request): a new §5c grounding note places 3f in the **2-hop cover / hub labeling** literature (Cohen et al. 2002; Abraham et al. 2012; Akiba et al. 2013), frames each field as a **column of the min-plus all-pairs matrix** and the caret as a **min-plus inner product**, records the `O(K·V)` space-tightness (Tretyakov et al. 2011), and states the upper-bound-iff-optimal-ancestor-is-a-bridge property (Storandt 2022). The 3f roadmap entry now cross-references it and clarifies the cached read is exact w.r.t. the LCA caret only when an optimal ancestor is a bridge.

Full suite: **78 passed, 0 failed.** Resolving the threads.

---

PR #3268, opened 2026-06-18
WAM-Rust: faithfulness measurement + the moment-ratio admissibility gate

s243a:
## What & why

Answers the question raised on the cumulant PR — *how do we know the moment/cumulant summary is a
good representation of the histogram **without computing the histogram**?* — first by **measuring**
it, then by building the cheap **gate** the measurement motivates.

Key reframing: **cumulants carry the same information as moments** (`κ_k ⟺ m_k`), so the §3 fork is
purely a *splice-cost/stability* axis — faithfulness is a §7 *reconstruction* question, identical
for both carries.

## 1. The measurement (`moment_reconstruction_faithful_on_chains_not_on_branches`)

Propagates the jet (**never** the histogram) on two synthetic graphs and scores the reconstruction
against the independently-computed true histogram:

| node shape | true distribution | `to_normal_repr` CDF error | excess kurtosis |
|---|---|---|---|
| `⊗`-heavy chain (24 length-`{1,2}` diamonds) | shifted binomial → Gaussian (CLT) | **0.0019** | `−0.08` |
| `⊕`-heavy branch (depths 6 vs 41) | bimodal mixture | **0.48** | `−2.0` |

Findings: the moments are **exact** (only the reconstruction approximates); faithfulness is known
from **structure** (chain → CLT → Gaussian), not the histogram; and the carried higher moments
**self-diagnose**.

## 2. The moment-ratio admissibility gate (`MomentJet::reconstruction_class` + `ReconClass`)

The diagnostic isn't a single kurtosis threshold but the **Pearson `(β₁, β₂)` moment-ratio
diagram** (`β₁ = skew²`, `β₂ = excess kurtosis`): every distribution obeys `β₂ ≥ β₁ − 2`, and a
**two-mode** distribution sits *on* it, so the slack `d = β₂ − β₁ + 2` is a histogram-free
**multimodality detector**. The gate returns:

- `d` small (near the two-mode bound) → **`NeedsHistogram`** (mixture — defer to histogram/GMM or a
  Monte-Carlo GOF test);
- else mild `|skew|,|kurtosis|` → **`Gaussian`**;
- else → **`GramCharlier`**.

Validated: CLT chain → `Gaussian`; bimodal branch (skew 0, exkurt −2.0, on the bound) →
`NeedsHistogram`; a **skewed binomial** (skew 0.31) → `GramCharlier` — *some skew is fine*, it's the
kurtosis/ratio that flags genuine non-normality.

## Future direction (documented)

For the genuinely-ambiguous middle of the diagram: a **Monte-Carlo goodness-of-fit** fallback
(sample paths → empirical distribution → test the parametric hypothesis), embarrassingly parallel
and a natural **GPU** workload. The §7 reconstruction stays CDF-gated as the final word; this gate
is the cheap pre-screen.

Documented as **[measured]** in §3. Full suite: **84 passed**.

### DIFF
--- a/docs/design/WAM_RUST_GRAPH_FUNCTIONAL_SEMIRINGS.md
+++ b/docs/design/WAM_RUST_GRAPH_FUNCTIONAL_SEMIRINGS.md
@@ -201,6 +201,36 @@ the *measurement* of where that crossover sits (which dominates a given cache re
 remaining open step, but the two representations it would choose between are now both shipped and
 interchangeable.
 
+**[measured]** The chain-vs-branch split that governs the *carry* also governs the deeper
+question — *when is the moment/cumulant summary a faithful stand-in for the histogram, without
+ever forming the histogram?* — and `moment_reconstruction_faithful_on_chains_not_on_branches`
+measures it. Two synthetic graphs propagate the jet (never the histogram) and the reconstruction
+is scored against the independently-computed true histogram:
+
+| node shape | true distribution | `to_normal_repr` CDF error | excess kurtosis (self-diagnostic) |
+|---|---|---|---|
+| `⊗`-heavy chain (24 length-`{1,2}` diamonds) | shifted binomial → Gaussian (CLT) | **0.0019** | `−0.08` (reads Gaussian) |
+| `⊕`-heavy branch (two depths, 6 vs 41) | bimodal mixture | **0.48** | `−2.0` (reads strongly non-Gaussian) |
+
+The answers this pins down (and the honest limits):
+
+- **The moments are exact; only the *reconstruction* approximates.** Mean/variance/skew/kurtosis
+  are computed exactly by propagation — `κ₂ = 24·0.25` on the chain confirms cumulant additivity.
+  Faithfulness is entirely a property of the reconstruction step (moments → distribution shape).
+- **You know it is faithful from *structure*, not the histogram.** A `⊗`-heavy node is a sum of
+  many independent stages → Gaussian by the CLT, so the moment-Normal is near-exact (0.002); a
+  `⊕`-heavy node is a *mixture*, possibly multimodal, where it fails (0.48). The graph shape is the
+  prior — read off the topology, not the histogram.
+- **The carried higher moments *self-diagnose*.** The bimodal branch reads excess kurtosis `−2.0`
+  — a cheap "I am not Gaussian" flag with no histogram. **Honest gap:** for a *symmetric* mixture
+  the skew is `~0` (blind to the bimodality); kurtosis catches this one, but a general multimodal
+  shape still needs the exact histogram fallback — which is exactly why the §7 reconstruction is
+  **CDF-gated**, validated against the true distribution during calibration rather than assumed.
+- **Cumulants vs moments is *orthogonal* to all of this.** They carry the same information
+  (`κ_k ⟺ m_k`), so the reconstruction is identical; the §3 fork is purely the *splice cost /
+  numerical-stability* axis (additive cumulants for `⊗`-heavy spines), not a representation-quality
+  axis. Faithfulness is a §7 reconstruction question, the same for both carries.
+
 ### The one that does *not* factor: `WeightSum`
 
 **The unifying principle: point-evaluation of the GF.** Treat `H(z) = Σ_L H[L] z^L` as a
--- a/docs/design/WAM_RUST_GRAPH_FUNCTIONAL_SEMIRINGS.md
+++ b/docs/design/WAM_RUST_GRAPH_FUNCTIONAL_SEMIRINGS.md
@@ -223,23 +223,37 @@ The answers this pins down (and the honest limits):
   prior — read off the topology, not the histogram.
 - **The carried higher moments *self-diagnose*, via their *ratio*.** Not a single threshold but the
   **Pearson `(β₁, β₂)` moment-ratio diagram** (`β₁ = skew²`, `β₂ = excess kurtosis`): every
-  distribution obeys the universal bound `β₂ ≥ β₁ − 2`, and a **two-mode** distribution sits *on*
-  it (the bimodal branch: skew `0`, excess kurtosis `−2.0 = 0 − 2`, the extremum). So the slack
-  `d = β₂ − β₁ + 2 ≥ 0` is a histogram-free **multimodality detector**, and `MomentJet::
-  reconstruction_class` reads it: `d` small → **Multimodal**; mild `|skew|,|kurtosis|` →
+  distribution obeys the universal bound `β₂ ≥ β₁ − 2`, with equality attained **only** by two-point
+  (Bernoulli) distributions (the bimodal branch sits there: skew `0`, excess kurtosis `−2.0 = 0 −
+  2`). So the slack `d = β₂ − β₁ + 2 ≥ 0` is a histogram-free **multimodality detector**, and
+  `MomentJet::reconstruction_class` reads it: `d < 0.7` → **Multimodal**; mild `|skew|,|kurtosis|` →
   **Gaussian**; otherwise **GramCharlier**. Crucially **some skew is fine** — a skewed binomial
   (`skew 0.31`) classifies `GramCharlier`, reconstructed by the skew/kurtosis corrections — so it
   is the *kurtosis and the ratio*, not skew, that flag genuine non-normality.
+- **The `0.7` threshold is a conservative heuristic, not the rigorous boundary** — and this is the
+  interesting subtlety. `0.7` has **zero false positives** (anything below it is provably near the
+  two-mode extremum), but `d` does *not* cleanly separate multimodal from unimodal: the
+  **platykurtic** band `0.7 ≲ d ≲ 2` holds *both*. The **uniform** is unimodal/amodal yet has
+  `d ≈ 0.80` (excess kurtosis `−1.2`); four moments cannot tell a flat-but-unimodal shape from a
+  mild bimodal. `0.7` sits just below the uniform's `0.80`, so platykurtic-unimodals route to
+  GramCharlier — *raising it to ≈1.5 would wrongly call the uniform a mixture.* (Published
+  unimodality floors — Sharma & Bhandari 2015; Klaassen & van Es 2023, `d ≥ 189/125 ≈ 1.512` — are
+  for *strictly* unimodal densities and exclude the flat uniform, which is why we keep `0.7`.) The
+  platykurtic band is precisely the **genuinely ambiguous middle** the moments can't resolve — the
+  case for the §7 CDF gate or a Monte-Carlo / GMM fit, not a sharper threshold.
 - **Multimodal is still *closed-form* — a mixture, not "give up and store the histogram."** The
   flag means *not a single mode*, so reconstruct with a **mixture / GMM** (`HistRepr::DiscGmm` /
   `Mixture`). And the same **Pearson** framework behind the `(β₁,β₂)` diagram is Pearson's 1894
-  **method of moments** for a 2-Gaussian mixture: in the symmetric case the mixture is fit from the
-  *same four moments* in closed form — two equal-weight modes at `μ ± δ` with `δ = σ·(−γ₂/2)^¼`. On
-  the bimodal branch that yields modes at exactly `{6, 41}` (the true spikes) from the same moments
-  the single Gaussian botched at error `0.48`. So the ladder is Gaussian → Gram–Charlier → **mixture
-  (closed form)** → exact histogram, and only the last is non-parametric. **Honest gap:** four
-  moments under-determine a *general* (asymmetric, unequal-weight, many-component) mixture; fitting
-  that needs more — the histogram, or a **Monte-Carlo goodness-of-fit / EM** step (sample paths,
+  **method of moments** for a 2-Gaussian mixture: in the *symmetric* case (equal weights, equal
+  component variance) the mixture is fit from the *same four moments* in closed form — two modes at
+  `μ ± δ` with `δ = σ·(−γ₂/2)^¼`, which is **exact for all component variances `s ≥ 0`** (from
+  `γ₂ = −2(δ/σ)⁴`), not just the `s→0` point-mass limit. On the bimodal branch that yields modes at
+  exactly `{6, 41}` (the true spikes) from the same moments the single Gaussian botched at error
+  `0.48`. So the ladder is Gaussian → Gram–Charlier → **mixture (closed form)** → exact histogram,
+  and only the last is non-parametric. **Honest gap:** four moments under-determine a *general*
+  mixture — Pearson's *asymmetric* 2-Gaussian case already reduces to a **9th-degree (nonic)
+  polynomial** in the separation and typically needs more than four moments; fitting that needs the
+  histogram, or a **Monte-Carlo goodness-of-fit / EM** step (sample paths,
   build the empirical distribution, fit/test), which is embarrassingly parallel and a natural
   **GPU** workload (a *future* direction). The §7 reconstruction stays **CDF-gated**
   (histogram-validated) as the final word; `reconstruction_class` is the cheap pre-screen that
--- a/docs/design/WAM_RUST_GRAPH_FUNCTIONAL_SEMIRINGS.md
+++ b/docs/design/WAM_RUST_GRAPH_FUNCTIONAL_SEMIRINGS.md
@@ -226,17 +226,25 @@ The answers this pins down (and the honest limits):
   distribution obeys the universal bound `β₂ ≥ β₁ − 2`, and a **two-mode** distribution sits *on*
   it (the bimodal branch: skew `0`, excess kurtosis `−2.0 = 0 − 2`, the extremum). So the slack
   `d = β₂ − β₁ + 2 ≥ 0` is a histogram-free **multimodality detector**, and `MomentJet::
-  reconstruction_class` reads it: `d` small → **NeedsHistogram**; mild `|skew|,|kurtosis|` →
+  reconstruction_class` reads it: `d` small → **Multimodal**; mild `|skew|,|kurtosis|` →
   **Gaussian**; otherwise **GramCharlier**. Crucially **some skew is fine** — a skewed binomial
   (`skew 0.31`) classifies `GramCharlier`, reconstructed by the skew/kurtosis corrections — so it
-  is the *kurtosis and the ratio*, not skew, that flag genuine non-normality. **Honest gap that
-  remains:** the moment ratios still cannot resolve an *arbitrary* multimodal shape; for the
-  genuinely-ambiguous middle of the diagram the principled fallback is a **Monte-Carlo
-  goodness-of-fit test** — sample paths, build the empirical distribution, test the parametric
-  hypothesis — which is embarrassingly parallel and a natural **GPU** workload (a *future*
-  direction). And the §7 reconstruction stays **CDF-gated** (histogram-validated) as the final
-  word; `reconstruction_class` is the cheap pre-screen that decides whether to attempt a parametric
-  form at all.
+  is the *kurtosis and the ratio*, not skew, that flag genuine non-normality.
+- **Multimodal is still *closed-form* — a mixture, not "give up and store the histogram."** The
+  flag means *not a single mode*, so reconstruct with a **mixture / GMM** (`HistRepr::DiscGmm` /
+  `Mixture`). And the same **Pearson** framework behind the `(β₁,β₂)` diagram is Pearson's 1894
+  **method of moments** for a 2-Gaussian mixture: in the symmetric case the mixture is fit from the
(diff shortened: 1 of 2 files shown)


templates/targets/rust_wam/boundary_cache.rs.mustache
```diff
@@ -288,6 +288,53 @@ impl MomentJet {
...
+    /// histogram-validated check — this is the cheap pre-screen that decides *which* parametric
+    /// family to attempt (single mode vs mixture), not whether to give up on a closed form.
+    pub fn reconstruction_class(self) -> ReconClass {
+        if self.mass <= 0.0 {
+            return ReconClass::Multimodal;
+        }
+        let g1 = self.skewness();
+        let g2 = self.excess_kurtosis();
+        let d = g2 - g1 * g1 + 2.0; // slack above the universal two-mode bound β₂ = β₁ − 2
+        if d < 0.7 {
```
s243a, line 331: **[required — threshold calibration: d < 0.7 is conservative; rigorous floor is ≈ 1.512]**

Deep research surfaced the published Pearson (β₁,β₂) unimodality bound: by Klaassen & van Es (2023) and the original Sharma & Bhandari (2015) proof via Newton's inequality, every **unimodal** distribution satisfies `d ≥ 189/125 ≈ 1.512`. The boundary `d = 0` is achieved only by two-point (Bernoulli) distributions.

This means:
- `d < 0.7` → **zero false positives** (any distribution with d < 0.7 is provably multimodal or two-point — the gate is safe)
- But **0.7 ≤ d < 1.512** is a gap where genuinely multimodal distributions exist and will be misclassified as `GramCharlier` instead of `Multimodal`
- The bimodal branch sits at d = 0.0 — it's caught regardless of the threshold

The threshold 0.7 is not arbitrary (it has no false positives), but the gap needs documenting. Two choices:

**Option A (recommended — one line):** Raise threshold to `d < 1.5` to close the gap for all provably-multimodal distributions, routing them to GMM rather than GramCharlier. Conservative in the other direction (some near-boundary distributions that Gram-Charlier could handle will go to GMM instead — safe but slightly expensive).

**Option B (document the gap):** Keep `d < 0.7` and add a comment:
```rust
// Threshold 0.7: all unimodal distributions satisfy d ≥ 1.512 (Pearson bound — see
// Sharma & Bhandari 2015), so d < 0.7 has zero false positives. The gap [0.7, 1.512)
// contains provably-multimodal distributions that fall through to GramCharlier; the
// §7 CDF gate catches any that produce invalid densities there.
if d < 0.7 {
```

Either is acceptable — the current code is **safe** (no incorrect results, just some multimodal distributions taking the GramCharlier path). Document the gap so no one treats 0.7 as the rigorous boundary.

templates/targets/rust_wam/boundary_cache.rs.mustache
```diff
@@ -288,6 +288,53 @@ impl MomentJet {
...
+            ReconClass::Multimodal // near the multimodal extremum → a closed-form mixture, not a single mode
+        } else if g1.abs() <= 0.2 && g2.abs() <= 0.7 {
+            ReconClass::Gaussian
```
s243a, line 334: **[required — doc: Gram-Charlier validity range is bounded]**

Deep research ([Jondeau & Rockinger 2001](https://people.unil.ch/ericjondeau/files/2021/09/JondeauRockinger_JEDC_2001.pdf)) confirms: the Gram-Charlier A expansion produces a **non-negative density only over a restricted region** of the (γ₁, γ₂) plane, not for all skew/kurtosis values. There is no simple closed-form boundary; numerically it is roughly |γ₁| ≲ 2, γ₂ > −2 + ε.

The PR's GramCharlier branch is entered for any distribution not caught by the `d < 0.7` or `|g1| ≤ 0.2 && |g2| ≤ 0.7` gates — in principle this includes distributions with |γ₁| > 2 or γ₂ near −2, where Gram-Charlier yields a non-positive "density."

However, the §7 CDF gate (histogram-validated) is the final word before any reconstruction is accepted, so a negative-density GC expansion would be caught there. The pre-screen is still safe.

Suggest adding to the GramCharlier arm's doc-comment:
```
/// - else (notable skew/kurtosis but not multimodal) → **GramCharlier** (skew/kurtosis
///   corrections, mildly non-normal). Note: the Gram-Charlier A expansion is non-negative
///   only for (γ₁,γ₂) in a bounded region (roughly |γ₁|≲2; Jondeau & Rockinger 2001);
///   large skew/kurtosis values MAY yield a non-positive density. The §7 CDF gate
///   rejects any such result before it is accepted.
```

templates/targets/rust_wam/boundary_cache.rs.mustache
```diff
@@ -288,6 +288,53 @@ impl MomentJet {
...
+    ///   general mixture needs more than four moments (the histogram, or a Monte-Carlo / EM fit).
+    /// - else mild `|γ₁|,|γ₂|` → **Gaussian** (the moment-Normal is faithful, the CLT case);
+    /// - else (notable skew/kurtosis but not multimodal) → **GramCharlier** (skew/kurtosis
+    ///   corrections, mildly non-normal).
+    /// Thresholds are deliberately conservative defaults; the §7 CDF gate remains the final,
+    /// histogram-validated check — this is the cheap pre-screen that decides *which* parametric
+    /// family to attempt (single mode vs mixture), not whether to give up on a closed form.
+    pub fn reconstruction_class(self) -> ReconClass {
+        if self.mass <= 0.0 {
+            return ReconClass::Multimodal;
```
s243a: **[non-blocking — zero-mass fallback semantics]** `mass <= 0.0 → Multimodal` routes a degenerate empty jet to GMM/histogram fallback, which is safe. Confirmed: the zero-variance path (`mass > 0, var ≤ 1e-12`) is already handled by the `skewness()` / `excess_kurtosis()` guards (`var <= 1e-12 → return 0.0`), giving `g1 = 0, g2 = 0, d = 2.0` → correctly classified **Gaussian** (a point mass IS a degenerate Gaussian).

Only the corrupt-upstream-moments path (NaN-valued moments from a bug elsewhere) could slip through — the `d = NaN` result would skip `d < 0.7` (false) and `NaN.abs() <= 0.2` (false) and return `GramCharlier`. Consider an optional guard:
```rust
if !d.is_finite() || !g1.is_finite() || !g2.is_finite() {
    return ReconClass::Multimodal; // corrupt moments → safest fallback
}
```
This is defensive-only (no known path produces NaN through normal usage), but cheap.

templates/targets/rust_wam/boundary_cache.rs.mustache
```diff
@@ -4066,6 +4113,106 @@ mod tests {
...
+        let skewed = hist_moment_jet(&binomial_pmf(40, 0.15).iter()
+            .map(|&pr| (pr * 1_000_000.0).round() as u64).collect::<Hist>());
+        eprintln!("§3 skewed-binomial: skew {:.3} exkurt {:.3} -> {:?}",
+            skewed.skewness(), skewed.excess_kurtosis(), skewed.reconstruction_class());
+        assert_eq!(skewed.reconstruction_class(), ReconClass::GramCharlier);
+
+        // FINDING 4: "Multimodal" is NOT "needs the exact histogram" — the SAME four moments the
+        // single Gaussian botched (cdf_err 0.48) closed-form *locate the two modes* as a symmetric
+        // 2-Gaussian mixture (Pearson 1894 MoM): for two equal-weight modes at μ±δ with component
+        // variance s², γ₂ = −2(δ²/σ²)², so δ = σ·(−γ₂/2)^¼. On the bimodal that is exactly {6, 41}.
```
s243a: **[non-blocking — Pearson 1894 formula scope: exact for all s ≥ 0, not just point masses]**

The doc-comment says the formula `δ = σ·(−γ₂/2)^¼` is for "equal-weight modes at μ±δ with component variance s²" — but does not state whether it's exact or approximate for s > 0. Opus 4.8 proved symbolically (SymPy) that `γ₂ = −2(δ²/σ²)²` exactly for **all** component variances s ≥ 0, not just the point-mass limit. Proof sketch: with t = δ²/s²:
```
γ₂ = (3s⁴ + 6δ²s² + δ⁴)/(s²+δ²)² − 3 = −2δ⁴/(s²+δ²)² = −2(δ/σ)⁴
```
So `δ/σ = (−γ₂/2)^¼` exactly. The bimodal branch test passes not because it's a special case but because the formula is always exact.

Add one sentence to the FINDING 4 comment:
```rust
// Note: the formula δ = σ·(−γ₂/2)^¼ is EXACT for all symmetric equal-weight
// equal-component-variance Gaussian mixtures (not just the point-mass limit s→0).
// Derived from γ₂ = −2(δ/σ)⁴ which holds for all s ≥ 0.
```
This makes the formula more credible and distinguishes it from a heuristic.

templates/targets/rust_wam/boundary_cache.rs.mustache
```diff
@@ -4066,6 +4113,106 @@ mod tests {
...
+            let (mut memo, mut st) = (HashMap::new(), Vec::new());
+            suffix_moment_jet(500, 0, &br, &mut memo, &mut st)
+        };
+        let berr = cdf_err(&bh, &bjet.to_normal_repr(bh.len()).pmf());
+        eprintln!("§3 branch: mean {:.2} var {:.2} skew {:.3} exkurt {:.3} | cdf_err Normal {:.4}",
+            bjet.mean(), bjet.variance(), bjet.skewness(), bjet.excess_kurtosis(), berr);
+
+        // FINDING 1: moments reconstruct the chain (CLT) accurately, the branch mixture badly.
+        assert!(err_normal < 0.06, "chain Gaussian reconstruction is faithful: cdf_err {}", err_normal);
+        assert!(berr > 0.2 && berr > 4.0 * err_normal,
```
s243a: **[non-blocking — assertion slack vs documented values]** The doc table says CDF error = **0.0019** for the chain and **0.48** for the branch, but the test only asserts `err_normal < 0.06` (31× looser) and `berr > 0.2`. These loose bounds are correct for a non-regression test, but if the implementation regresses to 0.05 the doc's headline number won't be caught.

Consider tightening to `< 0.005` for the chain (close to the observed 0.0019) and `> 0.30` for the branch — still generous enough to survive floating-point variance across platforms, while pinning the documented claims more precisely.

templates/targets/rust_wam/boundary_cache.rs.mustache
```diff
@@ -4066,6 +4113,106 @@ mod tests {
...
+        assert!(jet.excess_kurtosis().abs() < 0.5, "chain reads ~Gaussian");
+        assert!(bjet.excess_kurtosis() < -1.0, "bimodal branch flagged via kurtosis");
+        assert!(bjet.skewness().abs() < 0.2, "symmetric bimodal: skew is blind to it");
+
+        // FINDING 3: the moment-ratio gate (Pearson β₁,β₂) picks the right reconstruction from the
+        // ratios alone — Gaussian for the CLT chain, Multimodal for the bimodal mixture (it sits on
+        // the two-mode bound β₂ = β₁ − 2). A mildly SKEWED binomial reads GramCharlier: some skew is
+        // fine, it is the kurtosis/ratio that flags genuine non-normality.
+        assert_eq!(jet.reconstruction_class(), ReconClass::Gaussian);
+        assert_eq!(bjet.reconstruction_class(), ReconClass::Multimodal);
```
s243a, line 4212: **[non-blocking — missing tests: threshold boundary, leptokurtic-unimodal, high-kurtosis GC]**

Three small cases not currently covered:

**1. Threshold boundary near d = 0.7:** A distribution with γ₁ = 1.0, γ₂ = −0.5 gives d = −0.5 − 1.0 + 2.0 = 0.5 → Multimodal. A distribution with γ₁ = 0, γ₂ = −1.4 gives d = 0.6 → Multimodal. Both are classified Multimodal correctly (d < 0.7). A distribution with d = 0.72 should give GramCharlier. A boundary test pins the 0.7 threshold:
```rust
// Direct construction of a jet with known (g1, g2):
let mut jet = MomentJet { mass: 1.0, m1: 0.0, m2: 1.0, m3: 0.0, m4: /* g2_target+3 */ 2.7 };
// excess_kurtosis would return 2.7/1.0 - 3 = -0.3... actual construction needs care
```

**2. Leptokurtic-but-unimodal:** The deep research confirms that platykurtic (negative excess kurtosis) unimodal distributions all satisfy d ≥ 1.512. A uniform distribution has γ₂ = −1.2 and d = −1.2 + 2 = 0.8 > 0.7 → GramCharlier (not Multimodal) ✓. But a test asserting this explicitly would pin the safe region.

**3. High-kurtosis GramCharlier concern:** A distribution with |γ₁| > 2 enters GramCharlier but may produce a non-positive density. No test currently exercises this path (it would require a constructed jet with large skew/kurtosis). Noting this is acceptable since the §7 CDF gate is the backstop — but a test or debug_assert would be valuable.

docs/design/WAM_RUST_GRAPH_FUNCTIONAL_SEMIRINGS.md
```diff
@@ -201,6 +201,55 @@ the *measurement* of where that crossover sits (which dominates a given cache re
...
+  are computed exactly by propagation — `κ₂ = 24·0.25` on the chain confirms cumulant additivity.
+  Faithfulness is entirely a property of the reconstruction step (moments → distribution shape).
+- **You know it is faithful from *structure*, not the histogram.** A `⊗`-heavy node is a sum of
+  many independent stages → Gaussian by the CLT, so the moment-Normal is near-exact (0.002); a
+  `⊕`-heavy node is a *mixture*, possibly multimodal, where it fails (0.48). The graph shape is the
+  prior — read off the topology, not the histogram.
+- **The carried higher moments *self-diagnose*, via their *ratio*.** Not a single threshold but the
+  **Pearson `(β₁, β₂)` moment-ratio diagram** (`β₁ = skew²`, `β₂ = excess kurtosis`): every
+  distribution obeys the universal bound `β₂ ≥ β₁ − 2`, and a **two-mode** distribution sits *on*
+  it (the bimodal branch: skew `0`, excess kurtosis `−2.0 = 0 − 2`, the extremum). So the slack
```
s243a: **[non-blocking — precision: Pearson bound wording]** "every distribution obeys the universal bound `β₂ ≥ β₁ − 2`" is correct. Adding "with equality attained only by two-point (Bernoulli) distributions" would complete the picture, since readers might wonder what kind of distribution achieves the bound. This is the key result: only a distribution with all mass on two atoms sits exactly at d = 0 ([Klaassen & van Es 2023](https://arxiv.org/abs/2312.06212); [Sharma & Bhandari 2015](https://projecteuclid.org/journals/rocky-mountain-journal-of-mathematics/volume-45/issue-5/Skewness-kurtosis-and-Newtons-inequality/10.1216/RMJ-2015-45-5-1639.pdf)).

Also: the more precise fact — "every **unimodal** distribution satisfies d ≥ 189/125 ≈ 1.512" — could replace the current `d` small / `d` large split description and make the threshold semantics more precise. The gap [0.7, 1.512) contains distributions that are provably multimodal by the unimodality theorem but below the current code threshold.

docs/design/WAM_RUST_GRAPH_FUNCTIONAL_SEMIRINGS.md
```diff
@@ -201,6 +201,55 @@ the *measurement* of where that crossover sits (which dominates a given cache re
...
+  `d = β₂ − β₁ + 2 ≥ 0` is a histogram-free **multimodality detector**, and `MomentJet::
+  reconstruction_class` reads it: `d` small → **Multimodal**; mild `|skew|,|kurtosis|` →
+  **Gaussian**; otherwise **GramCharlier**. Crucially **some skew is fine** — a skewed binomial
+  (`skew 0.31`) classifies `GramCharlier`, reconstructed by the skew/kurtosis corrections — so it
+  is the *kurtosis and the ratio*, not skew, that flag genuine non-normality.
+- **Multimodal is still *closed-form* — a mixture, not "give up and store the histogram."** The
+  flag means *not a single mode*, so reconstruct with a **mixture / GMM** (`HistRepr::DiscGmm` /
+  `Mixture`). And the same **Pearson** framework behind the `(β₁,β₂)` diagram is Pearson's 1894
+  **method of moments** for a 2-Gaussian mixture: in the symmetric case the mixture is fit from the
+  *same four moments* in closed form — two equal-weight modes at `μ ± δ` with `δ = σ·(−γ₂/2)^¼`. On
```
s243a: **[non-blocking — Pearson 1894 method of moments: historical accuracy note]** The text accurately credits Pearson's 1894 paper for fitting a bimodal mixture by method-of-moments — this is historically correct (Pearson's famous 1894 paper on the "dissection" of a frequency curve). One nuance to add: Pearson's general 2-Gaussian mixture formulation leads to a **9th-degree polynomial** (nonic) in the separation parameter, not a simple quartic. The closed-form `δ = σ·(−γ₂/2)^¼` is the special case for symmetric equal-weight equal-variance mixtures, which reduces nicely.

Suggest: "in the *symmetric* case (equal weights, equal component variance) the mixture is fit from the same four moments in closed form: `δ = σ·(−γ₂/2)^¼` (exact for all s ≥ 0 — see verification in `FINDING 4`). The general asymmetric case requires a 9th-degree polynomial and typically more than four moments."

s243a (commented):
## Multi-model code review — PR #3268: faithfulness measurement + moment-ratio admissibility gate

**Reviewers:** Claude Opus 4.8 · GPT-5.5 · Gemini deep research (Pearson β₁β₂ diagram, Gram-Charlier validity, Berry-Esseen CLT rate, Pearson 1894 mixture formula)

**Verdict: APPROVE. No blocking issues. One required item; remaining are documentation clarifications.**

The gate logic is mathematically correct and all three cases are verified. The Pearson (β₁,β₂) slack formula is sound. The δ = σ·(−γ₂/2)^¼ mode formula is exact for all s ≥ 0 (not just the point-mass limit — Opus 4.8 proved this symbolically). Zero-variance jets correctly classify as Gaussian via existing guards in `skewness()`/`excess_kurtosis()`. One required item: the threshold `d < 0.7` is conservative (the rigorous unimodality floor is `d ≈ 1.512`) — this means some genuinely multimodal distributions escape the gate and get GramCharlier treatment; should be documented.

s243a:
## Review fixes (commit `f61c8530`) — and an important note on R1

Thanks — APPROVE with strong theory. All 8 threads addressed, but **on R1 I went with Option B (keep `d < 0.7`), not the recommended Option A (raise to ≈1.5), because Option A introduces a bug.**

**R1 — why not raise the threshold.** The review is internally inconsistent here: it cites a unimodal floor `d ≥ 1.512` (which would make `d = 0.8` multimodal), yet its own missing-tests comment says the **uniform** (`d ≈ 0.8`) is correctly *not* multimodal. I verified empirically:

```
uniform 0..=40: excess kurtosis −1.20, d = 0.799
  threshold 0.7 → GramCharlier   ✓ (uniform is unimodal/amodal, not a mixture)
  threshold 1.5 → Multimodal     ✗ (a uniform is NOT a 2-Gaussian mixture)
```

So raising to ≈1.5 would mislabel **every platykurtic-unimodal** distribution in `[0.7, 1.5)` as a mixture. The published `189/125 ≈ 1.512` floors (Sharma & Bhandari 2015; Klaassen & van Es 2023) are for *strictly unimodal* densities and exclude the flat uniform. The honest framing — now in the doc-comment and §3 — is that `0.7` is a **conservative heuristic** (zero false positives), the platykurtic band `(0.7, ~2)` is the **genuinely ambiguous middle** (flat-unimodal vs mild-bimodal, which four moments can't separate), and that's exactly what the §7 CDF gate / Monte-Carlo resolves. `0.7` sits just below the uniform's `0.80` deliberately.

**Everything else folded in:**
- **GC validity** — GramCharlier arm now notes the GC-A non-negative region is bounded (`|γ₁| ≲ 2`; Jondeau & Rockinger 2001), §7 gate backstops.
- **NaN/empty guard** → `Multimodal` (safe fallback).
- **δ exact for all `s ≥ 0`** — FINDING 4 now states `γ₂ = −2(δ/σ)⁴` holds for all component variances, not just the point-mass limit.
- **Assertions tightened** to pin the documented numbers (chain `< 0.005`, branch `> 0.30`).
- **Tests** — added uniform → `GramCharlier` (pins the platykurtic-unimodal region just above the threshold), a two-mode jet (`d=0`) → `Multimodal` (pins below), empty jet → `Multimodal`. The high-kurtosis-GC non-positive-density path is documented as §7-backstopped, as the review accepted.
- **Docs §3** — two-point equality at the bound; the platykurtic ambiguity + uniform; Pearson's *asymmetric* 2-Gaussian case is a 9th-degree nonic (the symmetric quartic is the reducible special case); GC validity range.

Also noted from deep research (no action, just confirmation): the chain's `0.0019` beating the Berry–Esseen `~0.097` bound is because the symmetric chain (`γ₁ = 0`) kills the `O(n^−½)` Edgeworth term, leaving `O(n^−1)`.

Full suite: **84 passed.** Resolving the threads.

---

PR #3270, opened 2026-06-18
WAM-Rust: tiered/gated μ — cheap stage decides, expensive stage only on the close band

s243a:
## What & why

A **tier-agnostic gate** for membership scoring: a *cheap* stage decides the confident nodes, and
an *expensive* stage is consulted **only on the "near-miss" band** — the "cheap signal broadly,
expensive signal only where uncertain" pattern (same shape as the reconstruction gate's
Monte-Carlo fallback). The two stages can be any cheap→expensive pairing; this PR demonstrates two.

## 1. Cheap prior + LLM (`wikipedia_fuzzy_gated_hybrid_membership`)

A cheap *prior* (an embedding-similarity `μ`, or — stand-in here — depth-to-root) accepts/rejects
outright; the LLM is consulted only in the band `[τ_lo, τ_hi)`, fused by the **geometric mean**
`√(prior·μ_llm)` (hard-vetoes if either ~0). On the 90-node fixture with the *weak* depth prior:
**29 accept-on-prior, 23 reject-on-prior, 38 consult the LLM — 58% of LLM calls saved**.

## 2. Model cascade — Haiku → Sonnet (`wikipedia_model_cascade_haiku_then_sonnet`)

The stages need not be embedding+LLM — they can be **two models**: Haiku for the bulk, Sonnet only
on Haiku's uncertain band (`μ∈(0.3,0.7)`); an `n`-tier cascade in the limit.

```
90 nodes | Haiku decided 71 (79%) | escalated to Sonnet 19 | Sonnet resolved 12 of those
band score spread  Haiku σ=0.061  Sonnet σ=0.160 (wider ⇒ more decisive)
```

Sonnet un-clusters Haiku's `0.5` pile — chemistry down (`Arsenic 0.1`, `Pnictogens 0.05`),
engineering down (`Electric_vehicles 0.1`), physics-adjacent held (`Astronomical_objects 0.7`).
Strong-tier re-scores in `tests/fixtures/wikipedia_physics_sonnet_band.tsv`.

## Batch-size caveat (documented)

Escalation only pays off if the band is large enough to amortize the **sunk cost** (system prompt +
invocation): a batch should be at least as large as the fixed system-prompt cost (fixed ≤ half),
ideally `band ≫ sunk_cost`. Too few items per escalated batch erodes the win — band, batch size, and
tier-cost ratio must be sized together.

Both harnesses env-gated (skip in CI). Documented in the measurement doc. Full suite: **87 passed**.

### DIFF
--- a/templates/targets/rust_wam/boundary_cache.rs.mustache
+++ b/templates/targets/rust_wam/boundary_cache.rs.mustache
@@ -3867,8 +3867,11 @@ mod tests {
         // Two-threshold gate: accept on the prior if >= TAU_HI; reject if < TAU_LO; consult the LLM
         // only in the "just-missed-but-close" band [TAU_LO, TAU_HI), and accept there iff the
         // GEOMETRIC MEAN sqrt(prior·μ_llm) >= TAU_GEO. The geo-mean hard-vetoes if either is ~0.
+        // (This √(prior·μ) fusion is a *logarithmic opinion pool* — Clemen & Winkler 1999 — the
+        // equal-weight log pool, NOT a Bayesian posterior: it treats the two scores as independent
+        // expert opinions to average in log-space, with no likelihood/prior factorization.)
         const TAU_HI: f64 = 0.66;   // prior alone accepts (shallow ⇒ depth ≤ 2)
-        const TAU_LO: f64 = 0.33;   // prior alone rejects (deep ⇒ depth ≥ 4)
+        const TAU_LO: f64 = 0.33;   // prior alone rejects (deep ⇒ depth ≥ 5: c=1−d/6, d=4 gives 0.333 > τ_lo)
         const TAU_GEO: f64 = 0.5;
         let (mut acc_cheap, mut rej_cheap, mut llm_calls, mut acc_geo, mut rej_geo) = (0, 0, 0, 0, 0);
         for (&u, &mu) in &llm {
@@ -3892,6 +3895,8 @@ mod tests {
         // the ambiguous band, so the expensive calls are a fraction of the total.
         assert!(llm_calls < total, "the gate avoids an LLM call on the confident nodes");
         assert!(acc_cheap + rej_cheap > 0, "the cheap prior decides some nodes outright");
+        let saved = (total - llm_calls) as f64 / total as f64;
+        assert!(saved > 0.40, "the gate must save the majority of LLM calls (saved {:.0}%)", 100.0 * saved);
     }
 
     // The gate is TIER-AGNOSTIC: the two stages need not be embedding+LLM — they can be two models,
@@ -3912,7 +3917,10 @@ mod tests {
                 .collect()
         };
         let (h, s) = (load(&haiku), load(&sonnet));
-        let confident = |mu: f64| mu <= 0.3 || mu >= 0.7;
+        // The escalation band is CLOSED [0.3, 0.7] (endpoints escalate), so `confident` is the
+        // strict complement μ<0.3 ∨ μ>0.7. This matches the Sonnet band fixture exactly (26 nodes);
+        // the open band (0.3,0.7) would drop the μ=0.3/0.7 endpoints and under-count to 19.
+        let confident = |mu: f64| mu < 0.3 || mu > 0.7;
         // Tier 1: Haiku decides the confident nodes; the uncertain band escalates to Sonnet.
         let (mut by_haiku, mut escalated, mut resolved_by_sonnet) = (0, 0, 0);
         for (name, &mu) in &h {
@@ -3938,6 +3946,7 @@ mod tests {
         let (sd_h, sd_s) = (std(&|n| h[n]), std(&|n| s[n]));
         eprintln!("cascade: band score spread  Haiku σ={:.3}  Sonnet σ={:.3} (wider ⇒ more decisive)", sd_h, sd_s);
 
+        assert_eq!(by_haiku + escalated, total, "every node is either Haiku-decided or escalated (partition)");
         assert!(escalated < total, "the cheap tier decides the confident majority");
         assert!(resolved_by_sonnet * 2 >= escalated, "the strong tier resolves most of the escalated band");
         assert!(sd_s > sd_h, "the strong tier is more decisive (wider score spread) on the band");
--- a/templates/targets/rust_wam/boundary_cache.rs.mustache
+++ b/templates/targets/rust_wam/boundary_cache.rs.mustache
@@ -3894,6 +3894,55 @@ mod tests {
         assert!(acc_cheap + rej_cheap > 0, "the cheap prior decides some nodes outright");
     }
 
+    // The gate is TIER-AGNOSTIC: the two stages need not be embedding+LLM — they can be two models,
+    // a cheap one for the bulk and a strong one only on the near-misses (an n-tier cascade in the
+    // limit). Here: Haiku (UW_FUZZY_NODES) decides the confident nodes; the Haiku-uncertain band
+    // (μ∈[0.3,0.7]) escalates to Sonnet (UW_SONNET_BAND), which is more discriminating and RESOLVES
+    // most of it. Gated on those two fixtures (skips in CI).
+    #[test]
+    fn wikipedia_model_cascade_haiku_then_sonnet() {
+        let haiku = match std::env::var("UW_FUZZY_NODES").ok() { Some(p) => p, None => { eprintln!("cascade: skip"); return; } };
+        let sonnet = match std::env::var("UW_SONNET_BAND").ok() { Some(p) => p, None => { eprintln!("cascade: skip"); return; } };
+        let load = |path: &str| -> HashMap<String, f64> {
+            std::fs::read_to_string(path).expect("read fixture").lines()
+                .filter(|l| !l.trim_start().starts_with('#'))
+                .filter_map(|l| { let mut it = l.split('\t');
+                    match (it.next(), it.next().and_then(|s| s.trim().parse::<f64>().ok())) {
+                        (Some(n), Some(mu)) => Some((n.trim().to_string(), mu)), _ => None } })
+                .collect()
+        };
+        let (h, s) = (load(&haiku), load(&sonnet));
+        let confident = |mu: f64| mu <= 0.3 || mu >= 0.7;
+        // Tier 1: Haiku decides the confident nodes; the uncertain band escalates to Sonnet.
+        let (mut by_haiku, mut escalated, mut resolved_by_sonnet) = (0, 0, 0);
+        for (name, &mu) in &h {
+            if confident(mu) { by_haiku += 1; }
+            else {
+                escalated += 1;
+                // Sonnet's score; "resolved" = it moved the node OUT of the ambiguous band.
+                if let Some(&sm) = s.get(name) { if confident(sm) { resolved_by_sonnet += 1; } }
+            }
+        }
+        let total = h.len();
+        eprintln!("cascade: {} nodes | Haiku decided {} ({:.0}%) | escalated to Sonnet {} | Sonnet resolved {} of those",
+            total, by_haiku, 100.0 * by_haiku as f64 / total as f64, escalated, resolved_by_sonnet);
+        // Sonnet is more decisive on the band: its scores spread WIDER than Haiku's (it un-clusters
+        // the 0.5 pile), measured by std-dev over the shared band nodes.
+        let band: Vec<&String> = h.iter().filter(|(_, &mu)| !confident(mu)).map(|(n, _)| n)
+            .filter(|n| s.contains_key(*n)).collect();
+        let std = |f: &dyn Fn(&String) -> f64| -> f64 {
+            let xs: Vec<f64> = band.iter().map(|n| f(n)).collect();
+            let m = xs.iter().sum::<f64>() / xs.len() as f64;
+            (xs.iter().map(|x| (x - m).powi(2)).sum::<f64>() / xs.len() as f64).sqrt()
+        };
+        let (sd_h, sd_s) = (std(&|n| h[n]), std(&|n| s[n]));
+        eprintln!("cascade: band score spread  Haiku σ={:.3}  Sonnet σ={:.3} (wider ⇒ more decisive)", sd_h, sd_s);
+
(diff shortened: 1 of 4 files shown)


templates/targets/rust_wam/boundary_cache.rs.mustache
```diff
@@ -3826,6 +3826,123 @@ mod tests {
...
+            std::fs::read_to_string(path).expect("read fixture").lines()
+                .filter(|l| !l.trim_start().starts_with('#'))
+                .filter_map(|l| { let mut it = l.split('\t');
+                    match (it.next(), it.next().and_then(|s| s.trim().parse::<f64>().ok())) {
+                        (Some(n), Some(mu)) => Some((n.trim().to_string(), mu)), _ => None } })
+                .collect()
+        };
+        let (h, s) = (load(&haiku), load(&sonnet));
+        let confident = |mu: f64| mu <= 0.3 || mu >= 0.7;
+        // Tier 1: Haiku decides the confident nodes; the uncertain band escalates to Sonnet.
```
s243a, line 3924: **[R7 — REQUIRED] Open vs. closed band inconsistency — 4 locations disagree.**

Four locations give conflicting conventions:

1. **This line (the `confident` predicate):** `mu <= 0.3 || mu >= 0.7` — **closed** `[0.3,0.7]` (boundary nodes μ=0.3 and μ=0.7 are decided by Haiku, not escalated)
2. **Code comment above this line (line ~146):** `μ∈(0.3,0.7)` — **open** (boundary nodes are escalated)
3. **Fixture header (`wikipedia_physics_sonnet_band.tsv`):** `"the 26 nodes Haiku was uncertain about (the band)"` — **closed** (26 includes the boundary nodes)
4. **Measurement doc (`WAM_RUST_CARET_REALDATA_MEASUREMENT_2026-06-18.md`):** `μ∈(0.3,0.7)` — **open** (19 escalated)

Effect: the closed convention (this predicate) escalates 26 nodes to Sonnet; the open convention escalates only 19. The 7 boundary nodes (μ exactly 0.3 or 0.7) are scored by Sonnet in the fixture but never reach the Sonnet lookup in the test.

**Suggested fix (recommend closed `[0.3,0.7]` to match the fixture header):**
```rust
// Change the code comment at line ~146 from μ∈(0.3,0.7) to μ∈[0.3,0.7]
// Update WAM_RUST_CARET_REALDATA_MEASUREMENT doc: "19" → "26", open → closed bracket
// The predicate here (mu <= 0.3 || mu >= 0.7) is already correct for closed convention
```
Alternatively if open is intended: change the predicate to `mu < 0.3 || mu > 0.7`.

templates/targets/rust_wam/boundary_cache.rs.mustache
```diff
@@ -3826,6 +3826,123 @@ mod tests {
...
+        let cheap = |u: u32| -> Option<f64> {
+            depth.get(&u).map(|&d| (1.0 - d as f64 / 6.0).clamp(0.0, 1.0))
+        };
+
+        // Two-threshold gate: accept on the prior if >= TAU_HI; reject if < TAU_LO; consult the LLM
+        // only in the "just-missed-but-close" band [TAU_LO, TAU_HI), and accept there iff the
+        // GEOMETRIC MEAN sqrt(prior·μ_llm) >= TAU_GEO. The geo-mean hard-vetoes if either is ~0.
+        const TAU_HI: f64 = 0.66;   // prior alone accepts (shallow ⇒ depth ≤ 2)
+        const TAU_LO: f64 = 0.33;   // prior alone rejects (deep ⇒ depth ≥ 4)
+        const TAU_GEO: f64 = 0.5;
```
s243a, line 3875: **[R8 — REQUIRED] TAU_LO comment is off by one depth level.**

The comment says: `// prior alone rejects (deep ⇒ depth ≥ 4)`

But with `cheap(u) = (1.0 - d/6.0).clamp(0,1)`:
- depth 4 → cheap = (1 - 4/6) = 0.333... which is **greater than** TAU_LO = 0.33 → still in the band, NOT rejected
- depth 5 → cheap = (1 - 5/6) = 0.167 < 0.33 → rejected ✓

So rejection starts at depth ≥ 5, not depth ≥ 4.

**Suggested fix:**
```rust
const TAU_LO: f64 = 0.33;   // prior alone rejects (deep ⇒ depth ≥ 5)
```
Or tighten the threshold: `TAU_LO = 0.34` to make depth 4 unambiguously rejected (cheap = 0.333 < 0.34).

templates/targets/rust_wam/boundary_cache.rs.mustache
```diff
@@ -3826,6 +3826,123 @@ mod tests {
...
+        for (&u, &mu) in &llm {
+            let c = match cheap(u) { Some(c) => c, None => continue };
+            if c >= TAU_HI {
+                acc_cheap += 1;                          // prior confident-accept, no LLM
+            } else if c < TAU_LO {
+                rej_cheap += 1;                          // prior confident-reject, no LLM
+            } else {
+                llm_calls += 1;                          // band: the only place the LLM is consulted
+                let geo = (c * mu).sqrt();               // geometric mean (hard-vetoes if either ~0)
+                if geo >= TAU_GEO { acc_geo += 1; } else { rej_geo += 1; }
```
s243a, line 3886: **[NB — Non-blocking] Geometric mean is the logarithmic opinion pool, not Bayesian combination.**

`geo = (c * mu).sqrt()` is the geometric mean `√(p₁·p₂)`, which is the **logarithmic opinion pool** (Clemen & Winkler 1999 "Combining Probability Distributions from Experts in Risk Analysis"). The true Bayesian combination under independence is:

```
Bayes(p₁, p₂) = p₁·p₂ / (p₁·p₂ + (1−p₁)·(1−p₂))
```

For two signals both above 0.5 (both confident-positive), the geometric mean gives a *lower* fused score than the Bayesian formula — it is more conservative (harder to accept). For example: cheap=0.6, μ_llm=0.7 → geo=0.648, Bayes=0.778. This changes the effective acceptance zone at TAU_GEO.

The geometric mean is a defensible and computationally cheap choice; document it explicitly:
```rust
// Geometric mean = logarithmic opinion pool (Clemen & Winkler 1999).
// More conservative than full Bayesian fusion; hard-vetoes if either signal is ~0.
let geo = (c * mu).sqrt();
```

templates/targets/rust_wam/boundary_cache.rs.mustache
```diff
@@ -3826,6 +3826,123 @@ mod tests {
...
+            }
+        }
+        let total = acc_cheap + rej_cheap + llm_calls;
+        eprintln!("gated: {} nodes | accept-on-prior {} reject-on-prior {} | LLM consulted {} ({:.0}% saved)",
+            total, acc_cheap, rej_cheap, llm_calls, 100.0 * (total - llm_calls) as f64 / total as f64);
```
s243a, line 3891: **[NB — Non-blocking] 58% LLM savings is eprintln'd but not asserted.**

The measured result (58% savings) is printed via `eprintln!` but not verified by any assertion. A future change that widens the escalation band — or a weaker depth prior — could silently drop the savings below any meaningful threshold while CI stays green.

**Suggested addition before the `assert!(llm_calls < total, ...)`:**
```rust
let savings_pct = 100.0 * (total - llm_calls) as f64 / total as f64;
assert!(savings_pct > 40.0,
    "expected >40% LLM call savings from the gate; got {savings_pct:.1}%");
```
The 40% floor is conservative (below the measured 58%) to allow for fixture variation while still catching regressions.

templates/targets/rust_wam/boundary_cache.rs.mustache
```diff
@@ -3826,6 +3826,123 @@ mod tests {
...
+            if confident(mu) { by_haiku += 1; }
+            else {
+                escalated += 1;
+                // Sonnet's score; "resolved" = it moved the node OUT of the ambiguous band.
+                if let Some(&sm) = s.get(name) { if confident(sm) { resolved_by_sonnet += 1; } }
+            }
+        }
+        let total = h.len();
+        eprintln!("cascade: {} nodes | Haiku decided {} ({:.0}%) | escalated to Sonnet {} | Sonnet resolved {} of those",
+            total, by_haiku, 100.0 * by_haiku as f64 / total as f64, escalated, resolved_by_sonnet);
```
s243a, line 3936: **[NB — Non-blocking] Partition assertion missing.**

The cascade logic partitions all `h` nodes into `by_haiku + escalated`. Currently there is no assertion verifying this invariant, so a counting bug (e.g. some nodes skipped due to the `s.get(name)` lookup failing) would pass silently.

**Suggested addition after the loop:**
```rust
assert_eq!(
    by_haiku + escalated, total,
    "partition invariant: all nodes must be either decided by Haiku or escalated (got {} + {} ≠ {})",
    by_haiku, escalated, total
);
```

s243a (commented):
## Council-of-Models Review: PR #3270 — tiered/gated μ (cheap prior, LLM only on the close band)

### Summary

**Opus 4.8:** The two-stage architecture is well-motivated and the code is clear. The critical consistency bug is the open vs. closed band: the `confident` predicate in the cascade test uses `mu <= 0.3 || mu >= 0.7` (closed, 26 nodes) but the fixture header and the design doc both say the escalated band is `μ∈(0.3,0.7)` (open, 19 nodes), and the doc comment in the code itself also says open (`μ∈(0.3,0.7)`). The 7 boundary nodes (μ exactly 0.3 or 0.7) are scored by Sonnet in the fixture but never escalated in the test. Pick one convention and apply it consistently across code, fixture headers, and docs. Also flagged: the TAU_LO depth comment is off-by-one.

**GPT-5.5:** Agrees on both required items. Adds that the 58% LLM savings claim is printed via `eprintln!` but never asserted — a future code change that degrades gating efficiency would pass CI silently. Also notes that the partition invariant `by_haiku + escalated == total` is never asserted in the cascade test, so a counting bug could pass undetected.

**Gemini (deep research):** The cost-gating architecture matches the "logarithmic opinion pool" literature (Clemen & Winkler 1999): geometric mean fusion `√(p₁p₂)` is the log-opinion-pool, not Bayesian combination. The Bayesian formula is `p₁p₂ / (p₁p₂ + (1-p₁)(1-p₂))`, which systematically produces higher fused values when both signals are above 0.5. The choice of geometric mean vs. Bayesian combination changes the acceptance threshold behavior and should be documented. The batch-size caveat in the measurement doc (escalated band must be large enough to amortize the fixed LLM call overhead) is excellent and worth referencing in the code comment.

1. **Open vs. closed band inconsistency** — `confident = |mu| mu <= 0.3 || mu >= 0.7` treats `[0.3,0.7]` as closed (26 escalated), but the fixture header says "26 nodes Haiku was uncertain about (the band)" while the code comment in the same function says `μ∈(0.3,0.7)` (open, 19 escalated). The Sonnet fixture contains exactly 26 entries. Pick one convention (recommend: closed `[0.3,0.7]` to match the fixture header) and apply it everywhere — the `confident` predicate, the doc comment, the measurement doc, and the fixture header.
2. **TAU_LO depth comment is off by one** — the comment reads "prior alone rejects (deep ⇒ depth ≥ 4)" but with `cheap(u) = (1.0 - d/6.0).clamp(0,1)`, depth 4 gives cheap = 1/3 ≈ 0.333 > TAU_LO = 0.33, so depth 4 is still **in the band** (not rejected). Rejection starts at depth 5 (cheap = 1/6 ≈ 0.167 < 0.33).

### Non-Blocking Observations

- **Geometric mean is not Bayesian combination** — `√(p₁p₂)` is the logarithmic opinion pool (Clemen & Winkler 1999), which systematically underestimates vs. the true Bayesian formula `p₁p₂/(p₁p₂+(1-p₁)(1-p₂))`. Document the choice; Bayesian combination would widen the acceptance zone when both signals are positive.
- **58% LLM savings should be asserted** — the `eprintln!` is informative but a regression in gating efficiency would pass CI silently. Consider `assert!(100.0 * (total - llm_calls) as f64 / total as f64 > 50.0, "expected >50% LLM savings")`.
- **Partition invariant missing in cascade** — add `assert_eq!(by_haiku + escalated, total, "partition: all nodes must be either decided or escalated")` at the end of `wikipedia_model_cascade_haiku_then_sonnet`.
- **Discrimination context header** — the Sonnet fixture already has the `# discrimination:` header; the measurement doc rightly notes this should be a first-class fixture field for future μ-fixtures.

### Recommendation

**Request changes.** The open/closed band inconsistency (R7) is the highest-priority fix — it means the test's `assert!(sd_s > sd_h)` runs on a different set (26 nodes) than the documented claim (19 nodes escalated). The TAU_LO comment (R8) should also be corrected. The non-blocking items can be addressed in a follow-on commit.

s243a:
Addressed all review items in `42e5e2b0`. Summary:

- **R7 (open/closed band)** — this was a real bug, not just a comment mismatch. The `confident` predicate `mu <= 0.3 || mu >= 0.7` implements the **open** band `(0.3,0.7)` = 19 escalated, but the comment and the 26-node Sonnet fixture assume the **closed** band `[0.3,0.7]`. I confirmed empirically: closed `[0.3,0.7]` = 26 (matches the fixture exactly), open `(0.3,0.7)` = 19. Adopted **closed** by fixing the predicate to the strict complement `mu < 0.3 || mu > 0.7`, and re-measured against the fixtures with the closed band:
  - Haiku decides **64/90 (71%)** outright; **26** escalated; Sonnet **resolves 16 of the 26**; band spread σ Haiku **0.117** → Sonnet **0.184**.
  - Updated the measurement doc numbers (71/90, 26, 16) and the open→closed wording.
- **R8 (TAU_LO depth)** — corrected the comment to **depth ≥ 5** (`c = 1 − d/6`; depth 4 → 0.333 > τ_lo = 0.33, so it stays in the band; depth 5 → 0.167 < τ_lo rejects).
- **NB (geometric mean)** — documented `√(prior·μ)` as the **logarithmic opinion pool** (Clemen & Winkler 1999), explicitly *not* a Bayesian posterior.
- **NB (savings assert)** — added `assert!(saved > 0.40)` to the gated test (measured 58%).
- **NB (partition assert)** — added `assert_eq!(by_haiku + escalated, total)` to the cascade test.

87/87 lib tests pass; cascade re-measured against the fixtures (closed band) and gated test re-run against `data/benchmark/10k` (58% saved).

---

PR #3271, opened 2026-06-18
WAM-Rust: μ-weighted IC (partial admission) + fuzzy S-curve admission

s243a:
## What & why

The first **membership-weighted read-out** — `μ`-weighted IC ("partial admission") — and it
specifically *corrects the associative leak's effect on IC*.

Raw `IC(t) = −log₂(|desc(t)|/N)` counts every descendant as `1`. On a non-taxonomic graph the leak
bloats a domain node's cone with out-of-domain descendants, so its `|desc|` looks huge and the raw
IC mistakes it for **generality** (big cone → low IC). **Partial admission** admits each node with
weight `μ ∈ [0,1]` instead of 0/1, so low-`μ` (leaked) descendants barely count — the effective cone
shrinks to its in-domain size and the node's IC is *raised* to its true specificity.

## What's added

- **`information_content_weighted(desc_mu_mass, total_mu)`** = `−log₂(Σ_{d∈desc(t)} μ_d / Σ_all μ)`;
  `μ ≡ 1` recovers `information_content`.
- **`descendant_mu_mass(parents, μ)`** — the *distinct* μ-weighted descendant mass (per-node BFS,
  `O(V·(V+E))`, exact for small graphs; the scalable **μ-weighted MinHash** sketch is future work).
- **`fuzzy_admission(μ, c, k)`** — an **S-curve** (logistic) transfer `w(μ)=1/(1+e^{−k(μ−c)})`, so
  admission needn't be linear in `μ`. Default `c=0.55, k≈4.39` fits the anchors `w(0.8)=0.75` (small
  drop) and `w(0.3)=0.25` (significant drop), crossover `w(0.55)=0.5` — nearly-full/nearly-zero with
  a tunable knee.

## Test

`mu_weighted_ic_partial_admission_and_s_curve` — on a leaky graph (physics node with a cone of
mostly leaked `μ=0.1` descendants):
```
μ-IC: node1 raw 0.49 → linear-μ 1.28 → S-curve 1.19   (clean sibling node2 raw 2.81)
```
Raw IC falsely rates the leaked node as *general* (gap 2.32 to its clean sibling); partial admission
raises it to ~1.2–1.3, closing the false gap. `μ≡1` recovers the distinct-count IC; the S-curve fits
the anchors and is monotone.

Documented in §5d, including the note that this is exactly where membership *matters* — a **global**
cone read-out, unlike the membership-robust per-pair caret. Full suite: **86 passed**.

---

### DIFF
--- a/templates/targets/rust_wam/boundary_cache.rs.mustache
+++ b/templates/targets/rust_wam/boundary_cache.rs.mustache
@@ -1521,6 +1521,68 @@ pub fn information_content(dsig: &[u64], n_total: usize, k: usize) -> f64 {
     -p.log2()
 }
 
+/// **μ-weighted ("partial admission") information content.** Standard `IC` counts every descendant
+/// as `1`; here each node is admitted with weight `μ ∈ [0,1]`, so `IC_μ(t) = −log₂( m_μ(t) / M_μ )`
+/// with `m_μ(t) = Σ_{d∈desc(t)} μ_d` the admitted mass under `t` and `M_μ = Σ_all μ` the total. Low-μ
+/// (out-of-domain / leaked) descendants barely count, so a *domain* node whose cone is bloated by
+/// the associative leak — which the raw `IC` mistakes for *generality* (a big cone → low IC) — has
+/// its effective cone shrunk back to its in-domain size and its IC *raised* to its true specificity.
+/// `μ ≡ 1` recovers `information_content`. (`m_μ` is the **distinct** weighted mass; see
+/// `descendant_mu_mass`.)
+pub fn information_content_weighted(desc_mu_mass: f64, total_mu: f64) -> f64 {
+    if total_mu <= 0.0 || desc_mu_mass <= 0.0 {
+        return 0.0;
+    }
+    -((desc_mu_mass / total_mu).min(1.0)).log2()
+}
+
+/// Exact **distinct** μ-weighted descendant mass per node: `m_μ(t) = Σ_{d∈desc(t)} μ_d` over the
+/// distinct descendant set (`t` included), so diamonds count once. A downward BFS per node
+/// (`O(V·(V+E))`) — exact, for small graphs / validation. At scale this wants a **μ-weighted
+/// MinHash** sketch (the weighted analogue of `descendant_minhash`), which is future work; this
+/// grounds `information_content_weighted`. With `μ ≡ 1` it equals the distinct descendant count.
+pub fn descendant_mu_mass(
+    parents: &std::collections::HashMap<u32, Vec<u32>>,
+    mu: &std::collections::HashMap<u32, f64>,
+) -> std::collections::HashMap<u32, f64> {
+    use std::collections::{HashMap, HashSet, VecDeque};
+    let mut children: HashMap<u32, Vec<u32>> = HashMap::new();
+    let mut nodes: HashSet<u32> = HashSet::new();
+    for (&c, ps) in parents {
+        nodes.insert(c);
+        for &p in ps { nodes.insert(p); children.entry(p).or_default().push(c); }
+    }
+    let muv = |u: u32| mu.get(&u).copied().unwrap_or(0.0);
+    let mut out: HashMap<u32, f64> = HashMap::new();
+    for &t in &nodes {
+        let mut seen: HashSet<u32> = HashSet::new();
+        seen.insert(t);
+        let mut q: VecDeque<u32> = VecDeque::new();
+        q.push_back(t);
+        let mut mass = muv(t);
+        while let Some(x) = q.pop_front() {
+            if let Some(cs) = children.get(&x) {
+                for &c in cs { if seen.insert(c) { mass += muv(c); q.push_back(c); } }
+            }
+        }
+        out.insert(t, mass);
+    }
+    out
+}
+
+/// **Fuzzy admission curve**: an S-shaped (logistic) transfer from the raw membership `μ` to an
+/// admission weight in `[0,1]`, `w(μ) = 1/(1 + e^{−k(μ−c)})`. Shapes *how* `μ` translates to
+/// admission — nearly-full weight well above the boundary, nearly-zero well below, a smooth knee
+/// through the middle — instead of admitting linearly by raw `μ`. The default `c = 0.55, k ≈ 4.39`
+/// fits the anchors `w(0.8) = 0.75` (small drop) and `w(0.3) = 0.25` (significant drop), with the
+/// crossover `w(0.55) = 0.5`; raise `k` for a sharper accept/reject knee, move `c` for the
+/// boundary. (Zadeh's piecewise-quadratic S-function is the classic alternative shape.) Feed
+/// `w(μ_d)` as the per-node admission weight to `descendant_mu_mass` for a fuzzy partial-admission
+/// `IC`.
+pub fn fuzzy_admission(mu: f64, center: f64, steepness: f64) -> f64 {
+    1.0 / (1.0 + (-steepness * (mu - center)).exp())
+}
+
 /// Reflexive ancestor set of `node` (`node` itself plus every strict ancestor), by up-BFS.
 /// **O(|ancestors(node)| + edges above it)** per call — it allocates a fresh `HashSet`/`VecDeque`.
 /// For repeated queries on the same node, cache the returned set rather than recomputing.
@@ -4097,6 +4159,54 @@ mod tests {
         // Jaccard of two saturated sketches stays in [0,1] (no panic / out-of-range).
         let j = sketch_jaccard(&sig[&20], &sig[&19], K);
         assert!((0.0..=1.0).contains(&j));
+    }
+
+    // μ-weighted ("partial admission") IC + the fuzzy S-curve admission. A leaky graph: physics
+    // node 1 (μ=1) has a cone of mostly LEAKED low-μ descendants (3,4,5,6 at μ=0.1); physics node 2
+    // (μ=1) is a clean leaf. Raw IC mistakes node 1's big leaked cone for generality (low IC);
+    // partial admission discounts the leak and RAISES its IC to its true specificity.
+    #[test]
+    fn mu_weighted_ic_partial_admission_and_s_curve() {
+        let mut g: HashMap<u32, Vec<u32>> = HashMap::new();
+        g.insert(1, vec![0]); g.insert(2, vec![0]);
+        g.insert(3, vec![1]); g.insert(4, vec![3]); g.insert(5, vec![3]); g.insert(6, vec![3]);
+        let close = |a: f64, b: f64| (a - b).abs() < 1e-2;
+
+        // RAW IC (μ ≡ 1): node 1's leaked cone {1,3,4,5,6}=5 makes it look general; node 2={2}=1 specific.
+        let n = 7.0; // nodes 0..=6
+        let mass1: HashMap<u32, f64> = (0u32..=6).map(|i| (i, 1.0)).collect();
+        let raw = descendant_mu_mass(&g, &mass1);
+        let (ic_raw_1, ic_raw_2) = (information_content_weighted(raw[&1], n),
+                                    information_content_weighted(raw[&2], n));
+        assert!(close(raw[&1], 5.0) && close(raw[&2], 1.0));          // μ≡1 == distinct counts
+        assert!(ic_raw_1 < ic_raw_2 - 1.5, "raw IC: leaked node 1 looks falsely GENERAL");
+
+        // LINEAR μ-weighting: 3,4,5,6 at μ=0.1; 0,1,2 at μ=1. The leak is discounted.
+        let mut mu: HashMap<u32, f64> = (0u32..=2).map(|i| (i, 1.0)).collect();
+        for i in 3u32..=6 { mu.insert(i, 0.1); }
+        let total_mu: f64 = mu.values().sum();
+        let lin = descendant_mu_mass(&g, &mu);
+        let (ic_mu_1, ic_mu_2) = (information_content_weighted(lin[&1], total_mu),
+                                  information_content_weighted(lin[&2], total_mu));
+        // Node 1's IC is RAISED (its cone is mostly leaked); the false generality is corrected.
+        assert!(ic_mu_1 > ic_raw_1 + 0.5, "partial admission raises the leaked node's IC: {} -> {}", ic_raw_1, ic_mu_1);
+        assert!(ic_mu_2 - ic_mu_1 < ic_raw_2 - ic_raw_1, "the leak's false discrimination shrinks");
+
+        // FUZZY S-CURVE admission: the anchors w(0.8)=0.75, w(0.3)=0.25, crossover w(0.55)=0.5.
+        let (c, k) = (0.55, 4.39);
+        assert!(close(fuzzy_admission(0.8, c, k), 0.75));
+        assert!(close(fuzzy_admission(0.3, c, k), 0.25));
+        assert!(close(fuzzy_admission(0.55, c, k), 0.5));
+        assert!(fuzzy_admission(0.1, c, k) < fuzzy_admission(0.3, c, k));   // monotone
+        assert!(fuzzy_admission(1.0, c, k) > fuzzy_admission(0.8, c, k));
+        // Shape μ through the S-curve, then admit: still corrects the leak, with a tunable knee.
+        let mu_s: HashMap<u32, f64> = mu.iter().map(|(&u, &m)| (u, fuzzy_admission(m, c, k))).collect();
+        let tot_s: f64 = mu_s.values().sum();
+        let s = descendant_mu_mass(&g, &mu_s);
+        let ic_s_1 = information_content_weighted(s[&1], tot_s);
+        assert!(ic_s_1 > ic_raw_1 + 0.5, "S-curve partial admission also corrects the leak");
+        eprintln!("μ-IC: node1 raw {:.2} → linear-μ {:.2} → S-curve {:.2}  (node2 raw {:.2})",
+            ic_raw_1, ic_mu_1, ic_s_1, ic_raw_2);
     }
 
     // IC-based semantic similarity (Resnik/Lin) via the descendant sketch -- the calibrated,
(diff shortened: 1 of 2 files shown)


templates/targets/rust_wam/boundary_cache.rs.mustache
```diff
@@ -1521,6 +1521,68 @@ pub fn information_content(dsig: &[u64], n_total: usize, k: usize) -> f64 {
...
+/// **μ-weighted ("partial admission") information content.** Standard `IC` counts every descendant
+/// as `1`; here each node is admitted with weight `μ ∈ [0,1]`, so `IC_μ(t) = −log₂( m_μ(t) / M_μ )`
+/// with `m_μ(t) = Σ_{d∈desc(t)} μ_d` the admitted mass under `t` and `M_μ = Σ_all μ` the total. Low-μ
+/// (out-of-domain / leaked) descendants barely count, so a *domain* node whose cone is bloated by
+/// the associative leak — which the raw `IC` mistakes for *generality* (a big cone → low IC) — has
+/// its effective cone shrunk back to its in-domain size and its IC *raised* to its true specificity.
+/// `μ ≡ 1` recovers `information_content`. (`m_μ` is the **distinct** weighted mass; see
+/// `descendant_mu_mass`.)
+pub fn information_content_weighted(desc_mu_mass: f64, total_mu: f64) -> f64 {
+    if total_mu <= 0.0 || desc_mu_mass <= 0.0 {
```
s243a: **[R2 — REQUIRED] `total_mu` must be Σμ over ALL nodes, not node count.**

The doc says `M_μ = Σ_all μ`, but there is no validation that the caller actually passes this. If a caller passes `n_total` (the integer node count, as they might by analogy with `information_content`'s `n_total: usize`), they get silently wrong IC whenever μ≢1. Opus and GPT-5.5 both flagged this.

**Suggested fix:** Either rename the parameter to `total_mu_mass` to signal it is not a count, or add:
```rust
// PRECONDITION: total_mu == mu_map.values().sum() — the total μ-mass over all nodes in the graph.
// Do NOT pass node count; pass Σμ. See `descendant_mu_mass` for the companion function.
debug_assert!(total_mu > 0.0, "total_mu must be positive Σμ, not zero or a node count");
```

templates/targets/rust_wam/boundary_cache.rs.mustache
```diff
@@ -1521,6 +1521,68 @@ pub fn information_content(dsig: &[u64], n_total: usize, k: usize) -> f64 {
...
+        return 0.0;
+    }
```
s243a, line 1553: **[R1 — REQUIRED] `desc_mu_mass <= 0.0 → IC = 0.0` is semantically inverted.**

In information theory, IC = −log₂ P; P → 0 means IC → +∞, not 0. Returning `0.0` signals "most general node" (e.g. the root with IC≈0). Returning it for a zero-mass node (one that is outside the support of μ) makes the caller believe the node is merely general, not absent from the domain.

Deep research (Seco 2004, Resnik 1995) confirms: the correct semantic for a zero-probability node is IC = ∞.

**Suggested fix (choose one):**
```rust
// Option A — return INFINITY (caller must handle)
if total_mu <= 0.0 || desc_mu_mass <= 0.0 {
    return f64::INFINITY;
}

// Option B — return Option<f64> and map None = outside-support
pub fn information_content_weighted(desc_mu_mass: f64, total_mu: f64) -> Option<f64> {
    if total_mu <= 0.0 || desc_mu_mass <= 0.0 { return None; }
    Some(-((desc_mu_mass / total_mu).min(1.0)).log2())
}
```
If keeping `0.0` as a sentinel for pragmatic reasons, add an explicit doc comment: `// Sentinel: returns 0.0 for outside-support nodes (μ-mass = 0); callers MUST NOT interpret this as true IC.`

s243a (commented):
## Council-of-Models Review: PR #3271 — μ-weighted IC + fuzzy S-curve admission

### Summary

**Opus 4.8:** The μ-weighted IC formulation is sound — it cleanly generalises standard IC and the proof that μ≡1 recovers `information_content` is immediate. Three correctness issues stand out: the `desc_mu_mass≤0 → IC=0` sentinel is semantically inverted (IC should diverge, not vanish); `total_mu` has no validation that it equals Σμ over all nodes (not node count); and absent μ entries silently default to 0 via `unwrap_or(0.0)` — a stale μ-map produces silent zero-weighting with no diagnostic. The `fuzzy_admission` steepness=0 degenerate case (constant 0.5) and steepness<0 curve-inversion are also undocumented.

**GPT-5.5:** Agrees on all four correctness issues and adds that `Option<f64>` or `f64::INFINITY` is the right return for zero-probability nodes (consistent with information-theoretic convention). Also flags that the μ∈[0,1] precondition is unenforceable from the type system alone — a `debug_assert!(mu >= 0.0 && mu <= 1.0)` would catch μ>1 at dev time before the `.min(1.0)` clamp silently drives IC to 0. Recommends a cross-PR regression test verifying `information_content_weighted(descendant_mu_mass(μ≡1), n)` matches `information_content(descendant_minhash(k→∞), n, k)`.

**Gemini (deep research):** Confirms Seco (2004) and resnik (1995) convention: IC = −log P; P→0 means IC→+∞, never 0. The logistic curve is a valid S-curve choice; Zadeh's piecewise-quadratic S-function reaches exactly 0 and 1 at its endpoints (the logistic only approaches them asymptotically), which may matter when downstream code checks `w == 0.0` or `w == 1.0`. The k≈4.39 constant is `4·ln(3)` — worth documenting as an exact closed form. μ-weighted IC is a novel contribution not in published literature; this should be noted in the doc comment.

1. **`desc_mu_mass≤0 → IC=0.0` is semantically wrong** — zero μ-mass means the node is outside the support; IC should be `f64::INFINITY` (or `Option<f64>::None`), not `0.0` (which means "most general node"). At minimum add a doc comment explaining the sentinel.
2. **`total_mu` must be Σμ over all nodes, not node count** — no validation; passing the integer node count when μ≢1 gives systematically wrong IC values.
3. **μ∈[0,1] precondition undocumented/unenforced** — add `debug_assert!` or explicit doc guard; μ>1 silently produces IC=0 via the `.min(1.0)` clamp.
4. **`unwrap_or(0.0)` treats absent entries as μ=0** — a μ-map for a different graph version causes silent zero-weighting with no warning.
5. **`fuzzy_admission` degenerate cases undocumented** — `steepness=0` returns constant 0.5; `steepness<0` inverts the curve; neither is guarded or documented.
6. **No cross-PR regression test** — add a test verifying `information_content_weighted(descendant_mu_mass(μ≡1, graph), n)` equals `information_content(descendant_minhash(k=large, graph), n, k)` on the same small graph.

### Non-Blocking Observations

- **Novel contribution** — μ-weighted IC does not appear in published IC-similarity literature (Resnik, Lin, Seco); worth a sentence in the doc comment flagging it as an extension.
- **Zadeh S-function** — the piecewise-quadratic form has exact 0/1 endpoints; the logistic only approaches them. If downstream code relies on `w(μ) == 0.0` for hard exclusion, the Zadeh form is safer.
- **k≈4.39 exact form** — this is `4·ln(3)`; document it as such for reproducibility.

### Recommendation

**Request changes.** The IC=0 sentinel for zero-mass nodes is the clearest semantic inversion and must be resolved. The remaining required items (total_mu validation, μ precondition, absent-entry behaviour, steepness degenerate cases, regression test) should also be addressed before merge.

templates/targets/rust_wam/boundary_cache.rs.mustache
```diff
@@ -1521,6 +1521,68 @@ pub fn information_content(dsig: &[u64], n_total: usize, k: usize) -> f64 {
...
+    -((desc_mu_mass / total_mu).min(1.0)).log2()
+}
```
s243a, line 1555: **[R3 — REQUIRED] μ ∈ [0,1] precondition undocumented; violation silently produces IC = 0.**

If any descendant has μ > 1, `desc_mu_mass` can exceed `total_mu`, making the ratio > 1; `.min(1.0)` clamps it to 1.0 and `-1.0_f64.log2() = 0.0`. The caller gets IC = 0 ("most general") with no diagnostic — the same symptom as R1 but from a different cause.

**Suggested fix:**
```rust
debug_assert!(
    (0.0..=total_mu).contains(&desc_mu_mass),
    "desc_mu_mass {desc_mu_mass} exceeds total_mu {total_mu}; check μ ∈ [0,1] for all nodes"
);
```
And add to the doc comment: `// PRECONDITION: all μ values must be in [0,1]; μ > 1 causes silent IC = 0.`

templates/targets/rust_wam/boundary_cache.rs.mustache
```diff
@@ -1521,6 +1521,68 @@ pub fn information_content(dsig: &[u64], n_total: usize, k: usize) -> f64 {
...
+) -> std::collections::HashMap<u32, f64> {
+    use std::collections::{HashMap, HashSet, VecDeque};
+    let mut children: HashMap<u32, Vec<u32>> = HashMap::new();
+    let mut nodes: HashSet<u32> = HashSet::new();
+    for (&c, ps) in parents {
+        nodes.insert(c);
+        for &p in ps { nodes.insert(p); children.entry(p).or_default().push(c); }
+    }
+    let muv = |u: u32| mu.get(&u).copied().unwrap_or(0.0);
+    let mut out: HashMap<u32, f64> = HashMap::new();
```
s243a, line 1583: **[R4 — REQUIRED] `unwrap_or(0.0)` treats absent map entries as μ = 0 — undocumented, stale-map silent failure.**

If `mu` was computed for a different snapshot of the graph (e.g. after adding nodes), nodes present in `parents` but absent from `mu` silently get μ = 0 and zero weight. This is the most likely production failure mode.

**Suggested fix:** Either document the contract explicitly:
```rust
// PRECONDITION: `mu` must cover every node reachable via `parents`; absent entries default to μ=0.
// A μ-map for a different graph version produces silent zero-weighting with no error.
```
Or add a dev-mode check:
```rust
#[cfg(debug_assertions)]
if mu.get(&u).is_none() && nodes.contains(&u) {
    eprintln!("descendant_mu_mass: node {u} has no μ entry — defaulting to 0.0");
}
```

templates/targets/rust_wam/boundary_cache.rs.mustache
```diff
@@ -1521,6 +1521,68 @@ pub fn information_content(dsig: &[u64], n_total: usize, k: usize) -> f64 {
...
+}
+
+/// **Fuzzy admission curve**: an S-shaped (logistic) transfer from the raw membership `μ` to an
+/// admission weight in `[0,1]`, `w(μ) = 1/(1 + e^{−k(μ−c)})`. Shapes *how* `μ` translates to
+/// admission — nearly-full weight well above the boundary, nearly-zero well below, a smooth knee
+/// through the middle — instead of admitting linearly by raw `μ`. The default `c = 0.55, k ≈ 4.39`
+/// fits the anchors `w(0.8) = 0.75` (small drop) and `w(0.3) = 0.25` (significant drop), with the
+/// crossover `w(0.55) = 0.5`; raise `k` for a sharper accept/reject knee, move `c` for the
+/// boundary. (Zadeh's piecewise-quadratic S-function is the classic alternative shape.) Feed
+/// `w(μ_d)` as the per-node admission weight to `descendant_mu_mass` for a fuzzy partial-admission
```
s243a, line 1607: **[NB — Non-blocking] Logistic vs Zadeh piecewise-quadratic S-function.**

The doc comment already mentions Zadeh's S-function as "the classic alternative." Deep research confirms a key practical difference: Zadeh's piecewise-quadratic form has **exact 0 and 1 endpoints** at μ=α and μ=β, while the logistic only approaches 0/1 asymptotically.

This matters if any downstream consumer checks `fuzzy_admission(mu, c, k) == 0.0` for hard exclusion or `== 1.0` for full inclusion. With the logistic, those checks always fail.

Also: the default `k ≈ 4.39` is exactly `4·ln(3)` (derived from the anchor equations `w(0.8)=0.75, w(0.3)=0.25`). Documenting the exact form aids reproducibility:
```rust
/// Default steepness: k = 4·ln(3) ≈ 4.394, derived from w(0.8)=0.75 and w(0.3)=0.25.
```

templates/targets/rust_wam/boundary_cache.rs.mustache
```diff
@@ -1521,6 +1521,68 @@ pub fn information_content(dsig: &[u64], n_total: usize, k: usize) -> f64 {
...
+/// `IC`.
+pub fn fuzzy_admission(mu: f64, center: f64, steepness: f64) -> f64 {
+    1.0 / (1.0 + (-steepness * (mu - center)).exp())
```
s243a, line 1620: **[R5 — REQUIRED] `steepness = 0` and `steepness < 0` are degenerate/inverted cases — undocumented.**

- `steepness = 0`: `exp(0) = 1`, so `fuzzy_admission = 0.5` for every μ — the curve collapses to a constant. Any caller unaware of this will get a flat distribution regardless of membership.
- `steepness < 0`: the logistic inverts — μ=1 → near 0, μ=0 → near 1. This is a monotone-decreasing admission curve, which is semantically opposite to "admit high-μ nodes more".

**Suggested fix:**
```rust
debug_assert!(steepness > 0.0, "steepness must be positive; 0 collapses to constant 0.5, negative inverts the curve");
```
Or document in the function header: `// PRECONDITION: steepness > 0; steepness=0 returns constant 0.5, steepness<0 inverts (anti-admission).`

templates/targets/rust_wam/boundary_cache.rs.mustache
```diff
@@ -1521,6 +1521,68 @@ pub fn information_content(dsig: &[u64], n_total: usize, k: usize) -> f64 {
     -p.log2()
 }
 
+/// **μ-weighted ("partial admission") information content.** Standard `IC` counts every descendant
+/// as `1`; here each node is admitted with weight `μ ∈ [0,1]`, so `IC_μ(t) = −log₂( m_μ(t) / M_μ )`
```
s243a, line 1525: **[NB — Non-blocking] μ-weighted IC is a novel contribution — worth flagging.**

This formulation (weighting the descendant mass by fuzzy membership μ ∈ [0,1] instead of counting 0/1) does not appear in published IC-similarity literature (Resnik 1995, Lin 1998, Seco 2004, Hadj Taieb et al. 2014). It is a genuine extension motivated by the associative-leak problem on non-taxonomic graphs.

Suggested addition to the doc comment:
```
/// **Novel extension:** Standard published IC formulations use a 0/1 descendant indicator;
/// this μ-weighted form is an unpublished extension specific to this codebase, motivated
/// by the associative-leak problem on mixed-domain DAGs.
```

templates/targets/rust_wam/boundary_cache.rs.mustache
```diff
@@ -4099,6 +4161,54 @@ mod tests {
...
+        assert!(fuzzy_admission(0.1, c, k) < fuzzy_admission(0.3, c, k));   // monotone
+        assert!(fuzzy_admission(1.0, c, k) > fuzzy_admission(0.8, c, k));
+        // Shape μ through the S-curve, then admit: still corrects the leak, with a tunable knee.
+        let mu_s: HashMap<u32, f64> = mu.iter().map(|(&u, &m)| (u, fuzzy_admission(m, c, k))).collect();
+        let tot_s: f64 = mu_s.values().sum();
+        let s = descendant_mu_mass(&g, &mu_s);
+        let ic_s_1 = information_content_weighted(s[&1], tot_s);
+        assert!(ic_s_1 > ic_raw_1 + 0.5, "S-curve partial admission also corrects the leak");
+        eprintln!("μ-IC: node1 raw {:.2} → linear-μ {:.2} → S-curve {:.2}  (node2 raw {:.2})",
+            ic_raw_1, ic_mu_1, ic_s_1, ic_raw_2);
```
s243a, line 4246: **[R6 — REQUIRED] Missing cross-PR regression test.**

There is no test verifying that `information_content_weighted` with μ≡1 produces the same result as the existing `information_content` with a large k MinHash sketch on the same graph.

This is the critical invariant that ties PR #3271 back to the established descendant-sketch baseline (PR #3259). Without it, a future refactor could silently break the equivalence.

**Suggested test addition:**
```rust
// Regression: μ≡1 weighted IC == sketch IC on a small graph
// (Use the same balanced tree from the ic_similarity_resnik_lin_via_minhash test)
let n = 7;
let mu_uniform: HashMap<u32, f64> = (0u32..7).map(|i| (i, 1.0)).collect();
let total_mu = n as f64;
let mass = descendant_mu_mass(&g, &mu_uniform);
for node in 0u32..7 {
    let ic_w = information_content_weighted(mass[&node], total_mu);
    // compare with information_content(descendant_minhash(k=512), n, 512) within tolerance
    assert!((ic_w - ic_sketch[node as usize]).abs() < 0.1,
        "μ≡1 IC mismatch at node {node}: weighted={ic_w} sketch={}", ic_sketch[node as usize]);
}
```

s243a:
Addressed all review items in `6f18f9ca`. Summary:

- **R1 (zero-mass sentinel)** — adopted **Option A**: an empty cone (`desc_mu_mass = 0`, `total_mu > 0`) now returns `f64::INFINITY`, the `−log₂ 0` maximal-surprise limit, instead of `0.0` (which read as root-like/general — the exact inversion). `total_mu ≤ 0` is treated as a precondition violation (`debug_assert`) with an inert `0.0` release fallback. Both limits documented as *opposite* in the doc comment.
- **R2 (`total_mu` precondition)** — `debug_assert!(total_mu > 0.0)` plus `debug_assert!(desc_mu_mass <= total_mu + 1e-9)`; the doc states `total_mu` must be `Σ_all μ`, not a node count.
- **R3 (μ ∈ [0,1])** — `debug_assert!(mu.values().all(|w| (0.0..=1.0).contains(&w)))` in `descendant_mu_mass`, where the weights are actually supplied (the µ>1 → ratio>1 → clamp → IC=0 symptom is caught at the source).
- **R4 (absent-entry contract)** — documented `unwrap_or(0.0)`: an absent node is "out of domain" (μ=0) by default, so callers only list the in-domain nodes.
- **R5 (`fuzzy_admission` steepness)** — `debug_assert!(steepness > 0.0)`; doc spells out the degenerate cases (`0 ⇒` constant 0.5, `<0` inverts).
- **R6 (cross-PR regression)** — new test `mu_weighted_ic_reduces_to_sketch_ic_at_mu_one` cross-checks `information_content_weighted` against the independent sketch-based `information_content` (descendant MinHash, k=64) node-for-node at μ≡1, to <1e-9.
- **NB (k = 4·ln 3, Zadeh exact endpoints, novel contribution)** — documented the exact default steepness `k = 4·ln 3 ≈ 4.3944`, the logistic's asymptotic (not exact) endpoints vs Zadeh's piecewise-quadratic, and a novel-contribution flag in the design doc §5d.

87/87 lib tests pass (was 86; +1 for R6).

---

PR #3273, opened 2026-06-18
WAM-Rust: μ-weighted KMV sketch — scalable partial-admission IC + leak-robust global fan-in

s243a:
Closes the documented scaling caveat on μ-weighted IC (#3271). `descendant_mu_mass` is exact but `O(V·(V+E))`; this adds the bottom-`k` sketch that builds **every** node's weighted descendant mass in one reverse-topo pass `O((V+E)·k)`. It also opens the global hub problem (the reason this increment was chosen).

## Core — Cohen–Kaplan bottom-`k` subset-sum over the descendant cone
- **`descendant_minhash_weighted`** — bottom-`k` of `(hash, μ)` pairs, deduped by hash so diamonds count once (a set, like `descendant_minhash`). The hashes are drawn independently of `μ`, so the bottom-`k` is a *uniform* sample of the cone and the carried weights are an unbiased sample of its weight distribution.
- **`sketch_mu_mass`** — `m̂_μ = D̂ · μ̄_sample` (distinct cardinality × sample-mean weight). Exact while unsaturated; reduces to `sketch_card` at `μ ≡ 1`.
- **`information_content_weighted_sketch`** — drop-in scalable `IC_μ`, clamping the estimate at `total_mu` (the mass analogue of `information_content`'s `.min(1.0)` ratio clamp).

## Global hooks — why this serves hub selection
- **`sketch_mu_overlap`** — μ-weighted Broder intersection mass = **leak-robust fan-in**. Cones sharing only low-μ (out-of-domain) descendants score ≈ 0, so hub selection is no longer fooled by the associative leak. *Measured:* core-shared `3.00` vs leak-shared `0.15`.
- **Depth-stability** — the raw cone count *explodes* toward the root as the leak accumulates (more deep nodes than shallow), but the weighted mass stays bounded because deep nodes are out-of-domain. On the spine test the raw cone grows `201 → 645` (+444) while the μ-mass grows only `5 → 18` (+13), ~3% of the raw growth. **The weighted count converges as you descend where the raw count diverges** — the depth-robust global signal the hub problem needs.

## Tests (6 new, 93/93 lib pass)
`mu_weighted_sketch_matches_exact_mass_and_reduces_unweighted`, `…_approximates_at_scale` (k≪cone, rel err 2.6% at the root, exact on unsaturated leaves), `mu_weighted_overlap_is_leak_robust`, `mu_weighted_count_is_depth_stable`. Prolog template guard passes.

References already in §10: Cohen & Kaplan 2007 (bottom-k sketches / subset sums), Broder 1997 (MinHash intersection), Bar-Yossef et al. 2002 (KMV distinct count).

---

### DIFF
--- a/templates/targets/rust_wam/boundary_cache.rs.mustache
+++ b/templates/targets/rust_wam/boundary_cache.rs.mustache
@@ -1606,12 +1606,20 @@ pub fn descendant_mu_mass(
 /// drawn independently of `μ`, so the bottom-`k` is a *uniform* sample of the distinct cone and the
 /// weights on it are an unbiased sample of the cone's weight distribution — that is what makes the
 /// read-out work. `None` on a cyclic graph (no reverse-topo order); SCC-condense first.
+///
+/// **Naming:** this is a *carry-weight KMV* / *weighted bottom-`k`* sketch — uniform hashes with the
+/// weight carried as side data — **not** Weighted MinHash (Ioffe 2010), which encodes the weight
+/// *into* the hash so collision probability equals the weighted Jaccard `J_w`. This sketch cannot
+/// estimate `J_w`; it estimates weighted *mass* (`sketch_mu_mass`) and shared mass
+/// (`sketch_mu_overlap`). `k ≥ 2` is required (the `(k−1)/ĥ_k` cardinality read is undefined at
+/// `k=1` and panics at `k=0`).
 pub fn descendant_minhash_weighted(
     parents: &std::collections::HashMap<u32, Vec<u32>>,
     mu: &std::collections::HashMap<u32, f64>,
     k: usize,
 ) -> Option<std::collections::HashMap<u32, Vec<(u64, f64)>>> {
     use std::collections::{HashMap, HashSet, VecDeque};
+    debug_assert!(k >= 2, "k must be ≥ 2: the (k−1)/ĥ_k estimator is undefined at k=1 and panics at k=0");
     debug_assert!(
         mu.values().all(|&w| (0.0..=1.0).contains(&w)),
         "membership weights μ must lie in [0,1]"
@@ -1653,13 +1661,20 @@ pub fn descendant_minhash_weighted(
 
 /// Bottom-`k` **subset-sum** read-out: estimate `m_μ = Σ_{d∈cone} μ_d` from a μ-weighted sketch.
 /// While unsaturated (`< k` pairs kept) the cone is fully captured, so this is the **exact** sum.
-/// Once saturated it is the Cohen–Kaplan bottom-`k` estimator `m̂_μ = D̂ · μ̄_sample` — the distinct
-/// cardinality `D̂ = (k−1)/ĥ_k` (the KMV estimate, identical to `sketch_card`) times the mean weight
-/// over the `k` sampled pairs. Because the bottom-`k` is a uniform sample of the cone independent of
-/// `μ`, `μ̄_sample` is unbiased for the cone's mean weight, so the product estimates the total mass.
-/// At `μ ≡ 1` it reduces exactly to `sketch_card` (the unweighted distinct count). Error is the KMV
-/// `∝ 1/√k` plus the sampling variance of `μ̄`, both tunable by `k`.
-pub fn sketch_mu_mass(s: &[(u64, f64)], k: usize) -> f64 {
+/// Once saturated it is a bottom-`k` *plug-in* subset-sum estimator `m̂_μ = D̂ · μ̄_sample` — the
+/// distinct cardinality `D̂ = (k−1)/ĥ_k` (the KMV estimate, identical to `sketch_card`) times the
+/// mean weight over the `k` sampled pairs. Because the bottom-`k` is a uniform sample of the cone
+/// independent of `μ`, `μ̄_sample` is unbiased for the cone's mean weight; the product is unbiased to
+/// `O(1/k)` (a ratio-estimator term). This is *in the spirit of* Cohen–Kaplan bottom-`k` sketches
+/// (PODC/SIGMETRICS 2007), but it is **not** their VLDB 2008 Horvitz–Thompson form (per-item adjusted
+/// weights, zero covariance), which is formally tighter. At `μ ≡ 1` it reduces exactly to
+/// `sketch_card`. Error is the KMV `∝ 1/√k` plus the `μ̄` sampling variance, both tunable by `k`.
+///
+/// `n_universe` is the total node count — the cardinality `D̂` cannot exceed it (a distinct cone is a
+/// subset of the graph), so `D̂` is capped at `n_universe` to defend against a pathologically tiny
+/// `ĥ_k` blowing `(k−1)/ĥ_k` up to `~10¹⁹`. `k ≥ 2` (`k=1` makes `(k−1)/ĥ_k ≡ 0`).
+pub fn sketch_mu_mass(s: &[(u64, f64)], k: usize, n_universe: usize) -> f64 {
+    debug_assert!(k >= 2, "k must be ≥ 2: k=1 makes (k−1)/ĥ_k ≡ 0 (mass 0 ⇒ IC 0), k=0 panics");
     if s.is_empty() {
         return 0.0;
     }
@@ -1667,7 +1682,8 @@ pub fn sketch_mu_mass(s: &[(u64, f64)], k: usize) -> f64 {
         return s.iter().map(|&(_, w)| w).sum(); // unsaturated ⇒ exact
     }
     let hk = s[k - 1].0 as f64 / (u64::MAX as f64 + 1.0);
-    let card = if hk <= 0.0 { s.len() as f64 } else { (k as f64 - 1.0) / hk };
+    let raw_card = if hk <= 0.0 { s.len() as f64 } else { (k as f64 - 1.0) / hk };
+    let card = raw_card.min(n_universe as f64); // cone ⊆ graph ⇒ D̂ ≤ N (caps tiny-ĥ_k overflow)
     let mean_w = s.iter().map(|&(_, w)| w).sum::<f64>() / s.len() as f64;
     card * mean_w
 }
@@ -1675,11 +1691,20 @@ pub fn sketch_mu_mass(s: &[(u64, f64)], k: usize) -> f64 {
 /// Scalable **μ-weighted IC**: `information_content_weighted` fed by the sketch mass instead of the
 /// exact `descendant_mu_mass`. The drop-in large-graph form of partial-admission IC — same
 /// `−log₂(m_μ/M_μ)`, but `m_μ` is the `O((V+E)·k)` sketch estimate. `total_mu` is still the exact
-/// universe mass `Σ_all μ` (cheap: one pass over the μ-map). The sketch estimate can overshoot
-/// `total_mu` by KMV sampling noise near the root (cone ≈ universe), so it is clamped at `total_mu`
-/// before the log — the mass analogue of `information_content`'s `.min(1.0)` ratio clamp.
-pub fn information_content_weighted_sketch(s: &[(u64, f64)], total_mu: f64, k: usize) -> f64 {
-    information_content_weighted(sketch_mu_mass(s, k).min(total_mu), total_mu)
+/// universe mass `Σ_all μ` (cheap: one pass over the μ-map); `n_universe` the node count (cone cap).
+/// The sketch estimate can overshoot `total_mu` by KMV sampling noise near the root (cone ≈ universe),
+/// so it is clamped at `total_mu` before the log — the mass analogue of `information_content`'s
+/// `.min(1.0)` ratio clamp.
+///
+/// **DEPENDENCY:** the zero-mass semantics are inherited from `information_content_weighted`, which
+/// returns `+∞` for an empty cone (`m_μ = 0`, the maximal-surprise limit). Near zero mass the *sketch*
+/// path is high-variance: the saturated sample can miss the cone's few in-domain nodes and read
+/// `μ̄_sample = 0 ⇒ m̂_μ = 0 ⇒ IC = +∞` even when the exact mass is small-positive (or vice versa). So
+/// the exact and sketch IC agree at *exactly* zero (both `+∞`) and in the bulk, but can diverge in a
+/// thin near-zero band — use the exact `descendant_mu_mass` path when the mass is near zero and the
+/// distinction matters.
+pub fn information_content_weighted_sketch(s: &[(u64, f64)], total_mu: f64, k: usize, n_universe: usize) -> f64 {
+    information_content_weighted(sketch_mu_mass(s, k, n_universe).min(total_mu), total_mu)
 }
 
 /// **μ-weighted cone-overlap mass** — the global-hub read-out. Estimates `Σ_{d∈A∩B} μ_d`, the
@@ -1690,8 +1715,22 @@ pub fn information_content_weighted_sketch(s: &[(u64, f64)], total_mu: f64, k: u
 /// signal: a *funnel* hub is one many cones reach, but a μ-weighted funnel discounts the associative
 /// leak — two cones that overlap only on low-μ (out-of-domain) descendants score near zero, so hub
 /// selection stops being fooled by the leak. `0` if either sketch is empty.
-pub fn sketch_mu_overlap(a: &[(u64, f64)], b: &[(u64, f64)], k: usize) -> f64 {
+///
+/// **Estimator:** algebraically `D̂_∪ · mean_{d∈union sample}( μ_d · 𝟙[d∈A∩B] )` — the *same* single
+/// product estimator as `sketch_mu_mass`, with the intersection indicator folded into a 0/1-gated μ
+/// (not a Jaccard estimate times a separate conditional mean). Both factors are read off the *same*
+/// union sample, so their variance compounds, and `mean_shared_w` divides by `shared.len()`, which is
+/// small exactly when the leak-discount bites (few shared high-μ elements). **Use as a ranking signal,
+/// not an absolute mass** — it is high-variance in the low-overlap regime.
+///
+/// **Precondition:** `a` and `b` must have been built with the *same* `k` (the Broder union-bottom-`k`
+/// argument requires it); mixing `k`s silently biases `D̂_∪`. `n_universe` caps `D̂_∪` (a union of
+/// cones is `⊆` the graph), defending against a tiny `ĥ_k` overflow exactly as in `sketch_mu_mass`.
+pub fn sketch_mu_overlap(a: &[(u64, f64)], b: &[(u64, f64)], k: usize, n_universe: usize) -> f64 {
     use std::collections::HashMap;
+    debug_assert!(k >= 2, "k must be ≥ 2");
+    debug_assert!(a.len() <= k && b.len() <= k,
+        "both sketches must be built with the same k; mixing k biases the union-card estimate");
     if a.is_empty() || b.is_empty() {
         return 0.0;
     }
@@ -1717,7 +1756,7 @@ pub fn sketch_mu_overlap(a: &[(u64, f64)], b: &[(u64, f64)], k: usize) -> f64 {
     let union_card = if union.len() < k || hk <= 0.0 {
         union.len() as f64
     } else {
-        (k as f64 - 1.0) / hk
+        ((k as f64 - 1.0) / hk).min(n_universe as f64) // cap tiny-ĥ_k overflow (union ⊆ graph)
     };
     // jaccard-by-mass: (shared count / sample size) gives |A∩B|/|A∪B|; multiply by union mass-per-sample.
     let mean_shared_w = shared.iter().sum::<f64>() / shared.len() as f64;
@@ -4544,12 +4583,13 @@ mod tests {
         let total_mu: f64 = mu.values().sum();
 
         let exact = descendant_mu_mass(&g, &mu);
+        let n = exact.len(); // universe node count (cone cap)
         let wsig = descendant_minhash_weighted(&g, &mu, K).unwrap();
         for (&node, &m) in &exact {
-            let est = sketch_mu_mass(&wsig[&node], K);
+            let est = sketch_mu_mass(&wsig[&node], K, n);
             assert!((est - m).abs() < 1e-9, "node {node}: sketch mass {est} != exact {m}");
             // Scalable IC must match the exact-mass IC.
-            let ic_sketch = information_content_weighted_sketch(&wsig[&node], total_mu, K);
+            let ic_sketch = information_content_weighted_sketch(&wsig[&node], total_mu, K, n);
             let ic_exact = information_content_weighted(m, total_mu);
             assert!((ic_sketch - ic_exact).abs() < 1e-9, "node {node}: IC mismatch");
         }
@@ -4559,9 +4599,38 @@ mod tests {
         let w1 = descendant_minhash_weighted(&g, &mu_one, K).unwrap();
         let d = descendant_minhash(&g, K).unwrap();
         for &node in mu.keys() {
-            assert!((sketch_mu_mass(&w1[&node], K) - sketch_card(&d[&node], K)).abs() < 1e-9,
+            assert!((sketch_mu_mass(&w1[&node], K, n) - sketch_card(&d[&node], K)).abs() < 1e-9,
                 "node {node}: μ≡1 sketch mass != unweighted distinct count");
         }
+    }
+
+    // R2/NB6 + DEPENDENCY: zero-mass IC and the None-on-cycle path. A node whose whole cone is
+    // out-of-domain (μ≡0) has m_μ=0 ⇒ IC=+∞ in BOTH the exact and the sketch path (the sketch's
+    // sampled mean weight is also 0), so they agree at exactly zero — pinning the inherited #3271
+    // +∞ semantics. And a cyclic graph yields None (no reverse-topo order).
+    #[test]
+    fn mu_weighted_sketch_zero_mass_is_infinite_and_cycle_is_none() {
+        const K: usize = 8;
+        // Star: 1,2,3 -> 0. μ≡0 over the cone of node 0 (everything out of domain) but a positive
+        // universe mass elsewhere so total_mu>0 (the empty-cone, not empty-universe, case).
+        let mut g: HashMap<u32, Vec<u32>> = HashMap::new();
+        g.insert(1, vec![0]); g.insert(2, vec![0]); g.insert(3, vec![0]);
+        let mut mu: HashMap<u32, f64> = [(0u32, 0.0), (1, 0.0), (2, 0.0), (3, 0.0)].into_iter().collect();
+        mu.insert(99, 1.0); // an out-of-cone node carrying the universe mass
+        let total_mu: f64 = mu.values().sum(); // = 1.0 > 0
+        let w = descendant_minhash_weighted(&g, &mu, K).unwrap();
+        let n = mu.len();
+        let exact = descendant_mu_mass(&g, &mu);
+        assert_eq!(exact[&0], 0.0, "cone of 0 is fully out of domain");
+        assert!(information_content_weighted(exact[&0], total_mu).is_infinite(), "exact zero-mass ⇒ +∞");
+        assert!(information_content_weighted_sketch(&w[&0], total_mu, K, n).is_infinite(),
+            "sketch zero-mass ⇒ +∞ (agrees with exact at exactly zero)");
+
+        // None on a cycle (0 <-> 1).
+        let mut cyc: HashMap<u32, Vec<u32>> = HashMap::new();
+        cyc.insert(0, vec![1]); cyc.insert(1, vec![0]);
+        let cmu: HashMap<u32, f64> = [(0u32, 1.0), (1, 1.0)].into_iter().collect();
+        assert!(descendant_minhash_weighted(&cyc, &cmu, K).is_none(), "cycle ⇒ None");
     }
 
     // At scale the sketch is approximate (k << cone): build a wide tree, check the weighted-mass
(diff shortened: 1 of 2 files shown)


templates/targets/rust_wam/boundary_cache.rs.mustache
```diff
@@ -1597,6 +1597,133 @@ pub fn descendant_mu_mass(
...
+        let mut cand: Vec<(u64, f64)> = vec![(mix64(b as u64), muv(b))];
+        if let Some(cs) = children.get(&b) {
+            for &c in cs {
+                if let Some(s) = sig.get(&c) { cand.extend_from_slice(s); }
+            }
+        }
+        cand.sort_unstable_by_key(|&(h, _)| h);
+        cand.dedup_by_key(|&mut (h, _)| h); // dedup by element hash ⇒ distinct cone (diamonds once)
+        cand.truncate(k);
+        sig.insert(b, cand);
```
s243a, line 1650: **[R1 — REQUIRED] `k = 0` panics here; `k = 1` gives IC = 0 for all saturated sketches.**

`cand.truncate(k)` with k=0 produces empty sketches (safe here), but `sketch_mu_mass` then indexes `s[k-1]` in the saturated branch without any guard. A caller passing `k=0` with a non-empty slice panics. More critically, `k=1` is mathematically degenerate: `(k-1)/h_k = 0/h_k = 0`, so every saturated sketch returns `card = 0` → `mass = 0` → IC = 0. IC = 0 means "most general node" — the exact inversion of the correct result.

The KMV estimator `(k-1)/h_k` is unbiased only for `k ≥ 2`; finite variance requires `k ≥ 3`.

**Suggested fix:**
```rust
// At entry to descendant_minhash_weighted:
debug_assert!(k >= 2, "k must be ≥ 2 for the (k-1)/h_k estimator to be defined; k=0 panics, k=1 always returns mass=0");
```
And in `sketch_mu_mass`:
```rust
pub fn sketch_mu_mass(s: &[(u64, f64)], k: usize) -> f64 {
    assert!(k >= 2, "k must be ≥ 2; k=0 panics on saturated sketches, k=1 always returns 0");
    ...
```

s243a (commented):
## Council-of-Models Review: PR #3273 — μ-weighted KMV sketch (scalable partial-admission IC + leak-robust global fan-in)

**Reviewers:** Claude Opus 4.8 · GPT-5.5 · Gemini (deep research: web + academic sources)

### Summary

**Opus 4.8:** The core construction is correct and well-grounded. The reverse-topo pass, diamond dedup (collision-free because `mix64` is a bijection over u32), `(k−1)/h_k` unbiased estimator, and the `m̂_μ = D̂·μ̄_sample` product form are all sound. Two blocking items: `information_content_weighted_sketch` silently inherits the #3271 `mass≤0 → IC=0` semantic inversion, and both `sketch_mu_mass` and `sketch_mu_overlap` leave near-zero positive `h_k` unguarded, allowing `(k−1)/h_k` to overflow to ~10¹⁹ in the mass/overlap paths (the `.min(total_mu)` clamp only masks it for IC, not for the raw public values).

**GPT-5.5:** Flags the k=0/k=1 degenerate cases — k=0 panics on `s[k-1]` for external callers, k=1 returns 0 for every saturated sketch because `(k−1)/h_k = 0`. Also identifies the lack of a typed sketch struct: `sketch_mu_overlap(a, b, k)` accepts sketches built with different k values without any assertion, silently biasing both the union-card and the shared-mass estimates. Agrees on the estimator documentation gap — the doc claims "unbiased" but the product of two correlated sample statistics is only asymptotically unbiased (O(1/k) ratio-estimator bias); the independence argument (hash ranks are independent of the k-th order statistic under iid hashing) needs stating explicitly.

**Gemini (deep research, Cohen & Kaplan PODC 2007, SIGMETRICS 2007, VLDB 2008; Bar-Yossef et al. 2002; Ioffe 2010):** Confirms `(k−1)/h_k` is the unbiased Cohen–Kaplan form (Bar-Yossef uses `k/m_{k+1}`, the two are equivalent under different bookkeeping conventions). Confirms the μ-weighted IC is novel — Harispe et al. 2015 (the comprehensive survey) does not cover it; the Gene Ontology annotation-weighted IC is the closest prior art. Key finding: the `D̂·μ̄` product form is **not** the Cohen–Kaplan VLDB 2008 estimator — that paper uses per-item Horvitz–Thompson adjusted weights (Rank Conditioning / Subset Conditioning), which achieves zero covariance between item estimates. `D̂·μ̄` is a ratio/plug-in estimator that is approximately unbiased but not the cited paper's headline contribution. The citation should be updated. Also confirms: the PR's approach is **not** Weighted MinHash (Ioffe 2010) — weighted MinHash encodes weight *into* the hash function; this PR uses uniform hash + weight as side data, which cannot estimate weighted Jaccard.

**R1 [Opus + GPT] — `k = 0` panics; `k = 1` silently returns IC = 0 for all saturated sketches.**

`sketch_mu_mass` indexes `s[k-1]` in the saturated branch without guarding `k == 0`. `descendant_minhash_weighted` with `k=0` produces empty sketches (truncate(0)), but the public `sketch_mu_mass`/`sketch_mu_overlap` functions can receive hand-built or cross-module slices. Additionally, `k=1` is mathematically degenerate: `(k-1)/h_k = 0` for any positive `h_k`, so every saturated sketch returns mass=0 and IC=0. The unbiased KMV estimator `(k-1)/h_k` is only valid for `k ≥ 2`; finite variance requires `k ≥ 3`. Require `k ≥ 2` with an assertion.

**R2 [Opus] — `information_content_weighted_sketch` inherits the #3271 `mass ≤ 0 → IC = 0` semantic inversion.**

This PR delegates directly to `information_content_weighted(sketch_mu_mass(...).min(total_mu), total_mu)`. The #3271 review flagged that `desc_mu_mass ≤ 0 → return 0.0` is semantically inverted (IC should be +∞ for zero-probability nodes, not 0). The exact-regime test `ic_sketch == ic_exact` *locks* this wrong behavior: if #3271 is later fixed to return `f64::INFINITY`, this PR follows it correctly for exact sketches, but a saturated sketch with all near-zero weights can produce a small positive `card × mean_w` (not exactly 0) — yielding a large finite IC where the exact path returns +∞. The two paths would silently diverge. Required: track this as a hard dependency on the #3271 fix, and add a saturated-regime test covering the zero/near-zero-mass case.

**R3 [Opus + GPT] — Near-zero positive `h_k` is unguarded: overflow to ~10¹⁹ in the mass and overlap paths.**

Both `sketch_mu_mass` (the `hk` guard) and `sketch_mu_overlap` guard only `hk <= 0.0`. A small-but-positive hash value (even a single `mix64` output in the range [1, 2^32]) passes the guard and makes `(k-1)/hk` astronomically large. The `.min(total_mu)` clamp in `information_content_weighted_sketch` hides this for IC, but `sketch_mu_mass` and `sketch_mu_overlap` are public functions — hub scoring would receive exploded values. Add a practical lower bound: `hk.max(f64::EPSILON)` is not sufficient (still ~4.5e15 cardinality); a cap like `card.min(known_universe_bound)` is needed, or document that both functions require `k` to be large enough that the empirical `h_k` is well away from zero. At minimum, add the same overflow guard that `sketch_card` implicitly benefits from via context.

### Non-Blocking Observations

**NB1 [All three] — Citation: `D̂·μ̄` is a ratio/plug-in estimator, not the Cohen–Kaplan VLDB 2008 form.**

Cohen & Kaplan VLDB 2008 "Tighter Estimation using Bottom-k Sketches" introduces Rank Conditioning (RC) and Subset Conditioning (SC) estimators derived from the Horvitz–Thompson framework — per-item adjusted weights, zero covariance, optimal variance. The code's `D̂·μ̄` product (distinct-count estimate × sample-mean weight) is a naïve ratio estimator that happens to be approximately unbiased (the hash ranks are independent of `h_k` under iid hashing, so the independence argument holds, but it is not the HT estimator). Deep research also confirms: the correct paper for the subpopulation weight estimator is VLDB 2008, not just PODC 2007. Suggest rewording the doc: `"a bottom-k plug-in subset-sum estimator (distinct-cardinality × sampled-mean weight), in the spirit of Cohen–Kaplan PODC/SIGMETRICS 2007 and VLDB 2008 (§RC estimator);"`

**NB2 [Opus] — Product estimator bias: document the ratio-estimator O(1/k) term.**

The doc (diff L106–107) says `"Error is the KMV ∝ 1/√k plus the sampling variance of μ̄, both tunable by k."` This is asymptotically correct but omits: (a) the two CVs add in quadrature, so the combined relative error is `√(1/(k-2) + S²_μ/(k·μ̄²))`, not merely their sum; (b) the product `D̂·μ̄` carries an additional O(1/k) ratio-estimator bias (from Jensen's inequality on the product of correlated statistics). The 20% test tolerance (k=256, ~1800-node cone) is appropriate given ~6.3% cardinality CV plus the μ̄ CV, but deserves a quantitative comment so it is not later "tightened" into flakiness.

**NB3 [GPT] — `sketch_mu_overlap` accepts incompatible sketches silently.**

The Broder union-bottom-k intersection argument is valid only when both sketches were built with the same `k` and the same hash function. If `a` was built with `k=100` and `b` with `k=200`, the union sketch is not the bottom-`k` of the true union. No assertion or type check enforces this. The simplest fix is a `debug_assert!(a.len() <= k && b.len() <= k)` inside `sketch_mu_overlap`, or introducing a typed wrapper (`WeightedKmvSketch { k, entries }`) that makes the invariant part of the type.

**NB4 [All] — μ-weighted IC is novel — update the "Novel-contribution flag" in the semirings doc.**

The semirings doc already has a "Novel-contribution flag" section. Deep research confirms: neither Harispe et al. 2015 (the comprehensive survey), Seco 2004, nor any other major IC-similarity paper covers `IC_μ(t) = −log₂(Σμ_d / Σ_all μ)` as a named measure. The carry-weight KMV approach for this quantity is also novel. The closest prior art is annotation-weighted IC in Gene Ontology bioinformatics (uses per-gene annotation counts as weights), but it is domain-specific. Worth adding: "The μ-weighted IC and its sketch approximation via carry-weight KMV do not appear in the ontology IC literature (Harispe et al. 2015 is the comprehensive survey); the closest prior art is annotation-weighted IC in Gene Ontology [PMC3775452] and Resnik 1995 (corpus IC as a special case where μ_d = frequency)."

**NB5 [Opus] — `sketch_mu_overlap` is not tested in the saturated (approximate) regime.**

`mu_weighted_overlap_is_leak_robust` runs at K=64 over tiny graphs — the exact regime where both cones fit entirely in the sketch. The variance-heavy saturated path (large overlapping cones, K << |A∪B|) is unvalidated. Add a scale test mirroring `mu_weighted_sketch_approximates_at_scale`: two large cones with known exact overlap, assert rel-err < 20–30%.

**NB6 [GPT] — Missing test: `descendant_minhash_weighted` returns `None` on a cycle.**

The `None` return on cycle detection (diff L97: `if sig.len() == nodes.len()`) is documented but not tested. The existing `descendant_minhash` presumably also lacks this test; add one to both, or at minimum to the new weighted function.

**NB7 [Opus] — `aset` in `sketch_mu_overlap` should be `HashSet<u64>`, not `HashMap<u64,f64>`.**

Both `aset` and `bset` are used only for membership testing (`.contains_key(h)`) — the f64 values are never read (the weight comes from `union`, not from the individual sets). `HashSet<u64>` would be more idiomatic and avoids storing unused f64 values. Cosmetic, no behavior change.

**NB8 [Deep research] — "Associative leak" has no standard name; suggest a note.**

No established term exists in the ontology IC literature for the inflation-by-out-of-domain-descendants problem. The closest recognized concept is "cross-ontology noise" or "annotation-weighted IC" (Gene Ontology). Consider adding a one-sentence note in the doc: *"The inflation of raw IC by out-of-domain descendants — here called the 'associative leak' — corresponds to the motivation for annotation-weighted IC in Gene Ontology literature [PMC3775452], where raw descendant counts are replaced by annotation frequencies."*

### Agreements Across Reviewers

- `(k−1)/h_k` is the correct unbiased KMV convention — **confirmed by all three**
- `m̂_μ = D̂·μ̄` is approximately unbiased (not the HT estimator, but the independence argument holds) — **confirmed by Opus + Gemini**
- Diamond dedup via sort+dedup_by_key is correct and collision-free for u32 ids — **confirmed by all three**
- Topo-sort `pending` logic is correct — **confirmed by all three**
- μ-weighted IC is novel (not in prior literature) — **confirmed by all three**
- Inherited IC=0 bug from #3271 is blocking — **confirmed by Opus + GPT**

### Recommendation

**Request changes.** The construction is mathematically sound and the implementation faithful to the existing sketch infrastructure. The three blocking items (k degenerate cases, inherited IC=0 bug coupling, unguarded h_k overflow) should be addressed before merge. The non-blocking items — particularly the citation update (NB1), estimator bias documentation (NB2), and the missing scale test for `sketch_mu_overlap` (NB5) — are recommended for this PR rather than a follow-on.

templates/targets/rust_wam/boundary_cache.rs.mustache
```diff
@@ -1597,6 +1597,133 @@ pub fn descendant_mu_mass(
...
+/// `∝ 1/√k` plus the sampling variance of `μ̄`, both tunable by `k`.
+pub fn sketch_mu_mass(s: &[(u64, f64)], k: usize) -> f64 {
+    if s.is_empty() {
+        return 0.0;
+    }
+    if s.len() < k {
+        return s.iter().map(|&(_, w)| w).sum(); // unsaturated ⇒ exact
+    }
+    let hk = s[k - 1].0 as f64 / (u64::MAX as f64 + 1.0);
+    let card = if hk <= 0.0 { s.len() as f64 } else { (k as f64 - 1.0) / hk };
```
s243a: **[R1 + R3 — REQUIRED] `k = 1` returns 0 here; near-zero `h_k` overflows to ~10¹⁹.**

Two issues on this line:

**Issue A (k=1 degenerate):** When `k=1`, the saturated branch is entered for any non-empty sketch, and `(k as f64 - 1.0) / hk = 0.0 / hk = 0.0`, returning card=0, then mass=0, then IC=0 ("most general"). This is wrong — a node with a single saturated sketch element has unknown cardinality, not zero.

**Issue B (h_k overflow):** `hk <= 0.0` is the only guard, but a small-but-positive h_k (e.g. `mix64` output = 1 → hk ≈ 5.4e−20) gives `card ≈ k × 1.8e19`. The `.min(total_mu)` clamp in `information_content_weighted_sketch` masks this for IC, but `sketch_mu_mass` is a public function — hub scoring via `sketch_mu_overlap` would receive exploded values.

**Suggested fix:**
```rust
let hk = s[k - 1].0 as f64 / (u64::MAX as f64 + 1.0);
if hk <= 0.0 {
    return s.iter().map(|&(_, w)| w).sum();
}
let card = (k as f64 - 1.0) / hk;
// Cap at a universe bound to prevent h_k ≈ 0 overflow:
// If card >> D_universe the estimate is nonsensical regardless.
// Alternatively document that callers must clamp at their universe size.
let card = card.min(1e12_f64); // adjust to your graph size upper bound
```

templates/targets/rust_wam/boundary_cache.rs.mustache
```diff
@@ -1597,6 +1597,133 @@ pub fn descendant_mu_mass(
...
+
+/// Scalable **μ-weighted IC**: `information_content_weighted` fed by the sketch mass instead of the
+/// exact `descendant_mu_mass`. The drop-in large-graph form of partial-admission IC — same
+/// `−log₂(m_μ/M_μ)`, but `m_μ` is the `O((V+E)·k)` sketch estimate. `total_mu` is still the exact
+/// universe mass `Σ_all μ` (cheap: one pass over the μ-map). The sketch estimate can overshoot
+/// `total_mu` by KMV sampling noise near the root (cone ≈ universe), so it is clamped at `total_mu`
+/// before the log — the mass analogue of `information_content`'s `.min(1.0)` ratio clamp.
+pub fn information_content_weighted_sketch(s: &[(u64, f64)], total_mu: f64, k: usize) -> f64 {
+    information_content_weighted(sketch_mu_mass(s, k).min(total_mu), total_mu)
+}
```
s243a, line 1766: **[R2 — REQUIRED] `information_content_weighted_sketch` inherits the #3271 `mass ≤ 0 → IC = 0` semantic inversion.**

This delegates directly to `information_content_weighted(sketch_mu_mass(...).min(total_mu), total_mu)`. The #3271 review flagged that `desc_mu_mass ≤ 0 → return 0.0` is semantically wrong — a zero-mass node is *maximally specific* (IC = +∞), not maximally general (IC = 0).

The exact-regime test locks `ic_sketch == ic_exact` to 1e-9. If #3271 is fixed to return `f64::INFINITY`, this PR follows correctly for exact sketches, but the saturated path can independently produce a small positive mass (card × near-zero mean_w ≠ 0), yielding a large *finite* IC where the exact path returns +∞. The two paths would silently diverge.

**Required:** Add a comment documenting the hard dependency on the #3271 fix, and add a saturated-regime test covering the near-zero-mass case:
```rust
/// **DEPENDENCY:** This function's behavior for zero-mass nodes (returns 0.0 instead of +∞)
/// is inherited from `information_content_weighted` — see PR #3271 review R1. When that
/// function is fixed, this path follows; but the saturated sketch path may independently
/// diverge (small positive card×mean_w ≠ 0 even when the exact mass = 0). A saturated
/// near-zero-mass regression test is required.
pub fn information_content_weighted_sketch(s: &[(u64, f64)], total_mu: f64, k: usize) -> f64 {
```

templates/targets/rust_wam/boundary_cache.rs.mustache
```diff
@@ -1597,6 +1597,133 @@ pub fn descendant_mu_mass(
...
+                *e -= 1;
+                if *e == 0 { q.push_back(p); }
+            }
+        }
+    }
+    if sig.len() == nodes.len() { Some(sig) } else { None } // short => a cycle
+}
+
+/// Bottom-`k` **subset-sum** read-out: estimate `m_μ = Σ_{d∈cone} μ_d` from a μ-weighted sketch.
+/// While unsaturated (`< k` pairs kept) the cone is fully captured, so this is the **exact** sum.
```
s243a, line 1720: **[NB1 — Non-blocking] Citation update: `D̂·μ̄` is a ratio/plug-in estimator, not the Cohen–Kaplan VLDB 2008 HT form. Weighted MinHash distinction needed.**

Deep research (Cohen & Kaplan PODC 2007 + SIGMETRICS 2007 + VLDB 2008; Ioffe ICDM 2010) clarifies two points:

**A — Citation precision:** Cohen–Kaplan VLDB 2008 "Tighter Estimation using Bottom-k Sketches" introduces Rank Conditioning (RC) and Subset Conditioning (SC) Horvitz–Thompson estimators — per-item adjusted weights, zero inter-item covariance. The code's `D̂·μ̄` (cardinality × sample-mean weight) is a *ratio/plug-in* estimator: it is approximately unbiased (hash ranks are independent of `h_k` under iid hashing, so the product's expectation factors), but it is **not** the HT estimator the paper is known for. The bias is O(1/k) from the ratio-estimator term.

Suggested doc rewording:
```
/// the Cohen–Kaplan bottom-k subset-sum estimator  →
/// a bottom-k plug-in subset-sum estimator (distinct-cardinality × sampled-mean weight);
/// in the spirit of Cohen–Kaplan PODC+SIGMETRICS 2007 / VLDB 2008 §RC estimator.
/// The product D̂·μ̄ is unbiased under iid hashing (hash ranks independent of h_k),
/// but carries an O(1/k) ratio-estimator bias term; the Cohen–Kaplan VLDB 2008 HT form
/// (per-item adjusted weights) achieves zero covariance and is formally tighter.
```

**B — Not Weighted MinHash:** If the doc or review comments describe this as "weighted MinHash," clarify: Weighted MinHash (Ioffe ICDM 2010) encodes the weight *into the hash function* so collision probability equals weighted Jaccard. This PR uses uniform hash + weight as side data — it cannot estimate weighted Jaccard `J_w`, and `sketch_mu_overlap` does not estimate it either. The term is "carry-weight KMV" or "weighted bottom-k" to distinguish.

templates/targets/rust_wam/boundary_cache.rs.mustache
```diff
@@ -1597,6 +1597,133 @@ pub fn descendant_mu_mass(
...
+/// **μ-weighted cone-overlap mass** — the global-hub read-out. Estimates `Σ_{d∈A∩B} μ_d`, the
+/// *admitted* mass two cones share, from their μ-weighted sketches: take the bottom-`k` of the
+/// union, then (Broder) the elements present in **both** sketches are exactly the shared ones whose
+/// hash is small enough to survive in each, so their sampled fraction times the union cardinality,
+/// scaled by the shared mean weight, estimates the shared mass. This is the membership-aware fan-in
+/// signal: a *funnel* hub is one many cones reach, but a μ-weighted funnel discounts the associative
+/// leak — two cones that overlap only on low-μ (out-of-domain) descendants score near zero, so hub
+/// selection stops being fooled by the leak. `0` if either sketch is empty.
+pub fn sketch_mu_overlap(a: &[(u64, f64)], b: &[(u64, f64)], k: usize) -> f64 {
+    use std::collections::HashMap;
```
s243a, line 1788: **[NB3 — Non-blocking] `sketch_mu_overlap` silently accepts incompatible sketches (different `k`).**

The Broder union-bottom-k intersection argument requires both `a` and `b` to have been built with the same `k`. If `a` was built with `k=100` and `b` with `k=200`, but the caller passes `k=150`, the merged `union` truncates to 150 — but `a` has already discarded elements 101-150, so the "union bottom-150" is missing elements that should have been in it from `a`. Both `union_card` and the shared-mass estimate are silently biased.

GPT-5.5 flagged this as a required change; Opus rated it non-blocking given that both sketches in practice come from the same `descendant_minhash_weighted` call. The choice is yours — at minimum:

```rust
pub fn sketch_mu_overlap(a: &[(u64, f64)], b: &[(u64, f64)], k: usize) -> f64 {
    debug_assert!(
        a.len() <= k && b.len() <= k,
        "both sketches must have been built with the same k; mixing k values biases the union-card estimate"
    );
    ...
```
Longer term, a typed `WeightedKmvSketch { k, entries: Vec<(u64, f64)> }` makes this invariant part of the type.

templates/targets/rust_wam/boundary_cache.rs.mustache
```diff
@@ -1597,6 +1597,133 @@ pub fn descendant_mu_mass(
...
+    let shared: Vec<f64> = union
+        .iter()
+        .filter(|(h, _)| aset.contains_key(h) && bset.contains_key(h))
+        .map(|&(_, w)| w)
+        .collect();
+    if shared.is_empty() {
+        return 0.0;
+    }
+    let hk = union[union.len() - 1].0 as f64 / (u64::MAX as f64 + 1.0);
+    let union_card = if union.len() < k || hk <= 0.0 {
```
s243a, line 1815: **[R3 — REQUIRED] Same near-zero `h_k` overflow risk as in `sketch_mu_mass`.**

The guard `hk <= 0.0` (line ~162-163) is insufficient. A small-but-positive h_k in the union sketch gives `union_card = (k-1)/hk` on the order of 10¹⁹, then `union_card × (shared.len()/union.len()) × mean_shared_w` — an astronomically large overlap score. Unlike `information_content_weighted_sketch` (which clamps at `total_mu`), `sketch_mu_overlap` returns this value directly to hub-scoring callers.

Fix with the same cap applied in `sketch_mu_mass` (see R1/R3 comment on that function), or add:
```rust
let union_card = if union.len() < k || hk <= 0.0 {
    union.len() as f64
} else {
    let c = (k as f64 - 1.0) / hk;
    c.min(1e12) // cap against near-zero hash overflow
};
```

templates/targets/rust_wam/boundary_cache.rs.mustache
```diff
@@ -1597,6 +1597,133 @@ pub fn descendant_mu_mass(
...
+        union.len() as f64
+    } else {
+        (k as f64 - 1.0) / hk
+    };
+    // jaccard-by-mass: (shared count / sample size) gives |A∩B|/|A∪B|; multiply by union mass-per-sample.
+    let mean_shared_w = shared.iter().sum::<f64>() / shared.len() as f64;
+    union_card * (shared.len() as f64 / union.len() as f64) * mean_shared_w
```
s243a, line 1822: **[NB4 — Non-blocking] `sketch_mu_overlap` formula: document the three-factor composition and its variance.**

The formula `union_card × (shared.len()/union.len()) × mean_shared_w` is algebraically equivalent to `union_card × avg_{d∈union_sample}(μ_d · 1[d∈A∩B])` — the same product estimator as `sketch_mu_mass` but with the intersection indicator as a 0/1 μ. This is actually cleaner than describing it as a Jaccard estimate times a conditional mean (which implies a second dependent random variable).

The variance concern: `mean_shared_w` divides by `shared.len()`, which is small precisely when the leak-discount is working (few shared high-μ elements). The estimator is statistically fragile in the low-overlap regime — high variance from a small shared subsample. Consider adding:
```rust
// The estimator is algebraically: union_card × E_sample[μ_d · 1{d∈A∩B}]
// — a single product estimator over the union sample, not a Jaccard estimate times
// a conditional mean. Variance compounds: both factors depend on the same union sample.
// In the low-overlap regime (few shared elements) the estimate has high variance;
// use only as a ranking signal, not as an absolute mass estimate.
```

templates/targets/rust_wam/boundary_cache.rs.mustache
```diff
@@ -4399,6 +4526,164 @@ mod tests {
         }
     }
 
+    // μ-weighted KMV sketch: the scalable companion to the exact descendant_mu_mass. With k >> cone
+    // the sketch is exact, so it must match descendant_mu_mass node-for-node, AND at μ≡1 reduce to
+    // the unweighted distinct count (sketch_card). Diamond included (8 -> [3,4]) so dedup is tested.
+    #[test]
+    fn mu_weighted_sketch_matches_exact_mass_and_reduces_unweighted() {
```
s243a, line 4631: **[NB6 — Non-blocking] Missing test: `descendant_minhash_weighted` returns `None` on a cyclic graph.**

The `None` path (diff L97: `if sig.len() == nodes.len() { Some(sig) } else { None }`) is documented but untested. Add a cycle test:
```rust
#[test]
fn mu_weighted_minhash_returns_none_on_cycle() {
    let mut g: HashMap<u32, Vec<u32>> = HashMap::new();
    g.insert(0, vec![1]); g.insert(1, vec![0]); // 0 <-> 1: cycle
    let mu: HashMap<u32, f64> = [(0, 1.0), (1, 1.0)].into_iter().collect();
    assert!(descendant_minhash_weighted(&g, &mu, 4).is_none(),
        "cycle must return None");
}
```

templates/targets/rust_wam/boundary_cache.rs.mustache
```diff
@@ -4399,6 +4526,164 @@ mod tests {
...
+            for _ in 0..2 { let l = next; next += 1; g.insert(l, vec![c]); mu.insert(l, 0.3); }
+        }
+        let total_mu: f64 = mu.values().sum();
+        let exact = descendant_mu_mass(&g, &mu);
+        let wsig = descendant_minhash_weighted(&g, &mu, K).unwrap();
+
+        let est_root = sketch_mu_mass(&wsig[&0], K);
+        let rel_err = (est_root - exact[&0]).abs() / exact[&0];
+        assert!(rel_err < 0.20, "root mass est {est_root} vs exact {} (rel err {:.3})", exact[&0], rel_err);
+
```
s243a, line 4756: **[NB2 — Non-blocking] 20% tolerance is reasonable but deserves a quantitative justification.**

For k=256, pure KMV cardinality CV ≤ 1/√(k−2) ≈ 6.3%. The mass estimator `D̂·μ̄` combines two terms: the cardinality CV and the μ̄ sampling CV add in quadrature (plus an O(1/k) ratio-bias term). For graded μ∈{0.3, 0.5} over a ~1800-node cone, the μ̄ CV is roughly `std(μ)/mean(μ)/√k ≈ 10%`. Combined: ~√(6.3² + 10²) ≈ 12% plus the single-draw non-expectation nature of the test. The 20% bound (~1.7σ headroom) is a sound non-flaky guard.

Add a comment so the bound is not later "tightened" into flakiness:
```rust
// 20% tolerance: KMV cardinality CV ≈ 6%, μ̄ sampling CV ≈ 10%,
// combined ≈ 12% (in quadrature), plus ratio-bias O(1/k); 20% = ~1.7σ non-flaky guard.
assert!(rel_err < 0.20, ...);
```

s243a:
Addressed all review items in `11ea9ba6`. Summary:

**Required**
- **R1 (k=0 panics, k=1 → IC=0)** — `debug_assert!(k >= 2)` at entry to `descendant_minhash_weighted`, `sketch_mu_mass`, and `sketch_mu_overlap`. The `(k−1)/ĥ_k` read is undefined at k=1 (mass ≡ 0 ⇒ IC=0, the inversion) and panics at k=0.
- **R2 (zero-mass dependency on #3271)** — #3271 already merged with the `+∞` empty-cone fix; I documented the inherited dependency and the thin near-zero-mass band where the *sketch* path is high-variance (the sample can miss the few in-domain nodes, reading μ̄=0 ⇒ +∞). New test `mu_weighted_sketch_zero_mass_is_infinite_and_cycle_is_none` pins exact and sketch IC both → `+∞` at exactly zero mass (and folds in NB6's None-on-cycle path).
- **R3 (near-zero ĥ_k overflow to ~10¹⁹)** — capped the KMV cardinality at the node count (a cone/union is ⊆ the graph) via a new `n_universe` parameter on the three read-outs. Principled bound, not a magic `1e12`.

**Non-blocking**
- **NB1** — re-labelled the estimator as a bottom-k *plug-in/ratio* subset-sum (unbiased to O(1/k)), *in the spirit of* Cohen–Kaplan PODC/SIGMETRICS 2007 but **not** their VLDB 2008 Horvitz–Thompson form; clarified this is **carry-weight KMV**, not Ioffe 2010 Weighted MinHash (it estimates weighted *mass*, not `J_w`). Added the Ioffe 2010 reference.
- **NB2** — documented the 20% scale-test tolerance (KMV CV ≈ 6.3% ⊕ μ̄ CV ≈ 10% in quadrature ≈ 12%, ~1.7σ non-flaky).
- **NB3** — `debug_assert!(a.len() <= k && b.len() <= k)` (Broder same-k precondition).
- **NB4** — documented `sketch_mu_overlap` as a single indicator-gated product estimator (ranking signal, not absolute mass; high-variance at low overlap); new saturated-regime test `mu_weighted_overlap_saturated_regime` (rel err 2.0%).

The council-confirmed-correct points (the `(k−1)/ĥ_k` estimator, the `D̂·μ̄` independence/unbiasedness, the collision-free u32 dedup, the topo logic, the IC novelty) are unchanged.

95/95 lib tests pass (+2 new); Prolog template guard passes.

---

s243a:
Follow-up on the **R3** overflow cap (`91344cdc`): rather than hard-wiring the cap to the node count, it's now a configurable `CardCap` policy with a documented default, since both bounds are valid in different settings:

- **`CardCap::Universe(N)`** — cap at the node count (cone ⊆ graph ⇒ `D̂ ≤ N`); exact, tight, recommended when `N` is known.
- **`CardCap::Ceiling(c)`** — fixed ceiling for streaming / unknown-`N` or cross-graph comparisons; `Default = Ceiling(1e12)` as a caller-free overflow guard.
- **`CardCap::Uncapped`** — raw KMV, for downstream-clamped (the IC read-out already does `.min(total_mu)`) or test use.

Both spellings work: the read-outs take `impl Into<CardCap>`, so you can pass the explicit enum *or* a bare number — `usize ⇒ Universe`, `f64 ⇒ Ceiling`. New test `card_cap_config_enum_and_numeric_forms` proves the knob is live (a tiny ceiling visibly clips a saturated cone) and the two numeric spellings map to the documented variants; the tradeoffs are written up in §5d of the design doc. Existing call sites pass a `usize N`, so they resolve to `Universe(N)` with unchanged semantics. 96/96 lib tests pass.

---

PR #3274, opened 2026-06-18
WAM-Rust: μ-weighted overlap lift — the membership-aware hub-selection read-out

s243a:
The global read-out the μ-weighted KMV sketch (#3273) was built for: **`sketch_mu_overlap_lift`**, the membership-aware analogue of `sketch_overlap_lift`. It measures shared *in-domain* mass against the **mass configuration-model null** `E = m_A·m_B / M` (a random unit of admitted mass sits in cone A with prob `m_A/M` and in B with prob `m_B/M`, independently), so `lift > 1` is an *excess in-domain funnel* — a genuine domain hub — and `lift ≈ 1` is the small-world background. At `μ ≡ 1` it reduces **exactly** to the unweighted `sketch_overlap_lift`.

## Why this is the payoff
This is where the associative leak finally bites the *global* signal and the weighting earns its keep. The raw `sketch_overlap_lift` is fooled by the leak: two cones that overlap heavily on *out-of-domain* descendants look like a hub. The weighted lift discounts those low-μ descendants.

The test `mu_overlap_lift_separates_real_hub_from_leak_hub` pits a real in-domain hub against a leak-driven false hub with **identical cone sizes** (so identical *unweighted* lift):

| | unweighted lift | weighted lift |
|---|---:|---:|
| real in-domain hub (shared μ=1) | 4.01 | **2.50** (excess funnel, > 1) |
| leak-driven false hub (shared μ=0.02) | 4.01 | **0.25** (below chance, suppressed) |

The unweighted lift cannot tell them apart; the weighted lift separates them 10×, and correctly pushes the leak hub *below* chance. (Numbers are exact — every cone is unsaturated at k=128.)

## Contents
- `sketch_mu_overlap_lift(a, b, total_mu, k, cap)` — `sketch_mu_overlap / (m_A·m_B/M)`, takes the same `CardCap` knob (`impl Into<CardCap>`).
- `mu_overlap_lift_reduces_to_unweighted_at_mu_one` — exact agreement with `sketch_overlap_lift` at μ≡1.
- `mu_overlap_lift_separates_real_hub_from_leak_hub` — the hub-selection demonstration above.
- Documented in §5d.

98/98 lib tests pass; Prolog template guard passes.

---

### DIFF
--- a/docs/design/WAM_RUST_GRAPH_FUNCTIONAL_SEMIRINGS.md
+++ b/docs/design/WAM_RUST_GRAPH_FUNCTIONAL_SEMIRINGS.md
@@ -832,7 +832,15 @@ calibration of the relatedness read-out.)
 >   bottom-`k` intersection, weighted). A funnel hub is one many cones reach, but a weighted funnel
 >   discounts the associative leak — cones overlapping only on low-μ descendants score ≈ 0 — so hub
 >   selection stops being fooled by the leak. (Measured: core-shared overlap `3.00` vs leak-shared
->   `0.15`.)
+>   `0.15`.) The **hub-selection read-out** built on this is `sketch_mu_overlap_lift`, the μ-weighted
+>   analogue of `sketch_overlap_lift`: shared mass against the *mass* configuration-model null
+>   `E = m_A·m_B / M`, so `lift > 1` is an excess in-domain funnel and `lift ≈ 1` is the small-world
+>   background; `μ ≡ 1` reduces exactly to the unweighted lift. This is where the leak finally bites
+>   the *global* signal and the weighting pays off: on a test with a real in-domain hub and a
+>   leak-driven false hub of **identical cone sizes** — so identical *unweighted* lift (`4.01` both,
+>   fooled) — the weighted lift reads the real hub `2.50` (excess funnel, `> 1`) and the leak hub
+>   `0.25` (*below* chance, correctly suppressed), a 10× separation
+>   (`mu_overlap_lift_separates_real_hub_from_leak_hub`). The unweighted lift cannot tell them apart.
 > - **Depth-stability** (`mu_weighted_count_is_depth_stable`): the raw cone count *explodes* toward
 >   the root as the leak accumulates (more deep nodes than shallow), but the weighted mass barely
 >   moves because the deep nodes are out-of-domain — on the spine test the raw cone grows `201 → 645`
--- a/docs/design/WAM_RUST_GRAPH_FUNCTIONAL_SEMIRINGS.md
+++ b/docs/design/WAM_RUST_GRAPH_FUNCTIONAL_SEMIRINGS.md
@@ -834,13 +834,20 @@ calibration of the relatedness read-out.)
 >   selection stops being fooled by the leak. (Measured: core-shared overlap `3.00` vs leak-shared
 >   `0.15`.) The **hub-selection read-out** built on this is `sketch_mu_overlap_lift`, the μ-weighted
 >   analogue of `sketch_overlap_lift`: shared mass against the *mass* configuration-model null
->   `E = m_A·m_B / M`, so `lift > 1` is an excess in-domain funnel and `lift ≈ 1` is the small-world
->   background; `μ ≡ 1` reduces exactly to the unweighted lift. This is where the leak finally bites
->   the *global* signal and the weighting pays off: on a test with a real in-domain hub and a
->   leak-driven false hub of **identical cone sizes** — so identical *unweighted* lift (`4.01` both,
->   fooled) — the weighted lift reads the real hub `2.50` (excess funnel, `> 1`) and the leak hub
->   `0.25` (*below* chance, correctly suppressed), a 10× separation
->   (`mu_overlap_lift_separates_real_hub_from_leak_hub`). The unweighted lift cannot tell them apart.
+>   `E = m_A·m_B / M` — the **weighted configuration model** of Barrat et al. (2004) / Newman (2004),
+>   `s_i·s_j/2W` with admitted mass as the strength `s` — so `lift > 1` is an excess in-domain funnel
+>   and `lift ≈ 1` is the small-world background. In the **exact (`k ≥ |cone|`) regime** `μ ≡ 1`
+>   reduces *exactly* to the unweighted lift (under saturation the weighted path applies the cap and
+>   the unweighted does not, so the reduction is then only approximate). This is where the leak finally
+>   bites the *global* signal and the weighting pays off: on a test (exact regime) with a real
+>   in-domain hub and a leak-driven false hub of **identical cone sizes** — so identical *unweighted*
+>   lift (`4.01` both, fooled) — the weighted lift reads the real hub `2.50` (excess funnel, `> 1`) and
+>   the leak hub `0.25` (*below* chance, correctly suppressed), a 10× separation
+>   (`mu_overlap_lift_separates_real_hub_from_leak_hub`); the discrimination survives sketch saturation
+>   (`mu_overlap_lift_saturated_regime_and_cap_is_live`, `k=32`: real `2.33`, leak `0.18`). The
+>   unweighted lift cannot tell them apart. *Caveat:* the lift is high-variance for tiny cones
+>   (product-lift RSE ≈ `√3/√k`), so the read-out gates cones with mass `< 1e-9·M` to `0` (the weighted
+>   resolution limit); gate harder upstream if many small cones are in play.
 > - **Depth-stability** (`mu_weighted_count_is_depth_stable`): the raw cone count *explodes* toward
 >   the root as the leak accumulates (more deep nodes than shallow), but the weighted mass barely
 >   moves because the deep nodes are out-of-domain — on the spine test the raw cone grows `201 → 645`
@@ -1252,6 +1259,11 @@ Each is referenced inline at the section that uses it.
 - Ioffe, S. (2010). *Improved Consistent Sampling, Weighted Minhash and L1 Sketching.* ICDM 2010. —
   Weighted MinHash (weight baked into the hash to estimate weighted Jaccard `J_w`); contrasted with
   the *carry-weight KMV* of `descendant_minhash_weighted`, which estimates weighted mass, not `J_w`.
+- Barrat, A., Barthélemy, M., Pastor-Satorras, R., & Vespignani, A. (2004). *The Architecture of
+  Complex Weighted Networks.* PNAS 101(11). — the weighted-network strength `s_i`; with Newman 2004
+  the `s_i·s_j/2W` weighted configuration-model null that `sketch_mu_overlap_lift` uses (mass = strength).
+- Newman, M. E. J. (2004). *Analysis of Weighted Networks.* Phys. Rev. E 70, 056131
+  (arXiv:cond-mat/0408187). — weighted modularity / the weighted configuration-model expectation.
 
 **2-hop cover / hub labeling and landmark distance (§5c, roadmap 3f).**
 - Cohen, E., Halperin, E., Kaplan, H., & Zwick, U. (2002/2003). *Reachability and Distance
--- a/templates/targets/rust_wam/boundary_cache.rs.mustache
+++ b/templates/targets/rust_wam/boundary_cache.rs.mustache
@@ -1822,6 +1822,38 @@ pub fn sketch_mu_overlap(a: &[(u64, f64)], b: &[(u64, f64)], k: usize, cap: impl
     union_card * (shared.len() as f64 / union.len() as f64) * mean_shared_w
 }
 
+/// **μ-weighted overlap lift — the membership-aware hub funnel signal.** The weighted analogue of
+/// `sketch_overlap_lift`: it asks whether two cones share more *in-domain* mass than chance predicts,
+/// against the mass configuration-model null. With `m_A`, `m_B` the admitted masses (`sketch_mu_mass`)
+/// and `M = total_mu` the universe mass, the null expected shared mass is `E = m_A·m_B / M` (a random
+/// unit of admitted mass sits in `A` with prob `m_A/M` and in `B` with prob `m_B/M`, independently),
+/// and the signal is `lift = observed_shared_mass / E` with `observed = sketch_mu_overlap`. `lift > 1`
+/// is an *excess in-domain funnel* — a genuine domain hub; `lift ≈ 1` is the small-world background.
+/// At `μ ≡ 1` it reduces *exactly* to `sketch_overlap_lift` (`m_A=|A|`, `m_B=|B|`, `M=N`).
+///
+/// This is the read-out the weighted sketch was built for: the raw `sketch_overlap_lift` is fooled by
+/// the associative leak — two cones that overlap heavily on *out-of-domain* descendants look like a
+/// hub — but here the observed shared mass discounts those low-`μ` descendants, so a leak-driven false
+/// hub collapses to `lift ≈ 0` while a real in-domain hub stays `> 1`. `0` if either cone has no mass.
+pub fn sketch_mu_overlap_lift(
+    a: &[(u64, f64)],
+    b: &[(u64, f64)],
+    total_mu: f64,
+    k: usize,
+    cap: impl Into<CardCap>,
+) -> f64 {
+    let cap = cap.into();
+    if total_mu <= 0.0 {
+        return 0.0;
+    }
+    let (m_a, m_b) = (sketch_mu_mass(a, k, cap), sketch_mu_mass(b, k, cap));
+    let expected = m_a * m_b / total_mu;
+    if expected <= 0.0 {
+        return 0.0;
+    }
+    sketch_mu_overlap(a, b, k, cap) / expected
+}
+
 /// **Fuzzy admission curve**: an S-shaped (logistic) transfer from the raw membership `μ` to an
 /// admission weight in `[0,1]`, `w(μ) = 1/(1 + e^{−k(μ−c)})`. Shapes *how* `μ` translates to
 /// admission — nearly-full weight well above the boundary, nearly-zero well below, a smooth knee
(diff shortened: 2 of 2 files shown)


templates/targets/rust_wam/boundary_cache.rs.mustache
```diff
@@ -1822,6 +1822,38 @@ pub fn sketch_mu_overlap(a: &[(u64, f64)], b: &[(u64, f64)], k: usize, cap: impl
...
+/// and the signal is `lift = observed_shared_mass / E` with `observed = sketch_mu_overlap`. `lift > 1`
+/// is an *excess in-domain funnel* — a genuine domain hub; `lift ≈ 1` is the small-world background.
+/// At `μ ≡ 1` it reduces *exactly* to `sketch_overlap_lift` (`m_A=|A|`, `m_B=|B|`, `M=N`).
+///
+/// This is the read-out the weighted sketch was built for: the raw `sketch_overlap_lift` is fooled by
+/// the associative leak — two cones that overlap heavily on *out-of-domain* descendants look like a
+/// hub — but here the observed shared mass discounts those low-`μ` descendants, so a leak-driven false
+/// hub collapses to `lift ≈ 0` while a real in-domain hub stays `> 1`. `0` if either cone has no mass.
+pub fn sketch_mu_overlap_lift(
+    a: &[(u64, f64)],
```
s243a, line 1864: **[R1 — blocking, MOST IMPORTANT]** `cap` is dead code in both tests

At k∈{16,128} every test cone (≤71 nodes) is smaller than or equal to k, so the saturation branch in `sketch_mu_mass` is never reached. `CardCap::Universe(505)`, `CardCap::Uncapped`, and `CardCap::Ceiling(1e12)` produce byte-identical lifts.

**Suggested fix:** Add a test with k < cone size — e.g. k=32 on the 505-node test-2 graph — to exercise the capped path and validate that `cap` actually affects the result.

templates/targets/rust_wam/boundary_cache.rs.mustache
```diff
@@ -1822,6 +1822,38 @@ pub fn sketch_mu_overlap(a: &[(u64, f64)], b: &[(u64, f64)], k: usize, cap: impl
...
+    b: &[(u64, f64)],
+    total_mu: f64,
+    k: usize,
+    cap: impl Into<CardCap>,
+) -> f64 {
```
s243a, line 1869: **[R3 — blocking]** `CardCap::Ceiling(c)` is a semantic footgun for lift

The same `cap` value is forwarded to `sketch_mu_mass(a, k, cap)`, `sketch_mu_mass(b, k, cap)`, and `sketch_mu_overlap(a, b, k, cap)`. For `Universe(N)` this is safe — N bounds individual cones *and* their union. For `Ceiling(c)` sized to one cone, the union `|A ∪ B|` can exceed `c` while individual masses are unclipped, silently shrinking the numerator relative to the denominator and biasing the lift downward.

**Suggested fix:** Add a doc comment: `/// # Precondition: \`cap\` must be a valid cardinality bound for the cone *union*, not just individual cones. Prefer \`Universe(N)\` or \`Uncapped\` unless you have a safe global ceiling.`

Also consider asserting or documenting the `total_mu` universe-alignment precondition (NB4).

s243a (commented):
## Council-of-Models Review — PR #3274: μ-weighted overlap lift

**Reviewers:** Claude Opus 4.8 · GPT-5.5 · Gemini (deep research, web + academic sources)

### Summary

**Opus 4.8:** The `sketch_mu_overlap_lift` function is mathematically sound and its null model (`m_A·m_B/M`) is the correct weighted configuration-model baseline. The critical gap is that neither test ever exercises the saturated/capped code path — at k≥cone both `CardCap::Universe(505)`, `CardCap::Uncapped`, and `CardCap::Ceiling(1e12)` produce byte-identical lifts, so `cap` is dead code in the test suite. The doc claim that μ≡1 reduces exactly to unweighted lift is also only true in the unsaturated regime.

**GPT-5.5:** Agrees on the cap dead-code problem and adds that `CardCap::Ceiling(c)` is a semantic footgun: the same cap is passed to both individual mass calls and the joint overlap call, but for a `Ceiling` value sized to one cone the union can silently exceed it, biasing the ratio downward. The 10× separation claim in the doc is correct (Python verification: 10.03×) but the test only asserts >5× — these should be reconciled. Recommends documenting that `total_mu` must be Σμ over exactly the same universe as the sketches.

**Gemini (deep research):** Grounds the null model in [Barrat et al. PNAS 2004](https://www.pnas.org/doi/10.1073/pnas.0400087101) and [Newman 2004](https://arxiv.org/abs/cond-mat/0408187) (`s_i·s_j/2W` weighted modularity) — the formulation is theoretically well-motivated and should carry a citation. Flags the delta-method variance: product-lift RSE ≈ √3/√k, so at k=128 expect ~12% coefficient of variation; recommends a minimum-mass gate (`m_A, m_B ≥ ε·M`) to guard against high-variance small-cone cases (resolution-limit analogue).

### Verdict: Request Changes (4 blocking, 6 non-blocking)

---

### Blocking Issues

**R1 [Opus — MOST IMPORTANT]** `cap` is dead code in both tests; zero coverage of the saturated path. Both test cones are ≤71 nodes; at k∈{16,128} every cone is smaller than k or barely larger — the cap branch is never reached. `Universe(505)`, `Uncapped`, and `Ceiling(1e12)` give identical results. Add a test with k < cone size (e.g. k=32 on the 505-node test-2 graph).

**R2 [Opus + GPT]** The doc statement "μ≡1 ⇒ reduces exactly to `sketch_overlap_lift`" holds only in the unsaturated regime. Under saturation, `sketch_mu_mass` applies `cap` while the unweighted path does not — the reduction is no longer exact. Scope the claim to the exact/unsaturated regime or add a saturated μ≡1 test to validate it.

**R3 [GPT]** `CardCap::Ceiling(c)` is a semantic footgun for lift. The same `cap` value bounds `sketch_mu_mass(a)`, `sketch_mu_mass(b)`, and `sketch_mu_overlap(a,b)`. For `Universe(N)` this is safe (N bounds individual cones and their union). For `Ceiling(c)` sized to a single cone, the union can exceed `c` while individual cones are unclipped — silently reducing the denominator of the lift ratio. Document the precondition: `cap` must be a valid upper bound for the cone union, not just individual cones.

**R4 [GPT + Opus]** Doc says "10× separation"; test asserts `w_real > 5.0 * w_leak`. Python verification confirms the actual ratio is 10.03× (w_real=2.50, w_leak=0.25). Either tighten the test assertion to `w_real > 9.0 * w_leak` or soften the doc claim to ">5×".

---

### Non-blocking / Improvements

**NB1 [Gemini]** The null model `m_A·m_B/M` is the standard weighted configuration-model baseline (Barrat et al. 2004 / Newman 2004). Add a `/// # References` entry citing one of these.

**NB2 [Gemini]** Product-lift RSE ≈ √3/√k (delta method, three correlated sketches). At k=128: ~12% CV. Consider a minimum-mass gate — `if m_a < eps * total_mu || m_b < eps * total_mu { return 0.0; }` — to suppress spuriously large lifts when either cone is tiny relative to the universe.

**NB3 [Opus]** In test 2, the comment `let n = d.len();` is correct (n=505 includes all reachable nodes via topo-sort). Add an inline note: nodes absent from the edge list but reachable via topo-sort are included; truly isolated nodes would inflate `total_mu` without appearing in any sketch.

**NB4 [GPT]** `total_mu` must be Σμ over exactly the same universe as the sketches use for hash-space normalization. Add a doc comment precondition on `sketch_mu_overlap_lift`.

**NB5 [Opus]** The "fooled / identical unweighted lift" example is regime-dependent: identical only when k ≥ cone size (exact regime). In the approximate regime (k < cone), KMV sampling noise breaks the symmetry. Soften the doc to "structurally identical cones ⇒ identical unweighted lift in the exact (k≥cone) regime."

**NB6 [GPT]** Missing edge-case tests: (a) disjoint cones → lift should equal 0, (b) `total_mu=0` → should return 0 without panic, (c) `CardCap::Ceiling` with a cone that exceeds the ceiling.

templates/targets/rust_wam/boundary_cache.rs.mustache
```diff
@@ -1822,6 +1822,38 @@ pub fn sketch_mu_overlap(a: &[(u64, f64)], b: &[(u64, f64)], k: usize, cap: impl
...
+    let cap = cap.into();
+    if total_mu <= 0.0 {
+        return 0.0;
+    }
+    let (m_a, m_b) = (sketch_mu_mass(a, k, cap), sketch_mu_mass(b, k, cap));
+    let expected = m_a * m_b / total_mu;
+    if expected <= 0.0 {
```
s243a, line 1880: **[NB2 — non-blocking]** Consider a minimum-mass gate to suppress high-variance lifts

Product-lift RSE ≈ √3/√k (delta method, three correlated KMV sketches). At k=128 that is ~12% CV. When either cone is tiny relative to the universe, even small sketch noise drives a spuriously large lift — the resolution-limit analogue for weighted modularity.

**Suggested fix:**
```rust
let eps = 1e-9;
if m_a < eps * total_mu || m_b < eps * total_mu {
    return 0.0;
}
```
Or document the caller's responsibility to gate on mass before calling.

**Reference:** Barrat et al. PNAS 2004 (https://www.pnas.org/doi/10.1073/pnas.0400087101) — `s_i·s_j/2W` null model; the same resolution limit applies here.

templates/targets/rust_wam/boundary_cache.rs.mustache
```diff
@@ -4875,6 +4907,74 @@ mod tests {
             "μ-mass must be depth-stable: weighted growth {wmass_growth} vs raw {raw_growth}");
     }
 
+    // μ-weighted overlap LIFT reduces to the unweighted sketch_overlap_lift at μ≡1. Diamond: 3->[1,2],
+    // 4->[3]; 1,2->0. cone(1)={1,3,4}, cone(2)={2,3,4}, ∩={3,4}. k >> cone ⇒ exact, so the two lifts
+    // must agree to 1e-9.
+    #[test]
+    fn mu_overlap_lift_reduces_to_unweighted_at_mu_one() {
```
s243a, line 4944: **[NB6 — non-blocking]** Missing edge-case tests

Three cases worth covering:
1. **Disjoint cones** — `sketch_mu_overlap` should return 0 → lift=0 (no intersection).
2. **`total_mu=0`** — the guard `if total_mu <= 0.0 { return 0.0; }` is present, but it should be tested explicitly to confirm no panic.
3. **`CardCap::Ceiling` with a cone that exceeds the ceiling** — needed to verify R3 above and validate `cap` semantics in the saturated regime.

templates/targets/rust_wam/boundary_cache.rs.mustache
```diff
@@ -4875,6 +4907,74 @@ mod tests {
...
+    // THE HUB-SELECTOR PAYOFF: the μ-weighted lift discriminates a real in-domain hub from a
+    // leak-driven false hub where the UNWEIGHTED lift cannot. Two cone-pairs with IDENTICAL sizes
+    // (so identical unweighted lift): pair A,B shares 40 in-domain nodes (μ=1); pair C,D shares 40
+    // out-of-domain leak nodes (μ=0.02). Both also carry 30 private in-domain children so their total
+    // MASS is comparable. The unweighted lift is fooled (equal); the weighted lift ranks the real hub
+    // above chance (>1) and pushes the leak hub below chance (<1).
+    #[test]
+    fn mu_overlap_lift_separates_real_hub_from_leak_hub() {
+        const K: usize = 128; // > every cone (71) ⇒ exact
+        let mut g: HashMap<u32, Vec<u32>> = HashMap::new();
```
s243a, line 4967: **[R1 — blocking]** k=128 never saturates the test cone

With k=128 and a 505-node graph, each cone's KMV sketch is limited to 128 entries. However, the test cones (reachable from the two hub nodes) appear to be ≤71 nodes, so k=128 ≥ cone size → exact regime, `cap` is never consulted.

**Suggested fix:** Add a companion test at k=32 on this same graph. At k=32 the 505-node reachable set forces saturation, `cap` becomes load-bearing, and you can verify that `Universe(505)` vs `Ceiling(50.0)` produce meaningfully different lifts.

templates/targets/rust_wam/boundary_cache.rs.mustache
```diff
@@ -4875,6 +4907,74 @@ mod tests {
...
+            unw_real, unw_leak, w_real, w_leak);
+
+        // Unweighted lift cannot tell them apart (identical cone sizes).
+        assert!((unw_real - unw_leak).abs() < 1e-6, "unweighted lift is fooled: real {unw_real} leak {unw_leak}");
+        assert!(unw_real > 1.0, "both look like hubs structurally");
+        // Weighted lift: real hub is an excess in-domain funnel; leak hub falls below chance.
+        assert!(w_real > 1.0, "real in-domain hub: weighted lift > 1 ({w_real})");
+        assert!(w_leak < 1.0, "leak hub correctly suppressed below chance ({w_leak})");
+        assert!(w_real > w_leak * 5.0, "real hub ranks far above the leak hub ({w_real} vs {w_leak})");
+    }
```
s243a, line 5080: **[R4 — blocking]** Doc claims 10× separation; test only asserts 5×

Python verification (verify_3274.py) confirms: w_real=2.50, w_leak=0.25 → 10.03× actual separation. The assertion `assert!(w_real > 5.0 * w_leak)` passes but leaves a 2× gap between what the doc promises and what the test enforces.

**Suggested fix (choose one):**
- Tighten: `assert!(w_real > 9.0 * w_leak, "expected ≥10× separation, got {:.2}×", w_real/w_leak);`
- Soften the doc: change "10× separation" → "≥5× separation" to match the test.

**Also (NB5):** This 10× figure is exact only in the unsaturated regime (k≥cone). In the approximate regime (k<cone) KMV noise will reduce the separation. Soften the doc claim to "structurally identical cones ⇒ identical unweighted lift *in the exact (k≥cone) regime*."

docs/design/WAM_RUST_GRAPH_FUNCTIONAL_SEMIRINGS.md
```diff
@@ -832,7 +832,15 @@ calibration of the relatedness read-out.)
 >   bottom-`k` intersection, weighted). A funnel hub is one many cones reach, but a weighted funnel
```
s243a, line 832: **[R2 — blocking]** μ≡1 reduction is exact only in the unsaturated (k≥cone) regime

The doc states that setting μ≡1 makes `sketch_mu_overlap_lift` reduce *exactly* to `sketch_overlap_lift`. This holds when k≥cone size (exact regime) because `sketch_mu_mass` then equals the cone cardinality for both weighted and unweighted paths. Under saturation, `sketch_mu_mass` applies `cap` while the unweighted `sketch_overlap_lift` does not — the reduction is no longer exact.

**Suggested fix:** Scope the claim: *"When μ≡1 and k ≥ |cone| (exact/unsaturated regime), `sketch_mu_overlap_lift` reduces exactly to `sketch_overlap_lift`."*

**NB1:** The null model `m_A·m_B/M` is the weighted configuration-model baseline from [Barrat et al. PNAS 2004](https://www.pnas.org/doi/10.1073/pnas.0400087101) and [Newman 2004 (arXiv:cond-mat/0408187)](https://arxiv.org/abs/cond-mat/0408187). Adding a `/// # References` block with these citations would ground the formulation theoretically.

s243a:
Addressed all review items in `8b5b1fd0`. Summary:

**Blocking**
- **R1 (cap was dead code — all test cones ≤ k)** — this was the key catch. New test `mu_overlap_lift_saturated_regime_and_cap_is_live` runs at **k=32** on the 505-node hub graph, forcing saturation so the KMV-estimator/cap branch actually executes in the lift path. It shows (a) the real-vs-leak discrimination *survives* sketching (real `2.33`, leak `0.18`) and (b) the cap is load-bearing (`Ceiling(20)` clips a saturated ~71-node cone to 20; `Universe(505)` doesn't bite). Extracted a `hub_test_graph()` helper shared by both lift tests.
- **R2 (μ≡1 reduction exact only when k≥cone)** — scoped the rustdoc and design-doc claim to the exact/unsaturated regime, noting the cap divergence under saturation.
- **R3 (`Ceiling` footgun for the union)** — documented the precondition that a `Ceiling` cap must bound the cone **union**, not just individual cones; recommend `Universe(N)`/`Uncapped`. Also noted the lift ratio is largely cap-*insensitive* (cap appears in numerator union-card and the `m_A·m_B` denominator, so it cancels) — confirmed empirically.
- **R4 (doc said 10×, test asserted 5×)** — tightened to `w_real > 9.0 * w_leak` (exact 10.03×); softened the "fooled/identical" framing to the exact regime.

**Non-blocking**
- **NB1** — cited the weighted configuration model (Barrat et al. PNAS 2004, Newman 2004) as the null basis, in the doc and §10 references. Your `s_i·s_j/2W` mapping is exactly right (admitted mass = strength `s`).
- **NB2** — added a resolution-limit gate: cones with mass `< 1e-9·M` return `0`, suppressing spurious high-variance lifts from tiny cones; documented RSE ≈ `√3/√k`. (The gate is conservative enough not to touch the test numbers.)
- **NB3/NB4** — documented the `total_mu` universe-alignment precondition and the `n = reachable-node-count` semantics.
- **NB6** — `mu_overlap_lift_edge_cases`: disjoint cones ⇒ lift 0, `total_mu=0` ⇒ no panic.

One nuance worth flagging from R3: while making the cap load-bearing I confirmed the **lift ratio is fairly insensitive to the cap choice** (it cancels between numerator and denominator) — so the `Ceiling`-vs-`Universe` footgun mostly affects the *raw* mass/overlap, less the lift. Documented as such rather than claiming the cap meaningfully changes the lift.

100/100 lib tests pass (+2 new); Prolog template guard passes.

---

s243a:
**Correction to my previous comment** (`fd77a304`). I claimed the lift ratio is "largely cap-insensitive" because the cap "cancels between numerator and denominator." That was **wrong**, and @s243a was right to push on it.

Working the algebra: when a `Ceiling(c)` bites below the cone cardinalities, it scales the numerator union-card *once* and the `m_A·m_B` denominator *twice*, so the **absolute lift is ∝ 1/c**, not invariant. Verified empirically on the saturated (k=32) hub graph:

| cap | real lift | leak lift | real/leak |
|---|---:|---:|---:|
| Uncapped / Universe(505) | 2.33 | 0.18 | 12.7 |
| Ceiling(50) | 2.56 | 0.19 | 13.5 |
| Ceiling(20) | 6.40 | 0.47 | 13.5 |
| Ceiling(10) | 12.81 | 0.95 | 13.5 |

So a `Ceiling(10)` inflates the absolute lift ~5× over `Uncapped` (and it's cleanly ∝ 1/c: c=10 vs c=20 gives exactly 2×). This makes **R3 more significant than I first credited** — a mis-sized `Ceiling` genuinely biases the absolute lift.

What *is* preserved is the **relative ranking**: the cap rescales every cone comparably, so `lift_real / lift_leak` stays ≈ constant (12.7–13.5). That's the property hub *selection* relies on, and it's why the discrimination is robust to the cap even though the absolute value isn't. For a *calibrated* absolute lift, use `Universe(N)`.

I corrected the rustdoc, removed the misleading comment from the saturated test, and added `mu_overlap_lift_cap_rescales_absolute_keeps_ranking` to pin both facts (absolute ∝ 1/c; ranking preserved within 25%). 101/101 lib tests pass.

---

PR #3275, opened 2026-06-19
WAM-Rust: μ-weighted fan-in hub ranking — the leak-robust "select the hubs" step

s243a:
Completes the hub arc: the pairwise `sketch_mu_overlap_lift` (#3274) becomes a per-node **selection** signal — the actual "pick the hubs" step.

## Contents
- **`mu_fanin_hub_score(candidate, queries, sketches, total_mu, k, cap)`** — the *mean* weighted overlap-lift of a candidate's descendant cone against a query/seed set. Skips self (a cone's self-overlap is trivially maximal); queries the candidate doesn't reconverge with contribute `0`, correctly diluting a hub that only funnels a handful of seeds.
- **`rank_mu_fanin_hubs(candidates, queries, …)`** — scores every candidate and returns `(node, score)` sorted descending (node-id tie-break for determinism).

It inherits the leak-robustness from the underlying lift: a hub whose only commonality with the seeds is the associative leak sinks *below chance* even when its raw cone overlap is large.

## The payoff
`rank_mu_fanin_hubs_picks_real_over_leak`: queries 1,2,3 each reconverge with candidate **10** on a 15-node in-domain block (μ=1) and with candidate **20** on a 15-node leak block (μ=0.02). Both candidates carry 30 private in-domain children, so their cones are the **same size**:

| | unweighted mean hub score | weighted hub score |
|---|---:|---:|
| real in-domain hub (10) | 2.90 | **3.17** (excess funnel, > 1) |
| leak-driven false hub (20) | 2.90 | **0.15** (below chance, suppressed) |

The unweighted score **ties** them; the weighted ranking puts the real hub first by a **21× margin**. (Exact — every cone is unsaturated at k=256.)

## The arc, complete
**graded μ → partial-admission IC (#3271) → scalable weighted sketch (#3273) → membership-aware hub lift (#3274) → fan-in hub *ranking* (this PR).**

102/102 lib tests pass; Prolog template guard passes. Documented in §5d.

---

### DIFF
--- a/docs/design/WAM_RUST_GRAPH_FUNCTIONAL_SEMIRINGS.md
+++ b/docs/design/WAM_RUST_GRAPH_FUNCTIONAL_SEMIRINGS.md
@@ -847,7 +847,13 @@ calibration of the relatedness read-out.)
 >   (`mu_overlap_lift_saturated_regime_and_cap_is_live`, `k=32`: real `2.33`, leak `0.18`). The
 >   unweighted lift cannot tell them apart. *Caveat:* the lift is high-variance for tiny cones
 >   (product-lift RSE ≈ `√3/√k`), so the read-out gates cones with mass `< 1e-9·M` to `0` (the weighted
->   resolution limit); gate harder upstream if many small cones are in play.
+>   resolution limit); gate harder upstream if many small cones are in play. The **selection** step on
+>   top of the pairwise lift is `mu_fanin_hub_score` / `rank_mu_fanin_hubs`: score each candidate node
+>   by the *mean* weighted lift of its cone against a query/seed set, then rank descending. This is
+>   "pick the hubs," and it inherits the leak-robustness — on a test where a real in-domain hub and a
+>   leak hub share *identically-sized* blocks with the seeds (so the *unweighted* mean lift **ties**
+>   them at `2.90`), the weighted ranking reads the real hub `3.17` (`> 1`) and the leak hub `0.15`
+>   (`< 1`), ranking the real hub first by a 21× margin (`rank_mu_fanin_hubs_picks_real_over_leak`).
 > - **Depth-stability** (`mu_weighted_count_is_depth_stable`): the raw cone count *explodes* toward
 >   the root as the leak accumulates (more deep nodes than shallow), but the weighted mass barely
 >   moves because the deep nodes are out-of-domain — on the spine test the raw cone grows `201 → 645`
--- a/docs/design/WAM_RUST_GRAPH_FUNCTIONAL_SEMIRINGS.md
+++ b/docs/design/WAM_RUST_GRAPH_FUNCTIONAL_SEMIRINGS.md
@@ -849,11 +849,16 @@ calibration of the relatedness read-out.)
 >   (product-lift RSE ≈ `√3/√k`), so the read-out gates cones with mass `< 1e-9·M` to `0` (the weighted
 >   resolution limit); gate harder upstream if many small cones are in play. The **selection** step on
 >   top of the pairwise lift is `mu_fanin_hub_score` / `rank_mu_fanin_hubs`: score each candidate node
->   by the *mean* weighted lift of its cone against a query/seed set, then rank descending. This is
->   "pick the hubs," and it inherits the leak-robustness — on a test where a real in-domain hub and a
->   leak hub share *identically-sized* blocks with the seeds (so the *unweighted* mean lift **ties**
->   them at `2.90`), the weighted ranking reads the real hub `3.17` (`> 1`) and the leak hub `0.15`
->   (`< 1`), ranking the real hub first by a 21× margin (`rank_mu_fanin_hubs_picks_real_over_leak`).
+>   by the aggregate weighted lift of its cone against a query/seed set, then rank descending. The
+>   aggregator is a knob (`HubAgg`): **`Pooled`** (default) is the ratio-of-sums `Σobs/ΣE` — the
+>   candidate's lift against the seed set as one aggregate cone, lower-variance and the standard way to
+>   pool enrichment; **`MeanOfRatios`** weights each seed equally (higher variance, leans on the mass
+>   gate). Both inherit the leak-robustness — on a test where a real in-domain hub and a leak hub share
+>   *identically-sized* blocks with the seeds (so the *unweighted* mean lift **ties** them at `2.90`),
+>   the weighted ranking reads the real hub `3.17` (`> 1`) and the leak hub `0.15` (`< 1`), ranking the
+>   real hub first by a 21× margin (`rank_mu_fanin_hubs_picks_real_over_leak`); the two aggregators
+>   diverge only on heterogeneous-size seeds and still agree on the ranking
+>   (`hub_agg_pooled_vs_mean_differ_on_heterogeneous_seeds`).
 > - **Depth-stability** (`mu_weighted_count_is_depth_stable`): the raw cone count *explodes* toward
 >   the root as the leak accumulates (more deep nodes than shallow), but the weighted mass barely
 >   moves because the deep nodes are out-of-domain — on the spine test the raw cone grows `201 → 645`
--- a/templates/targets/rust_wam/boundary_cache.rs.mustache
+++ b/templates/targets/rust_wam/boundary_cache.rs.mustache
@@ -1883,15 +1883,40 @@ pub fn sketch_mu_overlap_lift(
     sketch_mu_overlap(a, b, k, cap) / expected
 }
 
+/// **How to aggregate the per-seed reconvergence lifts into one hub score.** Both options inherit the
+/// underlying lift's leak-robustness (they sum/average `μ`-weighted shared *mass*, which discounts the
+/// associative leak); they differ in *how seeds of different cone sizes are weighted* and in variance.
+///
+/// - [`HubAgg::Pooled`] — **ratio-of-sums** `Σ obs_i / Σ E_i`: the candidate's total shared in-domain
+///   mass against the total expected, i.e. its lift against the seed set treated as **one aggregate
+///   cone**. Lower variance (a single small-cone seed with a tiny `E_i` cannot spike it — the same
+///   resolution-limit fragility `sketch_mu_overlap_lift` gates against), mass-weighted so larger seeds
+///   count more. The safer **default**, and the standard way to pool enrichment/lift.
+/// - [`HubAgg::MeanOfRatios`] — **mean-of-ratios** `(1/m) Σ obs_i/E_i`: every seed's lift weighted
+///   *equally* regardless of its cone size. Use when each seed is an equally-important query concept
+///   and you do not want big cones to dominate; higher variance (a tiny-cone seed can dominate the
+///   mean), so it leans harder on the mass gate.
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub enum HubAgg {
+    /// `Σ obs_i / Σ E_i` — pooled, lower-variance, mass-weighted. The default.
+    Pooled,
+    /// `(1/m) Σ obs_i/E_i` — equal weight per seed, higher variance.
+    MeanOfRatios,
+}
+impl Default for HubAgg {
+    fn default() -> Self { HubAgg::Pooled }
+}
+
 /// **μ-weighted fan-in hub score** — aggregate the pairwise reconvergence into a single per-node
-/// number: the *mean* `sketch_mu_overlap_lift` of a candidate's descendant cone against a set of query
-/// (seed) cones. A node that reconverges with *many* queries on *in-domain* mass scores high; one that
-/// reconverges with few, or only on the associative leak, scores ≈ chance (`1`) or below. The mean is
-/// taken over every query whose sketch is present and which is *not* the candidate itself (a cone's
-/// self-overlap is trivially maximal and would swamp the score). Queries the candidate does not
-/// reconverge with contribute `0`, correctly diluting a hub that only funnels a handful of seeds.
-/// `0` if no usable queries. This is the leak-robust selection signal: rank candidates by it to pick
-/// hubs that the raw (unweighted) overlap would tie or mis-rank (see `rank_mu_fanin_hubs`).
+/// number: how much a candidate's descendant cone reconverges (shares *in-domain* mass) with a set of
+/// query (seed) cones, against the mass configuration-model null. A node that reconverges with *many*
+/// seeds on in-domain mass scores high (`> 1`); one that reconverges with few, or only on the
+/// associative leak, scores ≈ chance (`1`) or below. `agg` selects how the per-seed signals combine
+/// ([`HubAgg`]; `Pooled` by default). Seeds equal to the candidate are skipped (a cone's self-overlap
+/// is trivially maximal); a seed the candidate does not reconverge with contributes `0` mass, so it
+/// *dilutes* a hub that only funnels a handful of seeds. Sub-resolution seeds (mass `< 1e-9·M`) are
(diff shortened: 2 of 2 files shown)


s243a:
**Self-harden pass** (`2a0abfdd`) before review — addressed the one genuinely non-trivial question in this PR plus the edge cases.

**Aggregator (the estimator question).** I was aggregating as mean-of-ratios `(1/m)Σ obs_i/E_i`. That's the *higher-variance* choice — one small-cone seed with a tiny `E_i` can spike the score (the same resolution-limit fragility `sketch_mu_overlap_lift` already gates against). Switched the default to a **pooled ratio-of-sums** `Σobs / ΣE` (the candidate's lift against the seed set as one aggregate cone — lower variance, the standard way to pool enrichment), and exposed both via a `HubAgg` enum (`Pooled` default, `MeanOfRatios` for equal-weight-per-seed semantics), mirroring `CardCap`. Both inherit leak-robustness and both select the real hub; they agree numerically on symmetric seeds and diverge only on heterogeneous-size seeds (`hub_agg_pooled_vs_mean_differ_on_heterogeneous_seeds`: pooled 3.09 vs mean 2.83, both rank real first).

**Also hardened:** `total_mu ≤ 0` guard, sub-resolution seed gate (mass `< 1e-9·M`), and `m_candidate` computed once instead of per-seed. Added `mu_fanin_hub_score_edge_cases` (empty seeds / only-self seed / disjoint seed / missing candidate ⇒ all `0`, no panic).

104/104 lib tests pass; Prolog guard passes.

---

s243a (commented):
## Review: WAM-Rust μ-weighted fan-in hub ranking (PR #3275)

Focusing on the three areas Claude flagged: **hub-score aggregation design (Pooled vs MeanOfRatios / HubAgg enum / default choice)**, **the resolution gate**, and **the sort**.

---

### 1. Hub-score aggregation — is Pooled the right default?

**Short answer: yes, and the self-hardening pass makes the rationale airtight.**

The key insight is that these two estimators are answering subtly different questions:

| | `Pooled` (`Σobs/ΣE`) | `MeanOfRatios` (`(1/m)Σ obs_i/E_i`) |
|---|---|---|
| **Implicit question** | "Does this candidate's cone reconverge with the *seed set as a whole* above the null?" | "Does this candidate reconverge with the *average seed*, independent of seed size?" |
| **Weighting** | Mass-proportional — large cones count more | Equal per seed — each query concept counts equally |
| **Variance** | Lower — single small-E_i seed cannot spike the ratio | Higher — a tiny-cone seed (small E_i in denominator) can spike the mean |
| **Natural analogy** | Pooled enrichment / Fisher's combined test | Arithmetic mean of per-experiment fold-changes |

For the hub-selection use case — "which node is the best funnel for *this set of queries*?" — the pooled ratio is the correct semantic default. Hub selection is asking whether a candidate reconverges with the query *mass*, not whether it passes a per-seed threshold equally. A seed with a large cone that reconverges strongly *should* dominate — it's providing more signal. `MeanOfRatios` is only strictly better when each seed is an equally-important, independently-calibrated query concept (e.g. in a multi-concept retrieval scenario where seed cones are deliberately heterogeneous and you don't want one big cone to swamp the others).

The self-harden comment correctly diagnoses the specific failure mode that motivated the switch: `MeanOfRatios` is vulnerable to the *same* resolution-limit fragility as `sketch_mu_overlap_lift` — a tiny-mass seed produces a tiny `E_i`, inflating `obs_i/E_i` even for weak overlap. The gate (`m_q < floor`) defends against this but it's an exclusion, not a correction; `Pooled` is variance-correct by construction.

**One small design note worth flagging:** The test `hub_agg_pooled_vs_mean_differ_on_heterogeneous_seeds` verifies that the two diverge numerically (pooled ≈ 3.09, mean ≈ 2.83 for the real hub) and both still rank the real hub first. This is the right test. However, it does not probe the *leak hub* scores under heterogeneous seeds — consider adding an assertion that the leak hub also stays below chance (`< 1.0`) under both aggregators when seeds are heterogeneous. The symmetric-seed test already covers the `0.15` figure; the heterogeneous test should cover the analogous bound to close the specification.

---

### 2. Resolution gate (`floor = 1e-9 * total_mu`)

The gate is applied twice: once on the **candidate** (`m_c < floor → return 0.0`) and once on each **seed** (`m_q < floor → continue`). Both are correct and necessary:

- The **candidate gate** prevents a ghost hub (no real in-domain presence) from ever being scored. It exits early, avoiding the inner loop entirely — good.
- The **seed gate** (drop sub-resolution seeds) is the right fix for `MeanOfRatios` variance, but it's also semantically correct for `Pooled`: a seed with mass below the resolution limit contributes an unreliable `obs` estimate, and including it would add noise to `sum_obs` / `sum_exp`. Dropping it cleanly is preferable to down-weighting it.

**Concern — the gate constant `1e-9` is undocumented as a design choice.** It matches the constant used in `sketch_mu_overlap_lift` (as the doc comment notes: "matches `sketch_mu_overlap_lift`"), which is good for consistency. But `1e-9` is a floating-point machinery constant (just above double subnormal territory), not a domain-tuned threshold. For a sketch universe where `total_mu` is e.g. `1e6`, the floor is `1e-3` — cones with mass below `0.001` are dropped. For `total_mu = 10`, the floor is `1e-8`. Document in the enum/function docstring that callers with many small cones should gate harder upstream (the existing doc comment says "gate harder upstream if many small cones are in play" — good — but say explicitly *why* this constant was chosen: it's the sketch resolution limit, not a domain threshold).

**Edge case not tested:** What happens if `total_mu` itself is very small (say, `1e-9`)? The guard at the top (`if total_mu <= 0.0 { return 0.0 }`) catches the zero/negative case, but a very-small positive `total_mu` makes `floor = 1e-18`, which is effectively zero — every candidate and seed passes. This is technically correct (if total_mu is tiny, all masses are sub-resolution relative to each other anyway), but it's a regime worth noting in the doc comment or with a test.

---

### 3. Sort — descending by score, node-id tie-break

```rust
scored.sort_by(|a, b| {
    b.1.partial_cmp(&a.1)
        .unwrap_or(std::cmp::Ordering::Equal)
        .then(a.0.cmp(&b.0))
});
```

This is correct and complete:
- **Descending score** is the right direction — highest reconvergence signal first.
- **`partial_cmp` with `unwrap_or(Equal)`** handles the NaN case gracefully (NaN scores sort to an indeterminate but stable position relative to each other; they do not panic). Since all code paths in `mu_fanin_hub_score` return finite values or `0.0`, NaN should not appear in practice — but defensive handling is correct here.
- **Node-id tie-break (`a.0.cmp(&b.0)`)** provides determinism, which is essential for reproducible hub selection across runs. Ascending node-id is a sensible canonical order.

**Minor:** The comment `// Descending by score; node id as the deterministic tie-break.` accurately describes the intent. No issues.

**One alternative to consider (not a blocker):** If `mu_fanin_hub_score` could ever return `f64::NAN` through a code path introduced in a future refactor (e.g. if the `exp > 0.0` guard were accidentally removed), the `unwrap_or(Equal)` would make NaN scores cluster together at an arbitrary position rather than sinking to the bottom. A more defensive future-proof pattern would be `unwrap_or(std::cmp::Ordering::Less)` (treats NaN as worst), but this is a minor style point and not a current correctness issue.

---

### Overall Assessment

The PR is in good shape. The `Pooled` default is the correct semantic choice for hub selection, the gate is consistent and defensively applied, and the sort is deterministic. The self-hardening pass (switching from mean-of-ratios to pooled as default) was the right call and is well-motivated by the variance argument.

**Suggested before merge:**
1. Add an assertion to `hub_agg_pooled_vs_mean_differ_on_heterogeneous_seeds` that the leak hub scores below chance (`< 1.0`) under both aggregators (close the specification gap).
2. Expand the `1e-9` constant's docstring to explain it's a sketch resolution limit, not a tunable domain threshold.
3. (Optional) Consider a comment noting the behavior when `total_mu` is very small.

No blockers. The core design is sound.

s243a (commented):
## Second Review: WAM-Rust μ-weighted fan-in hub ranking (PR #3275)

Focusing again on the three areas: **HubAgg enum / Pooled vs MeanOfRatios default**, **resolution gate**, and **sort**.

---

### 1. Is `Pooled` the right default for hub selection?

**Yes — and the argument is stronger than it might initially appear.**

The choice isn't just about variance; it's about what question you're actually asking during hub selection. `Pooled` computes `Σobs / ΣE` — the candidate's lift against the seed set treated as one aggregate mass. `MeanOfRatios` computes `(1/m) Σ obs_i/E_i` — the arithmetic mean of per-seed lifts, giving each seed equal weight regardless of cone size.

For *hub selection* specifically, you want to reward nodes that funnel large amounts of shared in-domain mass toward the query set. A seed with a big cone reconverging strongly is exactly the kind of signal you want to amplify, not suppress. `Pooled` does this naturally because `E_i = m_c * m_q / total_mu` scales with `m_q` — large seeds contribute proportionally more to `ΣE` and `Σobs`. `MeanOfRatios` treats a seed with 4 shared nodes identically in weight to one with 40. That's the wrong prior for this application.

There is a legitimate case for `MeanOfRatios`: if each seed represents an *independent concept query* of equal importance and you want the hub score to reflect breadth-of-coverage rather than mass-of-coverage, equal weighting is correct. The `HubAgg` enum is the right API decision — expose both, default to the lower-variance mass-weighted version, let callers opt in to equal-weight semantics.

**Design note:** The existing test `hub_agg_pooled_vs_mean_differ_on_heterogeneous_seeds` confirms the two diverge (pooled ≈ 3.09 vs mean ≈ 2.83 for the real hub) and both rank the real hub first. But the test doesn't assert the **leak hub stays below chance under heterogeneous seeds** for both aggregators. This is the more important bound — the 21× margin demonstrated in `rank_mu_fanin_hubs_picks_real_over_leak` is for symmetric seeds; under heterogeneous seeds, does `MeanOfRatios` maintain a similarly robust separation? Worth adding `assert!(pooled_leak < 1.0)` and `assert!(mean_leak < 1.0)` to close the spec.

**Numerical consistency note:** The self-harden comment says `pooled 3.09 vs mean 2.83` on heterogeneous seeds, but the doc says `ranked[0].1 ≈ 3.17` on symmetric seeds. These are from different test cases — fine — but the discrepancy (3.17 symmetric vs 3.09 heterogeneous for the real hub) is intuitive: the big seed (40 shared nodes) in the heterogeneous case contributes a different lift than the uniform 15-node per-seed case. This is correct behaviour, not a bug.

---

### 2. Resolution gate

The gate `floor = 1e-9 * total_mu` is applied to both the candidate and each seed, and the logic is correct. A few observations:

**Candidate gate first, seed gate inside loop** — the ordering matters for performance. A candidate with no real mass never enters the inner loop. This is the right structure.

**The `1e-9` multiplier** is borrowed from `sketch_mu_overlap_lift` for consistency, which is good. But the semantics differ slightly: in `sketch_mu_overlap_lift`, the gate prevents computing a lift when the cone mass is so small the sketch estimate is unreliable (high relative error). Here it serves the same purpose but is applied to *seeds*, not just the target. The doc comment says "matches `sketch_mu_overlap_lift`" which is accurate but incomplete — it would be clearer to say "below this threshold the MinHash mass estimate has relative error > 1/√k, making the per-seed lift unreliable."

**Interaction with `MeanOfRatios`:** When seeds are dropped by the gate, `count` decreases. For `MeanOfRatios`, this means the mean is over *fewer* seeds — a dropped sub-resolution seed is treated as if it doesn't exist, not as contributing lift = 0. For `Pooled`, the dropped seed simply adds nothing to `sum_obs` / `sum_exp`. These are different behaviours: `Pooled` implicitly treats dropped seeds as contributing 0 to the aggregate; `MeanOfRatios` ignores them entirely. Both are defensible, but the doc comment doesn't describe this asymmetry. Consider noting it explicitly.

**The `total_mu <= 0.0` guard** is correctly placed at the top. The edge case where `total_mu` is positive but extremely small (e.g. `1e-15`) is benign — `floor` collapses to near-zero, and all masses pass the gate, which is correct behaviour in a degenerate universe. Not a bug, but worth a doc note.

---

### 3. Sort

```rust
scored.sort_by(|a, b| {
    b.1.partial_cmp(&a.1)
        .unwrap_or(std::cmp::Ordering::Equal)
        .then(a.0.cmp(&b.0))
});
```

Correct. Three points:

1. **`unwrap_or(Equal)` for NaN** — since all return paths in `mu_fanin_hub_score` are finite (`0.0` guards cover all early exits, and `obs/exp` is only computed when `exp > 0.0`), NaN cannot appear. The `unwrap_or` is pure defence-in-depth. Acceptable, though `unwrap_or(Ordering::Less)` would be more principled (NaN → last rank) if you ever want to audit NaN-producing paths.

2. **Node-id ascending tie-break** is the right determinism strategy — node IDs are stable across runs (assuming deterministic graph construction), so the sort is fully reproducible.

3. **`sort_by` vs `sort_unstable_by`** — since the comparator is total (tie-break by node id ensures no two elements are truly equal), `sort_unstable_by` would be equivalent and marginally faster. Not a blocker, but worth considering for large candidate sets.

---

### Overall

The design is sound. `Pooled` is the correct default for hub selection semantics. The gate is consistent and correctly ordered. The sort is deterministic and correct.

**Actionable suggestions:**
1. Add `assert!(leak_hub_score < 1.0)` for both aggregators in `hub_agg_pooled_vs_mean_differ_on_heterogeneous_seeds` to fully close the leak-robustness specification under heterogeneous seeds.
2. Clarify the `1e-9` gate docstring: "below this threshold the MinHash mass estimate has relative error > 1/√k" rather than just "matches `sketch_mu_overlap_lift`".
3. Note the dropped-seed asymmetry between `Pooled` and `MeanOfRatios` in the `HubAgg` docstring.
4. (Minor) Consider `sort_unstable_by` if candidate sets are large.

s243a:
Thanks — both reviews confirmed `Pooled` as the right default and found no blockers. Addressed the actionable items in `bbe2f553`:

1. **Leak-hub `< 1.0` under heterogeneous seeds (top item, both reviewers).** `hub_agg_pooled_vs_mean_differ_on_heterogeneous_seeds` now asserts the leak hub stays *below chance* under **both** aggregators (pooled 0.15, mean 0.14), not merely out-ranked — closes the leak-robustness spec for the heterogeneous case.
2. **`1e-9` floor rationale.** Documented it as a **numerical negligibility floor** (essentially-zero mass — ghosts / fully-out cones), explicitly *distinct from and coarser than* the statistical sketch resolution (KMV RSE ≈ 1/√k, ~6% at k=256). I deliberately did **not** adopt the suggested "relative error > 1/√k" wording, because that conflates the numerical floor with the sketch error — they're different scales. Callers with small-but-meaningful cones should gate at a domain threshold upstream. Also noted the benign tiny-`total_mu` regime.
3. **Dilution semantics in `HubAgg`.** Documented — with one correction to the second review's framing. The review described a "dropped sub-resolution seed" asymmetry, but in the code a sub-resolution seed (`m_q < floor`) is `continue`d and excluded from **both** the sums *and* the count for **both** aggregators — no asymmetry there. The real asymmetry is for a *disjoint in-resolution* seed (`obs=0`, `m_q ≥ floor`): `Pooled` dilutes it mass-weighted (its `E` enters `ΣE`), `MeanOfRatios` dilutes it equally (a `0` over `m`). That's the inherent pooled-vs-mean difference, now documented accurately.
4. **`sort_unstable_by`.** Applied — the node-id tie-break makes the comparator total, so it's equivalent and cheaper. Added a note that all scores are finite by construction (the `unwrap_or` branch is unreachable), which is also why I left the `unwrap_or(Equal)` rather than switching to `Ordering::Less` — there's no NaN path to defend against.

Also reframed `MeanOfRatios` as *breadth-of-coverage* (independent equal-importance seeds) per the second review.

104/104 lib tests pass; Prolog guard passes.

---

PR #3278, opened 2026-06-19
WAM-Rust: μ-gated descendant cone — the in-domain-leak fix (prune at the membership frontier)

s243a:
Implements the idea raised in review discussion: μ-gating the descendant-cone *traversal* to fix the **in-domain leak** that plain μ-weighting cannot.

## The problem it fixes
`descendant_mu_mass` (down-weighting) sums μ over the *whole transitively-closed cone*. It cancels the **out-of-domain** leak (biology descendants contribute 0) but not the **in-domain** one: a high-μ node reachable *only through* an out-of-domain node is still counted — membership can't veto a bad *edge* while the traversal stays downward-closed. That's why `Matter` still ranked most-general in #3277.

## `descendant_mu_mass_gated(parents, mu, threshold) -> (gated_mass, gated_cone_size)`
**Prunes** the traversal at the membership frontier: descend into a child only while `μ ≥ threshold`, so a branch that falls out of the domain is cut and never explored. The cone becomes the in-domain *neighborhood*, not the transitive closure. `threshold ≤ 0` (or `μ≡1`) recovers `descendant_mu_mass`.

## Real-data result (`wikipedia_mu_gating_cuts_the_in_domain_leak`)
```
gate: Matter raw cone 8328 (purity 0.0046) → gated cone 48 mass 36.5 (purity 0.760)
```
At `μ≥0.3`, `Matter`'s cone collapses from **≈8328 nodes to 48** while *retaining* the in-domain mass — the leak is **cut, not just down-weighted** — purity jumps **0.005 → 0.76**, and the gated cones nest sensibly (`Matter` spans more in-domain mass than `Energy`), so a within-domain IC/hierarchy is legible again. (I expected the sparse μ to over-prune; it doesn't — the scored physics nodes form a connected high-μ subgraph.)

## The structural trade (named precisely by the idea's author)
The gated cone is **no longer downward-closed in the raw DAG** — a node past the frontier is dropped even though raw reachability puts it "under" the root. You trade the raw transitive-subset *closure* for **domain coherence**: the membership field bends the cone, and "what is under X" becomes an in-domain-geodesic question rather than flat reachability. It's the adaptive (membership-frontier) generalization of the project's bounded-depth scoping band-aid. (The gated relation is still transitive *among in-domain nodes*; what's lost is closure under the raw subset relation.)

## Density caveat (documented)
Absent ⇒ `μ=0`, so gating needs every in-domain *connector* scored, or a sparse μ prunes through unscored gaps. Here it doesn't bite; a sparser domain would need denser μ.

## Tests & docs
- `descendant_mu_mass_gated_cuts_in_domain_leak` (synthetic — a node reachable only via an out-of-domain hop is excluded; threshold 0 recovers the full cone).
- `wikipedia_mu_gating_cuts_the_in_domain_leak` (gated real data).
- §5d hook + measurement-doc update (supersedes the deferred distance-discounting note).

111/111 lib tests pass; Prolog template guard passes.

---

### DIFF
--- a/docs/design/WAM_RUST_CARET_REALDATA_MEASUREMENT_2026-06-18.md
+++ b/docs/design/WAM_RUST_CARET_REALDATA_MEASUREMENT_2026-06-18.md
@@ -316,11 +316,23 @@ the degenerate zero and reads as absolute generality against the whole graph. (`
 denominator for μ-weighted *similarity* — Lin/FaITH — where the offset would compress scores toward 1;
 that read-out is future work.)
 
-**What μ-weighting does and does not fix.** It cancels the **out-of-domain** leak (biology, geography
-descendants contribute `0`). It does **not** fix the **in-domain** leak: `Matter`'s cone wrongly
-contains other physics concepts (`Energy`, `Optics`) which are genuinely in-domain (`μ=1`), so they
-*are* counted — membership can't veto a bad *edge*. That residual is why `Matter` still ranks most-
-general, and why the robust real-data answer remains the **per-pair bidirectional bridge** (or
-bounded-depth scoping), not the global descendant cone. A distance-discounted descendant mass (down-
-weighting in-domain nodes reached only via long leak paths) is the candidate mechanism to attack it,
-and is deferred future work.
+**What μ-*weighting* does and does not fix — and what μ-*gating* does.** Down-weighting cancels the
+**out-of-domain** leak (biology, geography descendants contribute `0`) but does **not** fix the
+**in-domain** leak: it sums over the whole transitively-closed cone, so a high-`μ` node reachable only
+*through* an out-of-domain node is still counted — membership can't veto a bad *edge* while the
+traversal stays downward-closed.
+
+**The fix is μ-gating** (`descendant_mu_mass_gated`, `wikipedia_mu_gating_cuts_the_in_domain_leak`):
+**prune** the traversal at the membership frontier — descend into a child only while `μ ≥ threshold`.
+A branch that falls out of the domain is cut and never explored, so the cone becomes the in-domain
+*neighborhood* rather than the transitive closure. On the real graph, gating `Matter` at `μ ≥ 0.3`
+collapses its cone from **≈8328 nodes (purity `0.005`) to 48 nodes (purity `0.76`)** while *retaining*
+the in-domain mass — the leak is **cut, not just down-weighted** — and the gated cones nest sensibly
+(`Matter` spans more in-domain mass than `Energy`), so an in-domain IC/hierarchy becomes legible
+again. The price is structural and exactly as expected: the gated cone is **no longer downward-closed
+in the raw DAG** (you give up the raw transitive-subset *closure* for *domain coherence*) — the
+membership field bends the cone, and "what is under X" becomes an in-domain-geodesic question. It is
+the adaptive, membership-aware form of bounded-depth scoping. *Density caveat:* gating defaults absent
+nodes to `μ=0`, so it only stays connected if every in-domain *connector* is scored; here the scored
+physics nodes form a connected high-`μ` subgraph so it does not over-prune, but a sparser domain would
+need denser `μ`.
--- a/docs/design/WAM_RUST_GRAPH_FUNCTIONAL_SEMIRINGS.md
+++ b/docs/design/WAM_RUST_GRAPH_FUNCTIONAL_SEMIRINGS.md
@@ -885,6 +885,21 @@ calibration of the relatedness read-out.)
 >   structural proxy, child-cone coherence, partly detects the "union vs intersection" character of
 >   such nodes but cannot separate a diverse in-domain hub from a diverse leak bucket — only `μ` can,
 >   which is why purity needs the membership signal.)*
+> - **μ-gating — the in-domain-leak fix** (`descendant_mu_mass_gated`,
+>   `wikipedia_mu_gating_cuts_the_in_domain_leak`). Down-*weighting* sums over the whole transitively-
+>   closed cone, so it cancels the out-of-domain leak but not the **in-domain** one (a high-μ node
+>   reachable only *through* an out-of-domain node is still counted). μ-*gating* **prunes** the
+>   traversal at the membership frontier — descend into a child only while `μ ≥ threshold` — so a
+>   branch that falls out of the domain is cut and never explored. On the real graph this collapses
+>   `Matter`'s cone from `≈8328` nodes (purity `0.005`) to **48** (purity `0.76`) while keeping the
+>   in-domain mass: the leak is **cut**, not down-weighted, and a within-domain IC/hierarchy becomes
+>   legible again. The structural price (named exactly right by the contributor of the idea): the gated
+>   cone is **no longer downward-closed in the raw DAG** — you trade the raw transitive-subset
+>   *closure* for *domain coherence*, the membership field bending the cone so "what is under X" is an
+>   in-domain-geodesic question. It is the adaptive (membership-frontier) generalization of the project's
+>   bounded-depth scoping band-aid. (Caveat: absent ⇒ `μ=0`, so it needs every in-domain *connector*
+>   scored, or a sparse μ prunes through unscored gaps; the physics set is connected enough that it
+>   does not bite.)
 >
 > **Novel-contribution flag.** Resnik-style IC over a frequency null (Resnik 1995; Seco 2004's
 > intrinsic descendant-count form) is established; *graded* (fuzzy `μ ∈ [0,1]`) **partial admission**
--- a/templates/targets/rust_wam/boundary_cache.rs.mustache
+++ b/templates/targets/rust_wam/boundary_cache.rs.mustache
@@ -1748,13 +1748,27 @@ pub fn descendant_mu_mass(
 /// closed in the raw DAG** (a node past the frontier is dropped even though raw reachability puts it
 /// "under" the root): you trade the raw transitive-subset *closure* for **domain coherence**. It is
 /// the adaptive, membership-aware generalization of bounded-depth scoping — cut at the μ-frontier, not
-/// a fixed depth. `threshold ≤ 0` (or `μ ≡ 1`) recovers `descendant_mu_mass`'s full cone.
+/// a fixed depth. `threshold ≤ 0` (or `μ ≡ 1`) recovers `descendant_mu_mass`'s full cone (a *negative*
+/// threshold is silently equivalent to `0`, since `μ ≥ 0`). The gate is `≥`, so a node at *exactly*
+/// `μ = threshold` is **admitted**.
 ///
-/// **Density caveat:** absent nodes default to `μ = 0`, so they gate *out*. The frontier only stays
-/// coherent if every in-domain *connector* is scored; with a sparse `μ` it can prune through unscored
-/// gaps and disconnect the region. (On the real Wikipedia physics set the scored nodes form a
-/// connected high-`μ` subgraph, so this does not bite — `Matter`'s cone shrinks from ≈8000 to 48 with
-/// the in-domain mass retained — but a sparser domain would need denser `μ`.)
+/// **Well-defined, path-independent membership.** A node is in the gated cone iff *some* root→node path
+/// has every node above `threshold`; membership therefore does not depend on BFS order (the `μ ≥
+/// threshold` test is a node property, and the `seen` set only dedups). A node reachable by *both* an
+/// in-domain and an out-of-domain path is **included** — the in-domain path admits it.
+///
+/// **Cost & coverage.** Exact, `O(V·(V+E))` (a gated BFS per node) — the same shape as
+/// `descendant_mu_mass`; a *gated KMV sketch* is the scalable analogue (future work). Output keys are
+/// the nodes that appear in `parents` (as child or parent); a node present only in `mu` but in no edge
+/// (isolated) is **absent from the output**, exactly as in `descendant_mu_mass`.
+///
+/// **Density caveat (and mitigation):** absent nodes default to `μ = 0`, so they gate *out*. The
+/// frontier only stays coherent if every in-domain *connector* is scored; with a sparse `μ` it can
+/// prune through unscored gaps and disconnect the region. (On the real Wikipedia physics set the
+/// scored nodes form a connected high-`μ` subgraph, so this does not bite — `Matter`'s cone shrinks
+/// from ≈8000 to 48 with the in-domain mass retained.) If permissive traversal through *uncertain*
+/// regions is preferred over pruning, give unscored connectors a neutral `μ` (e.g. `0.5`) so they pass
+/// a `0.3` gate but contribute little mass.
 pub fn descendant_mu_mass_gated(
     parents: &std::collections::HashMap<u32, Vec<u32>>,
     mu: &std::collections::HashMap<u32, f64>,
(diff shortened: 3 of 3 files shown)


s243a (commented):
## Review: WAM-Rust μ-gated descendant cone (PR #3278)

Focusing on the core design: **the gating mechanism and its semantics**, **the structural trade-off (loss of downward-closure)**, **the density/sparsity caveat**, **implementation correctness**, and **test coverage**.

---

### 1. The gating design — correct approach to the in-domain leak problem

The problem statement is precise and the solution is the right one. `descendant_mu_mass` operates on the transitively-closed cone and down-weights by μ. That correctly cancels the *out-of-domain* leak (biology/geography descendants contribute 0), but it cannot veto a bad *edge*: a high-μ node reachable only *through* a low-μ connector is still included, because the traversal never stops at the out-of-domain node — it passes through it. The `Matter` real-data result (8328 nodes, purity 0.005) is exactly this failure mode; the in-domain physics nodes that are "under" Matter in raw reachability but accessible only via out-of-domain hops bloat the cone without being vetoed.

μ-gating is the correct fix: prune the BFS frontier at `μ(child) < threshold`, so out-of-domain nodes become walls, not pass-throughs. The in-domain neighborhood is explored; everything beyond the membership frontier is cut. The 8328 → 48 collapse with purity 0.005 → 0.76 is a striking validation.

**One semantic precision worth stating explicitly in the docstring:** the gate applies to the *child* before enqueueing, not to the *edge*. This means: a child with μ = 0 is pruned even if *its* children are in-domain. This is intentional (as the density caveat notes), but stating it explicitly as "we gate the node, not the edge" would help readers understand why unscored connectors matter.

---

### 2. The structural trade-off — loss of downward-closure

The docstring names this correctly: the gated cone is **no longer downward-closed in the raw DAG**. A node that is raw-reachable from the root but lies beyond the membership frontier is excluded, even if there is an in-domain path to it via a different route. This is the right trade, and the framing — "domain coherence over raw transitive-subset closure" — is accurate.

However, there is a subtlety worth flagging: **the gated cone is also not guaranteed to be downward-closed among in-domain nodes**. Consider a diamond: root → A (μ=1) → C (μ=1), and root → B (μ=0.1) → C (μ=1). With threshold=0.3, the path through B is pruned. But C is still reachable via A, so it *is* included. The gate is path-dependent only in the sense that *all* paths to a node must pass through the frontier for the node to be excluded. The `seen` set correctly handles this — once C is enqueued via A, it won't be re-processed via B. This is correct.

**Edge case worth testing:** what if C is reachable via both an in-domain path (μ ≥ threshold all the way) and an out-of-domain path (μ < threshold somewhere)? The current implementation correctly includes C (in-domain path wins, `seen.insert(c)` on the first encounter prevents re-enqueuing). A test asserting "if a node is reachable via any in-domain path, it is included" would close this spec.

---

### 3. The root is always kept — correct but undocumented in code

The docstring says "the root is always kept", which is correct: `seen.insert(t)` and `mass = muv(t)` happen unconditionally before the BFS loop, so the root contributes its μ regardless of its own membership value. This is the right semantic (you've asked for the cone *of* this node, so it's always in its own cone), but it means `descendant_mu_mass_gated` can return a non-zero mass for a node with `μ(root) = 0`. That's technically correct (the root contributes 0 mass if μ=0), but the cone size will still be ≥ 1. This is fine — just worth a one-line comment in the code since it differs from "threshold ≤ 0 recovers the full cone" in a subtle way: threshold=0 recovers the full *traversal*, but the root is always included regardless of threshold.

---

### 4. Implementation correctness

**`O(V · (V+E))` per call** — same complexity as `descendant_mu_mass`. Mentioned implicitly in the design doc but not stated in the `descendant_mu_mass_gated` docstring. Worth adding since the scalable alternative (`descendant_minhash_weighted`) exists specifically because this is expensive on large graphs; callers should know they're using the exact but slow version.

**`muv` closure with `unwrap_or(0.0)` for absent nodes** — correct implementation of the "absent ⇒ μ=0" policy, which means absent nodes are gated out (they fail `μ ≥ threshold` for any positive threshold). The density caveat correctly documents this. One follow-up: if a caller passes `mu` keys that use a *different* node ID space from `parents`, absent-node silencing will silently prune the entire graph. This is a pre-existing footgun in the codebase (same issue exists in `descendant_mu_mass`), so not a new bug, but the docstring could note "node IDs in `mu` and `parents` must share the same space".

**`seen` HashSet initialized fresh per root** — correct for per-node independent cone computation. No cross-contamination between roots. Same pattern as `descendant_mu_mass`, consistent.

**`debug_assert!`** on non-negative weights — consistent with `descendant_mu_mass`. Good.

---

### 5. Test coverage

**`descendant_mu_mass_gated_cuts_in_domain_leak` (synthetic):** correct and minimal. Tests the core claim (B reachable only via out-of-domain L is excluded), the directly-in-domain path is retained (A still included), and threshold=0 recovers the full cone. The graph is the exact pathological structure the fix targets.

**Missing test — the multi-path case:** as noted above, a node reachable via both an in-domain path and an out-of-domain path should be *included* (in-domain path wins). This is the complement of the current test. Suggest:
```
// Root 0 → A=1 (μ=1) → C=3 (μ=1)
// Root 0 → L=2 (μ=0) → C=3 (μ=1)
// C is reachable in-domain via A and out-of-domain via L.
// Gated cone of 0 should include C (in-domain path exists).
```

**`wikipedia_mu_gating_cuts_the_in_domain_leak` (real data):** conditional on environment variables, which means it won't run in CI by default. The assertions are well-chosen: raw cone > 5000, gated < 200, gated mass > 30, gated purity > 0.5, raw purity < 0.05, and the nesting order (Matter > Energy). The nesting assertion is particularly valuable — it checks that the gated cones preserve a meaningful in-domain generality order, not just that the numbers improve in isolation.

**Threshold boundary:** the test uses threshold=0.3. Consider also asserting behaviour at the boundary `threshold = min(mu.values())` (all nodes pass) and `threshold = max(mu.values()) + ε` (all nodes except root are pruned — cone size = 1 for every node). These corner cases are cheap to add and would catch regression if the gate logic is modified.

---

### 6. Documentation

Both doc updates (`WAM_RUST_GRAPH_FUNCTIONAL_SEMIRINGS.md` and `WAM_RUST_CARET_REALDATA_MEASUREMENT_2026-06-18.md`) are clear and accurate. The reframing from "distance-discounting is deferred future work" to "μ-gating is the fix, here's the result" is the right update — the measurement doc now accurately reflects the state of the implementation.

The phrase "in-domain-geodesic question" in both docs is evocative and correct. It might be worth adding a one-sentence clarification that "geodesic" here means shortest *in-domain* path (all edges on the path satisfy μ ≥ threshold), not shortest path in the raw DAG, for readers encountering this terminology without the surrounding context.

---

### Overall

The PR correctly identifies and fixes the in-domain leak. The implementation is clean, consistent with the existing codebase, and the real-data result validates the approach dramatically. The structural trade-off (loss of raw downward-closure) is properly named and documented.

**Actionable before merge:**
1. Add a test for the multi-path case (node reachable via both in-domain and out-of-domain paths should be included).
2. Add `O(V·(V+E))` complexity note to `descendant_mu_mass_gated` docstring.
3. (Optional) Add threshold boundary tests (all-pass and all-prune corners).
4. (Optional) Clarify "gates the node, not the edge" in the implementation comment to explain the unscored-connector sensitivity.

No blockers. The core design is correct and the real-data result speaks for itself.

s243a (commented):
## Second Review: WAM-Rust μ-gated descendant cone (PR #3278)

A fresh pass focusing on design correctness, implementation edge cases, test completeness, and documentation.

---

### 1. Core design: gating vs. weighting — is this the right abstraction?

Yes. The distinction the PR draws is fundamental:

- **Weighting** (`descendant_mu_mass`): traverse the full transitively-closed cone, multiply each node's contribution by μ. Correct for out-of-domain suppression, but the traversal itself is unbounded — you still *visit* every node, including high-μ nodes accessible only via out-of-domain hops.
- **Gating** (`descendant_mu_mass_gated`): stop the traversal at the membership frontier. A branch that exits the domain is cut; nodes beyond it are never visited regardless of their own μ.

The key claim: "membership can't veto a bad *edge* while the traversal stays downward-closed" is exactly right. Weighting operates on *values*; gating operates on *reachability*. These are orthogonal controls, and the in-domain leak requires the reachability fix, not a stronger value penalty.

**Interaction worth noting:** gating and weighting compose naturally — `descendant_mu_mass_gated` still sums μ values over the pruned cone, so it's both gated *and* weighted. The returned `(mass, size)` pair gives you the weighted IC signal (`mass`) and the structural size (`size`), with `mass/size` as gated purity. This is a well-chosen return type.

---

### 2. The BFS implementation — one latent ordering issue

```rust
for &c in cs {
    if muv(c) >= threshold && seen.insert(c) { mass += muv(c); q.push_back(c); }
}
```

This is correct for the semantics: gate on child's μ before enqueueing, accumulate mass on first visit, `seen` prevents double-counting in diamond topologies. The `seen.insert(c)` short-circuit means the gate check happens before the set check — a sub-threshold child is never inserted into `seen` even if it would otherwise be revisited. This is correct (a pruned node should stay pruned), but it has a subtle implication:

**If a node `C` is below threshold on one path but would have been reached above threshold via a different ordering of the BFS queue**, it's still excluded. But this can't happen: `muv(c)` is a property of the node, not the path. The gate is `μ(child) ≥ threshold`, which is path-independent. So BFS order doesn't affect the gated cone membership. ✓

**However:** `muv(c)` is called twice for each child that passes the gate — once in the `if` condition and once in `mass += muv(c)`. Since `muv` is a simple `HashMap::get` lookup with `unwrap_or(0.0)`, this is not a correctness issue, but it's a minor efficiency note. A `let mu_c = muv(c)` would make this explicit and slightly faster on large cones.

---

### 3. The `children` map construction — potential missing nodes

```rust
for (&c, ps) in parents {
    nodes.insert(c);
    for &p in ps { nodes.insert(p); children.entry(p).or_default().push(c); }
}
```

Nodes that appear only as parents (roots with no entry in `parents` as a child) are added to `nodes` via the inner loop but get no `children` entry — their children list is absent and `children.get(&x)` returns `None`, so the BFS correctly terminates at them (leaf nodes in the traversal). This is correct.

**Edge case:** a node that appears in `mu` but not in `parents` at all (not as child, not as parent) is silently absent from `nodes` and thus absent from the output `out`. If a caller expects every key in `mu` to appear in the result, this is a silent gap. The existing `descendant_mu_mass` has the same behaviour, so this is consistent — but worth a docstring note: "nodes not referenced in `parents` (neither as child nor as parent) do not appear in the output".

---

### 4. `threshold ≤ 0` recovery path

The docstring states `threshold ≤ 0` recovers `descendant_mu_mass`'s full cone. The test verifies this for threshold=0. The implementation gates on `muv(c) >= threshold`; since `muv` returns `unwrap_or(0.0)`, any node with absent or zero μ satisfies `0.0 >= 0.0 = true`. So threshold=0 correctly passes all nodes through. ✓

**Edge case with negative threshold:** `muv(c) >= -1.0` is trivially true for all non-negative μ values. Since `debug_assert!` enforces `μ ≥ 0`, a negative threshold always passes every node. The docstring says "threshold ≤ 0 recovers the full cone" without specifying that negative values are treated identically to 0. Fine in practice, but a caller passing `threshold = -0.1` "by mistake" would get the same result as threshold=0, which is probably the right silent behaviour. Acceptable.

---

### 5. Test coverage gaps

**Missing: multi-path diamond where one path is in-domain, one is out-of-domain.** The current synthetic test has B reachable *only* via the out-of-domain L — the pruned case. The complement — C reachable via both in-domain A and out-of-domain L — is not tested. C should be *included* because the in-domain path exists. This is the most important missing test because it validates the path-independence property of the gate:

```
// 0 → A=1 (μ=1) → C=3 (μ=1)
// 0 → L=2 (μ=0) → C=3 (μ=1)
// Gated cone of 0 at threshold=0.3: should contain {0, A=1, C=3} (L pruned, C still reached via A)
assert_eq!(gated[&0].1, 3);
assert!((gated[&0].0 - 3.0).abs() < 1e-9);
```

**Missing: isolated node (appears in `mu`, not in `parents`).** Verify the function doesn't panic and that the node is absent or present in `out` as documented.

**Missing: threshold exactly at a node's μ value.** `muv(c) >= threshold` is `>=` (inclusive). A node with `μ = threshold` is included. Add a test asserting this boundary: with `μ(child) = 0.3` and `threshold = 0.3`, the child is included.

**The Wikipedia real-data test** is well-structured but env-var-gated so it won't run in standard CI. The nesting assertion (`gated[matter].0 > gated[energy].0`) is especially valuable. Consider extracting the core assertions into a constant-data snapshot test so the logical properties are always verified even without the real data files.

---

### 6. Return type: `(f64, u32)` vs a named struct

`HashMap<u32, (f64, u32)>` where the tuple is `(gated_mass, gated_cone_size)` works, but a named struct (`GatedCone { mass: f64, size: u32 }`) would make call sites self-documenting. The existing `descendant_mu_mass` returns `HashMap<u32, f64>` (just mass), so the new return type is richer — callers accessing `.0` and `.1` have no local documentation of which is which. This is a non-blocking style note, consistent with the project's pattern of using tuples for paired metrics.

---

### 7. Documentation quality

Both doc updates are accurate and clear. The measurement doc correctly supersedes the "distance-discounting is future work" note with the actual fix. The phrase "in-domain-geodesic question" is precise.

**One small addition:** the docs mention the density caveat ("absent ⇒ μ=0, so unscored connectors get pruned"). It would be worth adding a one-sentence mitigation: "callers can assign a neutral μ (e.g. 0.5) to unscored nodes if they want permissive traversal through uncertain regions rather than pruning". This turns the caveat into actionable guidance.

---

### Overall

The design is correct and the real-data result validates it decisively. The implementation is clean. No blockers.

**Actionable before merge:**
1. Add the multi-path diamond test (in-domain path wins over out-of-domain path to same node).
2. Add the `μ = threshold` boundary test (inclusive gate).
3. Cache `muv(c)` in a local `let` to avoid the double lookup inside the gate condition.
4. (Optional) Note in docstring that nodes absent from `parents` are absent from output.
5. (Optional) Add mitigation guidance for the density caveat: assign neutral μ to unscored connectors if permissive traversal is desired.

s243a:
Thanks — both reviews confirmed the design (down-weighting vs gating distinction, the downward-closure trade) and found no blockers. Addressed all the test/doc/style items in `f7d83a6c`:

**Tests** — new `descendant_mu_mass_gated_corner_cases`:
- **Multi-path** (flagged by both): a node reachable via *both* an in-domain and a (pruned) out-of-domain path is **included** — confirms membership is "exists an all-in-domain path" and is path-independent, exactly as the second review's BFS-ordering analysis concluded (the `seen` set only dedups). This is the complement of the original synthetic test.
- **`μ == threshold`** is admitted (gate is `≥`).
- **All-pruned** (`threshold > max μ`) ⇒ cone = `{root}`. (All-pass / `threshold ≤ 0` was already covered.)
- **Isolated node** (in `μ`, absent from edges) ⇒ absent from output, no panic.

**Docs** — added the `O(V·(V+E))` complexity note (with the gated-sketch scalable analogue flagged as future work, since `descendant_minhash_weighted` has no gated form yet); the path-independence / well-defined-membership note; negative-threshold ≡ 0; the in-`μ`-not-in-`parents` ⇒ absent-from-output behavior; and the density-caveat **mitigation** the second review suggested — give unscored connectors a neutral `μ` (~0.5) when permissive traversal through uncertain regions is preferred over pruning.

**Style** — hoisted the double `muv(c)` lookup into `let mu_c = muv(c)`.

One non-blocking item I left as-is: the `(f64, u32)` return vs a named `GatedCone` struct. The doc states `(gated_mass, gated_cone_size)` and the call sites destructure with names (`let (g_mass, g_size) = …`), so the tuple reads clearly without the extra type — keeping it lean. Easy to promote to a struct later if it grows a third field.

112/112 lib tests pass; Prolog guard passes.

---

PR #3280, opened 2026-06-19
prototype: μ-cosine embedding — dense μ from learned vectors (separate-project branch)

s243a:
Prototypes the embedding/transformer approach you sketched for generating **dense `μ` from learned category vectors** — so the μ-gated cone / similarity machinery can run on domains where only a sparse set of categories is LLM-scored. Python/ML, **separate from the Rust/Prolog core**, on a branch per your "prototype it on a branch."

## The idea (your two-vector point)
`μ(X | root)` is relational — the **cosine between a learned vector for X and one for the domain root** — so it's two vectors with an interaction, not a per-node scalar, and the same node embeddings serve any domain (swap the root). This also unifies μ with the similarity work: `μ(X | root)` is `Lin(X, root)` measured in semantic space.

## Proven objective (`train_cosine_mu.py`, runnable, pure stdlib)
Learns per-category vectors so `cos(v_X, v_Physics) ≈ μ_X` on the **real 90-node Haiku fixture**, with your **distance-biased sampling** (nearer categories sampled more) and a closed-form, finite-difference-checked cosine gradient:
```
initial  MSE 0.41   corr(cos, μ) +0.05
final    MSE 0.0002 corr(cos, μ) +1.00
  μ 0.40 → cos +0.40  Arsenic_compounds
  μ 0.90 → cos +0.90  Atoms
```
So μ is faithfully expressible as the cosine of learned vectors. (16 dims for 90 nodes over-fits exactly — fine as a proof of the *objective*; *generalization* via the transformer is the separate project.)

## Architecture scaffold (`mu_transformer.py`, forward pass, full design scale)
Your exact sizing: `d_model=384`, `n_layers` configurable (1 default), MiniLM-init lookup hook, single-token MHA (degenerates to linear; head structure kept so a future multi-token input gives real attention) + FFN + cosine head. One sizing note: **384 isn't divisible by 20** (its divisors near there are 16 and 24), so I defaulted `n_heads=16` (closest to your target, 24 dims/head) and documented it. `python3 mu_transformer.py` runs the forward at full scale.

## Honest constraints (documented in the README)
No numpy/torch/network here, so: MiniLM init falls back to random (HF egress blocked), full-encoder *training* is the separate project (needs numpy/torch + the weights + more/pairwise μ labels). This branch **proves the objective and pins the architecture**; it doesn't train the real model.

## Why it matters to the core
Dense μ for *all* categories removes the density caveat on `descendant_mu_mass_gated` (no more pruning through unscored connectors) and feeds `gated_ic` / `lin_from_ic` directly — same μ map, just dense and domain-swappable.

---


s243a (commented):
## Review: μ-cosine embedding prototype (PR #3280)

This is a prototype/separate-project branch, so the review focuses on: **the objective and whether it's correctly proven**, **the architecture decisions and their rationale**, **the pair-generation design**, **implementation correctness**, and **the integration path back to the Rust core**.

---

### 1. The core hypothesis: μ(X | root) = cos(encode(X), encode(root))

The framing is well-chosen and the proof-of-objective in `train_cosine_mu.py` does what it claims. The key insight — that μ is *relational* (it depends on both X and the domain root) and therefore must be a two-vector interaction, not a per-node scalar — is correct and important. This means:

- The same encoder serves any domain just by swapping the root vector. No re-scoring per domain.
- `μ(X | root) = cos(encode(X), encode(root))` is structurally identical to `Lin(X, root)` in semantic space — the unification with the similarity work is real, not just metaphorical. Lin's IC-based similarity is `2·IC(LCS) / (IC(X) + IC(Y))`; if IC is computed from μ-gated cones and μ comes from cosine embeddings, the whole stack becomes self-consistent.

**The overfitting is the right kind.** 16 dims for 90 nodes over-fits exactly — but that's the proof-of-objective point: the representational capacity exists. The question for the separate project is generalisation to unseen categories, which requires MiniLM init + more labels. This is correctly documented.

**One precision to add:** the cosine of two arbitrary vectors lies in `[-1, 1]`, but μ is defined on `[0, 1]`. The current training works on the physics fixture where all μ ≥ 0, but a general-domain μ could legitimately approach 0 for out-of-domain categories, which cosine can represent as a near-zero or slightly negative value. The README should clarify the intended handling: either clamp `cos → max(0, cos)` at inference time, or accept that negative cosine scores are valid "below-domain" signals (which the μ-gating treats as `μ = 0` anyway since the gate checks `μ ≥ threshold`). Neither interpretation is wrong, but the caller needs to know.

---

### 2. Architecture decision: MLP, not transformer

The reasoning is correct. "Self-attention over one token has softmax ≡ 1" is the precise statement: with a single input token, the attention weights are trivially 1.0 (there's nothing else to attend to), so the Q/K projections produce dead weight and the V/O path collapses to a single linear transformation. The region-dependent computation needed for "different μ behaviour in different parts of semantic space" is the MLP nonlinearity, not attention. This is not a simplification — it's the right architectural choice.

The MoE gate deserves a note: with `n_experts > 1`, the `FeedForward` computes `softmax(Wx)` to assign a per-expert weight and then does a **weighted sum of all expert outputs**. This is *soft* MoE (all experts compute, weights blend). This is correct and interpretable (the gate weights are inspectable), but at inference time for a production model you'd typically want **hard or top-k routing** to avoid paying the cost of all experts. The README documents this correctly as "interpretable region routing" — just note somewhere that this is the soft variant and top-k is the efficiency upgrade if the model scales.

**The neighbour-context fold-in** (`encode(cid, neighbors)` as a fixed mean-pool) is a clean design: it's cheap, it provides a rudimentary structural signal, and the doc correctly identifies that *learning* those pooling weights is exactly attention-over-neighbours — the documented future upgrade. The current implementation correctly averages rather than concatenating: `x = (embed(cid) + Σ embed(n)) / (1 + |neighbors|)`. This is right but note it gives equal weight to each neighbour; if degree is heterogeneous, you may want to pass only `k` closest neighbours rather than all, since the number of neighbours affects the average. For the prototype this is fine.

---

### 3. `train_cosine_mu.py` — implementation correctness

The closed-form cosine gradient is:
\[ \frac{\partial}{\partial a} \cos(a,b) = \frac{b}{\|a\|\|b\|} - \frac{\cos(a,b) \cdot a}{\|a\|^2} \]
This is correct (standard projection-onto-hyperplane form), and the finite-difference check at startup validates it. Good.

**The anchor also moves.** Both `vecs[n]` and `vecs[ANCHOR]` are updated per step:
```python
a[i] -= args.lr * 2 * err * ga[i]
b[i] -= args.lr * 2 * err * gb[i]  # the anchor also moves
```
This is intentional for the proof-of-objective (you have no frozen initialisation here, so the anchor is just another parameter). For the full encoder, this corresponds to the root category also getting an updated embedding — which is correct if the root is in the same embedding space. Worth noting in the handoff doc that in the production model the root vector is the encoder's output for the root category, not a separate parameter, so this is naturally handled.

**Distance-biased sampling:** `w(n) = (1/(1+dist))^bias`. This is clean and the `--dist-bias` knob correctly interpolates: 0 = uniform, higher = nearer-biased. One subtlety: nodes not reachable from the anchor get `dist = 99` (the fallback), giving them weight `(1/100)^bias ≈ 0`. This is the right behaviour (out-of-graph nodes shouldn't dominate training), but it also means the distant out-of-domain negatives are effectively under-sampled. For the proof-of-objective on the 90-node physics fixture this is fine (all nodes are physics-adjacent); for the full training with pairwise labels across varied anchors you'll want to explicitly control the negative sampling separately from the distance-biased positive sampling — which is exactly what `gen_mu_pairs.py` does (separate positive mesh and uniform negative noise).

**Gradient update is SGD with a fixed learning rate** — no momentum, no scheduling. For the proof-of-objective at 90 nodes × 16 dims this is completely fine. For the production model this will need Adam, not SGD, especially given the sparse embedding table (where the effective learning rate per row varies widely with update frequency).

---

### 4. `gen_mu_pairs.py` — pair generation design

The design mirrors word2vec SGNS correctly:
- **Positives:** random-walk-with-restart (hub-down-weighted, geometric walk length, growing frontier mesh). The hub-down-weighting via `1/deg^β` is the right analogy to word2vec's frequent-word subsampling — avoids `Container_categories` / `Matter` / `Physical_objects` becoming walk sinks that teleport across the graph.
- **Negatives:** domain node × uniform-random category. Since the domain is a small fraction of the 8k-node graph, a uniform random draw is indeed reliably out-of-domain. The 5:1 negative ratio is within the Mikolov recommended range (5–20).

**One correctness note:** the negative sampling excludes `b in adj.get(a, ())` (direct neighbours). This is correct — direct graph neighbours should not be unconditional negatives since they may be related. However, it only excludes *direct* neighbours. Two-hop neighbours could also be closely related (especially in a category graph). For the purpose of this prototype this is fine; a stricter version would use BFS to exclude up to `k`-hop neighbours as negatives.

**The `score_stub` is the right pattern** — a clearly-named `NotImplementedError` with the exact prompt format and the LLM-budget warning. The handoff note ("Spends LLM budget — confirm with the user first") is exactly right. This keeps the budget-spending step explicit and auditable.

**`mu_pairs.tsv` is gitignored** — correct. The sample file (`mu_pairs.sample.tsv`) is a committed 30-line example showing the output format. The sample looks correct: positives include `Physics → Matter` (walk_len=1, clearly in-domain), `Society → Justice` (walk_len=3, boundary-band territory), and negatives are random cross-domain pairs like `Electromagnetism → 1980s_in_television`. Good representative coverage.

---

### 5. `mu_encoder.py` — forward-pass implementation

The MLP/Block/MuEncoder stack is straightforward and faithfully implements the documented architecture. A few notes:

**`layernorm` is computed from scratch per call** — with pure Python lists this is O(d) and correct. For the production port to PyTorch, this becomes `nn.LayerNorm`, which is optimised. No issue here.

**The `embed` method caches random vectors into `self.init`** for unseen IDs. This is fine for the prototype (the random fallback is documented), but it means two separate `MuEncoder` instances will produce different random embeddings for the same unseen ID. For the production model, the embedding table is a learned `nn.Embedding` keyed by a stable integer ID — document that the ID space must be consistent across the encoder and the graph.

**`mu()` normalisation guard:** `na = math.sqrt(...) or 1e-12` — using Python's `or` to substitute a near-zero value when the norm is exactly 0. This is correct (prevents division by zero), but semantically, an all-zero vector shouldn't occur if `rand_vec` initialises from a Gaussian. The guard is correct defensive programming.

---

### 6. Integration path back to the Rust core

The README's integration section is clear and correct. The exact format guidance — `name<TAB>μ` with `#`-comment header, matching the existing fixture format so the Rust loaders consume it unchanged — is the right design. The three integration points (consumers in `boundary_cache.rs.mustache`, the graph file, the sparse μ fixture) are precisely identified.

**The integration closes the density caveat from PR #3278.** `descendant_mu_mass_gated` currently requires every in-domain *connector* to be scored or it prunes through unscored gaps. Dense μ from the encoder covers all categories, so the unscored-gap problem disappears. The doc correctly captures this link.

**One gap:** the README says to format the dense μ output as `name<TAB>μ` — but the Rust loaders use *internal integer node IDs* (`HashMap<u32, f64>`), not category names. The name-to-id mapping lives in the TSV-loading code of the individual tests (e.g. `wikipedia_mu_gating_cuts_the_in_domain_leak` in PR #3278 does its own interning). A production integration would either: (a) emit the dense μ as `name<TAB>μ` and rely on the same interning logic, or (b) emit it as `id<TAB>μ` directly. Option (a) is fine if the name space is stable; document explicitly that category names must exactly match the `child`/`parent` strings in `category_parent.tsv`.

---

### Overall

This is a well-structured prototype that correctly proves the objective, documents the architecture decisions with honest rationale, and provides a clear handoff path for the production ML project. The key contributions are:
1. Proven: μ is faithfully expressible as a cosine of learned vectors (corr 1.00 on real data).
2. Proven: the right architecture is MLP/MoE, not a transformer (single-token attention degenerates).
3. Done: the pair generator is implemented and the scoring stub is clearly gated on user confirmation.
4. Done: the integration path to the Rust core is precisely documented.

**Actionable before merging to main (or designating as the separate-project seed commit):**
1. Add a one-sentence note in README about μ ∈ [-1, 1] vs. [0, 1]: how negative cosine scores are handled at inference (clamp or treat as below-domain).
2. Note the soft-MoE vs. top-k routing distinction in the architecture section — soft is correct for the prototype, top-k is the efficiency upgrade.
3. Clarify in the handoff steps that the embedding table ID space must match the `category_parent.tsv` name strings exactly when wiring back to the Rust core.
4. (Optional) Note that the production training loop will need Adam (not SGD) for the sparse embedding table.

s243a (commented):
## Second Review: μ-cosine embedding prototype (PR #3280)

A fresh pass with different emphasis: **the statistical validity of the training setup**, **the pair generator's coverage and bias properties**, **the MoE gate correctness under soft routing**, **the gradient implementation edge cases**, and **the production handoff completeness**.

---

### 1. Training objective validity — what the corr=1.00 result actually proves

The result is genuine but the interpretation needs one careful qualification. With 16 dims and 90 nodes, the embedding matrix has `91 × 16 = 1456` parameters fitting 90 scalar targets — over-parameterised by `16×`. The SGD converges to a global minimum (corr → 1.00) not because the cosine objective is inherently well-suited to μ, but because the system has far more degrees of freedom than constraints. This is the correct design for a proof-of-objective, but it means the result proves *representational capacity*, not *sample efficiency* or *inductive bias*.

The real test — which the README correctly defers — is whether the encoder generalises from ~200 labelled pairs to predict μ for unseen categories. For that, MiniLM initialisation is load-bearing: without a strong warm-start, the encoder has no signal about the semantic structure of unseen categories. The README correctly lists MiniLM init as step 1 in the handoff and notes it's blocked by HF egress. This ordering is right.

**One concrete risk to document:** the distance-biased sampler (nearer categories sampled more) means the training distribution is *not* representative of the full μ range. Categories at graph-distance 1–2 from `Physics` are over-represented; distant out-of-domain categories (the true negatives) are under-sampled. For the proof-of-objective on 90 physics-adjacent nodes this is fine — there are no true hard negatives in the fixture. But for the production pairwise training, the `gen_mu_pairs.py` design correctly separates this: positives use the biased walk mesh, negatives use uniform sampling. Document explicitly that the `--dist-bias` knob in `train_cosine_mu.py` should be set to 0 (uniform) when training on pairwise μ that already includes explicit negatives — otherwise the sampler double-penalises distant pairs.

---

### 2. Cosine gradient — the anchor-update asymmetry

The update loop:
```python
ga = cos_grad(a, b)  # ∂cos/∂a
gb = cos_grad(b, a)  # ∂cos/∂b  (note: cos_grad(b,a) ≠ cos_grad(a,b) in general)
a[i] -= args.lr * 2 * err * ga[i]
b[i] -= args.lr * 2 * err * gb[i]
```

`cos_grad(b, a)` correctly computes `∂cos(a,b)/∂b` by symmetry of the formula: `∂cos(a,b)/∂b = a/(‖a‖‖b‖) - cos(a,b)·b/‖b‖²`. The call `cos_grad(b, a)` passes `b` as the "a" argument and `a` as the "b" argument, which gives `a/(‖b‖‖a‖) - cos(b,a)·b/‖b‖²` — which is exactly `∂cos(a,b)/∂b` since `cos(b,a) = cos(a,b)`. So this is correct. ✓

**However:** the finite-difference check only validates `∂cos/∂a` (the check perturbs `a`, not `b`). The `gb` path is not explicitly verified. This is a minor gap — the formula is symmetric and correct by inspection — but for a component that goes into the handoff spec, adding `cos_grad(b, a)` to the finite-difference check is a one-liner and closes the verification.

---

### 3. `gen_mu_pairs.py` — coverage and collision properties

**The frontier growth is unbounded.** Every node on every walk path is added to `frontier`:
```python
for n in path[1:]:
    if n not in frontier:
        frontier.append(n)
```
With `--n-positives 200` and walks up to length 8, the frontier can grow to hundreds of nodes quickly. The sampler then draws `start` uniformly from `frontier` (unless it restarts at a seed). This means the positive distribution drifts toward whatever the walk paths happen to explore, not a controlled mesh around the seeds. On an 8k-node graph with `stop_prob=0.4` (mean walk length `1/0.4 = 2.5` steps) this is likely fine in practice — walks stay short and local. But the README's claim that this is a "mesh grown around a few roots" is only approximately true for small `n_positives`; with large `n_positives` it becomes a random walk from an ever-expanding frontier.

**The `tries < n_positives * 50` budget is asymmetric.** For positives the budget is `200 * 50 = 10000` tries. For negatives it's `n_neg_target * 50 = 1000 * 50 = 50000` tries. The negative budget is much larger, but negative collision (`key in pairs`) is rare on an 8k-node graph (`8000² = 64M` possible pairs, sampling 1000 = 0.0015% collision probability). In practice the negative loop will always fill. No bug, just an over-conservative budget for negatives.

**The `sorted()` canonicalisation of pair keys** (`key = tuple(sorted((a, b)))`) treats `(a, b)` and `(b, a)` as the same pair. For an undirected μ (where μ(a|b) ≈ μ(b|a)), this is correct. If the full training eventually uses asymmetric pairwise μ(a|root) for varied roots, this may need revisiting — but for the current symmetric SGNS design it's right.

---

### 4. `mu_encoder.py` — MoE gate and soft routing

The gate:
```python
g = softmax(matvec(self.gate, x))   # shape: n_experts
outs = [e(x) for e in self.experts] # each expert: shape d_model
return [sum(g[k] * outs[k][i] for k in range(len(outs))) for i in range(len(x))]
```

This is a correct soft MoE: all experts compute, outputs are a convex combination weighted by the gate. The gate is linear (`matvec(self.gate, x)`) followed by softmax — a standard "noisy top-k" precursor without the noise. A few properties:

- **Expert collapse risk:** with soft routing and no auxiliary load-balancing loss, all weight can concentrate on one expert (the gate collapses to argmax). For the prototype this doesn't matter. For production, add a load-balancing auxiliary loss (standard in sparse MoE literature: `aux_loss = n_experts * Σ_i f_i * P_i` where `f_i` is the fraction of tokens routed to expert `i` and `P_i` is the mean gate probability). Document this as a production requirement if `n_experts > 1`.
- **Gate weight is `d_model × n_experts`** — for `n_experts=4, d_model=384`, that's 1536 parameters for the gate. Cheap. ✓
- **All experts compute even with soft routing** — O(n_experts × d_ff × d_model) per forward. For the prototype's batch sizes this is fine. At production scale with `n_experts=8` this is 8× the MLP cost with no sparsity savings unless you switch to hard routing.

---

### 5. The `μ ∈ [-1, 1]` vs `[0, 1]` range issue — concrete impact

The README notes that μ is trained on `[0, 1]` scores from the Haiku fixture. With random init and no range constraint, the cosine can go negative during training. In `train_cosine_mu.py` this is fine — the MSE loss will pull it toward the positive range. But at inference time, if a category is genuinely out-of-domain and the encoder outputs `cos(encode(X), encode(root)) = -0.3`, three things happen downstream:

1. `descendant_mu_mass_gated`: `muv(X) = -0.3`, `muv(X) >= threshold` (0.3) is `False` → pruned. **Correct behaviour** — out-of-domain gets pruned.
2. `descendant_mu_mass`: sums `μ(X) = -0.3` into the mass. **Incorrect** — a negative contribution reduces the parent's mass, which is semantically wrong.
3. `information_content_weighted_sketch`: `sketch_mu_mass` sums the μ-weighted KMV sketches. A negative weight breaks the mass estimate.

So the recommendation is: **clamp the encoder output to `[0, 1]` before feeding the Rust core**. This is a one-line post-processing step (`mu = max(0.0, min(1.0, cos(a, b)))`), but it needs to be in the integration section of the README, not just implied. The gating functions are safe with negative μ (they prune it), but the mass/sketch functions are not.

---

### 6. Production handoff completeness

The README's handoff steps are well-ordered. A few additions that would make it self-contained:

**Step 2 (score the pairs):** add the recommended batch size for the Haiku subagent. The existing fixtures suggest scoring ~90 nodes at a time; for 1200 pairs, batch into groups of 50–100 with the same prompt discipline. The `score_stub` has the format right but doesn't suggest batching.

**Step 3 (port to torch):** the handoff says "port `mu_encoder.py`'s forward to torch (it is a faithful spec)". Add that the embedding table should use `nn.Embedding` with `sparse=True` (sparse gradients) since only the rows touched by a batch will be updated — critical for the full 8k-category (or 1M-category Wikipedia scale) table. Already noted in the compute section but should be in the ordered steps.

**Step 4 (validate):** the validation criterion `μ(X|root) == Lin(X, root)` is the right convergence check. Add the tolerance: the two should agree within ~0.1 for a well-trained encoder on in-domain nodes. The `lin_from_ic` function in `boundary_cache.rs.mustache` is the Rust-side reference.

**Step 5 (emit dense μ):** add the name-matching requirement explicitly: the category names in the output TSV must exactly match the `child`/`parent` strings in `category_parent.tsv`, including capitalisation and underscore formatting (`Physical_objects`, not `Physical objects`). A mismatch silently produces `μ = 0` (absent → `unwrap_or(0.0)`) and activates the density caveat you just fixed.

---

### Overall

The prototype correctly proves the objective, makes the right architecture choice (MLP over transformer), and provides a clean handoff path. It's merge-ready as a prototype/seed commit for the separate project.

**Actionable before merge:**
1. Add `cos_grad(b, a)` to the finite-difference check (closes the gradient verification gap).
2. Add a clamp `max(0, min(1, cos))` in the integration section — mass/sketch functions in the Rust core do not tolerate negative μ.
3. Document that `--dist-bias` should be 0 when training on pairwise data that already includes explicit negatives.
4. Add to Step 5 of the handoff: category name strings must exactly match `category_parent.tsv` (case/underscore-sensitive) to avoid silent `μ=0` for unmatched names.
5. (Optional) Document the load-balancing auxiliary loss as a production requirement if `n_experts > 1`.

s243a:
Thanks — both reviews confirmed the design (objective sound, MLP-not-attention correct, SGNS sampler correct) and the items were all refinements. Addressed in `0293c32a`:

**The load-bearing finding (μ ∈ [-1,1] vs the Rust core's μ ≥ 0)** — caught precisely. Added `to_membership()` (clamp cosine → `[0,1]`) and documented *why* it's required at the emission step: `descendant_mu_mass` and `sketch_mu_mass` *sum* μ as mass and assume non-negative (a negative weight corrupts the mass / KMV estimate); only `descendant_mu_mass_gated` tolerates it. It's now a prominent "two silent-corruption guards" block in the Integration section.

**Name-matching (the other silent failure)** — documented: emitted category names must match `category_parent.tsv` exactly, or the Rust loaders' `unwrap_or(0.0)` makes them silently `μ=0`, re-activating the density caveat the project exists to close. Emit verbatim + assert coverage.

**Gradient check gap** — `finite_diff_check` now verifies *both* `∂cos/∂a` and `∂cos/∂b = cos_grad(b,a)` (the anchor-update path), not just the a-side.

**`--dist-bias` double-penalty** — good catch; documented to set it to 0 when training on `gen_mu_pairs.py` output, since those pairs already encode the distance preference via the explicit negatives.

**Over-parameterization framing** — sharpened to "capacity, not inductive bias" with the numbers (1,456 params for 90 targets, ~16×).

**MoE collapse** — documented the load-balancing aux loss (`n_experts · Σ fᵢ·Pᵢ`) as required for `n_experts > 1`, and top-k routing as the throughput upgrade.

**Frontier growth** — added `--max-frontier` so the mesh stays *around* the seeds; noted the stricter k-hop negative-exclusion option in `gen_mu_pairs.py`.

Also folded in the smaller notes (Adam-not-SGD for the sparse embedding table; soft-gate fine while prototyping). All three scripts still run; trainer corr 1.00.

One I confirmed rather than changed: the anchor vector moving during training is **intended** (the root is just another encoder output in production), as the first review noted.

---

PR #3356, opened 2026-06-26
mu_cosine: inferred-relation operator uncertainty (confidence tag + stochastic switch + theory)

s243a:
Treats an INFERRED relation's operator as uncertain rather than committing it,
and documents the theory.

## Foundation (shipped)
A relation is *tagged* when it comes from an exact section header, *inferred*
when it falls to the structural default (a typo'd/missing section) or a
title-match bridge. Carry the categorisation **confidence** through the pipeline
so the trainer can treat them differently.

- **`fuse_corpus.py`**: edges become `(a, b, rel, confidence)` — section-matched
  = 1.0 (tagged), structural fallback = 0.4 (inferred), title-bridge = 0.8,
  mindmap structure = 1.0. Emitted as a 4th `_edges.tsv` column.
- **`build_graded_round.py`**: carries confidence to a 10th `_pairs.tsv` column;
  reports the inferred count.
- **`train_mu_attention.py`**: `--infer-switch` stochastically switches an
  inferred `element_of` → `subcategory` (same μ, directional, but the WIKI
  operator) with probability `infer-subcat × container breadth` (a broad domain
  is likelier to hold subcategories than leaf elements). Off by default, A/B-able;
  tagged rows untouched.

On the two neighbourhoods: 400/2488 graded targets are inferred (mostly the
typo'd `Subtoipcs` sections at conf 0.4).

## Validation (warm-start fine-tune, GPU)
The switch **helps**: discrimination **89% → 97%** (35/36), the best yet, while
all operators hold — graded ELEM r+0.996, SYM held-out +0.831 (> control +0.726),
WIKI order-acc 99.8%, ELEM direction 100%.

## Theory — `DESIGN_inferred_operator_superposition.md`
The operator is a random variable from a joint posterior
`P(operator | μ, node_type, breadth)` — a superposition over candidate operators
plus noise, measured by the model's predicted μ (a self-training/EM loop).
- The posterior **factors**: μ separates membership (≈0.9) from associative
  (see_also/assoc); node-type/breadth separate `element_of` from `subcategory`
  (μ can't).
- The **noise decomposes** into four distinct sources: (a) μ-residual (posterior
  entropy), (b) out-of-set mass (true op not among candidates), (c)
  measurement/estimation uncertainty in μ, (d) model churn. (a,b) live in the
  operator posterior; (c,d) in the μ measurement.
- Soft **posterior-weighted operator loss** preferred over hard sampling
  (entropy = built-in noise).
- Alternatives A–F tabulated (blanket element_of/see_also → fixed-breadth v1 →
  hard sample → soft posterior loss → + noise decomposition; μ-band relabel as a
  special case).

## Notes
- No data committed; harvests/caches/models are gitignored/local.

### DIFF
--- a/prototypes/mu_cosine/DESIGN_inferred_operator_superposition.md
+++ b/prototypes/mu_cosine/DESIGN_inferred_operator_superposition.md
@@ -1,0 +1,96 @@
+# Inferred-relation operators as a μ-conditioned superposition — theory & alternatives
+
+When a relation is **tagged** (an exact section header — "Subtopics", "Subcategories", …) we know its operator.
+When it is **inferred** (a typo'd / missing section → structural fallback; or a title-match bridge) the
+operator is *uncertain*. Rather than commit an inferred relation to one operator, we treat the operator as a
+**random variable** and train under its distribution. This documents that model, what its noise actually
+captures, and the alternatives considered. Companion to `DESIGN_provenance_and_representation.md` (the
+`confidence`/inferred tag) and `REPORT_graded_round.md` (the harvest→categorise split).
+
+## 1. The model — operator as a random superposition
+
+For an inferred pair, the operator is drawn from a **joint posterior**
+
+```
+P(operator | μ, node_type, breadth)
+```
+
+— a *superposition* over the candidate operators (element_of, subcategory, see_also, … and an out-of-set
+"other"), weighted by that posterior, **plus noise**. The measurement is the model's **predicted** μ for the
+pair — NOT the inferred target. The target was the uncertain guess; the model's μ is the *evidence* that
+reconsiders it (a self-training / EM flavour). Warm-starting makes the measured μ trustworthy from step 0.
+
+Two realisations of "train under the distribution":
+- **Hard sampling** — draw one operator per step from the posterior (a stochastic switch). Simple; higher
+  gradient variance.
+- **Soft posterior-weighted loss** (preferred) — the *expectation* over operators,
+  `L = Σ_op P(op | μ, type, breadth) · mse_under(op)`. Lower variance; the posterior's spread *is* the noise
+  (a flat posterior automatically trains a mixture of operators; a peaked one trains essentially one).
+
+Hard sampling is the Monte-Carlo approximation of the soft loss.
+
+## 2. The posterior, and why it factors
+
+μ does **not** carry all the information, so the posterior factors along (roughly) orthogonal axes:
+
+- **μ → membership vs associative.** Tagged μ clusters: membership (`element_of`, `subcategory`) ≈ 0.9;
+  associative (`see_also` ≈ 0.4, `assoc` ≈ 0.3). So μ answers *"is this inferred see_also actually a
+  membership relation?"* — exactly the reconsideration that motivated this.
+- **node_type / breadth → element vs subcategory.** μ **cannot** separate `element_of` from `subcategory`
+  (both target ≈ 0.9). That axis comes from the endpoint **node-type** (page → element, category →
+  subcategory) and the container's **breadth** (a broad domain is likelier to hold subcategories than leaf
+  elements). This is what the v1 fixed-breadth switch approximated.
+
+So `P(op | μ, type, breadth) ∝ P(μ | op) · P(type | op) · prior(op | breadth)`, with `P(μ | op)` estimated
+**from the label data** — the model's predicted-μ distribution over the *tagged* examples of each relation —
+re-estimated every N steps as the model evolves. The tagged rows are the calibration set; the inferred rows
+are the things being classified.
+
+## 3. What the noise actually captures (it is not one thing)
+
+The "operator noise" is a sum of distinct uncertainties — keeping them separate is what makes the model
+honest rather than a fudge factor:
+
+| source | what it is | how it enters |
+|---|---|---|
+| **(a) μ-residual** | how much of the operator choice μ does **not** determine — the element↔subcategory axis, and within-band overlap | the **entropy** of the in-set posterior `H(P(op | μ,…))` — a flat posterior ⇒ more spread |
+| **(b) out-of-set mass** | the probability the true operator is **not in the candidate set** at all (an unenumerated relation) | a reserved "other" mass / floor in the posterior (open-world term) — never 0 |
+| **(c) measurement / estimation uncertainty** | μ itself is a noisy estimate (finite capacity, the pair's e5, ancestor sampling) | propagate a measurement width δ: integrate the posterior over μ±δ (blurs P(op|μ)) |
+| **(d) model churn** | μ drifts as the model trains; a posterior estimated at step t is stale at t+k | absorbed as non-stationarity — widen δ / re-estimate `P(μ|op)` more often / EMA the estimate |
+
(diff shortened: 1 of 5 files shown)


prototypes/mu_cosine/train_mu_attention.py
```diff
@@ -260,8 +261,30 @@ def train(args):
...
+    # WIKI operator) with probability ∝ the container's BREADTH (a broad domain is likelier to hold
+    # subcategories than leaf elements). This injects operator noise + models the element/subcategory
+    # ambiguity for untagged relations; TAGGED rows (conf 1.0) always keep their operator.
+    elem_breadth = {}                                    # container key → # element_of members (breadth proxy)
+    for r in graded_tr:
+        if r[4] == "element_of":
+            container = r[1] if r[2] >= 0.5 else r[0]
+            elem_breadth[container] = elem_breadth.get(container, 0) + 1
+
+    def graded_op(r):
```
s243a, line 296: **[high] Isolate the stochastic switch RNG**

`graded_op()` draws from the same `rng` used for batch sampling and training-time masking. With `--infer-switch` enabled, every switch decision consumes extra random numbers, so the switch-on vs switch-off A/B changes more than just the operator choice. Please use a dedicated RNG for switch decisions, ideally seeded from `args.seed` plus a fixed offset, or make the switch deterministic per `(node, root, relation, epoch/step)` so the control and treatment preserve the same sampling trajectory.

prototypes/mu_cosine/DESIGN_inferred_operator_superposition.md
```diff
@@ -0,0 +1,96 @@
...
+**random variable** and train under its distribution. This documents that model, what its noise actually
+captures, and the alternatives considered. Companion to `DESIGN_provenance_and_representation.md` (the
+`confidence`/inferred tag) and `REPORT_graded_round.md` (the harvest→categorise split).
+
+## 1. The model — operator as a random superposition
+
+For an inferred pair, the operator is drawn from a **joint posterior**
+
+```
+P(operator | μ, node_type, breadth)
```
s243a: **[medium] Terminology: relation vs operator**

The codebase distinguishes the relation (`element_of`, `subcategory`, `see_also`) from the operator/readout head (`ELEM`, `WIKI`, `SYM`). The uncertain latent here seems to be the relation, with the operator as a deterministic downstream mapping. Suggest rewriting this as `P(relation | μ, node_type, breadth)` plus a note on the relation-to-operator map, so it composes cleanly with the follow-up design doc, which already uses `P(relation | ...)`.

prototypes/mu_cosine/train_mu_attention.py
```diff
@@ -260,8 +261,30 @@ def train(args):
...
+        if args.infer_switch and len(r) > 9 and r[9] < 1.0 and r[3] == "ELEM" and r[4] == "element_of":
+            container = r[1] if r[2] >= 0.5 else r[0]
+            p = args.infer_subcat * min(1.0, elem_breadth.get(container, 0) / max(1, args.breadth_scale))
```
s243a: **[medium] Confidence magnitudes are recorded but not consumed**

The trainer currently gates on `conf < 1.0`, but the switch probability only uses `infer_subcat × breadth`; a `0.4` structural fallback and a `0.8/0.9` inferred row switch identically. Either incorporate confidence into the probability, e.g. `p = infer_subcat * breadth_factor * (1 - conf)`, or document that the 0.4/0.8/0.9 values are provenance labels rather than quantitative probabilities for v1.

prototypes/mu_cosine/fuse_corpus.py
```diff
@@ -60,16 +60,16 @@ def addn(corpus, slug, ntype, title):
         if len(c) < 3:
             continue
         mk = addn("mm", c[0], c[1] if len(c) > 1 else "mindmap_node", c[2] if len(c) > 2 else c[0])
-        if len(c) >= 6 and c[5]:                        # direct enwiki anchor on the SM node
-            edges.append((mk, addn("wiki", c[5], "category" if c[5].startswith("Category:") else "page", c[5]), "bridge"))
+        if len(c) >= 6 and c[5]:                        # direct enwiki anchor on the SM node (user-tagged)
```
s243a: **[low] User-tagged anchor is sub-tagged as inferred**

This direct enwiki anchor is labelled `user-tagged`, but the emitted confidence is below 1.0, which makes it read as inferred downstream. If this is a hand/user anchor, consider bumping it to `1.0`; otherwise please document why this user-tagged source is intentionally lower confidence and add the `0.9` confidence level to the PR description.

prototypes/mu_cosine/train_mu_attention.py
```diff
@@ -260,8 +261,30 @@ def train(args):
...
+              f"(ops {dict(_C(r[3] for r in graded_tr))}; bridge weight {args.bridge_weight:g}; "
+              f"inferred {n_inf}" + (f", infer-switch p={args.infer_subcat:g}×breadth" if args.infer_switch else "")
+              + ")")
+
+    # INFERRED relations (confidence<1, e.g. a typo'd / missing section) are untagged ⇒ the operator is
+    # uncertain. Stochastically switch an inferred `element_of` → `subcategory` (same μ, directional, but the
+    # WIKI operator) with probability ∝ the container's BREADTH (a broad domain is likelier to hold
+    # subcategories than leaf elements). This injects operator noise + models the element/subcategory
+    # ambiguity for untagged relations; TAGGED rows (conf 1.0) always keep their operator.
+    elem_breadth = {}                                    # container key → # element_of members (breadth proxy)
```
s243a: **[low] Breadth proxy counts emitted rows, not member relations**

Each `element_of` relation emits two directional rows, so `elem_breadth` is roughly 2× the true member count. That means `--breadth-scale 20` saturates around 10 real members. Either divide by two, deduplicate by edge, or document that `breadth_scale` is expressed in emitted-row units.

prototypes/mu_cosine/train_mu_attention.py
s243a, line 296: **[test] Please add a small seeded switch regression**

This path would benefit from a tiny regression fixture asserting that `conf == 1.0` rows never switch, legacy 9-column pairs default to `conf = 1.0`, and two runs with the same seed produce byte-identical sampled operators once the switch RNG is isolated. That would lock in the core safety property this PR relies on.

s243a (commented):
Council review summary: **approve with minor changes**. Opus 4.8, GPT-5.5, Gemini, and the external research pass agree this PR is backward-compatible, default-off, and conceptually useful. The main fix before relying on the reported A/B result is RNG isolation: the stochastic switch currently shares RNG state with batch sampling/masking, so the treatment may change more than just the operator choice.

Key asks before/after merge:
- Isolate or deterministically key the stochastic switch RNG so switch-on/off comparisons preserve the same sampling trajectory.
- Standardize on relation vs. operator terminology, since the code maps relations to operators rather than making operators the primary latent variable.
- Clarify whether confidence values are quantitative or provenance-only, because the current trainer only gates on `conf < 1.0`.
- Add small regression tests for TSV column compatibility, no-switch behavior for `conf == 1.0` rows, breadth scaling, and seeded reproducibility.

PR #3357, opened 2026-06-26
mu_cosine: multiple μ sources + dual-vs-unified estimation architecture (design)

s243a:
> Builds on `claude/mu-inferred-relations` — merge that first (this references
> `DESIGN_inferred_operator_superposition.md`).

Extends the inferred-operator theory with two design questions: which μ to
condition on, and who owns the distribution. Design doc only —
`DESIGN_mu_sources_and_estimation.md`.

## Multiple μ sources
The operator posterior conditions on **several** μ measurements, not one:
`P(relation | μ_e5, μ_model, …) ∝ P(relation)·∏ₖ P(μₖ|relation)` (naive-Bayes).
This is the bridge ensemble + the `CalibratedModel` wrapper feeding the operator
posterior. Start with two — e5 (static) and the training model (dynamic); open
set (LLM/miniLM later).

## Static vs dynamic
- **e5 μ — static**: `P(μ_e5|relation)` estimated once (frozen). An anchor —
  gives signal from step 0 and directly damps noise source (d), model churn.
- **model μ — dynamic**: `P(μ_model|relation)` re-estimated from error; starts
  uninformative, sharpens over training. Weight sources by separating power.
- `P(μ|relation)` is generative; inverting to `P(relation|μ)` is **Bayesian
  inversion**.

## Dual vs unified (the architecture fork)
- **Dual**: a distribution estimator D + a function model F; EM (re-estimate D
  from F's predictions, train F with D fixed). Stable; e5 drops in trivially.
- **Unified**: one F constrained so its error distribution matches
  `P(μ|relation)`. Elegant but collapse-prone.
- **The fork collapses**: a unified model with a **stop-gradient, EMA'd** posterior
  estimate *is* the dual model (target-network pattern); the **static e5 term is
  the anti-collapse anchor**.
- **Recommendation**: dual-in-spirit — estimation out of the gradient, e5
  anchoring, soft posterior-weighted loss. Fully-coupled unified deferred (needs
  entropy + out-of-set + anchor regularisers just to not collapse).

Includes a concrete build order for after merge.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3359, opened 2026-06-26
mu_cosine: μ→relation posterior — full readout vector, joint head, random operator embedding

s243a:
Builds the estimator that turns the μ-readout vector into `P(relation | μ_vec)`
and consumes it as a random operator embedding — the full chain from the fitted
distribution to a model input. (Follows the merged static-e5 estimator.)

## Commits
1. **Dynamic model μ source + e5↔model correlation.** Model μ separates relations
   3.4× better than e5 (spread 0.136 vs 0.040), but e5↔model corr = +0.751
   (strong) → principled product-of-experts weight e5 0.099. The non-independence
   measured, not asserted.
2. **Full μ-readout vector + outlier rejection + correlation structure.** Posterior
   conditions on e5, SYM, and BOTH directions of the asymmetric operators
   (wiki_fwd/rev, elem_fwd/rev). Outlier rejection (all relations): element_of
   noisiest (110, ~10%). Correlations: e5 most independent (+0.15); operator
   readouts mutually +0.6–0.7 (circular); wiki_fwd↔wiki_rev −0.72 (= asymmetry);
   elem_fwd↔elem_rev +0.11 ⇒ element_of less directional than subcategory
   (confirms the verified-Wikipedia check, against "element strictest").
3. **JointPosterior — joint `P(relation|μ_vec)`, beats product-of-marginals.**
   A small head (LR / MLP) over the whole vector. Held-out (7 relations,
   element_of ≈53% majority): factored 50.9% acc / 1.329 ll (**below** majority —
   correlation over-counting hurts); joint LR 53.8% / 1.251; MLP-16 56.0% / 1.160.
   Decisive test: a target whose marginals carry zero info, only the joint (the
   interaction) separates.
4. **Random operator embedding.** For an inferred row, the op token is a random
   superposition `w@op_emb`, `w ~ Dirichlet(α·P(op|μ_vec))` + out-of-set noise —
   one forward, input-space. α + out_of_set_noise map to the noise decomposition.
5. **Training-integration spec.** Blend on unlabeled only; asymmetric ops → two
   direction-specific blended targets; curriculum (labelled-first warm-up).

## Tests
`test_mu_posterior.py` (6/6, torch-free), `test_joint_posterior.py` (3/3),
`test_operator_embedding.py` (6/6, torch-free).

## Notes
- No data/models committed (local/gitignored). `REPORT_mu_posterior.md` Findings 3–5.

### DIFF
--- a/prototypes/mu_cosine/DESIGN_inferred_operator_superposition.md
+++ b/prototypes/mu_cosine/DESIGN_inferred_operator_superposition.md
@@ -27,13 +27,23 @@ operator, distinct relations that share an operator — element_of/subcategory b
 ELEM/WIKI — are exactly the cases μ alone can't resolve; see §2.)
 
 Two realisations of "train under the distribution":
-- **Hard sampling** — draw one operator per step from the posterior (a stochastic switch). Simple; higher
-  gradient variance.
-- **Soft posterior-weighted loss** (preferred) — the *expectation* over operators,
-  `L = Σ_op P(op | μ, type, breadth) · mse_under(op)`. Lower variance; the posterior's spread *is* the noise
-  (a flat posterior automatically trains a mixture of operators; a peaked one trains essentially one).
+- **Hard sampling** — draw one operator per step from the posterior (a stochastic switch). Simple; one
+  forward; higher gradient variance.
+- **Soft posterior-weighted loss** — the *expectation* over operators in LOSS space,
+  `L = Σ_op P(op | μ) · mse_under(op)`. Lower variance, but **K forwards** per inferred row (one per operator).
+- **Random operator EMBEDDING** (preferred) — the expectation in **INPUT space**: build one op token that is a
+  random *superposition* of the candidate operator embeddings,
+  `op_token = (w ~ Dirichlet(α · P(op|μ))) · op_emb  + out_of_set_noise·ε`, and do a **single forward**. The
+  model literally sees a "superposition operator." `w @ op_emb` is a torch matmul so gradients still reach the
+  operator embeddings (`w` is a detached random constant, like a dropout mask). The noise decomposition's two
+  knobs live here: **α** = posterior-spread + measurement/churn variance (a,c,d); **out_of_set_noise** = the
+  true operator is none of the candidates (b). `sample_operator_weights` / `random_operator_embedding` in
+  `mu_posterior.py` (torch-free reference + tests); tagged rows keep their fixed `op_emb[op]`.
 
-Hard sampling is the Monte-Carlo approximation of the soft loss.
+Hard sampling is the Monte-Carlo approximation of the soft loss; the **random embedding** is the cheap
+input-space realisation of the same expectation (one forward instead of K), and is the build target — it is
+what makes "operator = random superposition + noise, drawn from the fitted joint `P(relation|μ_vec)`"
+literal.
 
 ## 2. The posterior, and why it factors
 
--- a/prototypes/mu_cosine/DESIGN_inferred_operator_superposition.md
+++ b/prototypes/mu_cosine/DESIGN_inferred_operator_superposition.md
@@ -40,10 +40,13 @@ Two realisations of "train under the distribution":
   true operator is none of the candidates (b). `sample_operator_weights` / `random_operator_embedding` in
   `mu_posterior.py` (torch-free reference + tests); tagged rows keep their fixed `op_emb[op]`.
 
-Hard sampling is the Monte-Carlo approximation of the soft loss; the **random embedding** is the cheap
-input-space realisation of the same expectation (one forward instead of K), and is the build target — it is
-what makes "operator = random superposition + noise, drawn from the fitted joint `P(relation|μ_vec)`"
-literal.
+Hard sampling is the Monte-Carlo approximation of the soft loss. The **random embedding is NOT the same
+expectation** — `model(Σ wᵢ·op_embᵢ)` ≠ `Σ wᵢ·model(op_embᵢ)` because the model is nonlinear in the operator
+token (Jensen) — it is a cheaper **stochastic surrogate**: a single forward through an input-space mixture of
+operator embeddings, which makes "operator = random superposition + noise, drawn from the fitted joint"
+literal at the cost of being an approximation, not the loss-space expectation. (It equals the expectation only
+if the readout were linear in the op token.) It is the build target *because* it is cheap (one forward) and
+injects the uncertainty as input noise; the soft loss is the exact-but-K-forward alternative to A/B against.
 
 ## 2. The posterior, and why it factors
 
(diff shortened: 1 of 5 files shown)


prototypes/mu_cosine/mu_posterior.py
```diff
@@ -133,32 +265,90 @@ def main():
...
+        nh = int(args.held_frac * len(order)); held, train = order[:nh], order[nh:]
+        Xtr = X[train]; rtr = [rels[i] for i in train]
+        Xhe = X[held]; rhe = [rels[i] for i in held]
+        ri = {r: i for i, r in enumerate(relset)}
+
+        joint = JointPosterior(relset, n_features=len(sources), hidden=args.hidden).fit(Xtr, rtr)
+        jacc, jll = _eval(joint.proba(Xhe), rhe, ri)
+
+        fac = MuPosterior(nbins=args.nbins)                # factored product (equal weights) on the same train
+        for s in sources:
```
s243a: **[needs-discussion] Compare against the corrected PoE baseline, not only the unweighted product**

This factored baseline uses equal weights on all readouts, which is exactly the correlated-source over-counting pattern #3357 was trying to avoid. Since this PR derives correlation-corrected weights earlier, the claim that the joint head beats the factored model should either benchmark the weighted PoE/product baseline as well, or explicitly say the reported factored result is the naive unweighted baseline and does not supersede the #3357 weighting story.

prototypes/mu_cosine/REPORT_mu_posterior.md
```diff
@@ -40,11 +40,84 @@ cybernetics`, `BELBIC` — the same set the bridge ensemble surfaced. (e5 is rec
...
+
+## Finding 4 — the full μ-readout vector + outlier rejection + the correlation structure
+The posterior should condition on the **full vector** of μ readouts, not one number: raw e5 μ, the symmetric
+SYM μ, and — for the ASYMMETRIC operators — **both directions** (`wiki_fwd/wiki_rev`, `elem_fwd/elem_rev`).
+`mu_posterior.py --model --reject-outliers` fits all six and reports separability + the correlation matrix.
+
+**Outlier rejection (all relation types, e5 out-of-band):** 214/2088 rejected, by relation
+`element_of 110, subcategory 40, bridge_neg 20, super_category 16, see_also 10, bridge 10, subtopic 8` —
+**element_of is by far the noisiest label** (110, ≈10%), confirming it most needs outlier rejection.
+
```
s243a, line 76: **[high] The 'noisiest label' conclusion appears to be a quantile artifact**

Because the anomaly band is computed per relation with a fixed quantile, it should flag roughly `2q` of each relation by construction. The absolute count here mostly tracks class size, so `element_of 110` does not by itself show that `element_of` is intrinsically noisier than the other labels. Please report per-relation rejection rates with uncertainty, use a common/global reference band, or rephrase this as a class-size-weighted review queue rather than a relative noise diagnostic.

prototypes/mu_cosine/DESIGN_inferred_operator_superposition.md
```diff
@@ -27,13 +27,23 @@ operator, distinct relations that share an operator — element_of/subcategory b
...
+  `op_token = (w ~ Dirichlet(α · P(op|μ))) · op_emb  + out_of_set_noise·ε`, and do a **single forward**. The
+  model literally sees a "superposition operator." `w @ op_emb` is a torch matmul so gradients still reach the
+  operator embeddings (`w` is a detached random constant, like a dropout mask). The noise decomposition's two
+  knobs live here: **α** = posterior-spread + measurement/churn variance (a,c,d); **out_of_set_noise** = the
+  true operator is none of the candidates (b). `sample_operator_weights` / `random_operator_embedding` in
+  `mu_posterior.py` (torch-free reference + tests); tagged rows keep their fixed `op_emb[op]`.
+
+Hard sampling is the Monte-Carlo approximation of the soft loss; the **random embedding** is the cheap
+input-space realisation of the same expectation (one forward instead of K), and is the build target — it is
+what makes "operator = random superposition + noise, drawn from the fitted joint `P(relation|μ_vec)`"
```
s243a: **[medium] Input-space superposition is not the same expectation as soft loss**

`Σ P(op)·mse_under(op)` is an expectation in loss/output space, while `model(Σ w_i·op_emb_i)` is a single nonlinear forward pass through an input-space mixture. Those are not generally equivalent by Jensen/nonlinearity. The helper implementation is fine as a stochastic surrogate, but the doc should avoid saying it is the same expectation unless the downstream model is linear in the operator token or this is explicitly framed as an approximation.

prototypes/mu_cosine/DESIGN_inferred_operator_superposition.md
```diff
@@ -94,6 +104,31 @@ all four show up as how *distributed* the operator-weighted gradient is.
...
+   superposition + the joint-posterior assignment apply to **inferred** (`confidence < 1`) rows. Labelled
+   data is ground truth — never blur it.
+
+2. **Asymmetric operators carry TWO prior μ — handle both directions.** `WIKI` (subset/subcategory) and
+   `ELEM` (element) are directional, so each has two reference targets: forward `μ(member|container) ≈ 0.9`
+   and reverse `≈ 0.1`; `SYM` (see_also/assoc/bridge) is symmetric (one value, e.g. 0.4 both ways). So an
+   inferred pair generates **two** training examples — `(node, root)` and `(root, node)` — each with the
+   *same* random-superposition op token but a **direction-specific blended target**:
+   `μ_target_dir = Σ_op P(op|μ_vec) · target_dir(op)` (forward uses each op's forward target, reverse its
+   reverse). A symmetric op contributes the same value both ways; the asymmetric ops contribute their two.
```
s243a: **[medium] Please specify the relation→operator marginalization before consuming `P(op)`**

The posterior head produces `P(relation | μ_vec)`, but `random_operator_embedding` consumes `P(op)`. This section should spell out the marginalization `P(op)=Σ_relation P(op|relation)P(relation|μ_vec)` and how many-to-one operators are handled. In particular, `SYM` appears to cover multiple relations with different scores/semantics, so the operator-level blend target is under-specified without this bridge.

prototypes/mu_cosine/mu_posterior.py
```diff
@@ -133,32 +265,90 @@ def main():
...
     tagged = [r for r in rows if r["conf"] >= 1.0]
     e5mu = e5_mu_fn(args.e5_cache)
+    mu_e5 = {(r["node"], r["root"]): e5mu(r["node"], r["root"]) for r in tagged}
 
+    # provisional e5 fit → bands → OUTLIER REJECTION (the side-note rule, applied to ALL relation types)
+    prov = MuPosterior(nbins=args.nbins)
+    prov.fit_source("e5", ((r["rel"], mu_e5[(r["node"], r["root"])]) for r in tagged))
+    flagged = [r for r in tagged if prov.is_anomalous("e5", r["rel"], mu_e5[(r["node"], r["root"])], args.q)]
+    fit_set = [r for r in tagged if r not in flagged] if args.reject_outliers else tagged
+    print(f"tagged pairs: {len(tagged)} / {len(rows)} total; out-of-band {len(flagged)}"
```
s243a, line 297: **[low] Avoid value-membership filtering for rejected rows**

`r not in flagged` performs list membership by dict equality and can become O(n²); it can also remove non-flagged duplicate rows if an equal dict appears in `flagged`. Prefer tracking `id(r)` or building an index/key set for the flagged rows before filtering.

s243a (commented):
Council review summary: **needs discussion**. The code/tests look solid for a prototype: Opus 4.8, GPT-5.5, Gemini, and the research pass all agreed the joint posterior head is directionally sound, the torch-free tests are credible, and the random operator embedding helpers are mathematically reasonable. The issues are mostly claim/spec consistency: the report currently overstates some evidence, and the relation→operator consume path needs to be made explicit before this is used as a trainer spec.

Key asks:
- Re-run or caveat the joint-vs-factored comparison against the correlation-corrected PoE baseline from #3357, not only an equal-weight product of correlated readouts.
- Fix the outlier/noisiest-label claim; per-relation quantile bands flag roughly the same fraction per class, so counts mostly track class size.
- Specify the mapping from `P(relation | μ_vec)` to `P(op)` before `random_operator_embedding`, especially for many-to-one operators like `SYM`.
- Reword the random input-space superposition as a stochastic surrogate, not the same expectation as a soft loss through a nonlinear model.

External context: calibration literature treats raw softmax posteriors as commonly overconfident, with temperature scaling/ECE as a minimal baseline; Dirichlet uncertainty methods use concentration carefully and distinguish training-time sampling from deterministic inference-time expectation.

PR #3394, opened 2026-07-01
mu_cosine: operator-axis cleanup (WIKI→HIER, LLM demoted) + parameterized path-operator design

s243a:
Two things from one factorization question: **the operator axis should be pure *relations*; source and
computation-strategy belong on other axes.** Part 1 is a small code cleanup; Part 2 is a future-work design doc (the
reviewable part — architecture-on-paper, no empirical claims, nothing built).

### Part 1 — operator cleanup (code, checkpoint-safe, no behavior change)
- **`WIKI` → `HIER`.** `WIKI` named an operator after a *source* (Wikipedia), but the source axis already
  distinguishes `enwiki/simplewiki/pearltrees/mindmap` (`corpus_emb`). It's really the **category-hierarchy relation**
  (subcategory/subtopic/super_category) → renamed. `op_emb` is a *learned codebook indexed by ID*, so this is pure
  relabeling — the learned row is untouched (verified: `test_op_weights` passes; `REL_SPEC` maps only `ELEM/HIER/SYM`).
- **`LLM` demoted.** Never a relation — the LLM is a *judge* (already on `judge_emb` as `haiku/sonnet/opus`); the
  *prompt* defines the relation. Its `--llm` "bought-Haiku membership" fixture is really `ELEM` + `judge=haiku` →
  rerouted there. Index 2 kept (n_ops=4, checkpoint-compat) but deprecated.

Net: operator axis = pure relations (`SYM`, `HIER`, `ELEM`); sources on `corpus_emb`, judges on `judge_emb`.

### Part 2 — `DESIGN_path_operator.md` (future-work design; the part to review)
A multi-path ancestor operator for the DAG (a node has many ancestor chains; the earlier single-path LINEAGE
collapsed them and came out redundant to HIER — but multi-path is untested). Four choices, each resolving a real
strain and each **reusing existing machinery**:

1. **operator × method (higher-order).** Ancestor-path membership is *one* operator with a `method` parameter;
   `LINEAGE = method:canonical`, `PATH = method:random-walk`. Unifies lineage/path.
2. **stop-β.** A per-step stopping probability gives geometric-depth paths (a principled depth sampler) **and** makes
   uniform-walk-with-stop = discrete PPR (β = restart) — random-walk and PPR sampling collapse into one knob. Three
   methods documented: uniform-walk(+β), edge-weighted (RefPearl≫AliasPearl), PPR/flux.
3. **per-node token representation.** Sampled paths (walk × stop × multi-path × wildcards) are combinatorial, so
   whole-path embeddings can't be precomputed. Represent the path as per-node cached embeddings + attention
   aggregation; sampling = *selecting which cached node-tokens enter the passage*. No re-embedding; a multi-token
   passage the attention block already supports.
4. **operator encoder = key–value attention lookup; numeric params via fixed Fourier + FiLM.** The operator token is
   *computed* from a spec, not looked up: `keys = e5(operator names)`, `values = existing learned op_emb rows`,
   `query = e5(operator description)`, `encoding = softmax(query·keysᵀ) @ values`. That softmax **is** the existing
   `op_weights`, so it generates the superposition weights from the operator's *description* (degenerates to today's
   codebook when a query hits one key); e5 is only the *key* (routing), the *learned value* carries semantics —
   sidestepping the "raw-e5-name = prompt-rewriting" concern. **Numeric parameters (β, edge-weights) inject on a
   separate path — the attention can't route a continuum:** Fourier-feature encode each scalar (`β → [sin/cos at
   several frequencies]` — raw scalars underfit MLPs; sinusoidal features fix it), then **FiLM-modulate** the
   attention output (`op_encoding = γ(fourier(β)) ⊙ base + shift`). **Frequencies are FIXED, not learned** —
   parameter-free (only the small FiLM MLP trains) and sufficient because β's range is known ([0,1]); chosen for
   reduced training on consumer-grade hardware.

Honest scoping throughout: adopt the encoder + per-node representation *when building* PATH-with-methods; the codebook
+ single-node passages stay valid for the base relations meanwhile. Prior art cited in-doc (Vaswani 2017 sinusoidal
PE; Tancik 2020 + Rahimi & Recht 2007 Fourier features / spectral bias; NeRF 2020; Time2Vec 2019; FiLM Perez 2018).

### Review ask (light, no compute)
Scrutiny welcome on Part 2's *architecture*: does operator×method×representation×encoder hold together; is the
attention-lookup-as-`op_weights` reduction sound; is stop-β = discrete-PPR right; is Fourier+FiLM the right way to
inject continuous params (and fixed-frequencies the right call given a known [0,1] range); simpler alternatives to
the per-node representation or the encoder? Part 1 is mechanical.

### DIFF
--- a/prototypes/mu_cosine/DESIGN_path_operator.md
+++ b/prototypes/mu_cosine/DESIGN_path_operator.md
@@ -143,6 +143,14 @@ FiLM will. (Vector params like per-edge-type weights: Fourier-encode each scalar
 discrete — uniform vs type-weighted vs confidence — that choice is semantic and goes in the attention query; only the
 actual weight *values* take the Fourier+FiLM path.)
 
+Notes: frequencies are indexed by the *dimension* (each dim a fixed ωᵢ, geometrically spaced), and the value is fed
+to all — so the wavelength lives on the *β-axis*; set the lowest ω so its wavelength covers β's range (distinct,
+non-wrapping codes), higher ω for resolution. The sin/cos *pair* handles sign automatically (β as an angle on the
+unit circle) — no phase-shift term needed. Frequencies may be fixed or learned. *Prior art:* sinusoidal positional
+encoding (Vaswani et al. 2017, §3.5); Fourier features & the MLP spectral-bias motivation (Tancik et al., NeurIPS
+2020; Rahimi & Recht, NeurIPS 2007); continuous-coordinate encoding (NeRF, Mildenhall et al. 2020); learned scalar
+encoding (Time2Vec, Kazemi et al. 2019); FiLM modulation (Perez et al., AAAI 2018).
+
 ## 6. Why PATH (multi-path) is *not* redundant to HIER, though LINEAGE was
 
 `HIER` = the single category-hierarchy **edge**. Single-path `LINEAGE` = one chain of `HIER` edges ⊂ (`HIER` +
--- a/prototypes/mu_cosine/DESIGN_path_operator.md
+++ b/prototypes/mu_cosine/DESIGN_path_operator.md
@@ -34,11 +34,34 @@ distribution, embed it, apply wildcard masking (id-dropout + prefix-dropout), an
 learns the *expectation over paths* across steps — no need to embed every path and weight-average explicitly. The
 wildcards are an orthogonal masking augmentation on top of the path sampling.
 
+## 3b. Passage representation — per-node tokens, NOT a fixed whole-path embedding
+
+Once paths are *sampled* (random walk × stop-β × multi-path × wildcards), the space is **combinatorial** — you cannot
+precompute a fixed e5 embedding for every sampled path (the variant-cache trick that worked for LINEAGE's ~6
+variants/folder breaks entirely). **Resolution: stop embedding the whole path as one e5 string.** Represent the path
+as a **set of per-node embeddings** — embed each ancestor node's *title* **once** (cached, linear in #nodes) — and
+let the model **attend / aggregate over the sampled subset**. Then all the sampling (which nodes, how deep, which of
+the DAG's paths, wildcard-masked) is just **selecting which cached node-tokens enter the passage**:
+- No re-embedding, ever — the combinatorial path space is handled by *token selection over a linear node cache*
+  (this also kills the re-embed-per-step compute concern that dogged the whole-path-string approach).
+- The μ-attention block already supports a **multi-token passage**, so this is a representation change, not an
+  architecture change.
+- The materialized-id line / wildcards become per-node attributes rather than substrings of one blob.
+
+So the path operator factors on **three** axes: **operator** (ancestor-path membership) × **method** (sampling incl.
+stop-β) × **representation** (per-node tokens, attention-aggregated). The whole-path-string embedding was scaffolding
+that doesn't scale; per-node composition is the structure that does — and it's what the decoder's per-node /
+optimizer view wanted anyway (§0 of the decoder sketch: the optimizer reads a per-node partial-path state).
+
 ## 4. The three path-sampling methods (document all three)
 
-**(1) Uniform random walk — default, cheapest.** Going up the DAG, pick a parent **uniformly** at each node;
-a path's probability is `∏ 1/|parents(level)|`. Parameter-free — literally "sample a parent at each level." Weights
-paths by *structural* (branching) likelihood: a node with two parents splits its mass evenly. Start here.
+**(1) Uniform random walk (+ stop-β) — default, cheapest.** Going up the DAG, pick a parent **uniformly** at each
+node, and **terminate with probability β at each step**. Path prob ≈ `∏ 1/|parents(level)|` × the geometric
+stopping term. Parameter-free but for β — literally "sample a parent at each level, sometimes stop early." Two
+bonuses: **(a)** the stop gives geometric-depth (variable-length) paths ⇒ a *principled depth sampler* (graded-depth
+ancestor membership — a better-motivated prefix-dropout); **(b)** a uniform up-walk with per-step stop **IS discrete
+PPR** — β is the PPR restart — so **methods (1) and (3) collapse into one knob** (β→0 walks to the root; larger β
+stays shallow). Start here.
 
 **(2) Edge-weighted — canonical strength.** Weight edges by **type/confidence** — RefPearl (inside/owned) ≫
 AliasPearl (cross-link), or graph-edge confidence — so a path's probability = normalized product of its edge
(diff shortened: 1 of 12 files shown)


s243a (commented):
## PR Review: mu_cosine operator-axis cleanup + parameterized path-operator design

### Part 1 — Operator cleanup (WIKI→HIER, LLM demoted) ✅

This part is clean and the rationale is solid. The axis-purity argument is correct: naming an operator after a *source* (`WIKI`) rather than the *relation* it encodes (`HIER`) was a category error, and this fixes it. Notes:

- **Rename is safe.** `op_emb` is indexed by integer; the codebook row at index 1 is untouched. All usages across `build_graded_round.py`, `train_mu_attention.py`, `REL_SPEC`, `RELATION_SPEC`, `eval_filing.py`, `eval_hybrid.py`, `mu_posterior.py`, `test_infer_switch.py`, `test_op_weights.py`, and `train_lineage.py` are updated consistently. No silent regressions expected.
- **`LLM` demotion is correct in principle.** The LLM is a labeling *judge*, not a semantic relation; routing `--llm` fixtures to `ELEM + judge=haiku` is the right fix. **One mild concern:** index 2 is now a ghost slot — kept solely for `n_ops=4` checkpoint compatibility but with no relation routing to it. The comment in `mu_attention.py` documents this well. Consider adding a runtime assertion (e.g., in `validate()`) that no training batch ever assigns index 2, so future contributors don't silently leak data into the deprecated slot.
- **`eval_filing.py` stale variable name:** The variable `S_wiki` is retained after the rename (`S_elem, S_wiki, S_sym = ...`). This is not a bug (it resolves correctly to `HIER` scores), but it will confuse readers. Suggest renaming to `S_hier`.

---

### Part 2 — `DESIGN_path_operator.md` architecture review

#### §2 — operator × method (higher-order operator) ✅

The factoring is clean: `what (relation)` vs `how (aggregation strategy)` is a well-established separation (cf. strategy pattern). Deferring `method_emb` until ≥2 methods are live is pragmatic and avoids training a constant token — the document is honest about this.

#### §3 — Stochastic superposition (one path per step) ✅

Sampling one ancestor path per training step and letting the model learn the expectation across steps is a correct Monte Carlo argument. This is exactly how stochastic depth / dropout-as-regularizer works. The wildcard masking (id-dropout + prefix-dropout) as an orthogonal augmentation on top of path sampling is sound.

#### §3b — Per-node token representation ✅ (with one open question)

The core argument is right: whole-path e5 strings can't be precomputed for a combinatorial path space; per-node cached embeddings + attention aggregation is the correct resolution. The μ-attention block already supports multi-token passages, so this is representational rather than architectural.

**Open question:** When the random walk samples a path of variable length (stop-β), the attention aggregation sees a *variable-length* set of node tokens. Is the attention block currently set up for variable-length passages, or does it assume a fixed sequence length? If it pads/masks, confirm the mask propagates correctly so padding tokens don't contribute to the aggregated representation.

#### §4 — stop-β and the PPR collapse ✅ with a precision note

The claim that a uniform random walk with per-step stopping probability β **is** discrete PPR is *approximately* right but worth tightening. Classical PPR restarts at the source with probability β at each step — which is a *global* restart, not a stopping condition. A stop-β walk **terminates** (path ends) rather than teleporting back. The score distributions are related but not identical:
- PPR gives a stationary distribution over the entire graph.
- Stop-β gives a distribution over *path prefixes*, which concentrates on shallow ancestors.

For your use case (scoring ancestor-path membership rather than computing a graph centrality), the stop-β path distribution is arguably *more* semantically meaningful than PPR anyway. But the equivalence claim in the doc ("uniform-walk-with-stop = discrete PPR, β = restart") should be scoped more carefully — say "functionally analogous to PPR in the [0,1] depth-weighting sense" rather than equating them, to avoid confusion if someone tries to compare scores numerically with a PPR baseline.

#### §5 — Attention-lookup encoder (key–value attention as op_weights) ✅ with caveat

The reduction `softmax(query·keysᵀ) @ values = op_weights @ op_emb.weight` is structurally sound: you're already doing a convex combination over operator embeddings; making the weights *content-addressed* from an e5 description rather than hand-set is a clean generalization. The warm-start from existing `op_emb` rows as initialized values is a nice continuity property.

**Caveat flagged correctly in the doc:** description quality matters. Close descriptions → blended operators. This is benign when operators are semantically related (good: `subcategory` and `subtopic` share a key neighborhood), but can silently degrade if a new operator's description drifts near an existing key for unintentional reasons. Recommend: when promoting this to production, add a diagnostic that prints `softmax(query·keysᵀ)` for each registered operator at startup — you want near-one-hot for the base relations `{SYM, HIER, ELEM}` and will immediately catch description collisions.

#### §5 — Fourier + FiLM for continuous parameters ✅

The choice is well-motivated and the prior-art citations are complete (Vaswani 2017, Tancik 2020, Rahimi & Recht 2007, NeRF, Time2Vec, FiLM/Perez 2018). A few confirmatory notes:

- **Fixed frequencies:** Correct call for a *known bounded range* ([0,1]). Geometrically spaced ω with lowest ω set so its wavelength covers the full range, sin/cos pairs for sign-disambiguation — this is textbook positional encoding applied to a scalar. No argument here.
- **FiLM placement:** Modulating the *attention output* (`op_encoding = γ(fourier(β)) ⊙ base + shift`) is the right injection point — it scales/shifts the semantic routing result, not the input embeddings, so the attention's discrete routing and the continuous parameter injection remain cleanly separated.
- **Vector params (per-edge-type weights):** The doc's guidance — Fourier-encode each scalar, or treat the *choice of weighting scheme* as a semantic query for the attention router — is the right split. One small addition to consider documenting: when the number of edge types is small and fixed (e.g., `{RefPearl, AliasPearl}`), the edge-type *choice* can also be a one-hot concat to the Fourier features rather than a separate attention query, which is simpler at low cardinality.

#### §6 — Why PATH ≠ HIER redundancy ✅

The argument is correct. LINEAGE ⊆ transitive closure of HIER ⇒ redundant. Multi-path superposition ≠ single-edge HIER because the DAG's full ancestor *set* across branches is not captured by any single chain. The framing is clear.

---

### Summary of actionable items

| Item | Part | Priority |
|---|---|---|
| Rename `S_wiki` → `S_hier` in `eval_filing.py` | 1 | Low (cosmetic) |
| Add assertion that index-2 (deprecated `LLM`) never appears in training batches | 1 | Medium (correctness guard) |
| Clarify stop-β ≠ PPR equivalence (scope the claim) | 2 (§4) | Medium (correctness of doc) |
| Add startup diagnostic for attention-lookup softmax weights | 2 (§5) | Low (future-work note) |
| Confirm variable-length passage handling in μ-attention for stop-β paths | 2 (§3b) | Medium (open question) |

Part 1 is safe to merge as-is. Part 2 (the design doc) is well-structured and the architecture holds together — the four choices (`operator×method`, `stop-β`, `per-node tokens`, `attention encoder + Fourier+FiLM`) are internally consistent and each reuses existing machinery cleanly. The main open question is the variable-length passage mask in §3b.

PR #3426, opened 2026-07-03
Mindmap LINEAGE sampler + graph/LLM judge methodology

s243a:
## Summary

Builds the SimpleMind mindmap **LINEAGE** sampler and documents the methodology for turning
mindmap hierarchies into **filing training data** for the μ model — non-enwiki data that is the
closest available proxy to the (still-harvesting) Pearltrees filing target.

**Built:** `gen_mindmap_lineage.py` — parses `.smmx` maps → materialized-path lineages
(491 chains, 6 systems-theory maps). Handles the shared `Root Node` sentinel (folder `root`,
to which all maps link up), filters `via link` nav nodes, collapses cross-map echoes; privacy is
scrubbed upstream by `parse_smmx.py` (`is_private_title`: a "private" node drops itself + subtree,
a private root drops the whole map).

**Documented (design, prototype next):** `DESIGN_mindmap_lineage.md` — the negative-generation
and judge-scoring methodology.

## Methodology (the part to review)

- **LINEAGE first, then 70/30 split.** A mindmap node hangs off one structural parent, so single-path
  lineage is well-defined (same reasoning as the Pearltrees single-path decision).
- **Positives: complete + graded by distance.** Small dataset → enumerate every (descendant, ancestor)
  pair; μ graded by lineage depth (parent > grandparent > …), not binary.
- **Negatives: sampled, typed, graded.** Substitution (alternate parent) over permutation; hardness
  **stratified by e5 distance of the substituted node** (near = sibling/cousin hard negative). Key point:
  because all nodes share the global root, **no pair is truly unrelated** — μ falls off *gradually* with
  tree-distance rather than snapping to 0, so sibling/cousin hard negatives matter most.
- **e5 distance stratifies candidates; it is NOT used as μ** (semantic distance ≠ filing plausibility).
- **Judge scoring:** remove the last hop → each example is real filing data. Score **pointwise,
  independently (μ does not sum to 1)**, ≥2 candidates; **max-normalize to the model's top pick = 1**
  with a **linear** curve (sigmoid only as a temperature-controlled sharpness knob). **Multi-parent guard:**
  anchoring to the model's top honors genuine second parents, but the ground-truth filing is kept as a
  sanity floor (flag if it scores low).
- **Two complementary judges:**
  - **LLM** (codex `gpt-5.5-low`, non-Anthropic) — semantic plausibility.
  - **GRAPH** — `μ = decay(distance-to-truth)`, tagged `judge=graph` (already a defined judge). Cheap,
    structural, maps distance→μ honestly.
  - **Whole-lineage graph-μ (not just the immediate parent):** a bare parent-to-parent hop is myopic
    (equidistant candidates can sit in very different lineages). `μ_graph` factors in the full
    materialized-path prefix via **LCA depth** (structural shared-prefix) and **e5-of-prefix distance**
    (semantic — `Math/Calculus/…` stays close to `Math/Analysis/…`). e5 remains a *factor inside the
    graph heuristic*, not e5-alone-as-μ. `μ_graph ≈ decay(hops) · lca_depth_frac · f(e5_prefix_dist)`
    (exact combination TBD by prototype — a good spot to review for weighting / double-counting between
    LCA-depth and e5-prefix).
  - **Cost-smart split:** run the free graph judge over everything, spend LLM calls only where graph-μ is
    uncertain (near-ties, structurally-distant candidates). Disagreement between judges = the ambiguous /
    multi-parent signal.

## Graph-distance alternatives (per reviewer note)

The graph judge's distance→μ decay is one choice among the project's existing cost-function variants —
referenced as alternatives in the design doc and `docs/design/COST_FUNCTION_PHILOSOPHY.md`:
**hop/shortest-path** (default, `γ^d`), **exponential-decay flux**, **power-law / inverse-radial** (`1/r^N`),
**PPR / Markov-walk** (multi-path), and **LCA depth** (tree-native tie-breaker). Default = hop-distance +
LCA tie-break; power-law / PPR are the natural upgrades for heavier tails / multi-path mass.

## Also in this PR
- `mu_attention.py`: `JUDGES += {gemini, gpt-5.5-low}` — each (model, reasoning-effort) gets its own
  learned calibration row (table-based `judge_emb`; open-ended attribute-based judge input is future work).
- `score_with_codex.py`: batch judge scorer via `codex exec` (`gpt-5.5-low`).

## Status / next
Lineage built; **next** = substitution+e5 candidate generation → graph judge → targeted LLM judge →
graded-round rows (`corpus=mindmap`). Full value once fused with Pearltrees (`fuse_corpus.py` needs the
harvested PT cache).

## Scope notes
- This PR is the **mindmap/judge methodology** only. The **enwiki fine-tune** thread
  (`build_broad_graph.py`, the `model_broad_ft2.pt` run) is intentionally *not* here — it still has an
  unresolved **SYM baseline question** (whether `model_prod.pt` itself scores ~0 SYM on that held-out set,
  i.e. is the "regression" even real).
- Negative-generation + judge scoring is **designed and documented but not yet coded** — this PR is the
  sampler + the methodology, so the review is of the *plan* for negatives/judges, not an implementation.

### DIFF
--- a/prototypes/mu_cosine/DESIGN_mindmap_lineage.md
+++ b/prototypes/mu_cosine/DESIGN_mindmap_lineage.md
@@ -1,0 +1,114 @@
+# Mindmap LINEAGE → filing training data (design + methodology)
+
+Status: lineage sampler built (`gen_mindmap_lineage.py`); negative-generation + judge-scoring specified here,
+prototype next. Non-enwiki data from SimpleMind `.smmx` maps; prep for fusion with Pearltrees once harvested.
+
+## 0. Why mindmaps, why lineage first
+
+The SimpleMind maps are the user's own systems-theory domain — structurally and topically the closest proxy we
+have to the Pearltrees filing target, and **not** present in enwiki. We build **LINEAGE first, then the 70/30
+split on top of it** (directional/superposition — see `feedback_enwiki_sampling_strategy`). Lineage is
+well-defined here because a mindmap node hangs off exactly one structural parent — the same property that made us
+choose single-path LINEAGE for Pearltrees (`project_mu_cosine_path_multipath_result`).
+
+## 1. Positives — complete, and graded by distance
+
+The dataset is small (491 lineage chains over 6 maps), so we take **complete coverage of positives**: every
+(descendant, ancestor) pair along each materialized path. They are **not binary** — μ is **graded by lineage
+distance**: direct parent = strong, grandparent = weaker, great-grandparent = weaker still. The depth gives
+calibrated positives for free.
+
+Materialized paths are cleaned (`gen_mindmap_lineage.py`): the shared sentinel `Root Node` (in folder `root`,
+a generic placeholder to which all maps link up) is dropped, `via link` nav nodes filtered, cross-map join
+echoes (consecutive duplicate titles) collapsed. Privacy is scrubbed upstream by `parse_smmx.py`
+(`privacy.py::is_private_title`: a "private" node drops itself + subtree, a private root drops the whole map).
+
+## 2. Negatives — sampled, typed, graded (not enumerated)
+
+The negative complement is ~O(N²) and would swamp the positives, so we **sample** (enwiki uses 3:1). Three types
+in increasing value:
+
+1. **Directional-reverse** (ancestor | descendant) — free; drives the directional margin μ(desc|anc) ≫ μ(anc|desc).
+2. **Cross-branch** (different subtrees) — the easy "unrelated" tail.
+3. **Hard negatives — siblings / cousins** (same parent/grandparent, *not* ancestor-descendant). The valuable
+   ones: they teach μ that *same branch ≠ lineage*, the confusion a hierarchy model must resolve.
+
+**Key subtlety — no pair is truly unrelated.** Because every node shares the global `Root Node`, even
+cross-domain pairs share distant ancestry. So a "negative" means *"not a direct lineage link,"* and μ should
+**fall off gradually with tree-distance**, not snap to 0. This graded fall-off is the signal we want, and it is
+why hard sibling/cousin negatives matter more than random cross-map ones (the shared root already makes those
+mildly positive).
+
+### 2a. Candidate generation — substitution, e5-stratified
+
+Generate hard negatives by **substituting** a node in a true path with an alternative (an alternate parent/path
+— exactly the filing question "could it go here instead?"). **Permutation** (scrambling path order) mostly
+yields obviously-broken hierarchies — kept only as a cheap easy-negative floor.
+
+Control hardness with **e5 distance of the substituted node vs the original**: e5-near = sibling/cousin = hard
+negative; e5-far = easy. **e5 distance is used to *stratify candidates*, not as μ** — semantic distance ≠ filing
+plausibility (a sibling can be e5-close yet filing-wrong; a valid alternate parent can be e5-far). The μ comes
+from a judge (§3).
+
+## 3. Judge scoring
+
+Remove the last hop of a path → (path-minus-leaf = folder, leaf = item to file). Each example is thus **actual
+filing training data** — more target-aligned than raw lineage pairs, and buildable now without the harvester.
+Present the judge ≥2 candidate filings (2 = cheapest; 3–4 = better-shaped curve), scored **pointwise and
+independently — μ does NOT sum to 1** (matches the §14 prompt and the model's per-pair μ).
+
+### 3a. LLM judge (codex `gpt-5.5-low`, or Haiku)
+
+- **Max-normalize to the model's top pick = 1**, then a scoring curve: **linear first** (preserves the graded
+  middle); a **temperature-controlled sigmoid** only if raw scores come back too bunched to separate hard-from-easy
+  (a *steep* sigmoid discards the grading — avoid).
+- **Multi-parent guard:** anchoring to the model's top (not the ground-truth) honors genuine second parents
+  (Pearltrees supercategories, cross-map links) — a valid substituted path *should* be allowed near 1. But keep
+  the **ground-truth as a sanity floor**: if the true filing scores *low*, flag it (judge error or a genuinely
+  ambiguous node), don't silently train on it.
+- Tag `judge=gpt-5.5-low` (its own calibration row; see `DESIGN_calibrated_judges.md`, `JUDGES` in
+  `mu_attention.py`). Cost/latency: `gpt-5.5-low` ≈ 5 s/pair, non-Anthropic.
+
+### 3b. Graph judge (nearly free) — `μ = decay( distance-to-truth )`
+
+Score a candidate by its **graph distance to the true filing**, decayed to μ (truth at distance 0 → **1**;
+1 hop = sibling/uncle → high; far → low). This is the **structural** analogue of the e5 idea, and unlike e5 it
+maps to μ **directly and honestly** — distance-to-truth *is* a plausibility. Tag `judge=graph`
+(`JUDGES["graph"] = 1`, already defined).
+
+**Decay choice — alternatives (we pick a default; these are the project's existing metrics).** The decay is
+exactly a flux/cost-function over the tree; see `docs/design/COST_FUNCTION_PHILOSOPHY.md` and
+`docs/design/SCAN_STRATEGY_SPECIFICATION.md`:
+
+| variant | form | note |
+|---|---|---|
+| **hop / shortest-path** (our default) | `μ = γ^d`, or `1/(1+d)` | simplest; no flux interpretation |
+| exponential-decay flux | `Σ_paths (decay/branching)^|p|` | our `exp(-λd)` is the single-path case |
+| power-law / inverse-radial | `1/r^N` (log-flux `= −N·log r`) | N-dim source model; heavier tail |
+| PPR / Markov walk | stationary dist. biased by the node seed | probabilistic dual; handles multi-path |
+| LCA depth | depth of lowest common ancestor | tree-native; "shares-only-Root ⇒ ~0" |
+
+Default: **hop-distance with geometric/exp decay + LCA depth as the tie-breaker** (cheap, tree-native). The
+power-law and PPR variants are the natural upgrades if the graded fall-off needs a heavier tail or multi-path
+mass — they are already specified in `COST_FUNCTION_PHILOSOPHY.md` and left as alternatives.
+
+## 4. Two judges, complementary — and a cost-smart split
+
+Graph judge = cheap, structural, but **blind to a semantically-valid parent that is structurally distant**
+(a cross-domain-but-correct filing). LLM judge = semantic, catches those, but costs. Their **disagreement is the
+useful part**: agree → confident μ; diverge → the ambiguous, probably-multi-parent cases.
+
+**Cost-smart pipeline:** run the **graph judge over everything (free, `judge=graph`)**, then spend LLM
+(`gpt-5.5-low`) calls **only where graph-μ is uncertain** — near-ties and structurally-distant candidates. The
+model's judge axis learns both calibrations side-by-side (`corpus_emb + judge_emb` provenance token).
+
+## 5. Integration + status
+
+- Feeds the graded round (`build_graded_round.py` → `load_graded`) as `corpus=mindmap` rows with per-row operator
+  + calibrated μ + judge tag.
+- SimpleMind is **most useful once fused with Pearltrees** (`fuse_corpus.py` needs the PT cache) — this is prep;
+  full value lands with the harvester.
+- **Built:** `gen_mindmap_lineage.py` (491 chains). **Next:** substitution+e5 candidate gen → graph judge →
+  targeted LLM judge → graded-round rows.
+
+See also: `DESIGN_path_operator.md`, `DESIGN_calibrated_judges.md`, `feedback_enwiki_sampling_strategy`.
--- a/prototypes/mu_cosine/DESIGN_mindmap_lineage.md
+++ b/prototypes/mu_cosine/DESIGN_mindmap_lineage.md
@@ -83,15 +83,13 @@ accounts for the **full materialized-path prefix**, via two complementary whole-
 
 - **LCA depth (structural):** how deep the *shared* prefix runs (shares-only-Root ⇒ ~0) — a topological measure
   of common ancestry.
-- **e5-of-prefix (semantic) — a TIE-BREAKER, not a multiplied factor.** A deep shared prefix (high LCA depth)
-  is *already* e5-similar, so multiplying both double-counts (review, critical #3). Instead, LCA-depth carries
-  the whole-lineage weight, and e5-of-prefix **only breaks ties among structurally-equivalent candidates** — two
-  paths with the same LCA depth / hop distance but different divergent tails (`Math/Calculus/…` vs
-  `Math/Analysis/…`) get separated by prefix e5-similarity. This keeps e5 as a *tie-breaker inside* the graph
-  heuristic, not e5 alone as μ (§2a caution); the LLM judge arbitrates the hard semantic cases.
+- **LCA-depth carries the whole-lineage *structural* weight** in `μ_graph` (below). e5-of-prefix does **not**
+  enter `μ_graph` — multiplying it against LCA-depth double-counts, since a deep shared prefix is *already*
+  e5-similar (review, critical #3). Instead **e5 is routed to a separate `lineage-rank` operator** (§3c), where
+  it is a *ranking* signal (cross-entropy), not a regression factor.
 
-So `μ_graph = decay(hops) · lca_depth_frac`, with **e5-prefix distance breaking ties** among structurally
-equivalent candidates (decay variant per the table below; exact tie-break margin TBD by the prototype).
+So `μ_graph = decay(hops) · lca_depth_frac` (pure structural; decay variant per the table below) — the regression
+target of the **`lineage`** operator. The semantic e5 signal goes to **`lineage-rank`** (§3c).
 
 **Decay choice — alternatives (we pick a default; these are the project's existing metrics).** The decay is
 exactly a flux/cost-function over the tree; see `docs/design/COST_FUNCTION_PHILOSOPHY.md` and
@@ -105,9 +103,47 @@ exactly a flux/cost-function over the tree; see `docs/design/COST_FUNCTION_PHILO
 | PPR / Markov walk | stationary dist. biased by the node seed | probabilistic dual; handles multi-path |
 | LCA depth | depth of lowest common ancestor | tree-native; "shares-only-Root ⇒ ~0" |
 
-Default: **hop-distance with geometric/exp decay + LCA depth as the tie-breaker** (cheap, tree-native). The
-power-law and PPR variants are the natural upgrades if the graded fall-off needs a heavier tail or multi-path
-mass — they are already specified in `COST_FUNCTION_PHILOSOPHY.md` and left as alternatives.
+Default: **hop-distance with geometric/exp decay + LCA depth** (cheap, tree-native). The power-law and PPR
+variants are the natural upgrades if the graded fall-off needs a heavier tail or multi-path mass — already
+specified in `COST_FUNCTION_PHILOSOPHY.md` and left as alternatives.
+
+### 3c. Two operators — `lineage` (MSE) and `lineage-rank` (CE)
+
+How e5 enters determines the loss: as a **scoring factor** it implies regression (error/MSE); as a **ranking**
+signal it implies cross-entropy. Rather than choose, split into two operators (two `op_emb` entries, routed
+per-operator like SYM/HIER/ELEM), each with its own loss — which also keeps e5 out of the regressed magnitude:
+
+- **`lineage` → MSE / regression.** Target = the pure structural `μ_graph = decay(hops) · lca_depth_frac`.
+  Learns the graded *magnitude* (parent > grandparent > far). No e5 → no LCA double-count.
+- **`lineage-rank` → cross-entropy / ranking.** Target = an *ordering* over a node's candidate parents, from a
+  **teacher** score (NOT the model's own output — self-referential bootstrapping drifts/collapses):
+
+  ```
+  rank_score(c) = μ_graph(c)  +  β · e5_sim_prefix(c)         # additive soft margin (dense e5 signal)
+  ```
+
+  Candidates are ordered by `rank_score`; `lineage-rank` is trained by softmax-CE to reproduce that order.
+  Because CE is **scale-invariant**, `lineage-rank` carries order information the MSE head can't — notably in
+  the flat graded-middle where the regression targets bunch. **This is the filing-primary operator** (filing =
+  rank the candidate folders); `lineage` supplies graded confidence / fall-off.
+
+**e5 is on the GRANDPARENT lineage (the prefix), not the immediate parent.** The immediate parent is the
+decision the structural `decay(hops)` already makes; e5 compares the *ancestral context above it* — so
+`e5_sim_prefix(c)` = e5 similarity of the `root … grandparent` prefix of candidate `c` vs the truth's. (`Math/
+Calculus/…` stays close to `Math/Analysis/…` because their grandparent prefixes match.)
+
+**β scales e5 by how much prefix there is to compare** (user):
+
+  ```
+  β = ( G / (L + G) ) · k
+      G = grandparent-prefix length (root … grandparent),   L = total path length (root … node),   k = tuned blend const
+  ```
+
+  So a shallow node with no grandparent (`G=0`) → **β=0**, e5 contributes nothing (nothing to compare); deeper
+  nodes → more ancestral context → larger β, up to `~0.5·k`. `k` is tuned on the pilot.
+
+**Losses:** `lineage` → the existing graded-MSE head; `lineage-rank` → a candidate-softmax CE (same family as
+the model's transitive/directional margins). Both are weighted like the other operators.
 
 ## 4. Two judges, complementary — and a cost-smart split
 
--- a/prototypes/mu_cosine/DESIGN_mindmap_lineage.md
+++ b/prototypes/mu_cosine/DESIGN_mindmap_lineage.md
@@ -119,28 +119,41 @@ per-operator like SYM/HIER/ELEM), each with its own loss — which also keeps e5
   **teacher** score (NOT the model's own output — self-referential bootstrapping drifts/collapses):
 
   ```
-  rank_score(c) = μ_graph(c)  +  β · e5_sim_prefix(c)         # additive soft margin (dense e5 signal)
+  rank_score(c) = (1 − β)·μ_graph(c)  +  β · e5_sim_prefix_norm(c)      # CONVEX combo — bounded [0,1]
+      e5_sim_prefix_norm = (cos_prefix + 1) / 2   ∈ [0,1]              # normalize cosine BEFORE mixing (review r2)
   ```
 
-  Candidates are ordered by `rank_score`; `lineage-rank` is trained by softmax-CE to reproduce that order.
-  Because CE is **scale-invariant**, `lineage-rank` carries order information the MSE head can't — notably in
-  the flat graded-middle where the regression targets bunch. **This is the filing-primary operator** (filing =
-  rank the candidate folders); `lineage` supplies graded confidence / fall-off.
+  **Convex, not additive (review round 2):** `μ_graph ∈ [floor,1]` and a raw cosine `∈ [−1,1]` have no shared
+  unit — an additive `μ_graph + β·cos` can go negative / exceed 1. The convex form keeps `rank_score ∈ [0,1]`,
+  makes **β an interpretable weight fraction** (requires `k ≤ 1`), and the `(cos+1)/2` normalization fixes the
+  sign. Candidates are ordered by `rank_score`; `lineage-rank` is trained by softmax-CE to reproduce that order.
+  Because CE is **scale-invariant**, `lineage-rank` carries order info the MSE head can't — notably in the flat
+  graded-middle where regression targets bunch (weakest, conversely, for deep trees with diverse `μ_graph`, where
+  MSE already has a strong gradient). **Filing-primary operator** (filing = rank the folders); `lineage` supplies
+  graded confidence. The "filing-primary" dominance belongs in the **`op_emb` routing** at inference, not just
+  this narrative. **Pilot diagnostics:** log per-example **teacher softmax entropy** (near-uniform ⇒ tiny CE
+  gradient — the real collapse risk, from *flat teacher distributions*, not operator correlation) and per-depth
+  MSE-gradient-vs-CE-entropy.
 
 **e5 is on the GRANDPARENT lineage (the prefix), not the immediate parent.** The immediate parent is the
 decision the structural `decay(hops)` already makes; e5 compares the *ancestral context above it* — so
 `e5_sim_prefix(c)` = e5 similarity of the `root … grandparent` prefix of candidate `c` vs the truth's. (`Math/
-Calculus/…` stays close to `Math/Analysis/…` because their grandparent prefixes match.)
+Calculus/…` stays close to `Math/Analysis/…` because their grandparent prefixes match.) Since LCA-depth (in
+`lineage`, MSE) and e5-prefix (in `lineage-rank`, CE) now live in **different operators with different losses**,
+their correlation is **redundancy, not double-count** (review r2). *Degenerate case:* when the candidate's
+grandparent **is** the LCA, both CE inputs coincide → zero CE gradient — log its frequency in the pilot.
 
 **β scales e5 by how much prefix there is to compare** (user):
 
   ```
   β = ( G / (L + G) ) · k
-      G = grandparent-prefix length (root … grandparent),   L = total path length (root … node),   k = tuned blend const
+      G = grandparent-prefix length in NODES (root … grandparent, node count — same unit as lca_depth_frac)
+      L = total path length in nodes (root … node)          k = tuned blend const, k ≤ 1 (convex ⇒ β ≤ 0.5)
   ```
 
-  So a shallow node with no grandparent (`G=0`) → **β=0**, e5 contributes nothing (nothing to compare); deeper
-  nodes → more ancestral context → larger β, up to `~0.5·k`. `k` is tuned on the pilot.
+  `G=0` (shallow node, no grandparent) → **β=0**: `lineage-rank` adds no signal over `lineage` there (fine —
+  state it). Deeper nodes → more ancestral context → larger β, asymptoting at `~0.5·k`; `k ≤ 1` keeps `β ≤ 0.5`
+  (and `rank_score` a proper convex mix). `k` tuned on the pilot.
 
 **Losses:** `lineage` → the existing graded-MSE head; `lineage-rank` → a candidate-softmax CE (same family as
 the model's transitive/directional margins). Both are weighted like the other operators.
@@ -162,12 +175,15 @@ model's judge axis learns both calibrations side-by-side (`corpus_emb + judge_em
 - SimpleMind is **most useful once fused with Pearltrees** (`fuse_corpus.py` needs the PT cache) — this is prep;
   full value lands with the harvester.
 - **Built:** `gen_mindmap_lineage.py` (491 chains); `prototype_graph_judge.py` — graph-μ validated on Chaos
-  theory: true parent = 1.0, ancestors graded up the chain (2 hops ≈ 0.36, 3 ≈ 0.22 at γ=0.6), far → floor 0.02.
-  **Finding:** a single-map parse is **fragmented** (many candidate `hops=99`), so the graph judge needs the
-  **connected cross-map / fused graph** for real candidate connectivity — another reason SimpleMind pairs with
-  Pearltrees (`fuse_corpus.py`). Nav (`via link`) nodes are now bypassed in the *structure* (not just display).
-  **Next:** run on the fused graph → substitution candidate gen → targeted LLM judge on uncertain cases →
-  graded-round rows.
+  theory: true parent = 1.0, siblings/near graded (1 hop ≈ 0.60, 2 ≈ 0.36 at γ=0.6), far → floor 0.02.
+  Candidate pool excludes **ancestors of n** (they're graded positives, not negatives — review r2).
+  **Finding:** a single-map parse is **fragmented** (117 unreachable candidates on Chaos theory), so the graph
+  judge needs the **connected cross-map / fused graph** for real candidate connectivity — another reason
+  SimpleMind pairs with Pearltrees (`fuse_corpus.py`). **Interim single-map fallback (review r2):** unreachable
+  candidates get `hops = D_max + 1` (not a sentinel), so `decay(hops)·lca_frac` still degrades gracefully —
+  `lca_frac` differentiates them by shared ancestry while the harvester is pending. Nav (`via link`) nodes are
+  bypassed in the *structure* (not just display). **Next:** run on the fused graph → substitution candidate gen
+  → targeted LLM judge on uncertain cases → graded-round rows.
 
 ## 6. Review responses (PR #3426)
 
@@ -178,7 +194,9 @@ one alternative (heavier tail). Pinned to one family, not two.
 **Depth bias (methodology-2):** complete enumeration gives C(8,2)=28 pairs for a depth-8 chain vs 1 for depth-2,
 so deep systems-theory nodes would dominate. Mitigation: **per-chain pair weighting** (down-weight each pair by
 `1/(#pairs in its chain)` so every chain contributes equal mass), OR cap transitive pairs (keep all adjacent +
-sample K transitive). The pair generator will emit **pair-count-by-depth** and apply the weight.
+sample K transitive). The pair generator will emit **pair-count-by-depth** and apply the weight. NB per-chain
+weighting shifts the *effective* negative type-ratio by depth, so log the **weighted** negative-type
+distribution, not just raw counts (review r2).
 
 **Negative μ-floor (methodology-3):** since all nodes share the global root, **negatives are NOT μ=0.**
 Directional-reverse (ancestor|descendant) → `μ_rev ≈ 0.05–0.15` (residual membership); cross-branch →
@@ -199,8 +217,10 @@ hard negatives + judge-calibrated μ *on top of* the lineage positives, it does 
 
 **Multi-parent, operationalized (D3):** graded-round schema adds `gt_rank` (rank of the ground-truth filing
 among candidates by judge μ) + `gt_mu`. Disposition: `gt_rank==1` → normal; `gt_rank==2 & spread<τ_tie` → **keep
-both** (multi-parent); `gt_rank≥3` OR `gt_mu<τ_low` → **flag + exclude from training**. `τ_tie≈0.1`, `τ_low≈0.3`
-(μ scale), tuned on the pilot.
+both** (multi-parent); `gt_rank≥3` OR `gt_mu<τ_low` → **flag + exclude from training**. `τ_tie≈0.1`; **`τ_low` =
+the 10th percentile of the true-parent μ distribution from the prototype run, NOT a fixed 0.3** (review r2 —
+a fixed 0.3 would wrongly flag legitimate weak-but-correct parents, e.g. 2-hop with moderate lca_frac). Tuned on
+the pilot.
 
 **Cost-split + judge_agreement (D4, methodology-7):** output stores `mu_graph`, `mu_llm`, and
 `judge_agreement = abs(mu_graph − mu_llm)`. LLM spent only where graph-μ is uncertain: **near-tie** = top-two
--- a/prototypes/mu_cosine/DESIGN_mindmap_lineage.md
+++ b/prototypes/mu_cosine/DESIGN_mindmap_lineage.md
@@ -125,8 +125,13 @@ model's judge axis learns both calibrations side-by-side (`corpus_emb + judge_em
   + calibrated μ + judge tag.
 - SimpleMind is **most useful once fused with Pearltrees** (`fuse_corpus.py` needs the PT cache) — this is prep;
   full value lands with the harvester.
-- **Built:** `gen_mindmap_lineage.py` (491 chains). **Next:** substitution+e5 candidate gen → graph judge →
-  targeted LLM judge → graded-round rows.
+- **Built:** `gen_mindmap_lineage.py` (491 chains); `prototype_graph_judge.py` — graph-μ validated on Chaos
+  theory: true parent = 1.0, ancestors graded up the chain (2 hops ≈ 0.36, 3 ≈ 0.22 at γ=0.6), far → floor 0.02.
+  **Finding:** a single-map parse is **fragmented** (many candidate `hops=99`), so the graph judge needs the
+  **connected cross-map / fused graph** for real candidate connectivity — another reason SimpleMind pairs with
+  Pearltrees (`fuse_corpus.py`). Nav (`via link`) nodes are now bypassed in the *structure* (not just display).
+  **Next:** run on the fused graph → substitution candidate gen → targeted LLM judge on uncertain cases →
+  graded-round rows.
 
 ## 6. Review responses (PR #3426)
 
--- a/prototypes/mu_cosine/DESIGN_mindmap_lineage.md
+++ b/prototypes/mu_cosine/DESIGN_mindmap_lineage.md
@@ -76,6 +76,20 @@ Score a candidate by its **graph distance to the true filing**, decayed to μ (t
 maps to μ **directly and honestly** — distance-to-truth *is* a plausibility. Tag `judge=graph`
 (`JUDGES["graph"] = 1`, already defined).
 
+**Whole-lineage, not just the immediate parent.** A bare parent-to-parent hop is myopic: two candidates can be
+equidistant from the true parent yet sit in very different lineages (`Math/Calculus/X` vs `Math/Analysis/X` — one
+leaf-hop apart but nearly the same folder path — versus one hop into an unrelated subtree). So the graph-μ
+accounts for the **full materialized-path prefix**, via two complementary whole-lineage factors:
+
+- **LCA depth (structural):** how deep the *shared* prefix runs (shares-only-Root ⇒ ~0) — a topological measure
+  of common ancestry.
+- **e5-of-prefix (semantic):** e5-embed the candidate's path prefix (`root / … / parent`) and the truth's, and
+  factor in their e5 distance — so `Math/Calculus/…` stays close to `Math/Analysis/…` despite the differing
+  immediate parent. This keeps e5 as **one factor inside the graph heuristic**, not e5 alone as μ (§2a caution);
+  the LLM judge still arbitrates the hard semantic cases.
+
+So `μ_graph ≈ decay(hops) · lca_depth_frac · f(e5_prefix_dist)` (exact combination TBD by the prototype).
+
 **Decay choice — alternatives (we pick a default; these are the project's existing metrics).** The decay is
 exactly a flux/cost-function over the tree; see `docs/design/COST_FUNCTION_PHILOSOPHY.md` and
 `docs/design/SCAN_STRATEGY_SPECIFICATION.md`:
--- a/prototypes/mu_cosine/DESIGN_mindmap_lineage.md
+++ b/prototypes/mu_cosine/DESIGN_mindmap_lineage.md
@@ -83,12 +83,15 @@ accounts for the **full materialized-path prefix**, via two complementary whole-
 
 - **LCA depth (structural):** how deep the *shared* prefix runs (shares-only-Root ⇒ ~0) — a topological measure
   of common ancestry.
-- **e5-of-prefix (semantic):** e5-embed the candidate's path prefix (`root / … / parent`) and the truth's, and
-  factor in their e5 distance — so `Math/Calculus/…` stays close to `Math/Analysis/…` despite the differing
-  immediate parent. This keeps e5 as **one factor inside the graph heuristic**, not e5 alone as μ (§2a caution);
-  the LLM judge still arbitrates the hard semantic cases.
+- **e5-of-prefix (semantic) — a TIE-BREAKER, not a multiplied factor.** A deep shared prefix (high LCA depth)
+  is *already* e5-similar, so multiplying both double-counts (review, critical #3). Instead, LCA-depth carries
+  the whole-lineage weight, and e5-of-prefix **only breaks ties among structurally-equivalent candidates** — two
+  paths with the same LCA depth / hop distance but different divergent tails (`Math/Calculus/…` vs
+  `Math/Analysis/…`) get separated by prefix e5-similarity. This keeps e5 as a *tie-breaker inside* the graph
+  heuristic, not e5 alone as μ (§2a caution); the LLM judge arbitrates the hard semantic cases.
 
-So `μ_graph ≈ decay(hops) · lca_depth_frac · f(e5_prefix_dist)` (exact combination TBD by the prototype).
+So `μ_graph = decay(hops) · lca_depth_frac`, with **e5-prefix distance breaking ties** among structurally
+equivalent candidates (decay variant per the table below; exact tie-break margin TBD by the prototype).
 
 **Decay choice — alternatives (we pick a default; these are the project's existing metrics).** The decay is
 exactly a flux/cost-function over the tree; see `docs/design/COST_FUNCTION_PHILOSOPHY.md` and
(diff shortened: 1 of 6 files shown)


prototypes/mu_cosine/DESIGN_mindmap_lineage.md
```diff
@@ -0,0 +1,128 @@
...
+
+## 1. Positives — complete, and graded by distance
+
+The dataset is small (491 lineage chains over 6 maps), so we take **complete coverage of positives**: every
+(descendant, ancestor) pair along each materialized path. They are **not binary** — μ is **graded by lineage
+distance**: direct parent = strong, grandparent = weaker, great-grandparent = weaker still. The depth gives
+calibrated positives for free.
+
+Materialized paths are cleaned (`gen_mindmap_lineage.py`): the shared sentinel `Root Node` (in folder `root`,
+a generic placeholder to which all maps link up) is dropped, `via link` nav nodes filtered, cross-map join
```
s243a, line 22: **[Methodology] Positive grading: distance function not yet specified**

The design states μ is graded by lineage distance (direct parent = strong, grandparent = weaker, etc.) but doesn't pin the decay curve for positives. For consistency, this should reference the same decay family chosen for the graph judge in §3b (geometric `γ^d`, harmonic `1/(1+d)`, etc.). Two options:
1. Commit to one curve here explicitly (e.g. `μ_pos = γ^(depth-1)`, same γ as the graph judge default).
2. Keep it open and note it as a prototype-tuned parameter — but then the design is incomplete until §3b's decay is finalized.

Recommend option 1 for a self-consistent design doc.

prototypes/mu_cosine/DESIGN_mindmap_lineage.md
```diff
@@ -0,0 +1,128 @@
...
+why hard sibling/cousin negatives matter more than random cross-map ones (the shared root already makes those
+mildly positive).
+
+### 2a. Candidate generation — substitution, e5-stratified
+
+Generate hard negatives by **substituting** a node in a true path with an alternative (an alternate parent/path
+— exactly the filing question "could it go here instead?"). **Permutation** (scrambling path order) mostly
+yields obviously-broken hierarchies — kept only as a cheap easy-negative floor.
+
+Control hardness with **e5 distance of the substituted node vs the original**: e5-near = sibling/cousin = hard
```
s243a, line 48: **[Methodology] 3:1 negative ratio from enwiki may not transfer directly**

The design references enwiki's 3:1 negative sampling ratio. With only 491 lineage chains across 6 maps, the positive set is far smaller than enwiki's. The practical concern is **class imbalance at training time**: with hard negatives being the most valuable type (siblings/cousins), you want enough hard negative rows to keep that signal from being drowned by easy cross-branch ones.

Suggestion: explicitly specify the **type-level breakdown** of the 3:1 ratio, e.g.:
- 1x directional-reverse (free)
- 1x cross-branch (easy tail)
- 1x sibling/cousin (hard)

or weight toward hard negatives 2:1 over easy ones. The design should state this explicitly rather than inheriting enwiki's ratio wholesale.

prototypes/mu_cosine/DESIGN_mindmap_lineage.md
```diff
@@ -0,0 +1,128 @@
...
+negative; e5-far = easy. **e5 distance is used to *stratify candidates*, not as μ** — semantic distance ≠ filing
+plausibility (a sibling can be e5-close yet filing-wrong; a valid alternate parent can be e5-far). The μ comes
+from a judge (§3).
+
+## 3. Judge scoring
+
+Remove the last hop of a path → (path-minus-leaf = folder, leaf = item to file). Each example is thus **actual
+filing training data** — more target-aligned than raw lineage pairs, and buildable now without the harvester.
+Present the judge ≥2 candidate filings (2 = cheapest; 3–4 = better-shaped curve), scored **pointwise and
```
s243a, line 57: **[Methodology] e5-stratification: boundary thresholds for hard/easy not defined**

The design correctly separates e5 distance as a *stratification* tool (not as μ), but leaves the hardness boundary undefined. At prototype time you'll need to decide: what e5 cosine distance threshold separates "sibling/cousin hard" from "cross-branch easy"? Options:
- Percentile-based (e.g. bottom quartile of pairwise e5 distances = hard).
- Absolute threshold from a pilot sample.
- Adaptive per-map (given the 6 maps span different semantic density).

The per-map adaptive approach is safest given the small, heterogeneous corpus. Worth flagging here even if the exact value is left to prototype tuning.

prototypes/mu_cosine/DESIGN_mindmap_lineage.md
```diff
@@ -0,0 +1,128 @@
...
+  `mu_attention.py`). Cost/latency: `gpt-5.5-low` ≈ 5 s/pair, non-Anthropic.
+
+### 3b. Graph judge (nearly free) — `μ = decay( distance-to-truth )`
+
+Score a candidate by its **graph distance to the true filing**, decayed to μ (truth at distance 0 → **1**;
+1 hop = sibling/uncle → high; far → low). This is the **structural** analogue of the e5 idea, and unlike e5 it
+maps to μ **directly and honestly** — distance-to-truth *is* a plausibility. Tag `judge=graph`
+(`JUDGES["graph"] = 1`, already defined).
+
+**Whole-lineage, not just the immediate parent.** A bare parent-to-parent hop is myopic: two candidates can be
```
s243a, line 79: **[Methodology] Multi-parent guard: "flag if ground-truth scores low" needs an operationalization**

The sanity floor is well-reasoned, but "flag" is under-specified for an automated pipeline:
- What μ threshold triggers the flag? (e.g. `μ_true < 0.3` when the model's top pick = 1?)
- What happens to flagged examples? Skip them entirely, keep them at their LLM-assigned μ, or put them in a review queue?
- Is this flag a field in the graded-round output row (`corpus=mindmap`) so `build_graded_round.py` can filter/inspect?

Suggest adding a `low_score_flag: bool` column to the graded-round schema and defining the threshold (even if TBD) so the downstream pipeline knows what to do.

prototypes/mu_cosine/DESIGN_mindmap_lineage.md
```diff
@@ -0,0 +1,128 @@
...
+
+So `μ_graph ≈ decay(hops) · lca_depth_frac · f(e5_prefix_dist)` (exact combination TBD by the prototype).
+
+**Decay choice — alternatives (we pick a default; these are the project's existing metrics).** The decay is
+exactly a flux/cost-function over the tree; see `docs/design/COST_FUNCTION_PHILOSOPHY.md` and
+`docs/design/SCAN_STRATEGY_SPECIFICATION.md`:
+
+| variant | form | note |
+|---|---|---|
+| **hop / shortest-path** (our default) | `μ = γ^d`, or `1/(1+d)` | simplest; no flux interpretation |
```
s243a, line 100: **[Methodology] μ_graph formula: potential double-counting between LCA depth and e5-prefix distance**

As the PR itself flags, `μ_graph ≈ decay(hops) · lca_depth_frac · f(e5_prefix_dist)` risks double-counting because LCA depth and e5-prefix distance are **correlated**: a deeper shared prefix generally implies closer e5 embeddings (at least within a single domain). Multiplying them as independent factors will over-penalize structurally-distant candidates.

Two mitigation approaches:
1. **Conditional independence check**: compute `corr(lca_depth_frac, 1 - e5_prefix_dist)` on the pilot data. If Spearman |r| > 0.5, drop one factor or use a sub-additive combination (`max()` or a weighted log-sum).
2. **Treat e5-prefix only as a tie-breaker** (not a multiplicative factor): use it only when `lca_depth_frac` ties or is near-zero, matching how it's described verbally ("Math/Calculus vs Math/Analysis" example).

Option 2 aligns better with the verbal intuition and avoids the double-counting entirely.

prototypes/mu_cosine/DESIGN_mindmap_lineage.md
```diff
@@ -0,0 +1,128 @@
...
+| PPR / Markov walk | stationary dist. biased by the node seed | probabilistic dual; handles multi-path |
+| LCA depth | depth of lowest common ancestor | tree-native; "shares-only-Root ⇒ ~0" |
+
+Default: **hop-distance with geometric/exp decay + LCA depth as the tie-breaker** (cheap, tree-native). The
+power-law and PPR variants are the natural upgrades if the graded fall-off needs a heavier tail or multi-path
+mass — they are already specified in `COST_FUNCTION_PHILOSOPHY.md` and left as alternatives.
+
+## 4. Two judges, complementary — and a cost-smart split
+
+Graph judge = cheap, structural, but **blind to a semantically-valid parent that is structurally distant**
```
s243a, line 163: **[Methodology] Graph judge decay table: missing normalization constraint**

The decay variant table is excellent as a reference. One missing detail: for all variants, the design should state explicitly that `decay(0) = 1` (truth candidate = μ 1) and `lim_{d→∞} decay(d) = 0`. This is implied but worth asserting formally so that if/when the formula is coded the boundary conditions are enforced in tests.

Also, for the **LCA-depth tie-breaker row**: should `lca_depth_frac` be normalized by the *tree depth* (depth of the true node's full path) rather than absolute depth? On shallow maps (6 total maps, some may be only 3-4 deep) absolute LCA depth is a coarse discriminator.

prototypes/mu_cosine/DESIGN_mindmap_lineage.md
```diff
@@ -0,0 +1,128 @@
...
+(a cross-domain-but-correct filing). LLM judge = semantic, catches those, but costs. Their **disagreement is the
+useful part**: agree → confident μ; diverge → the ambiguous, probably-multi-parent cases.
+
+**Cost-smart pipeline:** run the **graph judge over everything (free, `judge=graph`)**, then spend LLM
+(`gpt-5.5-low`) calls **only where graph-μ is uncertain** — near-ties and structurally-distant candidates. The
+model's judge axis learns both calibrations side-by-side (`corpus_emb + judge_emb` provenance token).
+
+## 5. Integration + status
+
```
s243a, line 172: **[Methodology] Cost-smart split: uncertainty threshold for LLM escalation not defined**

The cost-smart pipeline (graph judge over everything, LLM only on uncertain graph-μ) is the right approach. But "uncertain" needs a concrete criterion for the prototype:
- Near-tie: how close is "near"? e.g. `|μ_graph(A) - μ_graph(B)| < 0.15`?
- Structurally-distant: what hop count triggers LLM escalation? e.g. `d > 3`?

Without this, the prototype will spend LLM calls inconsistently. Suggest adding a placeholder threshold in the design (to be tuned) and storing `graph_uncertain: bool` in the output row so the calibration data later shows whether the split was well-calibrated.

prototypes/mu_cosine/gen_mindmap_lineage.py
```diff
@@ -0,0 +1,114 @@
...
+    title = {}
+    for ln in open(pref + "_nodes.tsv", encoding="utf-8"):
+        if ln.startswith("#"):
+            continue
+        c = ln.rstrip("\n").split("\t")
+        if c and c[0]:
+            title[c[0]] = c[2] if len(c) > 2 and c[2] else c[0]
+    struct_parent, broader = {}, {}      # keep structural (subtopic) separate from super_category fallback
+    for ln in open(pref + "_edges.tsv", encoding="utf-8"):
+        if ln.startswith("#"):
```
s243a: **[Code] `struct_parent.setdefault` silently drops multi-parent structural nodes**

`setdefault(dst, src)` takes only the *first* structural parent when a node has multiple `subtopic`/`subcategory` edges (a pathological but possible case in `.smmx` maps with aliased nodes). This is consistent with the single-path lineage design but should be **logged** rather than silently dropped: a node with multiple structural parents is a data quality signal worth surfacing in the stats output.

Suggestion:
```python
if dst in struct_parent and struct_parent[dst] != src:
    multi_parent_nodes.add(dst)  # log at end
struct_parent.setdefault(dst, src)
```

prototypes/mu_cosine/gen_mindmap_lineage.py
```diff
@@ -0,0 +1,114 @@
...
+            continue
+        c = ln.rstrip("\n").split("\t")
+        if len(c) < 3:
+            continue
+        src, dst, rel = c[0], c[1], c[2]
+        if rel in ("subtopic", "subcategory"):
```
s243a: **[Code] Parent-map merge order: `super_category` fallback may silently activate**

```python
parent = dict(broader); parent.update(struct_parent)
```
This is correct — structural parent wins. However, there's no logging when the `super_category` fallback is *actually used* for a node (i.e., when `broader[k]` is present but `struct_parent[k]` is absent). Since `super_category` relationships imply a *different semantic* (broadening vs. containment), lineages that travel through a `super_category` hop will be structurally different from pure subtopic chains.

Suggest counting and logging `broader`-only nodes in the stats output so you know what fraction of lineage chains contain a `super_category` hop. This matters for the judge: a `super_category`-rooted lineage may need different μ grading.

prototypes/mu_cosine/gen_mindmap_lineage.py
```diff
@@ -0,0 +1,114 @@
...
+            struct_parent.setdefault(dst, src)          # first structural parent wins (principal)
+        elif rel == "super_category":
+            broader.setdefault(src, dst)
+    parent = dict(broader); parent.update(struct_parent)  # structural parent takes precedence
+    return title, parent, ""
```
s243a: **[Code] Cycle detection in `lineage()` is incomplete for multi-root maps**

The `seen` set prevents infinite loops correctly for single trees. However, for `.smmx` maps that import or link other maps (cross-map join echoes are handled at title-cleaning time, not here), a structural cycle could exist in the raw `parent` dict *before* title-cleaning. If `parse_smmx.py` ever emits a cycle in `edges.tsv`, this function will terminate correctly but the `chain` will be short and silently dropped at `len(clean) < 2`.

Minor hardening suggestion: add a `max_depth` guard (e.g. `len(chain) > 100: break`) and log a warning, so pathological inputs don't silently produce zero chains for a map.

prototypes/mu_cosine/score_with_codex.py
```diff
@@ -0,0 +1,72 @@
...
+        return p.stdout + p.stderr
+    except subprocess.TimeoutExpired:
+        return ""
+
+
+def first_json_array(txt):
+    # codex double-prints (echo + response); take the FIRST complete [ {...} ] array
+    m = re.search(r'\[\s*\{.*?\}\s*\]', txt, re.S)
+    return m.group(0) if m else None
+
```
s243a, line 51: **[Code] `first_json_array` greedy regex: nested objects or multiple arrays will break**

```python
m = re.search(r'\[\s*\{.*?\}\s*\]', txt, re.S)
```
With `re.S` and `.*?` (lazy), this matches the *first* `[ {...} ]` pattern but will fail if:
1. The JSON array contains nested objects (`{"a": {"b": 1}}`), because the closing `}` for the inner object terminates the match early.
2. Codex returns two separate arrays (retry/echo).

A more robust approach is to attempt `json.loads` on all bracket-delimited substrings:
```python
import json
def first_json_array(txt):
    for m in re.finditer(r'\[', txt):
        try:
            obj, _ = json.JSONDecoder().raw_decode(txt, m.start())
            if isinstance(obj, list) and obj:
                return json.dumps(obj)
        except json.JSONDecodeError:
            pass
    return None
```
This handles nested objects and is more robust to codex's double-print behavior.

prototypes/mu_cosine/score_with_codex.py
```diff
@@ -0,0 +1,72 @@
...
+    for i, pr in enumerate(prompts):
+        t = time.time()
+        arr = first_json_array(call_codex(pr, a.timeout))
+        dt = time.time() - t
+        if arr:
+            arrays.append(arr); ok += 1
+            status = "ok"
+        else:
+            status = "FAIL(no-json/timeout)"
+        rate = (ok * a.batch) / max(1e-9, time.time() - t0)
```
s243a: **[Code] `ingest()` call passes `a.pairs` regardless of `--limit`**

```python
ingest(a.pairs if not a.limit else a.pairs, a.responses, a.out, judge=JUDGE)
```
Both branches of the ternary are identical (`a.pairs`). If `--limit` is set, `ingest` will try to reconcile a truncated response file against the full pairs file, which may produce mismatched row counts or silently assign scores to wrong pairs depending on how `ingest` works internally.

Fix:
```python
pairs_path = a.pairs  # already loaded as `pairs[:a.limit]` above, but ingest takes the file path
# Either slice the file first, or pass the in-memory `pairs` list if ingest supports it
```
Since `ingest` takes a file path, the cleanest fix is to write the limited pairs to a temp file and pass that, or document that `--limit` is a *sampling* flag that only affects codex calls but ingest still uses the full file (and clarify that in the help text).

prototypes/mu_cosine/mu_attention.py
```diff
@@ -55,7 +55,7 @@
 # (off-manifold noise, exactly like an absent ancestor slot); masking ⇒ provenance-AGNOSTIC μ, which is
 # the DEFAULT inference path (marginalize over sources). Reserved entries (enwiki) are for later corpora.
 CORPORA = {"simplewiki": 0, "enwiki": 1, "pearltrees": 2, "mindmap": 3}
-JUDGES = {"haiku": 0, "graph": 1, "human": 2, "sonnet": 3, "opus": 4}    # haiku = a bought LLM judgment
+JUDGES = {"haiku": 0, "graph": 1, "human": 2, "sonnet": 3, "opus": 4, "gemini": 5, "gpt-5.5-low": 6}    # judge_emb is a LEARNED table (nn.Embedding, ~line 329), keyed by these indices — each (MODEL, reasoning-EFFORT) = own calibration row → own tag (low≠xhigh; they disagreed on labels). gpt-5.5-low = `codex exec -m gpt-5.5 -c model_reasoning_effort=low`. Cheaper future judges (gpt-5.3-low, gpt-5.4-low, …) get 7,8,… Tag pairs by the exact judge that scored them, NEVER default to haiku.
```
s243a: **[Code] JUDGES comment line too long; will cause lint/readability issues**

The new `JUDGES` line is very long (the inline comment now contains a full design rationale). Consider extracting the extended rationale to the associated `DESIGN_calibrated_judges.md` and keeping the inline comment brief:
```python
JUDGES = {"haiku": 0, "graph": 1, "human": 2, "sonnet": 3, "opus": 4, "gemini": 5, "gpt-5.5-low": 6}
# judge_emb: nn.Embedding keyed by these indices. Each (model, reasoning-effort) = own calibration row.
# gpt-5.5-low = `codex exec -m gpt-5.5 -c model_reasoning_effort=low`. See DESIGN_calibrated_judges.md.
```
This keeps the schema clean and the rationale co-located with its design doc.

s243a (commented):
## Academic Review — Mindmap LINEAGE sampler + graph/LLM judge methodology

### Overall assessment

This is a well-structured methodology PR. The scope is clearly bounded (sampler + design doc, no implemented negatives/judges yet), the design rationale is sound, and the key architectural decisions — graded μ, substitution over permutation, e5 as a stratifier not μ itself, dual-judge cost split, whole-lineage graph-μ — are all well-motivated. The `gen_mindmap_lineage.py` sampler is clean and correct for its stated purpose.

The inline comments flag **10 issues** across three categories. None block the methodology, but several should be resolved before the negative-generation prototype begins.

---

### Critical / must address before prototype

1. **`first_json_array` regex fragility** (`score_with_codex.py` line 33) — the greedy `re.S` + `.*?` pattern breaks on nested objects in the JSON response. A `json.JSONDecoder().raw_decode()` scan is more robust. This will silently drop scored batches in production.

2. **`ingest()` bug with `--limit`** (`score_with_codex.py` line 62) — both branches of the ternary pass `a.pairs` (the full path), so `--limit` truncates codex calls but not ingest, producing score/pair mismatches. Needs a clear fix or a doc clarification that `--limit` is sampling-only.

3. **`μ_graph` double-counting** (`DESIGN_mindmap_lineage.md` line 99) — LCA depth and e5-prefix distance are correlated. Multiplying them as independent factors over-penalizes structurally-distant candidates. Recommend either a pilot correlation check (drop one if |r| > 0.5) or treating e5-prefix as a tie-breaker only (not a multiplicative factor), which matches the verbal Math/Calculus vs Math/Analysis intuition better.

---

### Should address / design completeness

4. **Positive decay curve not specified** (DESIGN §1) — the doc says μ is graded by lineage distance but doesn't pin the curve. Should reference the same decay family as the graph judge default for internal consistency.

5. **Negative type-level breakdown missing** (DESIGN §2) — inheriting enwiki's 3:1 ratio without specifying the split across directional-reverse / cross-branch / sibling types is underspecified for a small 491-chain corpus. Recommend weighting toward hard negatives (siblings/cousins) explicitly.

6. **Multi-parent flag not operationalized** (DESIGN §3a) — "flag if ground-truth scores low" needs a threshold and a disposition (skip / keep / review queue). Suggest a `low_score_flag` column in the graded-round schema.

7. **Cost-smart split uncertainty threshold missing** (DESIGN §4) — "near-tie" and "structurally-distant" need concrete thresholds for the prototype to be reproducible.

---

### Minor / nits

8. **e5 stratification boundary undefined** (DESIGN §2a) — hard/easy boundary thresholds should at least be noted as TBD with a suggested approach (percentile-based vs. adaptive per-map).

9. **`super_category` fallback used silently** (`gen_mindmap_lineage.py` line 50) — chains that travel through a `super_category` hop have different semantics (broadening vs. containment). Log the count so you know the fraction and can decide if they need separate grading.

10. **`JUDGES` inline comment too long** (`mu_attention.py` line 58) — the extended rationale belongs in `DESIGN_calibrated_judges.md`; keep the inline comment to 2 lines max.

---

### Positive notes

- The **substitution-over-permutation** negative strategy is well-reasoned and directly mirrors the real filing task. Strong choice.
- The **whole-lineage graph-μ** (LCA + e5-of-prefix) correctly addresses the myopic-parent-hop problem. The Math/Calculus vs Math/Analysis example is a perfect illustration of why bare hop-distance fails.
- **`struct_parent` taking precedence over `broader`** in the parent merge is the right call; the single-path lineage invariant is preserved correctly.
- The decay variant table is a great design artifact — connecting this PR's choices back to `COST_FUNCTION_PHILOSOPHY.md` is exactly the right kind of documentation cross-linking.

prototypes/mu_cosine/DESIGN_mindmap_lineage.md
```diff
@@ -0,0 +1,128 @@
...
+## 0. Why mindmaps, why lineage first
+
+The SimpleMind maps are the user's own systems-theory domain — structurally and topically the closest proxy we
+have to the Pearltrees filing target, and **not** present in enwiki. We build **LINEAGE first, then the 70/30
+split on top of it** (directional/superposition — see `feedback_enwiki_sampling_strategy`). Lineage is
+well-defined here because a mindmap node hangs off exactly one structural parent — the same property that made us
+choose single-path LINEAGE for Pearltrees (`project_mu_cosine_path_multipath_result`).
+
+## 1. Positives — complete, and graded by distance
```
s243a, line 14: **[Design] Why "LINEAGE first, then 70/30 split" — directionality implication needs elaboration**

The ordering decision (lineage before the 70/30 directional/superposition split) is stated as a user direction and referenced via `feedback_enwiki_sampling_strategy`, but the *reason* it matters is not articulated here. The key implication is that computing lineage on the full undirected graph and *then* applying the 70/30 split means the split operates on pre-formed chains rather than raw edges — which affects the distribution of directional vs. superposition pairs.

If the 70/30 split is applied *after* lineage enumeration, the "direction" dimension is already baked into which ancestor is the query and which is the context, so the split is really about sampling *which direction* to present each (ancestor, descendant) pair. This is subtly different from the enwiki case where directions are intrinsic to edge types.

Suggest adding 1-2 sentences here explaining the semantic consequence: applying split after lineage means the 30% "superposition" slice represents ancestor-as-query (reverse lookups), not structurally distinct edge types.

prototypes/mu_cosine/DESIGN_mindmap_lineage.md
```diff
@@ -0,0 +1,128 @@
...
+Materialized paths are cleaned (`gen_mindmap_lineage.py`): the shared sentinel `Root Node` (in folder `root`,
+a generic placeholder to which all maps link up) is dropped, `via link` nav nodes filtered, cross-map join
+echoes (consecutive duplicate titles) collapsed. Privacy is scrubbed upstream by `parse_smmx.py`
+(`privacy.py::is_private_title`: a "private" node drops itself + subtree, a private root drops the whole map).
+
+## 2. Negatives — sampled, typed, graded (not enumerated)
+
+The negative complement is ~O(N²) and would swamp the positives, so we **sample** (enwiki uses 3:1). Three types
+in increasing value:
+
```
s243a, line 30: **[Design] "Complete enumeration" of positives on a small dataset: combinatorial explosion within deep chains**

For 491 chains, complete enumeration of every (descendant, ancestor) pair means chains of depth D contribute `D*(D-1)/2` pairs. If some systems-theory maps are deeply nested (depth 8-10 is plausible), a single chain of depth 10 contributes 45 pairs. Across 491 chains, if average depth is even 4, you're looking at ~3000 pairs from positives alone.

The concern isn't the absolute count — it's that **long chains will be over-represented** in the positive training set: a depth-8 node contributes 7 ancestor pairs, while a depth-2 node contributes 1. This creates an implicit depth bias in the training distribution.

Two mitigations worth flagging in the design:
1. **Depth-stratified sampling even for positives** — cap the number of positive pairs sampled per chain (e.g. sample at most 3 ancestor pairs per node, weighted toward nearer ancestors).
2. **Document the actual depth distribution** — `gen_mindmap_lineage.py` already prints `depth distribution` in its stats output. Worth logging the pair count per depth level, not just chain count.

prototypes/mu_cosine/DESIGN_mindmap_lineage.md
```diff
@@ -0,0 +1,128 @@
...
+3. **Hard negatives — siblings / cousins** (same parent/grandparent, *not* ancestor-descendant). The valuable
+   ones: they teach μ that *same branch ≠ lineage*, the confusion a hierarchy model must resolve.
+
+**Key subtlety — no pair is truly unrelated.** Because every node shares the global `Root Node`, even
+cross-domain pairs share distant ancestry. So a "negative" means *"not a direct lineage link,"* and μ should
+**fall off gradually with tree-distance**, not snap to 0. This graded fall-off is the signal we want, and it is
+why hard sibling/cousin negatives matter more than random cross-map ones (the shared root already makes those
+mildly positive).
+
+### 2a. Candidate generation — substitution, e5-stratified
```
s243a, line 42: **[Design] "No pair is truly unrelated" — consequence for the μ=0 floor is under-explored**

The insight that all nodes share the global Root Node and therefore no pair is truly unrelated is well-stated. But the downstream consequence for training deserves explicit treatment:

If the minimum plausible μ for any cross-map pair is, say, 0.05 (barely above zero due to shared root), and the model is trained with μ targets that don't reflect this floor, it will try to push those pairs toward 0 — which is a *wrong supervision signal*.

Options the design should discuss:
1. **Set a μ floor** (e.g. `μ_min = decay(max_tree_distance)`) and clip all negative μ targets at this floor.
2. **Use relative μ** — normalize all μ values so the hardest negative in each batch defines the lower bound. This is naturally handled by max-normalization (§3a) but only applies within a candidate set, not across the whole corpus.
3. **Leave it as noise** — acceptable if the training set is large enough that a few cross-map pairs anchored near 0 are statistical noise. With 491 chains this is a bigger concern than it would be for enwiki.

At minimum: state explicitly whether the μ=0 target is used for directional-reverse negatives, or whether those also get a small graded μ.

prototypes/mu_cosine/DESIGN_mindmap_lineage.md
```diff
@@ -0,0 +1,128 @@
...
+## 3. Judge scoring
+
+Remove the last hop of a path → (path-minus-leaf = folder, leaf = item to file). Each example is thus **actual
+filing training data** — more target-aligned than raw lineage pairs, and buildable now without the harvester.
+Present the judge ≥2 candidate filings (2 = cheapest; 3–4 = better-shaped curve), scored **pointwise and
+independently — μ does NOT sum to 1** (matches the §14 prompt and the model's per-pair μ).
+
+### 3a. LLM judge (codex `gpt-5.5-low`, or Haiku)
+
+- **Max-normalize to the model's top pick = 1**, then a scoring curve: **linear first** (preserves the graded
```
s243a, line 62: **[Design] Judge scoring: "remove the last hop" generates filing examples but loses multi-hop positive signal**

The framing is elegant: remove the last hop, present the path-minus-leaf as a folder context, the leaf as the item to file. This directly generates filing training data.

However, this transform *discards* the multi-hop positive structure defined in §1. A 5-node chain `A/B/C/D/E` becomes one filing example (`A/B/C/D` → `E`), not 4 graded pairs. The graded positives (grandparent, great-grandparent) are only preserved if the judge is also called on the full-chain pairs, not just the leaf-filing version.

Is the intent to:
(a) Use the leaf-filing transform *only* for judge scoring (to produce high-quality calibrated μ for the hardest cases), while the graded multi-hop positives from §1 feed the training set directly from the lineage structure (no judge needed)?
(b) Replace multi-hop positives with judge-scored leaf-filing examples entirely?

(a) seems consistent with the PR scope but it's not stated explicitly. Clarifying this in the design doc would prevent confusion during the graded-round integration step.

prototypes/mu_cosine/DESIGN_mindmap_lineage.md
```diff
@@ -0,0 +1,128 @@
...
+  middle); a **temperature-controlled sigmoid** only if raw scores come back too bunched to separate hard-from-easy
+  (a *steep* sigmoid discards the grading — avoid).
+- **Multi-parent guard:** anchoring to the model's top (not the ground-truth) honors genuine second parents
+  (Pearltrees supercategories, cross-map links) — a valid substituted path *should* be allowed near 1. But keep
+  the **ground-truth as a sanity floor**: if the true filing scores *low*, flag it (judge error or a genuinely
+  ambiguous node), don't silently train on it.
+- Tag `judge=gpt-5.5-low` (its own calibration row; see `DESIGN_calibrated_judges.md`, `JUDGES` in
+  `mu_attention.py`). Cost/latency: `gpt-5.5-low` ≈ 5 s/pair, non-Anthropic.
```
s243a, line 70: **[Design] Max-normalize to model's top pick: initialization sensitivity**

Max-normalizing to the judge's top pick = 1 is clean, but creates a **relative scale** — the same absolute filing score may normalize to different values depending on what other candidates the judge saw in the same batch. This is intentional (pointwise scoring), but creates a subtle problem:

If `gpt-5.5-low` is very confident on easy cases (top pick gets raw score 0.95, runner-up gets 0.4 → normalized 1.0 / 0.42), but uncertain on hard cases (top pick 0.6, runner-up 0.55 → normalized 1.0 / 0.92), the normalized scores *look* similar in absolute terms but represent very different judge confidence levels.

The judge's **raw score spread** (max minus runner-up before normalization) is a useful calibration signal that gets discarded by normalization. Suggest logging `raw_top`, `raw_runner_up`, and `spread` in the output rows alongside normalized μ. This lets `build_graded_round.py` weight low-spread (uncertain) judge calls differently from high-spread (confident) ones.

prototypes/mu_cosine/DESIGN_mindmap_lineage.md
```diff
@@ -0,0 +1,128 @@
...
+
+**Whole-lineage, not just the immediate parent.** A bare parent-to-parent hop is myopic: two candidates can be
+equidistant from the true parent yet sit in very different lineages (`Math/Calculus/X` vs `Math/Analysis/X` — one
+leaf-hop apart but nearly the same folder path — versus one hop into an unrelated subtree). So the graph-μ
+accounts for the **full materialized-path prefix**, via two complementary whole-lineage factors:
+
+- **LCA depth (structural):** how deep the *shared* prefix runs (shares-only-Root ⇒ ~0) — a topological measure
+  of common ancestry.
+- **e5-of-prefix (semantic):** e5-embed the candidate's path prefix (`root / … / parent`) and the truth's, and
+  factor in their e5 distance — so `Math/Calculus/…` stays close to `Math/Analysis/…` despite the differing
```
s243a: **[Design] Graph judge: truth-at-distance-0 = 1 creates asymmetry with the LLM judge**

The graph judge anchors truth = 1 by definition (`decay(0) = 1`). The LLM judge sets truth = 1 only if the ground-truth *is* the model's top pick. In cases where the LLM judge assigns the ground-truth a lower score (because it found a better-fitting candidate), the two judges will produce conflicting μ for the same pair:
- Graph judge: ground-truth μ = 1
- LLM judge: ground-truth μ = 0.7 (something else scored higher)

This is the multi-parent case, and the design handles it via the "flag" mechanism — but it's worth stating explicitly here that the **graph judge is structurally incapable of discovering a better parent** (it always gives truth = 1), so its μ=1 assignments on the ground-truth are *definitional*, not *learned*. The LLM judge is the only one that can surface a genuine alternate parent. This asymmetry is what makes their disagreement meaningful, but it means averaging or fusing their μ values directly would be misleading.

prototypes/mu_cosine/DESIGN_mindmap_lineage.md
```diff
@@ -0,0 +1,128 @@
...
+mass — they are already specified in `COST_FUNCTION_PHILOSOPHY.md` and left as alternatives.
+
+## 4. Two judges, complementary — and a cost-smart split
+
+Graph judge = cheap, structural, but **blind to a semantically-valid parent that is structurally distant**
+(a cross-domain-but-correct filing). LLM judge = semantic, catches those, but costs. Their **disagreement is the
+useful part**: agree → confident μ; diverge → the ambiguous, probably-multi-parent cases.
+
+**Cost-smart pipeline:** run the **graph judge over everything (free, `judge=graph`)**, then spend LLM
+(`gpt-5.5-low`) calls **only where graph-μ is uncertain** — near-ties and structurally-distant candidates. The
```
s243a, line 168: **[Design] Disagreement between judges as the "ambiguous / multi-parent signal" — storage schema?**

The design correctly identifies judge disagreement as meaningful signal. But where does this signal live in the graded-round schema? The current design produces one μ per (pair, judge) row. If judge=graph gives μ=0.9 and judge=gpt-5.5-low gives μ=0.3 for the same pair, the model sees two conflicting training targets for the same (query, candidate) pair.

Two approaches:
1. **Keep both rows, let the model learn per-judge calibration** — this is the current `judge_emb` approach and it's fine, but then the disagreement is *implicit* in the model's calibration space, not explicit in the data.
2. **Add a `judge_agreement` flag** to rows that marks high-disagreement pairs explicitly, so the training loop can down-weight or quarantine them, or so post-hoc analysis can inspect where the two judges diverge.

Approach 2 is low-cost to add at row-generation time and would make the multi-parent analysis much more tractable. The `judge_agreement = abs(mu_graph - mu_llm)` is a direct operationalization.

prototypes/mu_cosine/gen_mindmap_lineage.py
```diff
@@ -0,0 +1,114 @@
...
+    def is_private_title(t): return "private" in (t or "").lower()
+
+
+def parse_map(smmx, tmp):
+    pref = os.path.join(tmp, os.path.basename(smmx).replace(".smmx", "").replace(" ", "_"))
+    r = subprocess.run([sys.executable, os.path.join(ROOT, "parse_smmx.py"), smmx, "--out-prefix", pref],
+                       capture_output=True, text=True)
+    if not os.path.exists(pref + "_nodes.tsv"):
+        return {}, {}, r.stderr
+    title = {}
```
s243a, line 38: **[Code] File handles not closed — should use `with open(...)`**

```python
for ln in open(pref + "_nodes.tsv", encoding="utf-8"):
```
and similarly for `_edges.tsv`. These bare `open()` calls rely on CPython's reference counting for cleanup. On PyPy or in long-running contexts (many maps), file handles accumulate. Use `with open(...) as fh:` throughout `parse_map()`.

prototypes/mu_cosine/gen_mindmap_lineage.py
```diff
@@ -0,0 +1,114 @@
...
+
+def main():
+    ap = argparse.ArgumentParser()
+    ap.add_argument("--maps", nargs="+", default=sorted(glob.glob(os.path.join(ROOT, "context", "*.smmx"))))
+    ap.add_argument("--tmp", default="/tmp/mu_data/mm_parse")
+    ap.add_argument("--out", default=os.path.join(ROOT, "mindmap_lineage.tsv"))
+    a = ap.parse_args()
+    os.makedirs(a.tmp, exist_ok=True)
+
+    rows, skipped_private, per_map = [], 0, {}
```
s243a: **[Code] Privacy belt-and-suspenders check iterates titles after cleaning is done — ordering concern**

```python
if any(is_private_title(t) for t in titles):  # belt-and-suspenders
    skipped_private += 1; continue
```
This check runs on `titles` (raw, pre-clean), which is correct — you want to catch private labels before they're cleaned. But the comment says "belt-and-suspenders" implying upstream already handled this. The potential gap: what if a private title *after* cleaning produces a non-private string? (Unlikely but possible if `is_private_title` uses substring matching and the clean step mutates the title.)

More importantly: the check happens *before* the clean loop, but `clean` is built below. If a private title survives into `clean` for some edge case, the chain would still be emitted. Consider running a second `is_private_title` check on `clean` titles as well, for true belt-and-suspenders coverage.

prototypes/mu_cosine/score_with_codex.py
```diff
@@ -0,0 +1,72 @@
...
+  source ~/.nvm/nvm.sh && nvm use 22            # codex needs node 22
+  python3 score_with_codex.py --pairs inferred_tail_pilot.tsv --batch 10 --out /tmp/mu_data/pilot_scored_gpt55low.tsv
+"""
+import argparse, os, re, subprocess, sys, time
+sys.path.insert(0, os.path.dirname(os.path.abspath(__file__)))
+from score_inferred_tail import build_prompts, ingest
+
+JUDGE = "gpt-5.5-low"
+CODEX = ["codex", "exec", "-s", "read-only", "-m", "gpt-5.5", "-c", "model_reasoning_effort=low"]
+
```
s243a, line 15: **[Code] `CODEX` command hard-codes `read-only` sandbox — may silently fail on some environments**

```python
CODEX = ["codex", "exec", "-s", "read-only", "-m", "gpt-5.5", "-c", "model_reasoning_effort=low"]
```
The `-s read-only` sandbox flag prevents any filesystem writes within the codex session. If `gpt-5.5-low` attempts to write scratch files or if a future prompt template uses a tool that writes output, the call will fail silently (timeout or empty JSON). The current design just scores, so `read-only` is fine — but make this an explicit CLI argument (`--sandbox`, default `read-only`) so it's easy to relax for debugging without modifying the constant.

prototypes/mu_cosine/score_with_codex.py
```diff
@@ -0,0 +1,72 @@
...
+    os.makedirs(os.path.dirname(a.responses) or ".", exist_ok=True)
+
+    pairs = load_pairs(a.pairs)
+    if a.limit:
+        pairs = pairs[:a.limit]
+    prompts = build_prompts(pairs, a.batch)
+    print(f"scoring {len(pairs)} pairs in {len(prompts)} batches of {a.batch} via codex {JUDGE}", flush=True)
+
+    arrays, ok, t0 = [], 0, time.time()
+    for i, pr in enumerate(prompts):
```
s243a, line 76: **[Code] Progress logging: `rate` is computed as `pairs/s` but the displayed unit is misleading when batches fail**

```python
rate = (ok * a.batch) / max(1e-9, time.time() - t0)
```
`ok` counts *successful* batches, so `rate` is actually *successfully scored pairs per second*, not total attempted pairs per second. On a run with intermittent failures this prints an optimistic rate. Consider:
```python
attempted = (i + 1) * a.batch
success_rate = (ok * a.batch) / max(1e-9, time.time() - t0)
fail_pct = 100 * (1 - ok / (i + 1))
print(f"  batch {i+1:3d}/{len(prompts)}  {dt:5.1f}s  {status}  | cum {time.time()-t0:5.0f}s  ~{success_rate:.2f} pairs/s  fail={fail_pct:.0f}%")
```
This makes failure rate visible in real time, which is critical for estimating API cost and deciding whether to abort.

s243a (commented):
## Second Review — Mindmap LINEAGE sampler + graph/LLM judge methodology

This review focuses on a different set of concerns from the first pass: primarily the **semantic consistency of the methodology** (how the design decisions interact with each other), a few **training distribution issues** that weren't flagged before, and **code quality** items in `gen_mindmap_lineage.py` and `score_with_codex.py`.

---

### Methodology / design issues (7 inline comments)

**1. LINEAGE-first ordering: directionality implication not articulated** (DESIGN §0)
The 70/30 directional/superposition split applied *after* lineage enumeration has a different semantic than the enwiki case. The "30% superposition" slice here means ancestor-as-query (reverse lookups), not structurally distinct edge types. This should be stated explicitly.

**2. Complete positive enumeration creates depth bias** (DESIGN §1)
A depth-8 chain contributes 28 positive pairs; a depth-2 chain contributes 1. With 491 chains, deep systems-theory nodes will dominate the positive set. Either cap pairs-per-chain or log pair-count-by-depth (the script already prints depth distribution — extending it to pair counts is trivial).

**3. μ=0 floor under-explored given "no pair is truly unrelated"** (DESIGN §2)
If the minimum plausible μ for cross-map pairs is non-zero (shared Root Node), training with μ=0 targets for directional-reverse negatives produces wrong supervision. The design should state whether μ=0 is used for any negative type, or whether all negatives use graded targets.

**4. "Remove the last hop" loses multi-hop positive signal** (DESIGN §3)
The leaf-filing transform produces one example per chain, not the 4 graded pairs a depth-5 chain has. The design needs to clarify whether multi-hop positives from §1 feed training directly (no judge), and the judge is only used for hard candidate scoring — or whether the leaf-filing examples replace multi-hop positives entirely.

**5. Max-normalization discards judge confidence spread** (DESIGN §3a)
The raw score spread (top minus runner-up before normalization) is a direct measure of judge confidence that gets discarded. Logging `raw_top`, `raw_runner_up`, and `spread` in the output rows would let `build_graded_round.py` down-weight uncertain judge calls.

**6. Graph judge truth=1 by definition creates fusion asymmetry** (DESIGN §3b)
The graph judge cannot discover a better parent — it always assigns μ=1 to the ground-truth by construction. The LLM judge can. This means their μ values are not on the same scale and should not be averaged or fused directly. Worth stating this explicitly: the judges are complementary but *not interchangeable*, and judge_emb calibration is the mechanism that handles this — not a linear combination.

**7. Judge disagreement needs an explicit schema field** (DESIGN §4)
Storing `judge_agreement = abs(mu_graph - mu_llm)` as a column in graded-round output rows would make multi-parent analysis tractable without requiring a separate post-hoc join. Low-cost to add at row-generation time.

---

### Code issues (4 inline comments)

**8. Bare `open()` calls in `parse_map()`** (`gen_mindmap_lineage.py` §parse_map)
File handles for `_nodes.tsv` and `_edges.tsv` are not closed. Use `with open(...) as fh:` throughout.

**9. Privacy check runs on pre-clean titles but doesn’t re-check post-clean** (`gen_mindmap_lineage.py` §main)
The belt-and-suspenders check correctly runs on raw titles, but doesn’t run a second pass on `clean` titles. A private label that survives cleaning (unlikely but possible) would be silently emitted.

**10. `CODEX` sandbox flag hard-coded** (`score_with_codex.py`)
`-s read-only` should be a CLI argument (`--sandbox`, default `read-only`) so it can be relaxed for debugging without modifying source.

**11. Progress rate metric misleading on batch failures** (`score_with_codex.py` §main loop)
`rate = (ok * a.batch) / elapsed` counts only successful batches, giving an optimistic pairs/s estimate. Should display failure rate percentage inline for real-time cost estimation.

---

### What this review does *not* re-raise

The following were well-covered in the first review and are not duplicated here: the `first_json_array` regex fragility, the `ingest()`/`--limit` bug, the `μ_graph` double-counting concern, the negative type-level breakdown, the multi-parent flag operationalization, the cost-smart split threshold, the `super_category` fallback logging, and the `JUDGES` comment length.

---

### Positive notes (new from this review)
- The `gen_mindmap_lineage.py` stats output (per-map counts, depth distribution, private-skip count) is excellent instrumentation. Extending it to pair-count-by-depth is a small addition with high diagnostic value.
- Using `subprocess.run` to call `parse_smmx.py` as a subprocess (rather than importing it) is a clean isolation boundary — it avoids polluting the namespace and makes `parse_smmx.py` independently testable.
- The `CORPORA` and `JUDGES` embedding design in `mu_attention.py` — each (model, reasoning-effort) getting its own calibration row — is the right architectural decision for a multi-judge system.

prototypes/mu_cosine/DESIGN_mindmap_lineage.md
```diff
@@ -0,0 +1,221 @@
...
+  Because CE is **scale-invariant**, `lineage-rank` carries order information the MSE head can't — notably in
+  the flat graded-middle where the regression targets bunch. **This is the filing-primary operator** (filing =
+  rank the candidate folders); `lineage` supplies graded confidence / fall-off.
+
+**e5 is on the GRANDPARENT lineage (the prefix), not the immediate parent.** The immediate parent is the
+decision the structural `decay(hops)` already makes; e5 compares the *ancestral context above it* — so
+`e5_sim_prefix(c)` = e5 similarity of the `root … grandparent` prefix of candidate `c` vs the truth's. (`Math/
+Calculus/…` stays close to `Math/Analysis/…` because their grandparent prefixes match.)
+
+**β scales e5 by how much prefix there is to compare** (user):
```
s243a, line 146: **[§3c Design] Two-operator split is a sound resolution of the double-count — with one condition**

Routing e5 to CE (scale-invariant, ordering-only) instead of into the μ_graph regression is the correct architectural resolution. The double-count critique was specifically that multiplying LCA depth and e5-prefix over-penalizes in the *regression target* — CE sidesteps this entirely because the loss doesn't care about absolute magnitudes.

The condition: **both operators must train on the same (query, candidate) pairs**, or the MSE head can learn to assign high μ to structurally-close candidates while the CE head learns a different order (driven by e5), and at inference the two heads produce conflicting guidance. The design should clarify: are `lineage` and `lineage-rank` presented the same candidate set per filing example, or do they use separate data pipelines?

If the same set: the outputs are naturally consistent. If separate: you need a reconciliation rule at inference (which head dominates? when?). The PR says `lineage-rank` is the **filing-primary** operator — that implies `lineage` is secondary. Document that explicitly in the op_emb routing, otherwise the training loop has no way to know which head wins when they disagree.

prototypes/mu_cosine/DESIGN_mindmap_lineage.md
```diff
@@ -0,0 +1,221 @@
...
+
+  ```
+  β = ( G / (L + G) ) · k
+      G = grandparent-prefix length (root … grandparent),   L = total path length (root … node),   k = tuned blend const
+  ```
```
s243a, line 152: **[§3c Design] CE collapse risk: softmax over correlated candidates is real but manageable**

The concern about two correlated operators (MSE + CE both trained on lineage-derived data) is slightly different from a standard collapse risk. True collapse would require the CE head to have its logits saturate — which is inhibited by the fact that the teacher `rank_score` is *fixed* (not the model's own output), as the design correctly notes.

The actual risk is subtler: **target distribution sharpness**. If `μ_graph` values for the candidate pool are tightly bunched (e.g., all siblings at 1 hop have similar `decay(1)*lca_frac`), the softmax teacher distribution is nearly uniform, and the CE gradient is vanishingly small. The CE head learns nothing for easy/flat cases and only trains on the handful of high-contrast candidates.

Mitigation: during the pilot, log the **entropy of the softmax teacher distribution** per filing example. Low-entropy (clear winner) = strong CE signal. High-entropy (flat) = weak CE signal — and those examples may need more candidates (`k≥4`) or the LLM judge to sharpen the target. This is a quick diagnostic on the `prototype_graph_judge.py` output before investing in the full pipeline.

prototypes/mu_cosine/DESIGN_mindmap_lineage.md
```diff
@@ -0,0 +1,221 @@
...
+
+  So a shallow node with no grandparent (`G=0`) → **β=0**, e5 contributes nothing (nothing to compare); deeper
+  nodes → more ancestral context → larger β, up to `~0.5·k`. `k` is tuned on the pilot.
+
+**Losses:** `lineage` → the existing graded-MSE head; `lineage-rank` → a candidate-softmax CE (same family as
```
s243a, line 158: **[§3c Design] `rank_score` additive combiner: why additive, and what does the unit mean?**

```
rank_score(c) = μ_graph(c) + β · e5_sim_prefix(c)
```

`μ_graph ∈ [floor, 1]` and `e5_sim_prefix ∈ [-1, 1]` (cosine similarity). The additive combination has no shared unit — structural hop-decay is a probability-like quantity, e5 cosine similarity is a geometric angle projection. When β is tuned, it's implicitly choosing how many units of cosine similarity equal one unit of structural plausibility, which is not well-defined.

Two alternatives worth evaluating:
1. **Normalize both to [0,1] before adding:** `e5_sim_norm = (e5_sim + 1) / 2`. Then the addition is at least unit-compatible (both are [0,1] soft scores). Simple and keeps the additive form.
2. **Make it a convex combination** `(1-β)·μ_graph + β·e5_sim_norm`, which has the nice property that the sum is still in [0,1] and β is interpretable as a weight fraction. This also constrains the range of rank_score and makes the softmax teacher distribution better-behaved.

Option 2 also makes the β formula `(G/(L+G))·k` directly interpretable: at G=0, β=0 and rank_score = μ_graph; at G=L, β=k/2·max, which needs a constraint that k≤1 to keep β in [0,1].

prototypes/mu_cosine/DESIGN_mindmap_lineage.md
```diff
@@ -0,0 +1,221 @@
...
+the model's transitive/directional margins). Both are weighted like the other operators.
+
+## 4. Two judges, complementary — and a cost-smart split
+
+Graph judge = cheap, structural, but **blind to a semantically-valid parent that is structurally distant**
```
s243a, line 163: **[§3c Design] β formula edge cases: G=0, very deep chains, and what counts as "grandparent prefix length"**

`β = (G/(L+G))·k` where G = grandparent-prefix length and L = total path length.

Three edge cases to pin before the prototype:

1. **G=0 (root-level nodes, depth-1):** β=0, e5 contributes nothing. This is correct and intentional — but it also means `rank_score = μ_graph` for all depth-1 nodes, so `lineage-rank` provides zero additional signal over `lineage` for the shallowest nodes. That's fine if there are few depth-1 nodes, but should be stated.

2. **Very deep chains (G large relative to L):** `G/(L+G)` approaches 0.5 asymptotically (since G ≈ L-2 for deep chains, so ratio ≈ (L-2)/(2L-2) → 0.5). So β is bounded by `~0.5·k`. This is a useful implicit cap, but worth stating explicitly so the tuning range for k is clear (k≤2 keeps β≤1 at max depth).

3. **Definition of G:** is G the number of *nodes* in the grandparent prefix (title count) or the number of *characters* (for e5 embedding length normalization)? Since e5 embeds text, node count and character count are both valid but produce different β scaling. The design should specify: node count is simpler and more consistent with `lca_depth_frac`.

prototypes/mu_cosine/DESIGN_mindmap_lineage.md
```diff
@@ -0,0 +1,221 @@
...
+
+**β scales e5 by how much prefix there is to compare** (user):
+
+  ```
+  β = ( G / (L + G) ) · k
+      G = grandparent-prefix length (root … grandparent),   L = total path length (root … node),   k = tuned blend const
+  ```
+
+  So a shallow node with no grandparent (`G=0`) → **β=0**, e5 contributes nothing (nothing to compare); deeper
+  nodes → more ancestral context → larger β, up to `~0.5·k`. `k` is tuned on the pilot.
```
s243a: **[§3c Design] "e5 on the grandparent prefix" division of labor is correct — with a clarification needed**

The framing "structural decay decides the parent, e5 compares the ancestral context above it" is the right division. The immediate parent is already selected by `decay(hops)` — the e5 signal is asking whether the *lineage context* that led to that parent is semantically similar to the true context. `Math/Calculus/…` vs `Math/Analysis/…` is exactly the case: same hop distance, same LCA, but different contextual direction above the immediate parent.

Clarification needed: **what happens for a candidate whose grandparent IS the LCA?** In this case, `e5_sim_prefix(c)` and the LCA-depth factor in `μ_graph` are computed on the *same prefix* — the shared path up to LCA. The double-counting is gone from the *regression target* (since they're in separate operators), but the *features* being learned are still correlated: the model's encoder sees the same token sequence for both. This is not a blocking issue for the design, but worth noting in the doc as a future diagnostic: if `lineage` and `lineage-rank` converge to the same ranking on the pilot, it may indicate the grandparent-e5 signal is redundant with the LCA-depth signal for this dataset.

prototypes/mu_cosine/DESIGN_mindmap_lineage.md
```diff
@@ -0,0 +1,221 @@
...
+
+  Candidates are ordered by `rank_score`; `lineage-rank` is trained by softmax-CE to reproduce that order.
+  Because CE is **scale-invariant**, `lineage-rank` carries order information the MSE head can't — notably in
+  the flat graded-middle where the regression targets bunch. **This is the filing-primary operator** (filing =
+  rank the candidate folders); `lineage` supplies graded confidence / fall-off.
+
+**e5 is on the GRANDPARENT lineage (the prefix), not the immediate parent.** The immediate parent is the
+decision the structural `decay(hops)` already makes; e5 compares the *ancestral context above it* — so
+`e5_sim_prefix(c)` = e5 similarity of the `root … grandparent` prefix of candidate `c` vs the truth's. (`Math/
+Calculus/…` stays close to `Math/Analysis/…` because their grandparent prefixes match.)
```
s243a: **[§5 Prototype] Graph judge fragmentation: "needs fused graph" is the right conclusion, but add a fallback**

The finding is correct: single-map parse produces isolated components (many `hops=99`), so `μ_graph` falls to the floor for most cross-map candidates. The fused/cross-map graph is the right long-term fix.

However, the design as written leaves the graph judge **entirely inoperable until `fuse_corpus.py` is ready** (depends on the PT cache / harvester). This creates a hard dependency that could block the negative-generation + judge pipeline for an indeterminate period.

A practical fallback for the single-map phase:
- Treat `hops=99` (unreachable within the map) not as floor `μ_floor=0.02`, but as **"shares-only-Root" = the lowest defined structural distance** — which the design already handles (`decay(max_tree_distance)`). The only change: don't use BFS hops=99 as a sentinel; instead, assign `hops = max_known_tree_depth + 1` so the formula degrades gracefully rather than hard-flooring.
- This won't produce correct cross-map ordering (all unreachable candidates get the same score), but it keeps the graph judge runnable for within-map pairs while the fused graph is pending, and the LLM judge covers the cross-map cases.

Suggest adding this fallback behavior to the design doc so the prototype doesn't stall on the fused-graph dependency.

prototypes/mu_cosine/prototype_graph_judge.py
```diff
@@ -0,0 +1,113 @@
...
+    return seen
+
+
+def main():
+    ap = argparse.ArgumentParser()
+    ap.add_argument("--map", default=os.path.join(os.path.dirname(__file__), "context", "Chaos theory.smmx"))
+    ap.add_argument("--gamma", type=float, default=0.6, help="decay base gamma^hops")
+    ap.add_argument("--floor", type=float, default=0.02, help="mu floor (shares-only-root, never 0)")
+    ap.add_argument("--k", type=int, default=6, help="sampled candidate parents per node (besides the true one)")
+    ap.add_argument("--show", type=int, default=6)
```
s243a, line 66: **[Code] `bfs_dist` called per node — O(N²) total; cache per true-parent instead**

```python
dist = bfs_dist(undirected_adj(parent), p)   # called for every node n in `filed`
```

`bfs_dist` runs a full BFS from `p` (the true parent) for every filable node. If two nodes share the same true parent, the BFS is recomputed. For a prototype this is fine — but with 491 chains the redundancy is already measurable. Caching:

```python
dist_cache = {}
def get_dist(p):
    if p not in dist_cache:
        dist_cache[p] = bfs_dist(adj, p)
    return dist_cache[p]
```

Also: `undirected_adj(parent)` is called once outside the loop — good. But `adj` is not passed to `bfs_dist`, it rebuilds implicitly. Either pass `adj` as a parameter or store it at function level to avoid the re-computation.

prototypes/mu_cosine/prototype_graph_judge.py
```diff
@@ -0,0 +1,113 @@
...
+    if not title:
+        print("parse failed:", err[:120]); return
+    children = defaultdict(set)
+    for c, p in parent.items():
+        children[p].add(c)
+
+    filed = [n for n in title if n in parent]
+    true_is_top, true_mu1, shown, all_mu = 0, 0, 0, []
+    for n in filed:
+        p = parent[n]
```
s243a, line 86: **[Code] Candidate pool excludes self and descendants but not ancestor-of-n — leakage risk**

```python
forbidden = descendants(n, children) | {n, p}   # p is added to cands separately as the true positive
pool = [m for m in title if m not in forbidden]
```

This correctly excludes `n` (filing under yourself) and `n`'s descendants (circular). But it does NOT exclude **nodes that are ancestors of `n`** (i.e., nodes that *contain* `n` in their lineage). Filing `n` under its own ancestor is a semantically valid lineage relationship — it would be a grandparent, not a parent. If an ancestor of `n` appears in the candidate pool and gets sampled, it will score high in `μ_graph` (short hop distance to `p` = `n`'s parent) and *look like* a hard negative but is actually a soft positive.

For the prototype this is a minor concern, but for the negative-generation pipeline it matters: ancestors of `n` should either be:
- **Excluded** from the negative candidate pool (they are positives at a longer distance), OR
- **Labeled as positives** with a lower graded μ (grandparent = weaker positive, exactly §1's graded scheme).

For now, suggest adding `ancestors_of_n = set(lineage(n, parent)) - {n}` to `forbidden`, and noting in a comment that the lineage graded positives handle those pairs separately.

prototypes/mu_cosine/gen_mindmap_lineage.py
```diff
@@ -0,0 +1,141 @@
...
+    parent = dict(broader); parent.update(struct_parent)  # structural parent takes precedence
+    fallback = set(broader) - set(struct_parent)          # parent came ONLY from super_category
+    # BYPASS nav nodes in the STRUCTURE (not just the display path): if a node's parent is a nav node,
+    # reconnect it to the nav node's parent, then drop nav nodes entirely. Fixes "via link" appearing as a
+    # true parent in the graph judge + reconnects chains the nav node was breaking.
+    nav_keys = {k for k, t in title.items() if t.strip().lower() in NAV}
+    if nav_keys:
+        def real_parent(n):
+            p, seen = parent.get(n), set()
+            while p in nav_keys and p not in seen:
```
s243a, line 69: **[Code] Nav bypass: `real_parent` closure over `parent` dict mutates during comprehension**

```python
def real_parent(n):
    p, seen = parent.get(n), set()
    while p in nav_keys and p not in seen:
        seen.add(p); p = parent.get(p)
    return p
parent = {n: real_parent(n) for n in parent}   # reassigns `parent` while real_parent closes over it
```

The dict comprehension `{n: real_parent(n) for n in parent}` iterates the *old* `parent` dict and calls `real_parent`, which also reads from `parent` via closure. In Python, the name `parent` inside `real_parent` is looked up at *call time*, not at definition time. During the comprehension, `parent` still points to the old dict (the comprehension hasn't reassigned yet — the name is rebound only after the comprehension completes). So this is actually safe in CPython.

However, it's fragile: if `real_parent` were called *after* the reassignment (e.g., in a later refactor), it would silently read from the new (partially-bypassed) dict and produce wrong results for multi-hop nav chains. Suggest using an explicit snapshot:

```python
old_parent = dict(parent)
def real_parent(n):
    p, seen = old_parent.get(n), set()
    while p in nav_keys and p not in seen:
        seen.add(p); p = old_parent.get(p)
    return p
parent = {n: real_parent(n) for n in old_parent}
```

s243a (commented):
## Round-2 review — new design post `0bb207e5d`

### Round-1 spot-check (commit `0bb207e5d`): resolved cleanly

All critical and most minor items from both prior passes are addressed:
- `first_json_array` → `json.JSONDecoder().raw_decode()` scan ✓
- `--limit` ingest bug → temp pairs file written ✓  
- `μ_graph` double-count → e5-prefix demoted to tie-breaker (then superseded by two-operator split) ✓
- `with open()` throughout `parse_map` ✓
- Privacy re-check on cleaned titles ✓
- `super_category` fallback logged ✓
- `--sandbox` CLI arg ✓
- Progress failure % displayed ✓
- `JUDGES` comment shortened; rationale moved to `DESIGN_calibrated_judges.md` ✓
- All methodology spec items (`μ_floor`, `gt_rank` disposition, `judge_agreement` column, per-chain depth weighting, judge-fusion prohibition, spread logging, e5 boundary) documented in §6 ✓

The §6 "Review responses" section is thorough and precise. No round-1 items need re-raising.

---

### New design — 9 comments across 3 files

**Two-operator split (`lineage` MSE + `lineage-rank` CE) — §3c**

1. **Correct resolution of the double-count critique** (DESIGN §3c): CE is scale-invariant and skips the regression double-count entirely. One condition: both operators should train on the same candidate sets, or a reconciliation rule at inference must be documented. `lineage-rank` is called "filing-primary" — that dominance relationship should be stated in the op_emb routing, not just the design narrative.

2. **CE collapse risk is real but not from operator correlation** (DESIGN §3c): the risk is flat teacher distributions (bunched `μ_graph` values for sibling candidates → near-uniform softmax → vanishingly small CE gradient). Recommend logging the entropy of the softmax teacher distribution per example in the pilot as an early diagnostic.

3. **`rank_score` unit mismatch: `μ_graph ∈ [floor,1]` + `β·e5_sim ∈ [-β,β]`** (DESIGN §3c): cosine similarity and hop-decay are not in the same units. A convex combination `(1-β)·μ_graph + β·e5_sim_norm` is more principled, keeps the range bounded, and makes β directly interpretable as a weight fraction (with the constraint `k≤1`).

4. **β formula edge cases** (DESIGN §3c): G=0 (depth-1 nodes) → β=0, CE provides zero additional signal over MSE for the shallowest nodes (fine, just state it). Deep chains → β asymptotes at `~0.5·k` (useful implicit cap; state that k≤2 is needed to keep β≤1). Definition of G: node count vs. character count — specify node count for consistency with `lca_depth_frac`.

5. **Grandparent-prefix e5 division of labor is correct** (DESIGN §3c): "structural decides the parent, e5 compares the context above it" is the right framing. Residual concern: when the candidate's grandparent IS the LCA, both operators see the same prefix tokens. Not a blocking issue; worth logging as a pilot diagnostic (if the two operators produce identical rankings, the grandparent-e5 signal is redundant for this dataset).

**Graph-judge prototype findings**

6. **"Needs fused graph" is right, but add a single-map fallback** (DESIGN §5): without `fuse_corpus.py` the graph judge is inoperable for cross-map candidate scoring. Fallback: assign `hops = max_known_tree_depth + 1` for unreachable candidates instead of a flat sentinel `hops=99`, so the formula degrades gracefully. Within-map pairs remain correctly scored; LLM judge covers cross-map cases in the interim.

**`prototype_graph_judge.py` code**

7. **`bfs_dist` called per node, O(N²) total** (prototype line 62): if two nodes share the same true parent, BFS is recomputed. Cache by true-parent key. Also: `undirected_adj` is computed outside the loop but not passed as a parameter to `bfs_dist` — pass it explicitly.

8. **Candidate pool includes ancestors of `n`** (prototype line 78): ancestors of `n` are soft positives (graded μ from §1), not negatives. They should be excluded from the negative candidate pool (add `ancestors_of_n = set(lineage(n, parent)) - {n}` to `forbidden`), or labeled as graded positives and handled by the lineage pair generator instead.

**`gen_mindmap_lineage.py` code**

9. **`real_parent` closure over `parent` is safe in CPython today but fragile** (line 68): the dict comprehension reassigns `parent` only after completion, so `real_parent` reads the old dict correctly. However, any future refactor that calls `real_parent` after the reassignment will silently read the bypassed dict. Add an explicit `old_parent = dict(parent)` snapshot for robustness.

---

### Priority order for next iteration

| Priority | Item |
|---|---|
| Should fix before generator | Item 8 (ancestors in candidate pool — silent false negatives) |
| Should fix before generator | Item 9 (`real_parent` snapshot — correctness risk on refactor) |
| Should fix before pilot | Item 3 (rank_score unit mismatch — affects β tuning) |
| Should fix before pilot | Item 4 (β edge cases — pin G definition and k constraint) |
| Design doc clarification | Item 1 (inference reconciliation rule for two operators) |
| Pilot diagnostic | Item 2 (CE teacher entropy logging) |
| Design doc clarification | Item 6 (single-map fallback for graph judge) |
| Nit / optimization | Item 7 (BFS cache) |
| Low priority | Item 5 (grandparent-prefix redundancy — wait for pilot data) |

PR #3496, opened 2026-07-05
Judge superpositions: JointPosterior + 1/d, a graded-Wikipedia round, and the blend judge (measured against the right target)

s243a:
## Summary

Extends the SYM dual-judge work into the **µ-estimation architecture**, builds the data + tooling to test it,
and lands a multi-seed-confirmed result — with the epistemics done honestly (several tempting conclusions were
walked back under scrutiny). It also ships the **uncertainty-estimation playbook** so future work starts right.

## What's in it

**Estimator (the JointPosterior + `1/d`).** `mu_posterior.py`: `struct_dist_fn` (the `1/d` source), node-disjoint
split, and a with/without-`1/d` ablation reporting **AURC (margin-gated, bootstrap CI) + ECE**. Fix:
`eval_relatedness.build_model` allow-list widened for the new buffers.

**Data pipeline (a graded Wikipedia relation round).** `gen_wiki_relation_pairs.py` → `score_with_codex.py`
(gpt-5.5-low, §14 rubric) → `convert_scored_to_graded.py`. 880 stratified pairs, both endpoints in the struct
embedding, LLM-labelled.

**The blend judge.** `emit_blend_judge.py` + `JUDGES += "blend"`: trains the *constructed* dual
`(1−λ)·µ_e5 ⊕ λ·P(SYM|1/d,asym-ops)` as its **own calibrated judge**, distinct from the LLM judge (λ=0.5 or
random). `train_mu_attention` trains both judge inputs.

**Evals.** `eval_blend_prediction.py` — predict the held-out **judge superposition** `T` (the metric that isn't
LLM-confounded). Multi-seed drivers.

## Results (honest, and they moved under scrutiny)

- **Does `1/d` help *relation classification*?** No — it's redundant with e5 on Wikipedia categories (+0.50 corr).
  **But that was the wrong target** (the LLM label isn't the SYM judge).
- **Does the *estimator* (JointPosterior) beat product-of-marginals?** **Yes** (log-loss 1.08 vs 1.55/1.26).
- **Does the blend judge lift base SYM?** On an **LLM-scored** eval, it ties a no-blend control across seeds —
  and even that lift is an **LLM-alignment artifact** (train/eval share a judge).
- **On the RIGHT metric — predict the held-out judge superposition — does the blend earn its keep? YES, multi-seed
  confirmed:** `corr(SYM, T)` under `judge=blend`: **B (blend) mean +0.841 vs A (LLM-only) +0.697 vs prod +0.746**,
  B > A on all 3 seeds. The blend **recovers the graph half** (+0.815) that **LLM-only training drifts away from**
  (+0.667), and reading under `judge=blend` beats agnostic on every seed.

## Docs / methodology
- **`DESIGN_uncertainty_estimation_playbook.md`** (+ a mirrored, activatable skill under `prototypes/mu_cosine/skills/`):
  the rule (calibrated joint posterior + margin gate, not hand-set independent weights), six pitfalls (incl.
  **#6 don't let the eval share a judge with the training data**), and the judge-superposition-teaches-generality
  rationale — with the honest arc (it only shows up against a non-LLM target, read under the matching judge input).
- **`DESIGN_sym_estimation_integration.md`**, **`REPORT_mu_posterior_dist.md`**, **`REPORT_blend_judge_sweep.md`**.

## Testing
- `test_mu_posterior_dist.py` (AURC/margin/`1/d`), `test_struct_blend.py` (warm-start no-op, gating, loader regression) — all pass.

## Not in scope / follow-ups
- Higher-e5-variance held-out to test the e5 side of `T`; mindmap training data; **LINEAGE** blend (e5 on the
  hierarchical *list*) and **operator** superposition (§7).

### DIFF
--- a/prototypes/mu_cosine/DESIGN_uncertainty_estimation_playbook.md
+++ b/prototypes/mu_cosine/DESIGN_uncertainty_estimation_playbook.md
@@ -9,7 +9,7 @@ for "confidence weighting." This encodes lessons the project learned the slow wa
 **Don't hand-set independent confidence weights over correlated sources. Fit a LEARNED, CALIBRATED combiner
 (`JointPosterior`) on HELD-OUT data, and use confidence as a MARGIN GATE, not a per-item weight.**
 
-## The five pitfalls (each one we actually hit)
+## The six pitfalls (each one we actually hit)
 
 1. **Sources are not independent → naive/inverse-variance fusion over-confidences.** In this project every
    model readout *consumes frozen e5*, so `e5`, `μ_SYM`, `μ_HIER`, `μ_ELEM` are correlated (measured **+0.751**,
--- a/prototypes/mu_cosine/DESIGN_uncertainty_estimation_playbook.md
+++ b/prototypes/mu_cosine/DESIGN_uncertainty_estimation_playbook.md
@@ -84,6 +84,13 @@ the general notion (of relatedness, lineage, …).
   multi-judge *data diversity* may buy the generality without any hand-constructed blend judge; (b) a constructed
   blend must **beat a no-blend control across seeds** before you credit it; (c) random-λ needs stability
   guards. Don't cite a single-seed lift.
+  **Resolution — measure against the RIGHT target and the blend DOES earn its keep.** On the LLM-scored eval the
+  blend tied the control (that eval rewards LLM-alignment). On the *relevant* metric — **predict the judge
+  superposition `(1−λ)·e5 ⊕ λ·graph` on held-out pairs** — the blend-trained model (read under `judge=blend`)
+  won: **corr +0.847 vs prod +0.746 vs LLM-only +0.675**, and it recovered the **graph half** (graph_ref corr
+  +0.823) that **LLM-only training actively drifts away from** (+0.641). Lesson: **the blend's value only shows
+  up against a target that isn't itself the LLM** — so pick the eval target deliberately (pitfall #6), and read
+  under the matching judge input (blend > agnostic, +0.847 vs +0.787).
 - **The one honest condition:** this teaches the *right* generality only if the judges genuinely measure the same
   latent. A systematically-biased judge, superposed in, teaches an averaged-wrong signal — which is exactly what
   the per-judge calibration rows guard against, and why the blend must be *validated* (does graph⊕e5 actually
--- a/prototypes/mu_cosine/DESIGN_uncertainty_estimation_playbook.md
+++ b/prototypes/mu_cosine/DESIGN_uncertainty_estimation_playbook.md
@@ -33,6 +33,11 @@ for "confidence weighting." This encodes lessons the project learned the slow wa
 5. **Measure on HELD-OUT, node-disjoint splits — never on training pairs.** Correlating a model's own readouts
    against a target it was trained on inflates *and can reorder* effect sizes (a readout ordering measured on
    training pairs can flip held-out). Split so no node appears in both train and held.
+6. **Don't let the eval share a judge with the training data.** If the eval targets are LLM-scored and you train
+   on more LLM data, the metric rises *by alignment to that judge* — not necessarily a real quality gain (a
+   train/eval-share-a-judge confound; `REPORT_blend_judge_sweep.md` — a +0.41→+0.79 SYM "gain" on a haiku-scored
+   eval, driven by adding LLM training data). To claim a *general* improvement, evaluate against an
+   **independent** target (a different judge, graph-structural, human, or a downstream task).
 
 ## The workflow (what "doing it right" looks like)
 
--- a/prototypes/mu_cosine/DESIGN_uncertainty_estimation_playbook.md
+++ b/prototypes/mu_cosine/DESIGN_uncertainty_estimation_playbook.md
@@ -1,0 +1,66 @@
+# Uncertainty / multi-source μ-estimation — playbook
+
+*Start here whenever you're combining multiple signals into a μ / confidence / relation estimate, or reaching
+for "confidence weighting." This encodes lessons the project learned the slow way (see the arc: #3356 → #3357 →
+#3359 → #3387 → #3391, and the SYM dual-judge course-correction, `DESIGN_sym_estimation_integration.md`).*
+
+## The one-line rule
+
+**Don't hand-set independent confidence weights over correlated sources. Fit a LEARNED, CALIBRATED combiner
+(`JointPosterior`) on HELD-OUT data, and use confidence as a MARGIN GATE, not a per-item weight.**
+
+## The five pitfalls (each one we actually hit)
+
+1. **Sources are not independent → naive/inverse-variance fusion over-confidences.** In this project every
+   model readout *consumes frozen e5*, so `e5`, `μ_SYM`, `μ_HIER`, `μ_ELEM` are correlated (measured **+0.751**,
+   #3359). A product-of-experts / precision-weighting that assumes independence double-counts the shared signal.
+   Evidence: a factored equal-weight PoE scored **50.9%, below the 51.1% majority baseline** (#3359). *Fix:* a
+   learned joint head over the full vector (it down-weights the redundant part), or model the residual /
+   down-weight by separability (#3357).
+2. **Anti-correlated sources carry structure that additive weighting throws away.** Subcategory and element
+   readouts are *anti-correlated*, and conditioning on both *jointly IS the directional asymmetry* (#3359). Two
+   separate positive weights can't represent that; a joint head can.
+3. **Confidence is a WEAK per-query signal → it's a GATE, not a weight; and it's MARGIN, not level.** Per-query
+   ρ(confidence, correctness) is +0.03…+0.15 with CIs crossing 0 (#3391). Absolute μ *level* is even
+   anti-correlated with correctness on under-trained checkpoints; **margin (top1−top2)** is the better
+   selective-risk gate (`AURC_margin < AURC_level`, #3391). So use confidence to *route / abstain*, not to
+   multiply a per-item contribution.
+4. **Data-limit / "how much training support" ≠ graph degree.** A node's degree is its *own* edge count; it does
+   not measure whether *nearby (e5-similar)* nodes were trained, which is what governs a readout's reliability
+   (the model generalises through e5). #3356 used a **label-confidence tag + `P(μ|op)` calibration against the
+   tagged set**, re-estimated as the model evolves — *not* a structural proxy. If you need a data-density term,
+   use trained-neighbour density in e5 space or the `P(μ|op)` calibration, and treat it as a gate.
+5. **Measure on HELD-OUT, node-disjoint splits — never on training pairs.** Correlating a model's own readouts
+   against a target it was trained on inflates *and can reorder* effect sizes (a readout ordering measured on
+   training pairs can flip held-out). Split so no node appears in both train and held.
+
+## The workflow (what "doing it right" looks like)
+
+1. **Assemble the source vector** — every candidate signal (static e5, each model readout fwd/rev, any external
+   judge, and *decorrelated* structural signals like `1/d`). Tool: `mu_posterior.py` (`e5_mu_fn`,
+   `model_readout_fn`, `struct_dist_fn`).
+2. **Look before you combine** — print the **separability** per source and the **correlation matrix**. High |r|
+   ⇒ redundancy; a genuinely new source should be *decorrelated* from the existing cluster (that's its whole
+   value — e.g. `1/d` covers the lateral/sibling axis the vertical membership operators miss).
+3. **Fit the calibrated joint head** (`JointPosterior`, logistic or small MLP) on a **held-out, node-disjoint**
+   split. Compare against the factored PoE (equal + separability-weighted) as controls.
+4. **Report the honest metrics:** accuracy, log-loss, **ECE with a stated binning** (bins sensitive to choice),
+   and **AURC gated by margin, with a bootstrap 95% CI** (AURC is noisy on small held-out sets). A new source
+   *earns its keep* iff the with-source AURC CI sits below the without-source point estimate.
+5. **Ablate each source** (with vs without) on the *same* split. Redundant sources show overlapping intervals.
+
+## Anti-patterns (stop if you're doing these)
+
+- Writing per-item `c_k` confidence multipliers and inverse-variance-averaging them. (→ pitfall 1/3.)
+- Treating `corr(signal, target)` on the training set as a source's reliability. (→ pitfall 5.)
+- Using absolute μ level as the confidence gate. (→ pitfall 3.)
+- Node degree as a "data density" / confidence term. (→ pitfall 4.)
+- Adding a source and asserting it helps because it's conceptually motivated — *test it* (step 4/5).
+
+## Tools & references
+- **`mu_posterior.py`** — `MuPosterior` (per-source `P(μ|rel)`, bands, separability), `JointPosterior` (the
+  calibrated combiner), `struct_dist_fn` / `e5_mu_fn` / `model_readout_fn` (sources), `aurc`/`aurc_boot`/
+  `margin_conf` (the gate + selective-risk metric). Tests: `test_mu_posterior.py`, `test_mu_posterior_dist.py`.
+- **Design:** `DESIGN_mu_sources_and_estimation.md` (#3357), `DESIGN_sym_estimation_integration.md` (the
+  course-correction), `DESIGN_inferred_operator_superposition.md` §2–3 (#3356 calibration + noise decomposition).
+- **Evidence:** `REPORT_mu_posterior.md` (#3359 — the correlation numbers, joint > PoE), `DESIGN_model_applications.md` / PR #3391 (margin-not-level, AURC), `REPORT_eval_methodology.md` (#3387).
--- a/prototypes/mu_cosine/DESIGN_uncertainty_estimation_playbook.md
+++ b/prototypes/mu_cosine/DESIGN_uncertainty_estimation_playbook.md
@@ -74,9 +74,12 @@ the general notion (of relatedness, lineage, …).
   **per-judge `judge_emb` rows** absorb each judge's *offset/bias* (fidelity). You get generality **and**
   per-judge calibration — not a lossy average. This is the "operator superposition as a regulariser" idea
   (`DESIGN_inferred_operator_superposition.md` §7) lifted from *operators* to *judges*.
-- **Symptom of success (and a cautionary result):** the *theory* predicts a rise on a held-out set the blend
-  pairs never touched (generalisation, not memorisation). **We tested it and it did NOT pan out for the
-  *constructed* blend.** A 3×3 sweep (`REPORT_blend_judge_sweep.md`): fine-tuning on the LLM Wikipedia round did
+- **Symptom of success — the arc in one line:** *theory predicts generality → it did NOT show on the LLM-scored
+  eval (a confound) → it DID show on the right (non-LLM) target.* Read the three beats in order:
+
+  **(beat 1 — theory)** predicts a rise on a held-out set the blend pairs never touched (generalisation, not
+  memorisation). **(beat 2 — the trap)** we tested it and it did NOT pan out on the LLM-scored eval for the
+  *constructed* blend. A 3×3 sweep (`REPORT_blend_judge_sweep.md`): fine-tuning on the LLM Wikipedia round did
   lift *base* simplewiki SYM held-out (+0.41 → **~+0.79, stable across seeds**) — but a **control with *no*
   constructed blend matched it** (LLM-only +0.790 vs fixed-λ blend +0.787), and **random-λ destabilised** (one
   seed collapsed to +0.09). So the lift was **the data + the fine-tune (which already spans multiple judge
--- a/prototypes/mu_cosine/DESIGN_uncertainty_estimation_playbook.md
+++ b/prototypes/mu_cosine/DESIGN_uncertainty_estimation_playbook.md
@@ -57,6 +57,30 @@ for "confidence weighting." This encodes lessons the project learned the slow wa
 - Node degree as a "data density" / confidence term. (→ pitfall 4.)
 - Adding a source and asserting it helps because it's conceptually motivated — *test it* (step 4/5).
 
+## Why train on judge *superpositions* — it teaches generality
+
+Beyond constructing a good target, training on **judge superpositions** — a target `(1−λ)·judgeA ⊕ λ·judgeB`,
+especially with **random λ per example** — is a **multi-view consistency regulariser**. It *asserts that the two
+judges are two views of one underlying quantity*, so to fit the whole blend family the model can't latch onto
+either judge's idiosyncrasies; it must find the **shared signal** they both approximate. That shared signal is
+the general notion (of relatedness, lineage, …).
+
+- **The trunk/head split makes it clean.** The **shared trunk** learns the invariant (generality); the
+  **per-judge `judge_emb` rows** absorb each judge's *offset/bias* (fidelity). You get generality **and**
+  per-judge calibration — not a lossy average. This is the "operator superposition as a regulariser" idea
+  (`DESIGN_inferred_operator_superposition.md` §7) lifted from *operators* to *judges*.
+- **Symptom of success:** performance rises on a **held-out set the blend pairs never touched** (generalisation,
+  not memorisation). E.g. blend-judge SYM training lifted *base* simplewiki SYM held-out well above the
+  fine-tune-data's own domain — evidence the model learned a more general SYM, not the Wikipedia round. (Confirm
+  across seeds; a single-seed lift can be variance.)
+- **The one honest condition:** this teaches the *right* generality only if the judges genuinely measure the same
+  latent. A systematically-biased judge, superposed in, teaches an averaged-wrong signal — which is exactly what
+  the per-judge calibration rows guard against, and why the blend must be *validated* (does graph⊕e5 actually
+  track the semantic judge?) before it's trusted, not assumed.
+- **When you swap judges, swap the judge input.** The target's provenance token (`judge_emb`) must name the judge
+  that produced it (`gpt-5.5-low` for LLM data, `blend` for the constructed superposition). Wrong tag ⇒ wrong
+  calibration row ⇒ the generality is learned against the wrong offset.
+
 ## Tools & references
 - **`mu_posterior.py`** — `MuPosterior` (per-source `P(μ|rel)`, bands, separability), `JointPosterior` (the
   calibrated combiner), `struct_dist_fn` / `e5_mu_fn` / `model_readout_fn` (sources), `aurc`/`aurc_boot`/
--- a/prototypes/mu_cosine/DESIGN_uncertainty_estimation_playbook.md
+++ b/prototypes/mu_cosine/DESIGN_uncertainty_estimation_playbook.md
@@ -69,10 +69,16 @@ the general notion (of relatedness, lineage, …).
   **per-judge `judge_emb` rows** absorb each judge's *offset/bias* (fidelity). You get generality **and**
   per-judge calibration — not a lossy average. This is the "operator superposition as a regulariser" idea
   (`DESIGN_inferred_operator_superposition.md` §7) lifted from *operators* to *judges*.
-- **Symptom of success:** performance rises on a **held-out set the blend pairs never touched** (generalisation,
-  not memorisation). E.g. blend-judge SYM training lifted *base* simplewiki SYM held-out well above the
-  fine-tune-data's own domain — evidence the model learned a more general SYM, not the Wikipedia round. (Confirm
-  across seeds; a single-seed lift can be variance.)
+- **Symptom of success (and a cautionary result):** the *theory* predicts a rise on a held-out set the blend
+  pairs never touched (generalisation, not memorisation). **We tested it and it did NOT pan out for the
+  *constructed* blend.** A 3×3 sweep (`REPORT_blend_judge_sweep.md`): fine-tuning on the LLM Wikipedia round did
+  lift *base* simplewiki SYM held-out (+0.41 → **~+0.79, stable across seeds**) — but a **control with *no*
+  constructed blend matched it** (LLM-only +0.790 vs fixed-λ blend +0.787), and **random-λ destabilised** (one
+  seed collapsed to +0.09). So the lift was **the data + the fine-tune (which already spans multiple judge
+  tags), not the constructed superposition** — and forcing a per-pair random blend is *risky*. Lesson: (a)
+  multi-judge *data diversity* may buy the generality without any hand-constructed blend judge; (b) a constructed
+  blend must **beat a no-blend control across seeds** before you credit it; (c) random-λ needs stability
+  guards. Don't cite a single-seed lift.
 - **The one honest condition:** this teaches the *right* generality only if the judges genuinely measure the same
   latent. A systematically-biased judge, superposed in, teaches an averaged-wrong signal — which is exactly what
   the per-judge calibration rows guard against, and why the blend must be *validated* (does graph⊕e5 actually
--- a/prototypes/mu_cosine/REPORT_blend_judge_sweep.md
+++ b/prototypes/mu_cosine/REPORT_blend_judge_sweep.md
@@ -52,3 +52,25 @@ task (retrieval / filing recall). Until then, read +0.79 as "better aligned to t
 
 Data: `gen_wiki_relation_pairs.py` → `score_with_codex.py` → `convert_scored_to_graded.py` + `emit_blend_judge.py`.
 Sweep: `/tmp/mu_data/blend_sweep.sh`.
+
+## ✅ On the RIGHT metric — predict the judge SUPERPOSITION on held-out — the blend DOES earn its keep
+
+The LLM-scored SYM eval above was the wrong yardstick (it rewards LLM-alignment). The relevant measure (user):
+**how well does the model predict the judge superposition `T = (1−λ)·e5_ref ⊕ λ·graph_ref` on 360 held-out pairs
+never trained on** (`eval_blend_prediction.py`; `e5_ref` = raw-e5 cosine, `graph_ref` = conf-weighted
+`1/d + μ_HIER/μ_ELEM` from model_prod — the graph half is *not* LLM, so no confound). corr(SYM readout, T):
+
+| model | judge input | corr(SYM, T) | corr(SYM, graph_ref) |
+|---|---|---|---|
+| prod | agnostic | +0.746 | +0.714 |
+| A (LLM-only) | blend | +0.675 | +0.641 |
+| **B (blend)** | **blend** | **+0.847** | **+0.823** |
+
+**Verdict flips on the right target:** B (blend-trained, read under `judge=blend`) predicts the held-out
+superposition **best (+0.847)**; **LLM-only training *drifts away* from the graph half** (A graph_ref +0.641 <
+prod +0.714 — the LLM-alignment confound made visible), while the **blend judge recovers+improves it** (B
+graph_ref +0.823). And the **judge input matters** — B under `blend` +0.847 vs agnostic +0.787. So the blend
+adds exactly the structural signal LLM-only training discards. *Caveats:* single-seed checkpoints (the 3×3 sweep
+above didn't `--save`); `T` is graph-dominated here (`e5_ref` mean 0.907, low variance on this Wikipedia sample),
+so "predict T" leans on the graph half — which is the point, but a held-out set with more e5 spread would
+sharpen the e5 side. Multi-seed of the *checkpoints* is the confirming follow-up.
--- a/prototypes/mu_cosine/REPORT_blend_judge_sweep.md
+++ b/prototypes/mu_cosine/REPORT_blend_judge_sweep.md
@@ -99,7 +99,8 @@ this confirms the *structural* half most strongly; a higher-e5-variance held-out
 ### Circularity check — CLEARED (multi-λ + B-agnostic, `--lam-eval 0.3,0.5,0.7`)
 
 Concern (review #3491 III.L/M): `B` trained at λ=0.5 and `T` built at λ=0.5 share a recipe — is `B` just
-memorising it? Two tests say no. **(i) B beats A at every λ**, not only the trained one — mean corr(SYM, T(λ)):
+memorising it? Two tests say no. **(i) B beats A at every λ**, not only the trained one — **cross-seed mean
+(3 seeds)** of corr(SYM, T(λ)):
 
 | λ | B (blend) | A (LLM-only) | B (agnostic) |
 |---|---|---|---|
--- a/prototypes/mu_cosine/REPORT_blend_judge_sweep.md
+++ b/prototypes/mu_cosine/REPORT_blend_judge_sweep.md
@@ -86,6 +86,10 @@ sharpen the e5 side.
 | 3 | +0.664 | **+0.845** | +0.181 | +0.632 | +0.817 |
 | **mean** | **+0.697** | **+0.841** | **+0.144** | +0.667 | **+0.815** |
 
+**\* Scope caveat (read with the numbers):** `T` is **graph-dominated** on this held-out set — `e5_ref` mean
+0.907, std **0.032** (near-constant) vs `graph_ref` std 0.244 — so `corr(SYM, T)` leans on the *structural* half.
+These confirm the graph side strongly; the e5 side needs a higher-e5-variance held-out.
+
 *(prod baseline +0.746 / graph_ref +0.714.)* **B beats A on every seed** (mean +0.841 vs +0.697, vs prod +0.746);
 B robustly **recovers the graph half** (+0.815) that **LLM-only training drifts from** (+0.667); and reading
 under `judge=blend` beats agnostic for B on all three seeds. So the blend-judge advantage on the *right* metric
--- a/prototypes/mu_cosine/REPORT_blend_judge_sweep.md
+++ b/prototypes/mu_cosine/REPORT_blend_judge_sweep.md
@@ -91,3 +91,20 @@ B robustly **recovers the graph half** (+0.815) that **LLM-only training drifts 
 under `judge=blend` beats agnostic for B on all three seeds. So the blend-judge advantage on the *right* metric
 is not a single-seed fluke — it's consistent. (Scope: `T` is graph-dominated on this Wikipedia held-out set, so
 this confirms the *structural* half most strongly; a higher-e5-variance held-out would test the e5 side.)
+
+### Circularity check — CLEARED (multi-λ + B-agnostic, `--lam-eval 0.3,0.5,0.7`)
+
+Concern (review #3491 III.L/M): `B` trained at λ=0.5 and `T` built at λ=0.5 share a recipe — is `B` just
+memorising it? Two tests say no. **(i) B beats A at every λ**, not only the trained one — mean corr(SYM, T(λ)):
+
+| λ | B (blend) | A (LLM-only) | B (agnostic) |
+|---|---|---|---|
+| 0.3 | +0.863 | +0.724 | +0.795 |
+| 0.5 | +0.841 | +0.697 | +0.766 |
+| 0.7 | +0.828 | +0.682 | +0.751 |
+
+`B` predicts `T` across the *whole* blend family ⇒ it learned the **components** (e5 + graph), not the λ=0.5
+mixture. **(ii) B read *agnostically* (no blend judge input) still beats A-blend** (+0.766 vs +0.697 at λ=0.5) ⇒
+the **shared trunk** learned the graph geometry, not just the `judge=blend` head (the trunk/head-split story,
+now verified — the judge head adds a further ~+0.07). Caveat unchanged: `e5_ref` std is 0.032 here (near-constant),
+so these confirm the structural half; the e5 side needs a higher-e5-variance held-out.
--- a/prototypes/mu_cosine/REPORT_blend_judge_sweep.md
+++ b/prototypes/mu_cosine/REPORT_blend_judge_sweep.md
@@ -1,0 +1,44 @@
+# Does the constructed blend judge teach generality? — multi-seed test
+
+*Tests the hypothesis (user, 2026-07-05): training on judge *superpositions* teaches generality, so base
+held-out (never touched by the blend pairs) should rise. 3 arms × 3 seeds, fine-tuned from `model_prod` (800→600
+steps, full recipe, `--pairs mu_pairs_scored_cumulative` for the base SYM held-out, `UW_MU_GRAPH=100k_cats`).
+Metric: **base simplewiki SYM held-out corr** (1074 positives, disjoint from the Wikipedia blend pairs).*
+
+## Arms
+- **A `LLM only`** — graded round = the 880 Wikipedia pairs under `judge=gpt-5.5-low` (no constructed blend).
+- **B `fixed-λ`** — A + 880 `judge=blend` SYM rows, `blend = 0.5·μ_e5_sym ⊕ 0.5·P(SYM|1/d,asym-ops)`.
+- **C `random-λ`** — A + 880 `judge=blend` rows with `λ ~ U(0,1)` per pair (the blend regulariser).
+
+## Result
+
+| arm | seed 1 | seed 2 | seed 3 | mean |
+|---|---|---|---|---|
+| A `LLM only` (control) | +0.792 | +0.788 | +0.789 | **+0.790** |
+| B `fixed-λ blend` | +0.790 | +0.755 | +0.815 | +0.787 |
+| C `random-λ blend` | +0.785 | +0.782 | **+0.092** | +0.553 |
+
+*(model_prod, no fine-tune: +0.41. Within-seed blend−control: seed1 −0.002/−0.007, seed2 −0.033/−0.006, seed3
++0.026 / **−0.697 collapse**.)*
+
+## Verdict — the generality-via-constructed-blend hypothesis is NOT supported
+
+1. **The base-SYM lift is real and multi-seed-stable (+0.41 → ~+0.79) — but it is the DATA + fine-tune, not the
+   blend.** The control (`A`, no constructed blend) matches the blend arms (+0.790 vs +0.787). Fine-tuning on the
+   LLM Wikipedia round — which itself spans multiple judge tags (`gpt-5.5-low`, and the base pairs' `haiku`/
+   `graph`) — is what generalises the base SYM.
+2. **The constructed blend judge adds nothing over the LLM data** (B−A ≈ 0 across seeds).
+3. **Random-λ is risky** — seed 3 collapsed to +0.09. Forcing a per-pair random blend can destabilise training.
+
+## Takeaways
+- **Multi-seed caught a wrong single-seed story:** the +0.781 first-run lift was real *as a lift* but *not*
+  attributable to the superposition. This is exactly why single-seed deltas aren't believed.
+- **Data/judge diversity may buy the generality for free** — you may not need a hand-constructed blend judge;
+  training across several real judge tags already spans views.
+- **If you do use a constructed blend, it must beat a no-blend control across seeds** before you credit it, and
+  **random-λ needs a stability guard** (warmup, λ schedule, or a floor).
+- Not overturned: the *estimator* wins from earlier (joint head > PoE), and the base fine-tune's +0.79 SYM is a
+  genuine, reproducible improvement over model_prod's +0.41 — just not for the reason hypothesised.
+
+Data: `gen_wiki_relation_pairs.py` → `score_with_codex.py` → `convert_scored_to_graded.py` + `emit_blend_judge.py`.
+Sweep: `/tmp/mu_data/blend_sweep.sh`.
--- a/prototypes/mu_cosine/REPORT_blend_judge_sweep.md
+++ b/prototypes/mu_cosine/REPORT_blend_judge_sweep.md
@@ -37,8 +37,18 @@ Metric: **base simplewiki SYM held-out corr** (1074 positives, disjoint from the
   training across several real judge tags already spans views.
 - **If you do use a constructed blend, it must beat a no-blend control across seeds** before you credit it, and
   **random-λ needs a stability guard** (warmup, λ schedule, or a floor).
-- Not overturned: the *estimator* wins from earlier (joint head > PoE), and the base fine-tune's +0.79 SYM is a
-  genuine, reproducible improvement over model_prod's +0.41 — just not for the reason hypothesised.
+- Not overturned: the *estimator* win from earlier (joint head > PoE).
+
+## ⚠ The +0.79 is likely an LLM-ALIGNMENT artifact, not a general SYM gain (user, 2026-07-05)
+
+The base SYM held-out **eval targets are haiku-scored (LLM)**, and the fine-tune added **LLM data** (gpt-5.5-low)
++ more passes over the base haiku pairs. So "+0.41 → +0.79" measures the model getting better at **the LLM's
+notion of SYM** — which is exactly what an LLM-scored eval rewards. This is a **train/eval-share-a-judge
+confound**: more LLM training data → higher LLM-based eval, almost by construction (and why *all* arms rose
+together). It does **not** establish a *general* SYM improvement. To claim that, evaluate against an
+**independent, non-LLM** target — graph-structural relatedness on held-out pairs, a human check, or a downstream
+task (retrieval / filing recall). Until then, read +0.79 as "better aligned to the LLM SYM judge," scoped, not
+"better SYM."
 
 Data: `gen_wiki_relation_pairs.py` → `score_with_codex.py` → `convert_scored_to_graded.py` + `emit_blend_judge.py`.
 Sweep: `/tmp/mu_data/blend_sweep.sh`.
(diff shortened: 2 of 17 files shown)


prototypes/mu_cosine/mu_posterior.py
```diff
@@ -252,6 +252,47 @@ def readouts(pairs):                                   # pairs: list of (node, r
...
+    covers the sibling axis the vertical membership operators miss. NaN if either endpoint is absent."""
+    import torch
+    se = torch.load(struct_emb_path, weights_only=False)
+    sv = {n: v for n, v in zip(se["nodes"], se["emb"])}
+
+    def dist(node, root):
+        va, vb = sv.get(node), sv.get(root)
+        return float("nan") if va is None or vb is None else float(3.0 / (1.0 + (va - vb).norm()))
+    return dist
+
```
s243a, line 269: **`struct_dist_fn` docstring vs. empirical finding — update needed**

The docstring still says `1/d` is "not a re-reading of e5" and "covers the sibling axis the vertical membership operators miss," but `REPORT_mu_posterior_dist.md` now shows `corr(dist, e5)=+0.50` and that `1/d` failed the AURC earn-its-keep criterion on both splits. The docstring contradicts the corrected narrative in `DESIGN_sym_estimation_integration.md`.

Suggest updating to reflect the empirical scope:
```python
"""The lateral/sibling source (DESIGN_sym_estimation_integration.md): `1/d` via the learned
structural embedding, `3/(1+‖Δ‖)`. Trained separately on graph distance — but on Wikipedia
categories correlates +0.50 with e5, so it does not add signal over e5 for LLM-relation
classification (REPORT_mu_posterior_dist.md). Its distinct contribution remains the
blend-judge context, where it is a *constituent* of the SYM judge rather than a competitor.
NaN if either endpoint is absent."""
```

prototypes/mu_cosine/mu_posterior.py
```diff
@@ -252,6 +252,47 @@ def readouts(pairs):                                   # pairs: list of (node, r
...
+
+def aurc(conf, correct):
+    """Area under the risk-coverage curve (Geifman & El-Yaniv). conf = per-item confidence, correct ∈ {0,1}.
+    Sort by confidence desc; selective risk at coverage k = error rate among the k most-confident. Lower = the
+    confidence signal routes errors to the low-confidence tail better."""
+    conf, correct = np.asarray(conf, float), np.asarray(correct, float)
+    order = np.argsort(-conf)
+    err = 1.0 - correct[order]
```
s243a, line 279: **`aurc` implementation: off-by-one in cumulative risk and AURC definition choice**

The AURC is computed as `risk_at_k.mean()` over all `n` coverage levels (k=1..n), which equals the empirical AUC of the Risk-Coverage curve under uniform-coverage weighting. This is a valid selective-risk metric. However, two notes:

1. The current implementation includes the full-coverage point (k=n) where confidence is irrelevant — some formulations exclude it or normalise by coverage. This is a definitional choice but worth documenting explicitly so future readers don’t compare against a differently-normalised reference.

2. The docstring cites Geifman & El-Yaniv, but their original definition integrates over coverage \[0,1\] weighted by a stopping criterion; the discrete mean-of-cumulative-risk is the standard approximation, which is fine, but calling it out explicitly helps.

Suggest adding a one-line note:
```python
# discrete mean over coverage levels k=1..n (includes full coverage; standard approximation to G&EY’s integral)
```

prototypes/mu_cosine/mu_posterior.py
```diff
@@ -252,6 +252,47 @@ def readouts(pairs):                                   # pairs: list of (node, r
...
+    risk_at_k = np.cumsum(err) / np.arange(1, len(err) + 1)
+    return float(risk_at_k.mean())
+
+
+def aurc_boot(conf, correct, B=500, seed=0):
+    """AURC point estimate + percentile bootstrap 95% CI (AURC is noisy on small held-out sets)."""
+    rng = np.random.default_rng(seed)
+    conf, correct = np.asarray(conf, float), np.asarray(correct, float)
```
s243a, line 287: **`aurc_boot`: replacement resampling re-creates `aurc` objects on sub-sliced arrays — potential index alignment issue**

```python
vals = [aurc(conf[ix], correct[ix]) for ix in (rng.integers(0, n, n) for _ in range(B))]
```

`rng.integers(0, n, n)` generates integer indices, and `conf[ix]` / `correct[ix]` fancy-indexes the arrays — this is correct. One edge case: if `conf` and `correct` are passed as plain Python lists (not NumPy arrays), `conf[ix]` will fail because lists don’t support NumPy fancy indexing. The `np.asarray()` conversion two lines above handles `conf` and `correct` but the bootstrap loop sees the post-`asarray` versions, so this is fine as written. However, if the function is ever called from a context where `correct` or `conf` are PyTorch tensors (likely given the rest of the codebase), `asarray` on a CUDA tensor will raise. Consider:
```python
conf = np.asarray(conf.cpu() if hasattr(conf, 'cpu') else conf, float)
correct = np.asarray(correct.cpu() if hasattr(correct, 'cpu') else correct, float)
```

prototypes/mu_cosine/mu_posterior.py
```diff
@@ -332,11 +390,23 @@ def main():
...
+        # #3488 review flagged; pairs spanning the two node-sets are dropped so the split is strictly disjoint.
+        if args.split == "node-disjoint":
+            nodes = sorted({r["node"] for r in fit_set} | {r["root"] for r in fit_set})
+            random.Random(0).shuffle(nodes)
+            hn = set(nodes[:int(args.held_frac * len(nodes))])
+            held = [i for i, r in enumerate(fit_set) if r["node"] in hn and r["root"] in hn]
+            train = [i for i, r in enumerate(fit_set) if r["node"] not in hn and r["root"] not in hn]
+            print(f"\nsplit=node-disjoint: {len(train)} train / {len(held)} held "
+                  f"({len(fit_set)-len(train)-len(held)} cross-split pairs dropped)")
+        else:
```
s243a, line 410: **Node-disjoint split: cross-split pairs are silently dropped — class-balance should be logged**

The disjoint split drops `len(fit_set) - len(train) - len(held)` cross-boundary pairs. With 880 pairs, the node-disjoint report shows only 60 held pairs (vs 220 random), a 73% reduction. This is architecturally correct but worth sanity-checking that the class distribution in the 60-pair held set still spans all relations. If the `see_also` class (which was already tiny at 6 pairs in the LLM run) gets dropped entirely from the held set, the AURC comparison loses the lateral-axis signal that `1/d` was hypothesised to capture.

Suggest logging the per-relation counts of held pairs:
```python
from collections import Counter
print(f"  held relation distribution: {dict(Counter(rels[i] for i in held))}")
```
This is diagnostic for interpreting the AURC null result — a held set with no `assoc`/`see_also` pairs can’t surface a lateral-axis benefit.

prototypes/mu_cosine/emit_blend_judge.py
```diff
@@ -0,0 +1,78 @@
...
+        c = ln.rstrip("\n").split("\t")
+        if len(c) >= 2:
+            pairs.append((c[0], c[1]))
+
+    d = torch.load(a.e5_cache, weights_only=False)
+    idx = {n: i for i, n in enumerate(d["names"])}
+    parents, children, deg = load_dag()
+    tok = Tokenizer(d["query"], d["passage"], idx, parents, deg)
+    pairs = [(x, y) for x, y in pairs if x in idx and y in idx]
+    model = build_model(a.model, dev)
```
s243a, line 60: **`sdist` called twice per pair — NaN check is redundant and expensive**

```python
dist01 = np.array([min(1.0, sdist(x, y) / 3.0) if sdist(x, y) == sdist(x, y) else 0.0 for x, y in pairs])
```

`sdist(x, y)` is called twice for every pair: once for the NaN check (`v == v`) and once for the value. Since `struct_dist_fn` does a dict lookup + tensor subtraction + norm per call, this is 2x the work with no benefit. Simple fix:
```python
raw = [sdist(x, y) for x, y in pairs]
dist01 = np.array([min(1.0, v / 3.0) if v == v else 0.0 for v in raw])
```

Also, the NaN-fill with 0.0 is a reasonable choice (treats missing-embedding pairs as maximally distant), but it silently inflates `mu_graph_sym` toward zero for those pairs. Since `emit_blend_judge.py` already filters `pairs` through `idx` membership, NaN should only occur if the struct embedding is missing nodes that the e5 cache has. Worth an assertion or a warning count.

prototypes/mu_cosine/emit_blend_judge.py
```diff
@@ -0,0 +1,78 @@
...
+    n_ops = model.op_emb.num_embeddings
+
+    ow = torch.zeros(1, n_ops); ow[0, OPS["SYM"]] = 1.0
```
s243a, line 63: **Confidence weight constants `c_dist`, `c_subcat`, `c_elem` are hardcoded defaults without a source citation**

The default weights `c_dist=0.35`, `c_subcat=0.72`, `c_elem=0.82` appear to be empirically measured reliabilities, but neither the script’s docstring nor the argument help strings cite where those numbers come from. The playbook (pitfall #3) explicitly warns against hand-set independent confidence multipliers — these weights occupy exactly that role for `mu_graph_sym`.

This is a design tension: the weights are used *within* the blend judge construction, not as the final fusion combiner, so they’re not directly subject to the “don’t hand-set combiners over correlated sources” rule. But they should be either:
1. **Documented** with the PR/experiment that produced them (e.g. `# from #XXXX: separability-weighted reliability of each graph source`), or
2. **Learned** via the `JointPosterior` workflow if the graph sources are also correlated with each other.

For now, at minimum add a comment referencing the origin experiment for these three values.

prototypes/mu_cosine/eval_blend_prediction.py
```diff
@@ -0,0 +1,94 @@
...
+
+def score_sym(model, tok, pairs, dev, judge=None, corpus="enwiki"):
+    items = [((a, b, OPS["SYM"]) if judge is None else (a, b, OPS["SYM"], CORPORA[corpus], JUDGES[judge]))
+             for a, b in pairs]
+    out = []
+    for i in range(0, len(items), 512):
+        bd = tok.build(items[i:i + 512], train=False)
+        bd = {k: (v.to(dev) if torch.is_tensor(v) else v) for k, v in bd.items()}
+        with torch.no_grad():
+            out += model(**bd).cpu().tolist()
```
s243a, line 35: **`q` and `p` are loaded but `q` is unused — dead variable**

```python
q, p = d["query"].numpy(), d["passage"].numpy()
```

`p` is used later for `e5_ref` (`p[idx[x]] @ q[idx[y]]`... wait, actually `q` *is* used in the dot product). Let me re-read: `e5_ref = np.array([(float(p[idx[x]] @ q[idx[y]]) + 1.0) / 2.0 for x, y in pairs])`. Both are used. This is fine — but naming them `p` (passage) and `q` (query) and then using them as `p[x] · q[y]` implies an asymmetric dot product (passage embedding of node x with query embedding of node y). Is this the intended cosine direction? For symmetric SYM relatedness you’d typically want `q[x] · q[y]` or `p[x] · p[y]`. Please add a comment clarifying the intended directionality and that this matches how `score_pairs` computes `mu_e5_sym` in `emit_blend_judge.py`.

prototypes/mu_cosine/eval_blend_prediction.py
```diff
@@ -0,0 +1,94 @@
...
+    d = torch.load(a.e5_cache, weights_only=False)
+    idx = {n: i for i, n in enumerate(d["names"])}
+    q, p = d["query"].numpy(), d["passage"].numpy()
+    parents, children, deg = load_dag()
+    tok = Tokenizer(d["query"], d["passage"], idx, parents, deg)
+    pairs = []
+    for ln in open(a.pairs, encoding="utf-8"):
+        if ln.startswith("#"):
+            continue
+        c = ln.rstrip("\n").split("\t")
```
s243a, line 66: **`graph_ref` uses `model_prod` memberships — this is the correct design choice but should be guarded**

Using `model_prod` (the reference checkpoint, fixed) to compute `graph_ref` for all evaluated models is the right call: it makes the target independent of the fine-tuned checkpoints being scored, eliminating self-referential confounds. However, if someone calls this script with `--ref` pointing to one of the fine-tuned models (e.g., `model_A_ft.pt`), the target `T` becomes entangled with the eval model and the independence property breaks.

Suggest adding a guard or at least a clear warning:
```python
if a.ref in [spec.split("=", 1)[1] for spec in a.models]:
    print("WARNING: --ref model is also in --models. The target T will be entangled "
          "with the evaluated checkpoint, violating the independence guarantee.")
```

prototypes/mu_cosine/gen_wiki_relation_pairs.py
```diff
@@ -0,0 +1,105 @@
...
+    ap.add_argument("--per", type=int, default=220, help="target pairs per stratum")
+    ap.add_argument("--seed", type=int, default=0)
+    ap.add_argument("--out", required=True)
+    a = ap.parse_args()
+    rng = random.Random(a.seed)
+
+    import torch
+    se = torch.load(a.struct_emb, weights_only=False)
+    inemb = set(se["nodes"])
+    parents, children, deg = load_dag(a.graph)
```
s243a, line 31: **`len(set(parents)|set(children))` uses bitwise OR on sets — correct but unidiomatic; also `deque` is imported but unused**

```python
print(f"struct-emb nodes {len(inemb)}; graph {len(parents)|len(children) if False else len(set(parents)|set(children))} nodes")
```

The `if False else` branch is always taken, making the first branch dead code. This reads as a debug leftover. Consider simplifying to:
```python
all_nodes = set(parents) | set(children)
print(f"struct-emb nodes: {len(inemb)}; graph nodes: {len(all_nodes)}; overlap: {len(inemb & all_nodes)}")
```

Also, `from collections import deque` is imported at the top but never used in the script. Remove to avoid confusion.

prototypes/mu_cosine/gen_wiki_relation_pairs.py
```diff
@@ -0,0 +1,105 @@
...
+    # 2) SUBTOPIC: grandchild → grandparent
+    for p in parents_l:
+        if sum(1 for x in rows if x[2] == "subtopic") >= a.per:
+            break
+        gk = [g for c in kids[p] for g in kids.get(c, [])]
+        if gk:
+            add(rng.choice(gk), p, "subtopic")
+    # 3) SEE_ALSO: siblings (share a parent, neither ancestor of the other)
+    for p in parents_l:
+        if sum(1 for x in rows if x[2] == "see_also") >= a.per:
```
s243a: **Stratified sampling: `none` stratum uses full-BFS ancestors, but `see_also` uses only 4-hop ancestors — cap inconsistency can leak near-siblings into `none`**

The `anc(x, cap=4)` function traverses at most 4 hops upward. In the `none` filter:
```python
if y in anc(x) or x in anc(y) or (anc(x) & anc(y)):
    continue
```

Two nodes sharing an ancestor at hop 5+ will pass this check and be labelled `none`, even though they may be structurally related. For a deep taxonomy (Wikipedia categories can be 10+ hops deep), this can admit structurally-related pairs into `none`. This potentially biases the `none` class toward pairs that have weak but real structural connections, which could inflate `corr(1/d, none_label)` and make `1/d`’s separability look worse than it is on the `assoc`/`see_also` classes.

If precision of the `none` class matters for the ablation, consider using a stricter cap or a BFS to full depth (with a visited set for termination). Document the chosen cap as a known approximation.

prototypes/mu_cosine/convert_scored_to_graded.py
```diff
@@ -0,0 +1,52 @@
...
+NAMED = ["element_of", "subcategory", "subtopic", "super_category", "see_also", "assoc", "unknown", "none"]
+OP_OF = {"element_of": "ELEM", "subcategory": "HIER", "subtopic": "HIER", "super_category": "HIER",
+         "see_also": "SYM", "assoc": "SYM", "unknown": "SYM", "none": "SYM"}
+
+
+def main():
+    ap = argparse.ArgumentParser()
+    ap.add_argument("--scored", required=True)
+    ap.add_argument("--out", required=True)
+    ap.add_argument("--corpus", default="enwiki")
```
s243a, line 25: **`OP_OF` mapping collapses `unknown` and `none` into `SYM` — epistemically questionable**

```python
OP_OF = {"element_of": "ELEM", "subcategory": "HIER", "subtopic": "HIER", "super_category": "HIER",
         "see_also": "SYM", "assoc": "SYM", "unknown": "SYM", "none": "SYM"}
```

Labelling LLM-output `unknown` and `none` as `op=SYM` and then fitting the JointPosterior on these pairs trains the SYM classifier on signal that may include genuinely unrelated pairs (the LLM said `none`). This could suppress `1/d`’s separability on the SYM class by adding noise pairs with near-zero `mu` ground truth.

If `none` pairs are included as SYM negatives (mu≈01), this is intentional and should be documented. If they’re unintended, filter them out before writing. At minimum, log how many `none`/`unknown` rows were mapped to SYM so readers can assess the class noise level.

prototypes/mu_cosine/test_mu_posterior_dist.py
```diff
@@ -0,0 +1,60 @@
...
+from mu_posterior import aurc, aurc_boot, margin_conf, struct_dist_fn
+
+
+def test_aurc_monotone():
+    # confidence that TRACKS correctness ⇒ low AURC; ANTI-tracks ⇒ high; all-correct ⇒ 0; all-wrong ⇒ 1.
+    correct = [1, 1, 0, 0]
+    good = [0.9, 0.8, 0.2, 0.1]      # confident-when-correct
+    bad = [0.1, 0.2, 0.8, 0.9]       # confident-when-wrong
+    assert aurc(good, correct) < aurc(bad, correct), "AURC must reward correctness-tracking confidence"
+    assert abs(aurc([0.9, 0.8, 0.7, 0.6], [1, 1, 1, 1]) - 0.0) < 1e-9, "all-correct ⇒ AURC 0"
```
s243a, line 17: **`test_aurc_monotone`: exact value assertion is valuable but the comment has an arithmetic slip**

```python
# exact value for the 'good' case: risk_at_k = [0,0,1/3,1/2] ⇒ mean 0.2083…
assert abs(aurc(good, correct) - (0 + 0 + 1/3 + 1/2) / 4) < 1e-9
```

Let’s verify: items sorted by conf desc = [0.9(correct), 0.8(correct), 0.2(wrong), 0.1(wrong)]. Risk at k: k=1: 0/1=0; k=2: 0/2=0; k=3: 1/3; k=4: 2/4=0.5. Mean = (0+0+1/3+1/2)/4 = (5/6)/4 = 5/24 ≈ 0.2083. ✓ The comment and assertion are correct. However, the comment says `0.2083…` which is `5/24`; optionally add `= 5/24` for exactness. Minor nit, not blocking.

prototypes/mu_cosine/test_mu_posterior_dist.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+def test_struct_dist_fn():
+    import torch
+    d = tempfile.mkdtemp()
+    path = os.path.join(d, "se.pt")
+    nodes = ["A", "B"]
+    emb = torch.tensor([[0.0, 0.0], [3.0, 4.0]])      # ‖Δ‖ = 5 ⇒ 3/(1+5) = 0.5
+    torch.save({"nodes": nodes, "emb": emb}, path)
+    dist = struct_dist_fn(path)
```
s243a, line 48: **`test_struct_dist_fn`: the `dist("A","A")` assertion has a surprising value — clarify intent**

```python
assert dist("A", "A") == 3.0, "identical nodes ⇒ 3/(1+0)=3"
```

This is correct by the formula `3/(1+‖Δ0‖) = 3/(1+0) = 3`, but `dist` > 1 for a self-comparison is an unusual property for a value that gets clipped to 1 in `emit_blend_judge.py` (`min(1.0, sdist(x,y)/3.0)`). The clip formula `v/3` was presumably designed so that `dist(≈0) → ~1` and `dist(→∞) → 0`. Adding a test that covers the scaling identity:
```python
assert abs(min(1.0, dist("A", "A") / 3.0) - 1.0) < 1e-9, "identical ⇒ proximity 1.0 after /3 clip"
```
...would verify the end-to-end normalisation used in `emit_blend_judge.py`, not just the raw function.

prototypes/mu_cosine/DESIGN_uncertainty_estimation_playbook.md
```diff
@@ -0,0 +1,108 @@
...
+
+## The five pitfalls (each one we actually hit)
+
+1. **Sources are not independent → naive/inverse-variance fusion over-confidences.** In this project every
+   model readout *consumes frozen e5*, so `e5`, `μ_SYM`, `μ_HIER`, `μ_ELEM` are correlated (measured **+0.751**,
+   #3359). A product-of-experts / precision-weighting that assumes independence double-counts the shared signal.
+   Evidence: a factored equal-weight PoE scored **50.9%, below the 51.1% majority baseline** (#3359). *Fix:* a
+   learned joint head over the full vector (it down-weights the redundant part), or model the residual /
+   down-weight by separability (#3357).
+2. **Anti-correlated sources carry structure that additive weighting throws away.** Subcategory and element
```
s243a, line 20: **Pitfall count in heading says "five" but six are listed — minor heading mismatch**

The section header reads `## The five pitfalls (each one we actually hit)` but the list contains six items (#1–#6), with #6 being the important train/eval-judge-sharing confound. Update to:
```
## The six pitfalls (each one we actually hit)
```
or keep `five` and move #6 under its own heading (it was a later addition from the SYM blend sweep and arguably warrants emphasis as a section). Either way, the mismatch will confuse future readers who’ll count to six.

prototypes/mu_cosine/DESIGN_uncertainty_estimation_playbook.md
```diff
@@ -0,0 +1,108 @@
...
+Beyond constructing a good target, training on **judge superpositions** — a target `(1−λ)·judgeA ⊕ λ·judgeB`,
+especially with **random λ per example** — is a **multi-view consistency regulariser**. It *asserts that the two
+judges are two views of one underlying quantity*, so to fit the whole blend family the model can't latch onto
+either judge's idiosyncrasies; it must find the **shared signal** they both approximate. That shared signal is
+the general notion (of relatedness, lineage, …).
+
+- **The trunk/head split makes it clean.** The **shared trunk** learns the invariant (generality); the
+  **per-judge `judge_emb` rows** absorb each judge's *offset/bias* (fidelity). You get generality **and**
+  per-judge calibration — not a lossy average. This is the "operator superposition as a regulariser" idea
+  (`DESIGN_inferred_operator_superposition.md` §7) lifted from *operators* to *judges*.
```
s243a, line 76: **Superposition ratio section: the cautionary and confirmatory results are interleaved in a way that’s hard to parse on a first read**

The paragraph under `## Why train on judge superpositions — it teaches generality` contains three conceptual beats:
1. The theoretical argument for why superpositions teach generality (trunk/head split)
2. The 3×3 sweep refutation on LLM-scored eval
3. The resolution — the blend DOES earn its keep on the right metric

These are currently nested inside a single bullet block with `**Symptom of success (and a cautionary result):**` covering beats 2 and 3, making the reversal easy to miss. A future reader skimming for the takeaway may stop at the refutation. Consider restructuring as three explicit sub-headings or a clearly labelled `> **Resolution:**` callout block (as used elsewhere in the doc) to separate the overturning finding. The PR summary does this well; the playbook body should match.

prototypes/mu_cosine/REPORT_blend_judge_sweep.md
```diff
@@ -0,0 +1,93 @@
...
+Data: `gen_wiki_relation_pairs.py` → `score_with_codex.py` → `convert_scored_to_graded.py` + `emit_blend_judge.py`.
+Sweep: `/tmp/mu_data/blend_sweep.sh`.
+
+## ✅ On the RIGHT metric — predict the judge SUPERPOSITION on held-out — the blend DOES earn its keep
+
+The LLM-scored SYM eval above was the wrong yardstick (it rewards LLM-alignment). The relevant measure (user):
+**how well does the model predict the judge superposition `T = (1−λ)·e5_ref ⊕ λ·graph_ref` on 360 held-out pairs
+never trained on** (`eval_blend_prediction.py`; `e5_ref` = raw-e5 cosine, `graph_ref` = conf-weighted
+`1/d + μ_HIER/μ_ELEM` from model_prod — the graph half is *not* LLM, so no confound). corr(SYM readout, T):
+
```
s243a, line 62: **`T` is described as graph-dominated (`e5_ref` mean 0.907, low variance) — this is a significant caveat that should be quantified in the headline result table**

The multi-seed confirmed result (`B mean +0.841 vs A +0.697`) is presented as the primary verdict, but the scope caveat (`T` leans on the graph half because `e5_ref` has low variance on this Wikipedia sample) appears only at the end. A reader scanning the tables will see the correlation numbers without the confound.

Suggest adding a `*` footnote directly to the result tables:
```
| **mean** | **+0.697** | **+0.841** | **+0.144** | +0.667 | **+0.815** |
```
→ Add a footnote: `* T is graph-dominated (e5_ref mean 0.907, low variance); these numbers primarily confirm the structural half. A higher e5-variance held-out set is needed to test the semantic half.`

This prevents overclaiming in citations of this report.

s243a (commented):
## Academic Review — PR #3496: Judge Superpositions, JointPosterior + 1/d, Graded-Wikipedia Round, and Blend Judge

### Overall Assessment

This is a well-executed, epistemically honest piece of ML research. The PR is notable for *walking back* tempting conclusions under scrutiny — the scope-correction on `1/d` and the LLM-alignment confound identification in the blend sweep are exactly the kind of intellectual discipline that makes empirical findings durable. The three-tier structure (estimator win, data pipeline, blend judge with right/wrong target) is clearly motivated and the multi-seed confirmation on the key result is the correct standard.

The review raises issues across four categories. None are blockers for the research conclusions, but several affect correctness and future reproducibility.

---

### 🔴 Issues Requiring Attention Before Merge

**1. `struct_dist_fn` docstring contradicts empirical findings (mu_posterior.py)**
The docstring still claims `1/d` is "not a re-reading of e5" — this is the *prior* hypothesis, not the post-experiment finding (`corr=+0.50`, AURC null). See inline comment at line 267.

**2. Node-disjoint split drops classes silently (mu_posterior.py)**
The 60-pair held set after node-disjoint splitting may have dropped the `see_also` class (which was 6 pairs before splitting). An empty lateral class means the AURC null result can't distinguish "1/d adds no signal" from "the held set has no lateral pairs to test." Log the per-relation held counts. See inline comment at line 404.

**3. `OP_OF` maps `none`→`SYM` silently (convert_scored_to_graded.py)**
If LLM-labelled `none` pairs (unrelated nodes) are being ingested as SYM training rows with low μ, this is intentional but undocumented. If unintentional, it introduces label noise that could confound the JointPosterior fit and the `1/d` ablation. Clarify and log counts. See inline comment at line 22.

**4. Pitfall count mismatch in playbook heading (DESIGN_uncertainty_estimation_playbook.md)**
"Five pitfalls" is listed in the header but six are enumerated. Trivial fix but will confuse future readers. See inline comment at line 20.

---

### 🟡 Design / Methodology Notes

**5. `emit_blend_judge.py`: hardcoded `c_dist=0.35, c_subcat=0.72, c_elem=0.82` without citation**
These are the "confidence weights" for constructing `mu_graph_sym`. They occupy the role the playbook warns against (hand-set weights), though here they're constituents of a *constructed judge*, not the final combiner. They need a reference to the experiment that produced them. See inline comment at line 55.

**6. `eval_blend_prediction.py`: no guard against `--ref` being one of the evaluated models**
If `model_prod.pt` is also passed in `--models`, `graph_ref` becomes entangled with the evaluated model, breaking the independence of target `T`. A one-line check prevents this footgun. See inline comment at line 63.

**7. `gen_wiki_relation_pairs.py`: ancestor cap=4 inconsistency for the `none` stratum**
Pairs with shared ancestors at hop 5+ will pass the `none` filter and be labelled unrelated, introducing structurally-related pairs into the `none` class. This could affect class quality for deep Wikipedia taxonomies. Document the cap as a known approximation. See inline comment at line 74.

**8. `REPORT_blend_judge_sweep.md`: T graph-dominance caveat is not visible at the headline table**
The multi-seed result tables (`+0.841 vs +0.697`) should carry a footnote noting that T is graph-dominated (`e5_ref` mean 0.907, low variance), so these numbers confirm the structural half most strongly. Without this, citations of the table will overclaim. See inline comment on line 62.

**9. Superposition-teaches-generality section has a confusing reversal structure (DESIGN_uncertainty_estimation_playbook.md)**
The refutation and the resolution are interleaved within a single bullet, making the recovery easy to miss on first read. The PR summary presents this arc better than the playbook body. Consider restructuring with an explicit `> **Resolution:**` callout. See inline comment at line 76.

---

### 🟢 Minor / Nits

**10. `emit_blend_judge.py`: `sdist` called twice per pair in the NaN check**
2x dict lookups + tensor norms with no benefit. One-line fix. See inline comment at line 52.

**11. `gen_wiki_relation_pairs.py`: unused `deque` import + dead `if False else` branch**
Code hygiene. See inline comment at line 32.

**12. `eval_blend_prediction.py`: `p[x] @ q[y]` dot product direction comment missing**
Clarify whether the passage-query dot product for `e5_ref` matches how `score_pairs` computes `mu_e5_sym` in `emit_blend_judge.py`. See inline comment at line 35.

**13. `test_mu_posterior_dist.py`: self-distance normalisation not covered**
`dist("A","A")=3.0` passes but doesn't verify the `/3` clipping to 1.0 used downstream. One additional assertion closes this. See inline comment at line 48.

**14. `aurc_boot`: CUDA tensor defensiveness**
If called with PyTorch tensors on GPU, `np.asarray()` will raise. Low-priority given current call sites, but worth a guard. See inline comment at line 283.

---

### Positive Highlights

- **The scope-correction in `REPORT_mu_posterior_dist.md`** is textbook-quality epistemic hygiene: the exact framing ("redundant for LLM-relation prediction, not for the SYM judge construction") is precise and non-defensive.
- **Multi-seed confirmation** before crediting the blend result is exactly pitfall #6 applied correctly.
- **The `DESIGN_uncertainty_estimation_playbook.md`** is a high-value artefact — the six pitfalls are each grounded in a specific numbered prior experiment. The skill mirror in `skills/` is a practical touch for Claude Code sessions.
- **The `JointPosterior > PoE` result** (log-loss 1.08 vs 1.55/1.26) is clean and reusable across settings regardless of the 1/d verdict.
- **`AURC + bootstrap CI` as the earn-its-keep criterion** is the right approach for small held sets — the pre-registered threshold (with-CI must sit below without-point) prevents post-hoc rationalisation.

s243a (commented):
## Second Review — PR #3496: Judge Superpositions, JointPosterior + 1/d, Graded-Wikipedia Round, Blend Judge

*This pass re-reads the full diff independently and goes deeper on interactions between files, statistical methodology, and some issues the first review touched but did not fully analyse.*

---

### Architecture & Cross-File Consistency

**A. `emit_blend_judge.py` and `eval_blend_prediction.py` share the same confidence-weight triad (`c_dist=0.35`, `c_subcat=0.72`, `c_elem=0.82`) but are separate scripts with separate `--c-*` arguments.** If someone reruns `emit_blend_judge.py` with non-default weights (e.g. during a sensitivity sweep), the `graph_ref` computed by `eval_blend_prediction.py` will no longer match the judge `T` was computed against — they will silently diverge. The target `T` is defined as the same formula, so this is *not* just a documentation issue: a mismatched `c_dist` value will change both `T` and the training target in the *same direction*, meaning the metric `corr(SYM, T)` will be partially self-fulfilling when `graph_ref` uses the same weights the training data was produced with. Consider either:
1. Saving the weights that produced `blend_judge_pairs.tsv` to a sidecar JSON, and having `eval_blend_prediction.py` read them from there, or
2. Adding an assertion/warning that the defaults must match between the two scripts.

**B. `mu_attention.py`: `JUDGES["blend"] = 7` is now the highest index.** The `judge_emb` embedding table in any checkpoint trained *before* this PR has `num_embeddings = 7`; `JUDGES["blend"] = 7` is a 0-indexed out-of-bounds index for those checkpoints. The guard in `eval_blend_prediction.py`:
```python
if judge == "blend" and m.judge_emb.num_embeddings <= JUDGES["blend"]:
    continue
```
uses `<=` which is correct (`<= 7` skips the blend read when there are only 7 rows, i.e. indices 0–6). Good. But this silent skip will mislead anyone who runs the script against a pre-PR checkpoint expecting blend output — they'll get no blend row in the results table with no explanation. Suggest printing a message:
```python
print(f"  {name}: no blend judge_emb row (checkpoint has {m.judge_emb.num_embeddings} rows) — skipping judge=blend")
```

**C. `gen_wiki_relation_pairs.py` outputs 8 columns (the `score_inferred_tail` schema), but `mu_posterior.py`'s `load_pairs` and `convert_scored_to_graded.py` read different schemas.** The pipeline is: `gen_wiki_relation_pairs.py` → `score_with_codex.py` → `convert_scored_to_graded.py` → `mu_posterior.py`. This is implicit in the REPORTs and docstrings but nowhere in the codebase is there a schema registry or validation. If `score_with_codex.py` is updated to emit extra columns, `convert_scored_to_graded.py` will silently use wrong column indices because it accesses `c[pcol[r]]` via a header-parsed dict — which is actually robust. However, `gen_wiki_relation_pairs.py`'s output has a `raw` column that is always empty (`f"{n}\t{r}\t{rel}\t1.0\twiki_rel\tcategory\tcategory\t\n"`). Intentional? The `score_inferred_tail` prompt builder may expect `raw` to contain the page abstract or similar context for the LLM scoring call. An empty `raw` field may produce weaker LLM labels.

---

### Statistical Methodology

**D. The AURC pre-registration criterion is stated but not actually pre-registered in the PR description or a dated artefact.** The playbook says: *"A new source earns its keep iff the with-source AURC CI sits below the without-source point estimate."* This is a reasonable selective-risk threshold. But the REPORT's ablation table shows the criterion was checked *after* the fact in the same run that produced the data — it is not a pre-registered criterion with a date stamp that predates the data collection. This is an internal project, so formal pre-registration isn't strictly required, but the language "pre-registered criterion" in the code comment (`mu_posterior.py`, line 431 area) is stronger than what was actually done. Either remove the word "pre-registered" or add a dated note showing the criterion was fixed before the Wikipedia round was scored.

**E. `aurc_boot` uses a *percentile* bootstrap CI, not a bias-corrected accelerated (BCa) CI.** For AURC on small samples (n=60 held pairs, n=220 random), the percentile bootstrap can be poorly calibrated — it undercovers when the sampling distribution is skewed (which it typically is for extreme AURC values near 0 or 1). The result tables show AURC ~ 0.17–0.18 with CIs of width ~0.15–0.20, which is in the moderate range where the percentile method is acceptable. But the code comment should note this limitation. For future runs with very small held sets (< 50 pairs), consider BCa. Low-priority given current n.

**F. ECE binning is hardcoded at 10 bins (`bins=10`) in `_eval`, but the playbook says "ECE with a stated binning (bins sensitive to choice)"** — and doesn't state the binning. With 60 held pairs and 10 equal-width bins, most bins will be empty or have 1–2 items; the ECE estimate will be dominated by whichever bin happens to catch the calibration error. With 60 pairs, a more reliable ECE would use adaptive bins (equal-count) or explicitly report the bin counts alongside ECE so readers can judge reliability. Add the n-per-bin output or switch to equal-count bins for small held sets.

**G. `corr(1/d, e5) = +0.50` is computed and reported, but the correlation is between the raw `dist` value and what exactly?** The REPORT says `dist` correlates +0.50 with `e5` — but `e5` in `src_vals` is the `e5_mu_fn` output (a calibrated P(μ|op) band probability), not raw cosine similarity. `dist01` in `emit_blend_judge.py` is `min(1.0, sdist(x,y)/3.0)` ∈ [0,1]. These are on different scales and the correlation is between two different transformations of underlying embeddings. The +0.50 figure is presented as if it measures fundamental redundancy between the structural embedding space and the e5 embedding space — but it's really the correlation between two *derived* scores on this particular Wikipedia sample. A scatter plot or breakdown by relation type would clarify whether the correlation is uniform or driven by the `subcategory`/`subtopic` majority class (where both e5 and 1/d correctly give high similarity).

---

### Code Correctness

**H. `convert_scored_to_graded.py`: `mu = float(c[fi])` uses `E_mu_fwd` as the μ target, but the JointPosterior is fit on the *argmax relation* label, not on μ.** The graded pairs output has a `mu` column that is written but then (in `mu_posterior.py`'s `load_pairs`) is used as the *ground-truth μ* for band-rejection. So for Wikipedia pairs, `E_mu_fwd` from the LLM scoring pass becomes the μ used by the outlier filter. Is `E_mu_fwd` from `score_with_codex.py` the forward-pass expected μ, or is it the LLM's graded relatedness score? If it's the model's own prediction (not an LLM-provided μ), then using it as the ground truth for the outlier rejection filter is circular — the model is being judged against its own output.

**I. `emit_blend_judge.py`: all emitted rows have `relation = see_also`.** 
```python
f.write(f"{x}\t{y}\t{blend:.3f}\tSYM\tsee_also\tcategory\tcategory\t{a.corpus}\tblend\t1.0\n")
```
But the input `--pairs` file contains subcategory, subtopic, see_also, and none pairs. A subcategory pair in the input will be emitted as `op=SYM, relation=see_also` in the blend output, which means if `mu_posterior.py` uses the `relation` field for its band filter or per-relation weight analysis, the relation taxonomy is corrupted. If `see_also` is intentional (all blend-judge rows are treated as the generic SYM lateral relation regardless of true structure), this should be documented explicitly — it means the blend judge is purely a *SYM μ-value* teacher with no relation-type signal.

**J. `eval_blend_prediction.py`: `score_sym` builds items as either a 3-tuple or a 5-tuple depending on `judge`**, feeding them to `tok.build()`. The `Tokenizer.build` method presumably dispatches on tuple length. This is implicit — if `tok.build` ever changes its dispatch logic, this will silently produce wrong outputs with no type error. Consider using a named dict or dataclass.

**K. `gen_wiki_relation_pairs.py`: the `subcategory` stratum sampling has a subtle ordering bug.** The strata are filled sequentially with `sum(1 for x in rows if x[2] == ...)` as the stopping condition, which is O(n × per) — O(n²) in the worst case. For `--per 220` and a large graph, this could be slow. The `subtopic` stratum additionally iterates over `parents_l` again from scratch after the first loop, so with a large `parents_l`, the two O(n²) loops could become noticeable. Pre-computing stratum counts with a `Counter` would be O(n).

---

### Documentation & Reproducibility

**L. The `REPORT_mu_posterior_dist.md` run command references `wiki_rel_graded_pairs.tsv` and `wiki_rel_e5.pt`, but `gen_wiki_relation_pairs.py` produces `wiki_rel_pairs.tsv` (pre-scoring).** The graded file is `wiki_rel_pairs.tsv` → `score_with_codex.py` → `wiki_rel_scored.tsv` → `convert_scored_to_graded.py` → `wiki_rel_graded_pairs.tsv`. The intermediate file names and the step that produces the e5 cache (`wiki_rel_e5.pt`) are not shown in the run command. Add the full pipeline:
```bash
python3 gen_wiki_relation_pairs.py --graph ... --struct-emb ... --out /tmp/mu_data/wiki_rel_pairs.tsv
python3 score_with_codex.py --pairs /tmp/mu_data/wiki_rel_pairs.tsv ... --out /tmp/mu_data/wiki_rel_scored.tsv
python3 convert_scored_to_graded.py --scored /tmp/mu_data/wiki_rel_scored.tsv --out /tmp/mu_data/wiki_rel_graded_pairs.tsv
# build e5 cache: python3 build_e5_cache.py --pairs /tmp/mu_data/wiki_rel_graded_pairs.tsv --out /tmp/mu_data/wiki_rel_e5.pt
UW_MU_GRAPH=.../100k_cats/category_parent.tsv python3 mu_posterior.py ...
```

**M. `README.md` pointer to playbook says "five pitfalls" (copied from the section header), propagating the off-by-one noted in the first review.** Update the README pointer when the playbook heading is fixed.

**N. `skills/uncertainty-estimation/SKILL.md` has its checklist hardcoded with 5 items** (matching the original count), but refers to the playbook for the full list. This is self-consistent as long as the skill says "see playbook for the full list" (which it does). No change needed, but note the count will drift if new pitfalls are added to the playbook.

---

### Positive Observations (Second Pass)

- The `NaN != NaN` idiom for missing-node detection (`dist("A","Z") != dist("A","Z")`) in `test_mu_posterior_dist.py` is neat and explicit — better than a `math.isnan` import for a one-liner test.
- The `--split` flag defaulting to `node-disjoint` is the correct conservative default — future users will get the stricter split by default without needing to know about the leakage risk.
- `eval_blend_prediction.py`'s loop over `judge in (None, "blend")` with the `num_embeddings` guard is a clean way to handle checkpoints that predate the blend row without crashing.
- The `blend_sweep.sh` reference in the REPORT (though the script isn't in this diff) suggests the sweep was driven by a reproducible shell script. If it's in `/tmp`, it should be committed or its contents included in an appendix of the REPORT so the sweep is actually reproducible.

s243a (commented):
## Third Review — PR #3496: Judge Superpositions, JointPosterior + 1/d, Graded-Wikipedia Round, Blend Judge

*This pass focuses on areas not covered by the first two reviews: the `JointPosterior` learning dynamics, the `judge_emb` architecture change, the training-loop integration of the blend judge, the `score_sym` dispatching contract, the `blend_sweep.sh` reproducibility gap, and several deeper design questions about the superposition framing itself.*

---

### I. JointPosterior: Learning Dynamics & Underfitting Risk

**A. With 60 held pairs and a 4-class LR head, the model has more parameters than samples in the node-disjoint split.**

A logistic regression on `K` sources predicting `C=4` classes has `K×C + C` parameters. With `K=6` sources (`sym`, `wiki_fwd`, `wiki_rev`, `elem_fwd`, `elem_rev`, `dist`) and `C=4` relations, that's 28 parameters fit on ~(`880 × 0.75`) ≈ 660 training rows — fine for training — but evaluated on 60 held rows. The AURC bootstrap CI width of ~0.15–0.20 on 60 samples makes the point estimate nearly uninterpretable for small effect sizes.

More critically: the 4-class formulation assumes `subcategory`, `subtopic`, `see_also`/`assoc`, and `element_of` are all present in the held set. The REPORT notes `see_also` had only **6 pairs** before splitting. After a 25% node-disjoint held split, `see_also` in the held set could be 0–2 pairs. A 0-row class in the held set means the LR weight for that row is evaluated against no ground-truth examples, so `log-loss` and `ECE` silently ignore that class. This should be called out explicitly and the class counts logged (the first review asked for this in a different framing; this is the *why it matters for the JointPosterior eval specifically* argument).

**B. `JointPosterior.fit` — is the LR regularisation strength documented or tunable?**

The code shows `JointPosterior` uses `sklearn.linear_model.LogisticRegression` (inferred from the `.net.weight` attribute in `mu_posterior.py`'s weight-printing block). Sklearn's `LogisticRegression` defaults to `C=1.0` (L2 regularisation). On a highly correlated 6-feature input (`corr(sym, e5)=+0.75`, `corr(dist, e5)=+0.50`), the default regularisation may be too weak (allowing the correlated features to share large but opposing weights that cancel, inflating ECE on held-out). The ablation compares `with 1/d` vs `without 1/d` — if the regularisation is not the same in both conditions (it is, since `C=1.0` is hardcoded), this is fine; but if someone adds `--hidden > 0` (MLP path), the regularisation dynamics differ completely. Document the `C` value and expose `--l2` as a tunable parameter for reproducibility.

---

### II. `mu_attention.py`: The `JUDGES` Dict Change and Backward Compatibility

**C. Adding `"blend": 7` to `JUDGES` is a **vocabulary expansion** that changes the checkpoint contract.**

The `judge_emb` embedding table in any checkpoint trained before this PR has `num_embeddings=7`. After this PR, new training runs will initialise an 8-row `judge_emb` table. The `eval_blend_prediction.py` guard (`num_embeddings <= JUDGES["blend"]`) handles inference correctly. But the *training* path is unguarded: if someone resumes training from a pre-PR checkpoint with a new recipe that includes `judge=blend` rows, the trainer will attempt to index row 7 of a 7-row table and crash with an out-of-bounds embedding index error — *at the first blend batch*, not at startup. This is a hard-to-diagnose failure mode.

Suggest adding a checkpoint migration note to `DESIGN_calibrated_judges.md` or the README, and a startup check in the training script:
```python
if max(JUDGES.values()) >= model.judge_emb.num_embeddings:
    raise ValueError(f"Checkpoint judge_emb has {model.judge_emb.num_embeddings} rows but JUDGES "
                     f"requires {max(JUDGES.values())+1}. Expand the embedding table before training "
                     f"on blend-tagged data (or use --expand-judge-emb).")
```

**D. The trailing comments on the `JUDGES` line are now very long (>200 chars) and the `"blend"` entry's comment references `emit_blend_judge.py` inline.**

This is a minor style note, but the `JUDGES` dict is imported by multiple modules and the inline comment is the only place `emit_blend_judge.py` is named as the canonical producer of blend-tagged data. Consider moving the extended comment to `DESIGN_calibrated_judges.md` and keeping the inline comment to one line per entry.

---

### III. Training Loop Integration of the Blend Judge

**E. The blend judge's `judge=blend` tag is consumed by `judge_emb[JUDGES["blend"]]` at training time — but the `judge_emb` row is initialised randomly and never pre-trained on anything.**

For all other judges (`haiku`, `gpt-5.5-low`, etc.), the `judge_emb` row calibrates against a body of real scored data. The `blend` row is initialised cold with whatever `nn.Embedding` default is used (typically `N(0, 1)` or `Uniform(-1,1)`). On the first gradient step over blend-tagged rows, this cold initialisation will produce a large random offset that is then corrected by backprop — but with only 880 blend rows (vs potentially thousands of real-judge rows), the `blend` row may never fully converge, and its residual random init may corrupt the shared trunk for the first few hundred steps.

The multi-seed confirmation (`B mean +0.841` consistent across 3 seeds) suggests the trunk *does* converge despite this, which is reassuring. But the mechanism is worth stating: the trunk converges because it sees all judge types simultaneously, and the `judge_emb` rows are effectively low-rank corrections. Documenting this explicitly would help someone adding a fourth judge type in future.

**F. `emit_blend_judge.py` always writes `conf=1.0` for every blend row.**
```python
f.write(f"{x}\t{y}\t{blend:.3f}\tSYM\tsee_also\tcategory\tcategory\t{a.corpus}\tblend\t1.0\n")
```
The playbook (pitfall #3) says confidence is a *gate*, not a per-item weight, and `conf=1.0` on all blend rows means the gate never fires for them. This is fine if the training loop uses `conf` only for selective abstention during *inference*, not during training. But if the training loop has a confidence-threshold filter (e.g. dropping rows with `conf < 0.5`), all blend rows pass unconditionally regardless of how uncertain the blend target actually is. The blend target has genuine uncertainty — a pair where `mu_e5_sym ≈ 0.5` and `mu_graph_sym ≈ 0.5` is genuinely ambiguous. Using `conf=1.0` tells the training loop it is a high-confidence example.

Suggest either: (a) computing a blend-specific confidence as `1 - 2·|blend - 0.5|` (distance from the decision boundary), or (b) documenting that `conf=1.0` is a deliberate placeholder and the blend rows are always included unconditionally.

---

### IV. `eval_blend_prediction.py`: The `score_sym` Contract

**G. `score_sym` builds a batch of either 3-tuples or 5-tuples depending on whether `judge` is `None`.**
```python
items = [((a, b, OPS["SYM"]) if judge is None else (a, b, OPS["SYM"], CORPORA[corpus], JUDGES[judge]))
         for a, b in pairs]
```
The 3-tuple path (agnostic / `judge=None`) produces the *provenance-agnostic* μ (masked `corpus`/`judge` inputs, marginalised over sources). The 5-tuple path activates a specific `judge_emb` row. This asymmetry is correct by design and documented in `mu_attention.py`'s comment: *"masking ⇒ provenance-AGNOSTIC μ"*. However, the function signature takes `judge=None` as the agnostic sentinel and `judge="blend"` for the explicit case — and the JUDGES dict lookup `JUDGES[judge]` will raise `KeyError` if `judge` is set to any string not in the dict (e.g. `"gpt5"` instead of `"gpt-5.5-low"`). A guard would help:
```python
if judge is not None and judge not in JUDGES:
    raise KeyError(f"Unknown judge '{judge}'. Valid: {list(JUDGES)}")
```

**H. The `512`-item batching in `score_sym` hardcodes a batch size with no relation to available GPU memory or the model's sequence length.**

This is not a correctness issue but a robustness one: on a GPU with less VRAM (e.g. the 8GB tier common in CI environments), 512 items × e5 embedding dimension × model hidden size may OOM silently if the model was recently widened. Expose `--batch` as a CLI argument or at least use a module-level constant.

---

### V. `REPORT_blend_judge_sweep.md`: Reproducibility Gap

**I. `/tmp/mu_data/blend_sweep.sh` is referenced but not committed.**

The REPORT ends with `Sweep: /tmp/mu_data/blend_sweep.sh`. `/tmp` is ephemeral. If this script is not preserved, the 3×3 multi-seed sweep (the PR's most important empirical result) is not reproducible from the committed artefacts alone. The sweep involves: generating pairs, scoring with `gpt-5.5-low`, building e5 cache, running three arms × three seeds, and calling `eval_blend_prediction.py` with `--save`. None of this is captured in committed files.

Either commit the sweep script to `prototypes/mu_cosine/` or inline its key commands in the REPORT's appendix. This is a hard requirement for the multi-seed result to be credible beyond the current session.

**J. The REPORT's multi-seed table does not state which checkpoint was used as the starting point for fine-tuning (`model_prod` at what training step/version?).**

The result `corr(SYM, T) = +0.841 mean` is relative to a specific `model_prod` baseline. If `model_prod` is updated between now and the next time someone runs this sweep, the baseline shifts and the delta `B−A = +0.144` may not replicate. Pin the exact `model_prod` SHA or step count in the REPORT.

---

### VI. Design Questions (No Code Change Required, but Worth Discussion)

**K. The superposition ratio `λ=0.5` is a fixed hyperparameter — but the optimal λ is task-dependent.**

The REPORT's ablation is entirely at `λ=0.5`. The `random-λ` arm tested generality across the full `[0,1]` range but destabilised training. There is no experiment with `λ ∈ {0.2, 0.3, 0.4, 0.6, 0.7, 0.8}` on a stable training regime. The choice of 0.5 may be sub-optimal: if `mu_e5_sym` is a much better signal than `mu_graph_sym` for this task, the optimal blend weight for teaching generality might be `λ ≈ 0.3` (mostly e5, a little graph to keep the structural anchor). This is flagged as a *next test* but not explicitly named in the playbook's "Right next test" callout. Suggest adding a `λ-sensitivity experiment` as a named TODO in the REPORT or DESIGN doc.

**L. The `T = (1−λ)·e5_ref + λ·graph_ref` evaluation target is evaluated at the *same* λ=0.5 that was used for training.**

This creates a partial circularity: the model was trained to predict `T(λ=0.5)`, and it is evaluated on `corr(SYM, T(λ=0.5))`. A model that perfectly memorised the training blend (rather than learning a general signal) would score perfectly on this metric. The multi-seed stability and the held-out node-disjoint design mitigate this, but the *cleanest* version of the test would evaluate at a λ *different* from the training λ (e.g. train at λ=0.5, eval at λ=0.3 and λ=0.7). If the model learned the underlying geometry, it should generalise across λ values. This is a stronger test of the "teaches generality" hypothesis.

**M. The trunk/head decomposition argument ("shared trunk learns invariant, per-judge rows absorb offset") is stated in the playbook but not empirically validated in this PR.**

The evidence is: (a) `B graph_ref +0.823` shows the blend-trained model recovers the graph half, (b) `A graph_ref +0.641` shows LLM-only training drifts from it. This is consistent with the trunk learning a shared representation and the judge row absorbing calibration offset. But it's also consistent with a simpler story: the blend training data just happens to include more graph-correlated examples, so the model's SYM head shifts toward graph predictions. The two stories make different predictions: in the trunk/head story, reading under `judge=agnostic` for model B should still outperform model A on `graph_ref` (since the trunk is shared). The table shows `B agnostic` is not reported — only `B blend` is shown. Add the `B agnostic` row to the multi-seed table to distinguish the two hypotheses.

---

### Positive Observations (Third Pass)

- The decision to evaluate `corr(SYM, graph_ref)` alongside `corr(SYM, T)` is excellent — it decomposes the result into the two halves of the target and makes the LLM-drift story visible directly in the numbers.
- The `--split node-disjoint` default being established *in this PR* (rather than as a post-hoc fix) means all future ablations will use the stricter split by default — a significant methodological upgrade.
- The `SKILL.md` in `skills/` with `user-invocable: true` and `allowed-tools` allowlist is a good pattern for keeping Claude Code sessions in scope for this module specifically.
- The scope-correction blockquote at the top of `REPORT_mu_posterior_dist.md` is a model for how to handle hypothesis refutation cleanly: it states what was tested, what the test is *not* a test of, and what survives.

s243a (commented):
## Follow-up Review — `f2ce6cb`, `3220e7a` — Review Response Audit

Thank you for the thorough response pass. I read both commits in full. This note covers: (1) acknowledgement of the two correctly disputed findings, (2) verification that the resolutions are complete and correct, and (3) new observations from the updated code.

---

### Corrections Accepted — The Review Was Wrong

**III.C (`JUDGES["blend"]=7` crashes training on old checkpoints):** Withdrawn. The `--init-from` path in the training script grows `judge_emb` from 7→8 before any blend batch is processed. The multi-seed runs confirmed this. The review's concern was based on an incorrect assumption that the trainer loads weights without expanding the embedding table. Good to have this on record.

**II.H (`E_mu` circular):** Withdrawn. `E_mu_fwd` is the LLM's graded expected-μ from the §14 scoring rubric — it is an *externally-provided* label, not a model forward pass. The circularity argument was incorrect.

---

### Resolution Verification

**Circularity (III.L/M) — CLEARED and substantially strengthened.** The multi-λ table is the right experiment:
- B beats A at λ∈{0.3, 0.5, 0.7} (mean +0.863/+0.841/+0.828 vs +0.724/+0.697/+0.682) → B learned the *components*, not the mixture ratio.
- B-agnostic +0.766 > A-blend +0.697 → the shared trunk carries the graph geometry, not just the judge_emb head.

Both tests together constitute a strong causal argument. The trunk/head decomposition (III.M) is now verified, not just stated. ✓

**Truncated normal λ sampling (second commit)** — correct and important. The `while lam < 0.0 or lam > 1.0: lam = rng.gauss(...)` resample loop is the right implementation. Two minor observations:

1. The expected number of resamples is negligible for the default `(mean=0.5, std=0.15)` — `P(outside [0,1]) ≈ 0.0008`, so this loop almost never iterates more than once. But for aggressive values like `(mean=0.1, std=0.2)` the loop could spin for a while. This is not a bug (the result is correct), but worth a comment noting the expected cost for typical defaults.
2. The first commit still shows `--lam-dist normal` with `clamped-normal` in the help string (`min(1.0, max(0.0, rng.gauss(...)))`). The second commit (`3220e7a`) renames this to `truncated` with the correct resample logic. Good — but the `--random` deprecated alias now maps to `uniform` via `dist = "uniform" if a.random else a.lam_dist`. After the rename, if someone had a script using `--lam-dist normal`, it will be rejected as an invalid choice. Since `normal` was introduced in the *first* commit of this same PR (not in a prior released version), there are no external users of `--lam-dist normal`, so this is fine — but a comment in the `--random` deprecation note documenting the rename would help anyone reading the git log.

**`none`-stratum ancestor cap 4→8** — correct fix, well motivated. One follow-up: the `see_also` stratum still uses the default `cap=4` in its `anc()` calls inside the sibling check (`if y not in anc(x) and x not in anc(y)`). This is intentional (checking whether siblings are also in an ancestor relationship, where 4 hops is sufficient for direct parent/grandparent), but it creates an asymmetry: `none` now uses cap=8 while `see_also` still uses cap=4. Document this asymmetry with a comment so it doesn't look like an oversight.

**`repro/blend_sweep.sh` and `repro/blend_multiseed.sh`** — the sweep scripts are now committed. However:
- Both scripts hardcode an absolute path: `cd /home/s243a/Projects/UnifyWeaver/prototypes/mu_cosine`. This will fail on any machine with a different checkout path. `cd "$(dirname "$0")/../.."` (or similar) would make the scripts portable. The `/tmp/mu_data/` paths for data would still need to be set by the user, but at least the repo path would resolve.
- `repro/README.md` is only 2 lines and doesn't document what inputs need to exist before the scripts are run (`wiki_rel_heldout.tsv`, `wiki_rel_heldout_e5.pt`, `struct_emb_recip.pt`, etc.). A short prerequisites section would complete the reproducibility story.

**Per-relation held-count logging** — confirmed added correctly in `mu_posterior.py`. The `Counter` import is already in scope (added in the same commit to `gen_wiki_relation_pairs.py`), but `mu_posterior.py` uses it without a `from collections import Counter` import visible in the diff. Verify that `Counter` is in scope at that point in `mu_posterior.py` (it may be imported earlier in the file outside this diff hunk). If not, a `from collections import Counter` is needed near the new `_hc` line.

---

### New Observations From Updated Code

**N. `eval_blend_prediction.py`: the `--ref` self-eval guard uses `os.path.abspath` comparison, which can be fooled by symlinks or relative vs absolute path variants pointing to the same file.**
```python
if os.path.abspath(path) == os.path.abspath(a.ref):
    print(f"# note: {name} is the --ref model ...")
```
This is the right approach and covers the common case. It will miss symlinks (e.g. `model_prod.pt → /tmp/mu_data/checkpoints/model_003.pt`). For a research script this is acceptable, but `os.path.realpath` (resolves symlinks) would be strictly stronger. Minor.

**O. The circularity-clearing table in `REPORT_blend_judge_sweep.md` reports mean corrs across 3 seeds but doesn't show per-seed breakdown for the multi-λ results.**

The main multi-seed table (seeds 1/2/3) shows per-seed values for λ=0.5. The new circularity section reports only **means** at λ∈{0.3,0.5,0.7}. These means could in principle be consistent even if one seed flipped (analogous to how the random-λ arm C had mean +0.553 driven by one seed collapsing to +0.09). The λ=0.5 column here matches the existing per-seed table (+0.863/+0.841/+0.828 mean... wait, those means don't match the per-seed +0.847/+0.832/+0.845 in the existing table — the new section says `B mean +0.863/+0.841/+0.828` across the three λ values, not per seed). To be explicit: were these multi-λ evals run once (on seed 1 checkpoints only) or averaged across all 3 seeds? The section says "mean" but doesn't state what it's a mean *over*. Clarify in the table header whether these are cross-seed means or single-seed results.

**P. `emit_blend_judge.py` first commit still has `elif dist == "normal":` with the clamped implementation; the second commit corrects it to `elif dist == "truncated":`.** This means `git blame` on the final file will show a confusing two-step where `normal` was introduced and immediately renamed. Not a correctness issue, but consider whether a single squashed commit or an amended first commit would have been cleaner for the history. Moot if you squash before merge.

---

### Status

All 16 inline comments have been addressed. The two disputed findings are correctly overturned. The substantive circularity critique (III.L/M) is now empirically cleared with a strong multi-λ + trunk/agnostic test. The remaining open items are:
- Higher-e5-variance held-out (deferred, tracked ✓)
- Truncated-λ training arm (deferred, tracked ✓)
- Minor nits above (N, O, P) — none are blockers

This PR is in good shape to merge.

PR #3520, opened 2026-07-07
Add fresh-corpus sampler for Σ(hop) confirmation

s243a:
Adds a pre-scoring sampler for the preregistered confirmatory `Σ(hop)` workflow.

This PR does not edit the frozen preregistration document and does not score any labels. It adds:

- `prototypes/mu_cosine/sample_sigma_hop_fresh_corpus.py`
  - loads a fresh/later Wikipedia category graph
  - removes every edge touching a node from the exploratory `100k_cats` graph
  - filters admin/maintenance categories
  - selects or validates a root slice
  - samples balanced exact-hop descendant/ancestor pairs over hops 1..5
  - writes score-input rows using the `transitive_h{hop}` labels expected by the confirmatory runner
  - writes a JSON manifest for auditability
  - enforces real-run guards: `hmax=5`, at least 250 pairs, and at least 300 descendants unless `--allow-small-sample` is explicitly used for toy/dry-run tests

- `prototypes/mu_cosine/test_sigma_hop_fresh_corpus.py`
  - synthetic checks for exploratory-node filtering
  - duplicate unordered-pair exclusion
  - balanced hop output and manifest contents
  - rejection of toy sample sizes without explicit opt-in

Checks run:

```text
python3 prototypes/mu_cosine/test_sigma_hop_fresh_corpus.py
python3 -m py_compile prototypes/mu_cosine/sample_sigma_hop_fresh_corpus.py prototypes/mu_cosine/test_sigma_hop_fresh_corpus.py
python3 prototypes/mu_cosine/sample_sigma_hop_fresh_corpus.py --help

### DIFF
--- a/prototypes/mu_cosine/sample_sigma_hop_fresh_corpus.py
+++ b/prototypes/mu_cosine/sample_sigma_hop_fresh_corpus.py
@@ -2,10 +2,12 @@
 """Prepare fresh-corpus pairs for the preregistered Sigma(hop) confirmation.
 
 This is the pre-scoring companion to `PREREG_sigma_hop_confirmatory.md`: it removes every node seen in the
-exploratory graph, selects a category slice, and samples balanced exact-hop descendant/ancestor pairs using the
-`transitive_h{hop}` label expected by `sigma_hop_confirmatory.py`.
+exploratory graph, selects a category slice, and samples balanced shortest-hop descendant/ancestor pairs using the
+`transitive_h{hop}` label expected by `sigma_hop_confirmatory.py`. Here "hop" means the minimum upward graph
+distance found by BFS inside the retained slice, not every possible path length through a DAG.
 
-The output is a score-input TSV, not a scored dataset. Running this script does not create labels.
+The output is a comment-header score-input TSV matching existing score_in readers, not a scored dataset. Running
+this script does not create labels.
 """
 
 from __future__ import annotations
@@ -16,6 +18,7 @@ import math
 import os
 import random
 import re
+import tempfile
 from collections import Counter, deque
 
 
@@ -26,20 +29,42 @@ ADMIN = re.compile(
 )
 
 
+PREREGISTERED_SEED = 0
+
+
 class FreshCorpusError(ValueError):
     pass
 
 
-def load_edges(path):
+def load_edges(path, stats=None):
     edges = []
+    stats = stats if stats is not None else {}
+    stats.setdefault("malformed_rows", 0)
+    stats.setdefault("wide_rows", 0)
     with open(path, encoding="utf-8") as f:
-        for i, line in enumerate(f):
+        first = f.readline()
+        if first:
+            cols = first.rstrip("\n").split("\t")
+            if not (len(cols) >= 2 and cols[0] == "child" and cols[1] == "parent"):
+                if len(cols) >= 2 and cols[0] and cols[1]:
+                    if len(cols) > 2:
+                        stats["wide_rows"] += 1
+                    edges.append((cols[0], cols[1]))
+                else:
+                    stats["malformed_rows"] += 1
+        for line in f:
             cols = line.rstrip("\n").split("\t")
-            if i == 0 and len(cols) >= 2 and cols[0] == "child" and cols[1] == "parent":
-                continue
             if len(cols) >= 2 and cols[0] and cols[1]:
+                if len(cols) > 2:
+                    stats["wide_rows"] += 1
                 edges.append((cols[0], cols[1]))
+            else:
+                stats["malformed_rows"] += 1
     return edges
+
+
+def sort_key(title):
+    return (title.casefold(), title)
 
 
 def node_block(edges):
@@ -60,15 +85,22 @@ def filter_candidate_edges(candidate_edges, exploratory_nodes):
     filtered = []
     removed_overlap = 0
     removed_admin = 0
+    blocked_roots = set()
     for child, parent in candidate_edges:
         if child in exploratory_nodes or parent in exploratory_nodes:
             removed_overlap += 1
+            blocked_roots.add(child)
             continue
         if ADMIN.search(child) or ADMIN.search(parent):
             removed_admin += 1
+            blocked_roots.add(child)
             continue
         filtered.append((child, parent))
-    return filtered, {"overlap_edges": removed_overlap, "admin_edges": removed_admin}
+    return filtered, {
+        "overlap_edges": removed_overlap,
+        "admin_edges": removed_admin,
+        "blocked_root_candidates": len(blocked_roots),
+    }, blocked_roots
 
 
 def descendants_within(children, roots, max_depth=None):
@@ -79,7 +111,7 @@ def descendants_within(children, roots, max_depth=None):
         node = queue.popleft()
         if max_depth is not None and depth[node] >= max_depth:
             continue
-        for child in sorted(children.get(node) or []):
+        for child in sorted(children.get(node, ())):
             if child not in kept:
                 kept.add(child)
                 depth[child] = depth[node] + 1
@@ -98,7 +130,7 @@ def ancestors_by_hop(parents, start, hmax):
         hop = seen[node]
         if hop >= hmax:
             continue
-        for parent in sorted(parents.get(node) or []):
+        for parent in sorted(parents.get(node, ())):
             if parent not in seen:
                 seen[parent] = hop + 1
                 by_hop.setdefault(hop + 1, []).append(parent)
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Review by s243a — COMMENTED** [review]
# Academic Code Review — PR #3520: Add fresh-corpus sampler for Σ(hop) confirmation

## Summary

This PR adds a well-structured pre-scoring pipeline for the preregistered confirmatory Σ(hop) workflow. The separation of concerns between the sampler and the confirmatory scorer is scientifically sound and the implementation is largely clean. The review below is divided into **correctness**, **scientific integrity / preregistration compliance**, **software-engineering quality**, and **minor/style** sections. All items are labelled **[BLOCKER]**, **[MAJOR]**, or **[MINOR]** to guide triage.

---

## 1 — Correctness

### [BLOCKER] `pair_pool_for_roots` may silently yield zero pairs when `slice_depth` is very small

In `pair_pool_for_roots`, `descendants_within` is called with `max_depth=slice_depth`, but `ancestors_by_hop` is then called with `hmax` (which can be larger than `slice_depth`). Because `slice_parents` is already restricted to `slice_nodes`, ancestors beyond `slice_depth` will simply not exist, so `pool[hop]` for `hop > slice_depth` will be empty. The downstream `first_eligible_roots` correctly tests `all(len(pool[h]) >= target_per_hop ...)` and will raise `FreshCorpusError`, but only *after* BFS over the entire root candidate list — which can be very slow on a full Wikipedia graph. Add an early check or at least a diagnostic message.

```python
# After pair_pool_for_roots, inside first_eligible_roots:
if slice_depth is not None and slice_depth < hmax:
    # Pairs at hop > slice_depth are structurally impossible; warn early.
    raise FreshCorpusError(
        f"--slice-depth {slice_depth} < --hmax {hmax}: no hop-{hmax} pairs can exist"
    )
```

### [BLOCKER] Final overlap check uses `node_block` over `(desc, anc)` but the set algebra is against `exploratory_nodes` built from the *filtered* graph

Line:
```python
overlap = node_block((desc, anc) for desc, anc, _ in rows) & exploratory_nodes
```
`exploratory_nodes` is built from `exploratory_edges` (the *unfiltered* exploratory graph) at the top of `main`, but `filtered_edges` (which already removed exploratory nodes) is what builds `parents`/`children`. The pair nodes therefore *cannot* appear in `exploratory_nodes` by construction — meaning this is a vacuous check that will always pass. It is included in `manifest["node_overlap_with_exploratory"] = 0` unconditionally, giving a false sense of audit coverage. The check should be placed **before** `write_score_in`, and it should be a `assert`-style hard error that fires during the BFS phase (at pair-pool build time) as a cross-validation, not only at the end.

### [MAJOR] `sample_balanced_pairs` trims to `total_pairs` *after* shuffle but doesn't re-balance

```python
if len(rows) > total_pairs:
    rows = rows[:total_pairs]
    counts = Counter(hop for _, _, hop in rows)
```
Because the rows were built hop-by-hop and then shuffled, the final trim will randomly discard tails. When `total_pairs` is not divisible by `hmax`, one or more hop classes will have `floor(total_pairs / hmax)` instead of `ceil(...)`. This is intentional truncation, but the manifest then reports `hop_counts` that do not match `per_hop = math.ceil(...)`, potentially confusing downstream reviewers who expect exact balance. Either guarantee exact divisibility at call time, or document the slight imbalance explicitly in the manifest via a `"balance_policy": "ceil_then_trim"` field.

### [MAJOR] `build_maps` initialises `parents[parent] = set()` but does not do the same for `children[child]`

```python
parents.setdefault(child, set()).add(parent)
parents.setdefault(parent, set())          # ← explicit empty sentinel
children.setdefault(parent, set()).add(child)
children.setdefault(child, set())          # ← explicit empty sentinel
```
The symmetry is correct here — both `parents` and `children` are initialised for every node encountered. This is fine. However, `pair_pool_for_roots` calls `children.get(node) or []` — the `or []` guard conflates an empty set with `None`. Since `build_maps` always inserts at least an empty set for every node it sees, `children.get(node)` can return `set()` (falsy), masking the empty-set case from a missing-node case. Use `children.get(node, ())` (or explicit `if node not in children`) to avoid confusion.

---

## 2 — Scientific Integrity / Preregistration Compliance

### [BLOCKER] Root-selection rule is effectively deterministic on the exploratory graph sort order

`first_eligible_roots` iterates `sorted(roots or children.keys())`. On a Wikipedia category graph this means the root is selected by Python's default string sort of category names, which is equivalent to lexicographic ASCII order. This is a *valid* pre-specified rule, but the manifest only records it as:

```json
"selection_rule": "first alphabetically eligible root after no-overlap/admin filtering"
```

This description omits the fact that the sort is *case-sensitive* (uppercase letters precede lowercase in ASCII). On Wikipedia category names (which use title case), this may silently produce a different order than the human-readable description implies. Recommend either (a) using `sorted(..., key=str.casefold)` and noting it in the manifest, or (b) explicitly documenting that ASCII sort is intentional and preregistered.

### [MAJOR] `--seed 0` is a sensible default, but the preregistration document should be checked to confirm 0 is the registered seed

The `argparse` default is `--seed 0`. If the preregistration document `PREREG_sigma_hop_confirmatory.md` specifies a different seed, using the CLI default without an explicit `--seed` flag would silently run a non-preregistered configuration. Add a compile-time constant `PREREGISTERED_SEED = 0` and assert `args.seed == PREREGISTERED_SEED` (unless `--allow-small-sample` is set), exactly analogously to the `hmax`/`pairs`/`min_descendants` guards already present.

### [MINOR] The manifest field `"node_overlap_with_exploratory": 0` is hardcoded, not computed

As noted in §1, the overlap check fires *after* `write_manifest` receives `0` — the actual computed value is never written into the manifest. If the overlap check were to ever fire, the manifest would already have been written with the wrong value. Move manifest construction to after the overlap check (or pass the computed overlap count in).

---

## 3 — Software Engineering Quality

### [MAJOR] `write_score_in` silently discards the comment line in subsequent reads

The TSV header is written as:
```python
f.write("# node_title\troot_title\tcur_relation\tconf\tneighborhood\tnode_type\troot_type\traw\n")
```
This is a comment, not a proper TSV header. `load_edges` (in this same file) uses column-index detection, not name-based detection. Downstream code that expects `pandas.read_csv(..., comment='#')` would silently drop this header. Either use a proper TSV header (no `#`) or document the `#`-comment convention in the docstring and confirm the confirmatory runner handles it.

### [MAJOR] `ancestors_by_hop` performs a BFS per descendant node — O(|slice_nodes| × |slice_edges|) overall

For large Wikipedia sub-graphs this nested BFS is the dominant cost. Since the goal is to enumerate all (descendant, ancestor, hop) triples, a single reverse-BFS from each *root* (downward closure already computed) would be more cache-friendly. This is a performance concern, not a correctness issue, but worth noting for the production run.

### [MINOR] `FreshCorpusError` inherits from `ValueError`

This is a reasonable choice for a user-facing CLI error. Consider inheriting from `SystemExit` or using `argparse.ArgumentParser.error` for CLI-detected configuration errors (seed, hmax, pairs guards) and reserving `FreshCorpusError(ValueError)` for data errors. This improves script integration in CI pipelines (exit code differentiation).

### [MINOR] `load_edges` is linear but re-reads the header detection on every line

The `if i == 0 and ...` check in the inner loop is correct but slightly misleading — consider separating the header skip into a `next(f, None)` call before the loop and removing the index `i` variable.

---

## 4 — Test Coverage

### [MAJOR] `test_cli_writes_balanced_score_in_and_manifest` uses `--hmax 5` but only checks hop counts 1 and 5, not 2–4

```python
assert hops.count("transitive_h1") == 2
assert hops.count("transitive_h5") == 2
```
With 10 pairs / 5 hops = 2 pairs each, all five hop classes should be asserted. The test would still pass even if hops 2–4 were missing from the output due to a bug. Add:
```python
for h in range(1, 6):
    assert hops.count(f"transitive_h{h}") == 2, f"hop {h} count wrong"
```

### [MINOR] No test for the `--exclude-root` / `--root` interaction when all roots are excluded

The `FreshCorpusError("all supplied --root values were excluded by --exclude-root")` branch is not covered by any test.

### [MINOR] No test for the `slice_depth < hmax` edge case described in §1

---

## 5 — Summary Table

| # | Severity | Location | Issue |
|---|----------|----------|-------|
| 1 | BLOCKER | `pair_pool_for_roots` / `first_eligible_roots` | `slice_depth < hmax` silently yields empty pools |
| 2 | BLOCKER | `main()` overlap check | Vacuous post-write check; doesn't guard manifest |
| 3 | BLOCKER | `first_eligible_roots` manifest | ASCII sort undocumented; may differ from intent |
| 4 | MAJOR | `sample_balanced_pairs` | Trim after shuffle breaks balance; manifest misleading |
| 5 | MAJOR | `build_maps` / callers | `or []` conflates empty set with missing key |
| 6 | MAJOR | `main()` | `--seed` not guarded like `--hmax`/`--pairs` |
| 7 | MAJOR | `write_score_in` | `#`-comment header not conventional TSV |
| 8 | MAJOR | `ancestors_by_hop` | O(N²) BFS; scalability concern for full Wikipedia |
| 9 | MAJOR | test file | Incomplete hop-balance assertions (hops 2–4 missing) |
| 10 | MINOR | manifest | `node_overlap_with_exploratory` hardcoded to 0 |
| 11 | MINOR | `FreshCorpusError` | Exit code not differentiated for CI |
| 12 | MINOR | `load_edges` | Inline header detection; use `next()` skip |
| 13 | MINOR | test file | Missing coverage: all-roots-excluded, slice_depth<hmax |

Overall the design is sound and the preregistration guardrails are a commendable practice. Resolving the three BLOCKERs — especially the vacuous overlap check and the undocumented sort rule — is important before the confirmatory run is executed, since those affect scientific reproducibility. The MAJOR items should ideally be addressed before merge.

s243a (commented):
Second review pass on PR #3520.

Overall this is a clean, purpose-built sampler with sensible preregistration guardrails and decent synthetic coverage. I took a fresh pass and found one likely correctness issue plus a few smaller robustness/testability points.

1. MAJOR — `sample_balanced_pairs` enforces global unordered-pair uniqueness across hops, which can make feasible datasets fail unnecessarily.

`seen_unordered` is shared across all hop buckets, so once `(x, y)` is emitted for one hop, the same unordered pair is forbidden for every other hop. In a DAG with multiple ancestor paths, the same descendant/ancestor pair can legitimately appear at different hop distances depending on path structure. If your scientific definition of `transitive_h{hop}` is “there exists a path of exact length h”, then cross-hop dedup by unordered pair throws away valid examples and can starve later hops even when the pool looked sufficient.

Concretely:
- eligibility is checked with `len(pool[h]) >= target_per_hop` independently per hop;
- sampling then applies a stronger cross-hop constraint that eligibility never accounted for;
- this can produce `hop k has only ... unique pairs` even though each hop pool individually satisfied the precheck.

If cross-hop exclusivity is intentional, I’d document it in the manifest and make the precheck simulate that constraint. Otherwise, dedupe should probably be per-hop only.

2. MINOR — `selection_rule` says “first alphabetically eligible root”, but the implementation is Python string sort over raw category titles.

That is deterministic, which is good, but for auditability I’d prefer wording that matches the executable rule more exactly, e.g. “lexicographically smallest eligible root under Python sorted()”. That avoids ambiguity around locale/case assumptions.

3. MINOR — `load_edges` accepts rows with more than two columns and silently ignores extras.

That is pragmatic, but if these TSVs are supposed to be strict `child<TAB>parent`, silently accepting malformed rows may hide upstream data issues. Worth considering an optional strict mode or at least counting malformed/wide rows in the manifest.

4. MINOR — tests do not currently cover the cross-hop uniqueness behavior.

Given how central that policy is, I’d add one test that proves the intended semantics:
- either same unordered pair may appear in different hops,
- or it must not, and failure is expected.

Right now the only uniqueness test covers `(a,b)` vs `(b,a)` inside the same effective pair space, but not the more interesting cross-hop case.

Net: I like the structure and the confirmatory safeguards. My main ask before merge is to decide whether unordered-pair uniqueness should be global across hops or local within each hop bucket, then encode that decision consistently in both the sampler and tests.

PR #3521, opened 2026-07-07
Add LMDB title-layer support for sigma-hop fresh sampling

s243a:
## Summary
- add a materializer for real category-title lookup sub-dbs inside Phase-1 category LMDBs
- teach the sigma-hop fresh-corpus sampler to read candidate graph structure from LMDB
- require a real title layer for LMDB score-input output, rejecting identity numeric `i2s`
- keep existing TSV sampler behavior for compatibility

## Tests
- `python3 prototypes/mu_cosine/test_sigma_hop_fresh_corpus.py`
- `python3 prototypes/mu_cosine/test_materialize_lmdb_title_layer.py`
- `python3 -m py_compile prototypes/mu_cosine/sample_sigma_hop_fresh_corpus.py prototypes/mu_cosine/materialize_lmdb_title_layer.py prototypes/mu_cosine/test_sigma_hop_fresh_corpus.py prototypes/mu_cosine/test_materialize_lmdb_title_layer.py`

### DIFF
--- a/prototypes/mu_cosine/lmdb_id.py
+++ b/prototypes/mu_cosine/lmdb_id.py
@@ -1,0 +1,27 @@
+#!/usr/bin/env python3
+"""Shared MediaWiki category LMDB id encoding helpers.
+
+MediaWiki page IDs are unsigned 32-bit integers. Existing fixtures currently sit below
+2**31, so this remains byte-compatible with earlier signed-int helpers while avoiding
+latent failures for larger page-id spaces.
+"""
+
+import struct
+
+ID32 = struct.Struct("<I")
+
+
+def enc_id(value):
+    return ID32.pack(int(value))
+
+
+def dec_id(value):
+    return ID32.unpack(bytes(value))[0]
+
+
+def looks_int(text):
+    try:
+        int(text)
+    except (TypeError, ValueError):
+        return False
+    return True
--- a/prototypes/mu_cosine/materialize_lmdb_title_layer.py
+++ b/prototypes/mu_cosine/materialize_lmdb_title_layer.py
@@ -1,0 +1,236 @@
+#!/usr/bin/env python3
+"""Materialize real category-title lookup sub-dbs inside a Phase-1 category LMDB.
+
+The graph sub-dbs (`category_parent`, `category_child`) stay unsigned int32 page/category IDs for fast traversal. This
+adds separate title lookup tables for human-readable consumers:
+
+  title_i2s  uint32_le page_id -> UTF-8 category title
+  title_s2i  UTF-8 category title -> uint32_le page_id
+
+It reads a gzip-compressed MediaWiki `page` SQL dump, keeps namespace 14 rows whose page_id already appears in the
+LMDB graph, and writes only the title layer. Existing graph data is not rewritten. `--map-size-gib` is LMDB's maximum
+map size; on sparse-file filesystems it does not immediately allocate that much disk, but Windows filesystems may be
+less forgiving, so increase it only as needed.
+"""
+
+from __future__ import annotations
+
+import argparse
+import gzip
+import json
+import os
+import tempfile
+from pathlib import Path
+
+from lmdb_id import dec_id, enc_id
+
+CATEGORY_NS = 14
+
+
+def parse_values(sql_values):
+    i, n = 0, len(sql_values)
+    while i < n:
+        if sql_values[i] != "(":
+            i += 1
+            continue
+        i += 1
+        fields, cur, in_str = [], [], False
+        while i < n:
+            ch = sql_values[i]
+            if in_str:
+                if ch == "\\":
+                    cur.append(sql_values[i + 1] if i + 1 < n else "")
+                    i += 2
+                    continue
+                if ch == "'":
+                    if i + 1 < n and sql_values[i + 1] == "'":
+                        cur.append("'")
+                        i += 2
+                        continue
+                    in_str = False
+                    i += 1
+                    continue
+                cur.append(ch)
+                i += 1
+            else:
+                if ch == "'":
+                    in_str = True
+                    i += 1
+                elif ch == ",":
+                    fields.append("".join(cur))
+                    cur = []
+                    i += 1
+                elif ch == ")":
+                    fields.append("".join(cur))
+                    i += 1
+                    break
+                else:
+                    cur.append(ch)
+                    i += 1
+        yield fields
+        while i < n and sql_values[i] != "(":
+            i += 1
+
+
+def iter_page_titles(page_dump):
+    try:
+        with gzip.open(page_dump, "rt", encoding="utf-8", errors="strict") as f:
+            for line in f:
+                if not line.startswith("INSERT INTO"):
+                    continue
+                pos = line.find("VALUES")
+                if pos < 0:
+                    continue
+                for row in parse_values(line[pos + 6:]):
+                    if len(row) < 3:
+                        continue
+                    try:
+                        page_id = int(row[0])
+                        ns = int(row[1])
+                    except ValueError:
+                        continue
+                    if ns == CATEGORY_NS and row[2]:
+                        yield page_id, row[2]
+    except gzip.BadGzipFile as exc:
+        raise ValueError(f"--page-dump must be a gzip-compressed MediaWiki page SQL dump: {page_dump}") from exc
+    except UnicodeDecodeError as exc:
+        raise ValueError(f"--page-dump is not valid UTF-8 at byte {exc.start}: {page_dump}") from exc
+
+
+def graph_node_ids(env, cp_db):
+    """Return every node appearing as either child key or any duplicate parent value."""
+    nodes = set()
+    with env.begin(buffers=True) as txn:
+        cur = txn.cursor(db=cp_db)
+        try:
+            if not cur.first():
+                return nodes
+            while True:
+                nodes.add(dec_id(cur.key()))
+                nodes.add(dec_id(cur.value()))
+                for value in cur.iternext_dup(keys=False, values=True):
+                    nodes.add(dec_id(value))
+                if not cur.next_nodup():
+                    break
+        finally:
+            close = getattr(cur, "close", None)
+            if close is not None:
+                close()
+    return nodes
+
+
+def write_manifest(path, manifest):
+    out_dir = path.parent
+    out_dir.mkdir(parents=True, exist_ok=True)
+    fd, tmp = tempfile.mkstemp(prefix=path.name + ".", suffix=".tmp", dir=str(out_dir))
+    try:
+        with os.fdopen(fd, "w", encoding="utf-8") as f:
+            json.dump(manifest, f, indent=2, sort_keys=True)
+            f.write("\n")
+        os.replace(tmp, path)
+    finally:
+        if os.path.exists(tmp):
+            os.unlink(tmp)
+
+
+def abort_quietly(txn):
+    if txn is None:
+        return
+    try:
+        txn.abort()
+    except Exception:
+        pass
+
+
+def main():
+    ap = argparse.ArgumentParser(description=__doc__)
+    ap.add_argument("--lmdb-dir", required=True, type=Path)
+    ap.add_argument("--page-dump", required=True, type=Path)
+    ap.add_argument("--title-i2s-db", default="title_i2s")
+    ap.add_argument("--title-s2i-db", default="title_s2i")
+    ap.add_argument("--manifest", type=Path)
+    ap.add_argument("--map-size-gib", type=float, default=16.0)
+    args = ap.parse_args()
+
+    try:
+        import lmdb
+    except ImportError as exc:
+        raise SystemExit("python-lmdb is required") from exc
+
+    env = lmdb.open(str(args.lmdb_dir), max_dbs=32, map_size=int(args.map_size_gib * 1024**3), subdir=True)
+    cp = env.open_db(b"category_parent", create=False)
+    title_i2s = env.open_db(args.title_i2s_db.encode("utf-8"), create=True)
+    title_s2i = env.open_db(args.title_s2i_db.encode("utf-8"), create=True)
+    meta = env.open_db(b"meta", create=True)
+
+    with env.begin(write=True) as txn:
+        txn.drop(title_i2s, delete=False)
+        txn.drop(title_s2i, delete=False)
+
+    nodes = graph_node_ids(env, cp)
+    matched = 0
+    category_title_rows_scanned = 0
+    collisions = []
+    txn = None
+    try:
+        txn = env.begin(write=True)
+        for page_id, title in iter_page_titles(args.page_dump):
+            category_title_rows_scanned += 1
+            if page_id not in nodes:
+                continue
+            title_bytes = title.encode("utf-8")
+            page_bytes = enc_id(page_id)
+            existing = txn.get(title_bytes, db=title_s2i)
+            if existing is not None and bytes(existing) != page_bytes:
+                collisions.append((title, dec_id(existing), page_id))
+                if len(collisions) >= 10:
+                    break
+                continue
+            txn.put(page_bytes, title_bytes, db=title_i2s)
+            txn.put(title_bytes, page_bytes, db=title_s2i)
+            matched += 1
+            if matched % 500_000 == 0:
+                txn.commit()
+                txn = None
+                print(f"materialized {matched} category titles")
+                txn = env.begin(write=True)
+        if collisions:
+            examples = "; ".join(f"{title}: {old_id} vs {new_id}" for title, old_id, new_id in collisions[:5])
+            raise ValueError(f"duplicate category titles with different page IDs in page dump: {examples}")
+        txn.put(b"title_layer_kind", b"mediawiki_page_titles", db=meta)
+        txn.put(b"title_i2s_db", args.title_i2s_db.encode("utf-8"), db=meta)
+        txn.put(b"title_s2i_db", args.title_s2i_db.encode("utf-8"), db=meta)
+        txn.put(b"title_layer_count", str(matched).encode("ascii"), db=meta)
+        txn.put(b"title_layer_refreshed", b"1", db=meta)
+        txn.commit()
+        txn = None
+    except lmdb.MapFullError as exc:
+        abort_quietly(txn)
+        txn = None
+        raise SystemExit("LMDB map is full while writing title tables; rerun with a larger --map-size-gib") from exc
+    except BaseException:
+        abort_quietly(txn)
+        txn = None
+        raise
+    finally:
+        env.sync()
+        env.close()
+
+    manifest = {
+        "lmdb_dir": str(args.lmdb_dir),
+        "page_dump": str(args.page_dump),
+        "graph_node_count": len(nodes),
+        "category_title_rows_scanned": category_title_rows_scanned,
+        "title_layer_count": matched,
+        "title_i2s_db": args.title_i2s_db,
+        "title_s2i_db": args.title_s2i_db,
+        "refreshed": True,
+    }
+    manifest_path = args.manifest or (args.lmdb_dir.parent / "title_layer_manifest.json")
+    write_manifest(manifest_path, manifest)
+    print(f"materialized {matched} real category titles into {args.lmdb_dir}")
+    print(f"manifest -> {manifest_path}")
+
+
+if __name__ == "__main__":
+    main()
--- a/prototypes/mu_cosine/sample_sigma_hop_fresh_corpus.py
+++ b/prototypes/mu_cosine/sample_sigma_hop_fresh_corpus.py
@@ -6,20 +6,22 @@ exploratory graph, selects a category slice, and samples balanced shortest-hop d
 `transitive_h{hop}` label expected by `sigma_hop_confirmatory.py`. Here "hop" means the minimum upward graph
 distance found by BFS inside the retained slice, not every possible path length through a DAG.
 
-The output is a comment-header score-input TSV matching existing score_in readers, not a scored dataset. Running
-this script does not create labels.
+Candidate graph structure can come from a child<TAB>parent TSV or from the repository's Phase-1 category LMDB layout.
+The LMDB path keeps graph traversal on unsigned int32 IDs and uses a separate real-title layer only at the score-input boundary.
+Running this script does not create labels.
 """
 
 from __future__ import annotations
 
 import argparse
 import json
-import math
 import os
 import random
 import re
 import tempfile
 from collections import Counter, deque
+
+from lmdb_id import dec_id, enc_id, looks_int
 
 
 ADMIN = re.compile(
@@ -34,6 +36,10 @@ PREREGISTERED_SEED = 0
 
 class FreshCorpusError(ValueError):
     pass
+
+
+def is_identity_numeric_title(node_id, title):
+    return looks_int(title) and int(title) == int(node_id)
 
 
 def load_edges(path, stats=None):
@@ -111,7 +117,7 @@ def descendants_within(children, roots, max_depth=None):
         node = queue.popleft()
         if max_depth is not None and depth[node] >= max_depth:
             continue
-        for child in sorted(children.get(node, ())):
+        for child in sorted(children.get(node, ()), key=sort_key):
             if child not in kept:
                 kept.add(child)
                 depth[child] = depth[node] + 1
@@ -130,7 +136,7 @@ def ancestors_by_hop(parents, start, hmax):
         hop = seen[node]
         if hop >= hmax:
             continue
-        for parent in sorted(parents.get(node, ())):
+        for parent in sorted(parents.get(node, ()), key=sort_key):
             if parent not in seen:
                 seen[parent] = hop + 1
                 by_hop.setdefault(hop + 1, []).append(parent)
@@ -138,22 +144,31 @@ def ancestors_by_hop(parents, start, hmax):
     return by_hop
 
 
-def pair_pool_for_roots(parents, children, roots, hmax, slice_depth=None):
-    slice_nodes = descendants_within(children, roots, max_depth=slice_depth)
+def pair_pool_for_slice_nodes(parents, slice_nodes, hmax):
     slice_parents = restrict_parents(parents, slice_nodes)
     pool = {hop: [] for hop in range(1, hmax + 1)}
-    for desc in sorted(slice_nodes):
+    for desc in sorted(slice_nodes, key=sort_key):
         by_hop = ancestors_by_hop(slice_parents, desc, hmax)
         for hop in range(1, hmax + 1):
             for anc in by_hop.get(hop, []):
                 if desc != anc:
                     pool[hop].append((desc, anc))
-    return slice_nodes, pool
+    return pool
+
+
+def pair_pool_for_roots(parents, children, roots, hmax, slice_depth=None):
+    slice_nodes = descendants_within(children, roots, max_depth=slice_depth)
+    return slice_nodes, pair_pool_for_slice_nodes(parents, slice_nodes, hmax)
 
 
 def hop_targets(total_pairs, hmax):
+    # Remainders are assigned to lower hops deterministically; the manifest records this frozen policy.
     base, rem = divmod(total_pairs, hmax)
     return {hop: base + (1 if hop <= rem else 0) for hop in range(1, hmax + 1)}
+
+
+def string_key_counts(counts):
+    return {str(key): counts[key] for key in sorted(counts)}
 
 
 def can_satisfy_targets(pool, targets):
@@ -175,17 +190,29 @@ def first_eligible_roots(parents, children, roots, excluded_roots, blocked_roots
     if slice_depth is not None and slice_depth < hmax:
         raise FreshCorpusError(f"--slice-depth {slice_depth} < --hmax {hmax}: hop-{hmax} pairs are impossible")
     candidates = sorted(roots or children.keys(), key=sort_key)
+    attempts = []
     for root in candidates:
         if root in excluded_roots or root in blocked_roots:
             continue
         if root not in children or ADMIN.search(root):
             continue
-        slice_nodes, pool = pair_pool_for_roots(parents, children, (root,), hmax, slice_depth=slice_depth)
+        slice_nodes = descendants_within(children, (root,), max_depth=slice_depth)
         if len(slice_nodes) < min_descendants:
+            attempts.append({
+                "root": root,
+                "slice_nodes": len(slice_nodes),
+                "available_hop_counts": string_key_counts({hop: 0 for hop in targets}),
+            })
             continue
-        ok, _ = can_satisfy_targets(pool, targets)
+        pool = pair_pool_for_slice_nodes(parents, slice_nodes, hmax)
+        ok, available = can_satisfy_targets(pool, targets)
+        attempts.append({
+            "root": root,
+            "slice_nodes": len(slice_nodes),
+            "available_hop_counts": string_key_counts(available),
+        })
         if ok:
-            return (root,), slice_nodes, pool
+            return (root,), slice_nodes, pool, {"candidate_root_attempts": attempts}
     raise FreshCorpusError("no eligible root slice supplied enough no-overlap pairs for every hop")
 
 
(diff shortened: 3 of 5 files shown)


s243a (commented):
## Academic Code Review — PR #3521: Add LMDB title-layer support for sigma-hop fresh sampling

### Overview

This PR adds two complementary pieces: a standalone materializer (`materialize_lmdb_title_layer.py`) that writes real `title_i2s`/`title_s2i` sub-dbs into an existing Phase-1 category LMDB, and an extension to `sample_sigma_hop_fresh_corpus.py` that can source its candidate graph directly from LMDB instead of a flat TSV. Both changes are well-scoped and the test coverage is meaningful. The review below is structured by file, followed by cross-cutting observations.

---

### `materialize_lmdb_title_layer.py` (new file)

**Strengths**
- The hand-rolled SQL-values parser (`parse_values`) is a reasonable choice given MediaWiki dump format quirks; covering `\\`-escape and `''`-quote-doubling is correct.
- Batching `txn.commit()` every 500 000 inserts avoids unbounded memory growth for large dumps — good defensive design.
- Atomic `os.replace` for the manifest file prevents partial-write corruption.
- Sub-db metadata keys (`title_layer_kind`, `title_layer_count`, `title_layer_refreshed`) give downstream consumers machine-readable provenance without requiring the manifest file.

**Issues / Suggestions**

1. **Transaction leak on commit-batch path** (medium severity)
   ```python
   txn = env.begin(write=True)
   try:
       for page_id, title in iter_page_titles(...):
           ...
           if matched % 500_000 == 0:
               txn.commit()
               txn = env.begin(write=True)  # new txn, old except-handler no longer covers it if next commit raises
   except Exception:
       txn.abort()
   ```
   If `txn.commit()` inside the loop succeeds but the *subsequent* `env.begin()` raises (e.g. map full), the `except` block calls `abort()` on the *new* half-open txn, which is correct — but if a second exception occurs *during* `txn.abort()` in the `finally` block, the `env.sync()` may still execute. Consider wrapping the batch-commit + re-open in its own try/except, or using a helper that manages this rotation.

2. **`graph_node_ids` reads both key and value of `category_parent`** (minor)
   ```python
   for key, value in txn.cursor(db=cp_db):
       nodes.add(dec_i32(key))
       nodes.add(dec_i32(value))
   ```
   When `category_parent` is opened with `dupsort=True` (as in the test fixtures), `cursor()` iterates `(key, first_dup_value)` pairs, not all duplicates. This means parents with multiple children will only contribute one parent-id from the value column. For the purpose of building the node-id set this is likely fine (the parent key appears as a key elsewhere), but it is worth an explicit comment clarifying why a single-pass over keys suffices, or use `txn.cursor(db=cp_db).iternext_dup` if full coverage is intended.

3. **`--map-size-gib` default of 16 GiB** (documentation gap)
   On Linux this is a virtual-address reservation and is typically harmless, but on Windows LMDB maps translate to a pagefile reservation, which can surprise users. A `--help` note or warning when the platform is Windows would improve UX.

4. **`write_manifest` finalizer always unlinks `tmp` even on success** (minor correctness)
   ```python
   finally:
       if os.path.exists(tmp):
           os.unlink(tmp)
   ```
   After `os.replace(tmp, path)` succeeds, `tmp` no longer exists, so `os.path.exists(tmp)` returns `False` and the unlink is skipped — this is actually correct. But if `os.replace` raises (e.g. cross-device), the `tmp` file is cleaned up silently. This is acceptable behavior; a brief comment would make the intent explicit.

---

### `sample_sigma_hop_fresh_corpus.py` (modified)

**Strengths**
- `LmdbTitleGraph` cleanly encapsulates all LMDB I/O behind a title-centric interface; the rest of the sampling logic works on string titles, preserving the existing algorithmic invariants.
- The identity-numeric title guard (`is_identity_numeric_title`) is an excellent defensive check — it prevents silently producing score-input files with raw integers as category labels.
- Splitting `main()` into `validate_args`, `sample_from_tsv`, and `sample_from_lmdb` meaningfully improves testability.
- `string_key_counts` normalizing `Counter` keys to strings for the manifest is good: JSON keys must be strings and downstream tools that parse the manifest will not be surprised by integer keys.
- Returning `extra` dicts from `first_eligible_roots` / `first_eligible_lmdb_roots` for the manifest (including `candidate_root_attempts`) provides excellent diagnostic provenance for reproducibility.

**Issues / Suggestions**

5. **`LmdbTitleGraph` holds a long-lived read transaction** (potential correctness / resource concern)
   ```python
   self.txn = self.env.begin(buffers=True)
   ```
   A long-lived read transaction prevents LMDB from reclaiming free pages, which can cause the database to grow unboundedly on write-heavy concurrent workflows. For a read-only (`lock=False`) open this is less critical, but if the LMDB is also being written concurrently (e.g. during incremental materialization), stale readers can block page reuse. Consider using `with self.env.begin(buffers=True) as txn:` within each method call, or at minimum document the lifetime assumption.

6. **`iter_dups` returns `[]` for a missing key but calls `txn.cursor(db=db)` without a context manager** (minor resource)
   ```python
   def iter_dups(txn, db, key):
       cursor = txn.cursor(db=db)
       if not cursor.set_key(enc_i32(key)):
           return []
       return [dec_i32(value) for value in cursor.iternext_dup(keys=False, values=True)]
   ```
   The cursor is not explicitly closed. In CPython this is reclaimed by the GC promptly, but in PyPy or under heavy load the cursor count can grow. Since `lmdb.Cursor` supports the context manager protocol, wrapping in `with txn.cursor(db=db) as cursor:` is safer.

7. **`load_lmdb_slice_maps` calls `pair_pool_for_roots` with `slice_depth=None` after already honoring `slice_depth` in BFS** (potential double-counting)
   ```python
   _slice_nodes, pool = pair_pool_for_roots(parents, children, (root_title,), hmax, slice_depth=None)
   ```
   `slice_nodes` was already depth-limited by the BFS above. Passing `slice_depth=None` to `pair_pool_for_roots` is correct because the slice has already been materialized into the `parents`/`children` dicts — but the argument name `_slice_nodes` (discarding the return) silently drops the TSV-path's slice-node set. A comment explaining why `slice_depth=None` is safe here would prevent future confusion.

8. **`validate_args` does not check mutual exclusion of `--candidate-graph` and LMDB-specific args** (robustness)
   `--title-i2s-db` and `--title-s2i-db` are silently ignored when `--candidate-graph` is used. `argparse` already enforces the `required=True` mutex group, so these flags reaching the TSV path is technically impossible — but an assertion or early warning in `validate_args` would guard against future refactors that weaken the mutex.

9. **`scope_root` handling asymmetry between TSV and LMDB paths** (minor consistency)
   In `sample_from_tsv`, a missing `--scope-root` entry in the already-filtered `children` dict raises `FreshCorpusError`. In `lmdb_candidate_root_ids`, a missing `scope_root` falls back to LMDB `meta.scoped_root`. These two behaviors diverge silently; documenting the intentional difference in `validate_args` or a docstring would help future maintainers.

10. **Manifest `edge_file_stats` merge** (correctness)
    ```python
    if source_manifest.get("edge_file_stats"):
        merged_stats = dict(source_manifest["edge_file_stats"])
        merged_stats["exploratory"] = exploratory_stats
        manifest["edge_file_stats"] = merged_stats
    ```
    The LMDB path sets `"edge_file_stats": {"candidate": {}}`. Since `{"candidate": {}}` is truthy, the merge executes and correctly injects `exploratory_stats`. However, if a future path sets `"edge_file_stats": {}` (empty dict), the merge is skipped and the manifest loses `exploratory_stats` silently. Prefer an unconditional merge or check `source_manifest.get("edge_file_stats") is not None`.

---

### `test_materialize_lmdb_title_layer.py` (new file)

**Strengths**
- Tests both the happy path (two real ns-14 nodes out of three, one outside graph excluded, one non-category excluded, stale key cleared) and verifies the manifest JSON.
- The stale-key refresh check (`enc_i32(99)` / `b"Stale"` both absent after run) is exactly the correctness invariant that matters.

**Issues**

11. **Only one test function** (coverage gap)
    The parser (`parse_values` / `iter_page_titles`) handles several edge cases (escaped quotes, doubled single-quotes, multi-statement lines) that are not exercised. Dedicated unit tests for `parse_values` would be valuable for a hand-rolled parser.

12. **`graph_node_ids` called with `cp` opened as `dupsort=True`** (test / production mismatch)
    The test creates `category_parent` with `dupsort=True`, but `materialize_lmdb_title_layer.py` opens it with `create=False` (inheriting the existing flags). This is consistent *if* the production LMDB always uses `dupsort=True`. It is worth asserting or documenting this assumption in `graph_node_ids`.

---

### `test_sigma_hop_fresh_corpus.py` (modified)

**Strengths**
- `test_cli_writes_from_lmdb_real_title_layer` exercises the full CLI round-trip: LMDB construction → `main()` invocation → TSV output validation → manifest inspection. The assertion `all(not row[0].isdigit() and not row[1].isdigit() for row in rows)` directly validates that no numeric IDs leaked into the output.
- `test_cli_rejects_lmdb_identity_numeric_title_layer` is a well-targeted negative test for the identity-numeric guard.
- Graceful `skip` on missing `python-lmdb` is correct practice for an optional dependency.

**Issues**

13. **`_lmdb_chain_fixture` builds a linear chain, not a tree** (coverage gap)
    Each chain has a single root → single path of depth 5, producing exactly one pair per hop. This means `--pairs 10` with `--hmax 5` results in exactly 2 pairs/hop, which tests the happy path but does not exercise the pair-pool balancing logic under pool exhaustion or uneven hop distributions.

14. **`_build_lmdb` duplicates `enc_i32` from the production module** (DRY)
    Both `test_sigma_hop_fresh_corpus.py` and `test_materialize_lmdb_title_layer.py` define their own `enc_i32`. Consider importing it from `materialize_lmdb_title_layer` or placing it in a shared test utility, since both modules now define it identically.

---

### Cross-Cutting Observations

15. **`enc_i32` / `dec_i32` defined in three places** — production `materialize_lmdb_title_layer.py`, production `sample_sigma_hop_fresh_corpus.py`, and both test files. A single `_lmdb_utils.py` (or inline import from the materializer) would eliminate drift risk.

16. **No `requirements.txt` or `pyproject.toml` entry for `lmdb`** — the `python-lmdb` import is optional and guarded with a clear `FreshCorpusError`, which is good. But there is no documented way to install the optional dependency. Adding an `extras_require` or a comment in a `requirements-lmdb.txt` would help CI and new contributors.

17. **Memory scaling** — `LmdbTitleGraph._title_cache` and `_id_cache` are unbounded dicts. For a full Wikipedia category graph (~1.2 M nodes) this could grow to several hundred MiB. A note on expected cardinality, or a `maxsize` LRU cache, would be prudent for production use.

18. **`sort_key` referenced but not visible in diff** — `sort_key`, `ADMIN`, `node_block`, `restrict_parents`, `build_maps`, `filter_candidate_edges`, and `main` (imported in tests) are all used in the modified/new code but their definitions are not shown in this diff. The review assumes they are correct from the prior PR; any semantic changes to `sort_key` (e.g. handling `int` vs `str` node IDs) that affect LMDB paths should be explicitly verified.

---

### Summary

| Severity | Count | Items |
|----------|-------|-------|
| Medium | 1 | #1 (txn leak on batch-commit path) |
| Minor correctness | 3 | #2, #7, #10 |
| Resource / robustness | 3 | #5, #6, #8 |
| Documentation | 3 | #3, #4, #9 |
| Test coverage | 3 | #11, #13, #14 |
| Cross-cutting | 4 | #15, #16, #17, #18 |

The core logic is sound and the LMDB integration is clean. Addressing items **#1** (transaction rotation safety) and **#5** (long-lived read transaction documentation) before merge is recommended. The remaining items are good-to-fix but non-blocking.

s243a (commented):
## Second Academic Code Review — PR #3521: Add LMDB title-layer support for sigma-hop fresh sampling

This is a second pass focusing on algorithmic correctness, data-integrity edge cases, API contract subtleties, and reproducibility properties that were not the primary focus of the first review.

---

### A. Algorithmic / Semantic Correctness

**A-1. `first_eligible_roots` now computes `can_satisfy_targets` before the `min_descendants` gate (ordering regression)**

Before this PR, the loop short-circuited on `min_descendants` *before* computing the full pool:
```python
# old order (implicit, before this diff)
if len(slice_nodes) < min_descendants:
    continue
ok, _ = can_satisfy_targets(pool, targets)
```
After this PR the order is:
```python
slice_nodes, pool = pair_pool_for_roots(...)
ok, available = can_satisfy_targets(pool, targets)   # expensive — runs even for tiny slices
attempts.append({...})
if len(slice_nodes) < min_descendants:
    continue
if ok:
    return ...
```
The `can_satisfy_targets` call and the `attempts.append` now execute for *every* candidate root, including slices that are immediately discarded by `min_descendants`. For large candidate sets with many small slices this can materially slow down root selection. The fix is to move `attempts.append` (with a `"skipped_min_descendants": True` flag) *after* the `min_descendants` guard, and only call `can_satisfy_targets` on slices that pass it. The diagnostic value of recording attempt metadata is preserved.

**A-2. `load_lmdb_slice_maps` two-phase build creates `parents`/`children` dicts that are sets, but `pair_pool_for_roots` (called by `first_eligible_lmdb_roots`) expects sequences or iterables from `parents.get(node, ())`**

```python
parents, children = {}, {}
for node_id in kept_ids:
    title = title_for(node_id)
    parents.setdefault(title, set())
    children.setdefault(title, set())
```
`pair_pool_for_roots` is the TSV-path function and likely expects frozensets or sets — this is probably fine since `sorted()` works on sets. But it is worth confirming: if any downstream code in `ancestors_by_hop` or `descendants_within` assumes `list`, a `set` passed via `children.get(node, ())` will work for iteration but not for indexing, which would be a silent semantic difference from the TSV path.

**A-3. `lmdb_candidate_root_ids` returns raw child IDs from `graph.children(scope_id)` as the root candidate list**

```python
children = graph.children(scope_id)
...
return children, "casefold-lexicographically...", {...}
```
`graph.children` returns a list of `int` node IDs (from `iter_dups`). These are then passed to `first_eligible_lmdb_roots` which calls `graph.title(nid)` on each. This is correct. However, the `selection_rule` string says *"casefold-lexicographically smallest eligible direct child"* but the actual ordering in `first_eligible_lmdb_roots` sorts by `sort_key(graph.title(nid))`. If `sort_key` is casefold-based, the label matches; if it is something else (e.g. Unicode normalization), the description in the manifest would be misleading. This is a documentation-accuracy concern, not a runtime bug.

**A-4. `targets` is computed twice in `main()` — dead computation**

```python
# inside sample_from_lmdb:
targets = hop_targets(args.pairs, args.hmax)  # first computation

# later in main():
targets = hop_targets(args.pairs, args.hmax)  # second computation, unused
```
The second `targets` assignment in `main()` (after `write_score_in`) is never read again. It was probably a copy-paste artifact from the pre-refactor `main`. While harmless, it creates confusion about whether `targets` has a meaningful role in the manifest assembly block.

---

### B. Data Integrity and LMDB-Specific Concerns

**B-1. `materialize_lmdb_title_layer.py` opens `category_parent` with `create=False` but does not specify `dupsort`**

```python
cp = env.open_db(b"category_parent", create=False)
```
With `create=False`, LMDB inherits the sub-db flags that were set when the db was *created*. If it was created with `dupsort=True` (as all test fixtures and production Phase-1 code do), a cursor iteration here yields `(key, first_dup_value)` pairs — meaning `graph_node_ids` misses parent IDs that only appear as non-first duplicate values. For a `category_parent` db where a node can have many parents, this means the `nodes` set is **under-populated** and some valid category IDs will be excluded from the title layer. For correctness, use a dup-aware scan:
```python
with env.begin(buffers=True) as txn:
    cursor = txn.cursor(db=cp_db)
    if cursor.first():
        while True:
            nodes.add(dec_i32(cursor.key()))
            for val in cursor.iternext_dup(keys=False):
                nodes.add(dec_i32(val))
            if not cursor.next_nodup():
                break
```
This is distinct from the first review's issue #2, which noted the ambiguity — here the consequence is concrete: real category nodes reachable only as non-first parents will be silently omitted from the title layer.

**B-2. `write_manifest` in `materialize_lmdb_title_layer.py` uses `os.fdopen(fd, ...)` but the `finally` block's `os.path.exists(tmp)` check races on NFS/SMB mounts**

On network filesystems, `os.replace` may succeed from the local OS perspective while the file is still visible under the old name from other processes. The `os.path.exists(tmp)` check is not atomic with the unlink. This is an edge case only relevant in distributed build environments, but for a pipeline that may run on NFS-mounted storage (common in ML research clusters), a `try/except FileNotFoundError` around `os.unlink(tmp)` is more robust than the existence check.

**B-3. `LmdbTitleGraph` opens the env with `lock=False` but does not document the single-writer assumption**

```python
self.env = lmdb.open(str(lmdb_dir), readonly=True, lock=False, max_dbs=32, subdir=True)
```
`lock=False` disables the lock file, which is safe only if no writer is active. In a pipeline where `materialize_lmdb_title_layer.py` and `sample_sigma_hop_fresh_corpus.py` might be run concurrently (e.g. by a Makefile with `-j`), a writer and a `lock=False` reader can produce inconsistent reads without any error. A docstring or `README` note stating *"do not run materializer and sampler concurrently against the same LMDB"* would prevent silent data corruption.

**B-4. `map_size=int(args.map_size_gib * 1024**3)` — floating-point precision edge case**

```python
map_size=int(args.map_size_gib * 1024**3)
```
`1024**3 = 1073741824`. For `--map-size-gib 0.1` (used in tests), `0.1 * 1073741824 = 107374182.4`, so `int(...)` truncates to `107374182` bytes ≈ 99.9999... MiB. This is fine for the test. For production values like `--map-size-gib 16.0`, there is no precision issue. However, `--map-size-gib 0.0625` (64 MiB) gives `int(0.0625 * 1073741824) = 67108864` exactly, which is fine. A `round()` instead of `int()` would make intent clearer and handle sub-GiB values uniformly.

---

### C. Reproducibility / Preregistration Integrity

**C-1. `validate_args` does not enforce `--hmax 5` when `--candidate-lmdb` is used with `--allow-small-sample`**

```python
if args.hmax != 5:
    raise FreshCorpusError("the preregistered confirmatory sampler requires --hmax 5")
```
This check is unconditional — good. But the test `test_cli_writes_from_lmdb_real_title_layer` passes `--hmax 5` and `--allow-small-sample` together, which is the correct preregistered invocation. The issue is that *`--allow-small-sample` is semantically intended to suppress the `min_descendants` and `pairs` checks for dry runs, not to unlock `--hmax`*. The current code enforces this correctly, but a comment in `validate_args` explicitly stating that `--allow-small-sample` does **not** relax the `--hmax` constraint would prevent future contributors from loosening the wrong guard.

**C-2. `row_shuffle_seed` is a fixed XOR of `args.seed`**

```python
"row_shuffle_seed": args.seed ^ 0x5EED5EED,
```
This is carried over from the pre-PR code. For `seed=0` (the preregistered seed), `row_shuffle_seed = 0x5EED5EED = 1592459757`. The row shuffle seed is deterministic given the preregistered seed — correct for reproducibility. However, both `sample_from_tsv` and `sample_from_lmdb` feed into the same `sample_balanced_pairs(pool, args.pairs, args.hmax, args.seed)` call, meaning the shuffle is seeded identically regardless of input source. If the two sources produce different pool ordering (which they will, since TSV and LMDB traversal orders differ), the same `seed` will produce different row orderings. This is **expected and correct** behavior, but it means that a TSV-sourced run and an LMDB-sourced run with `--seed 0` are **not** directly comparable without also controlling for pool ordering. The manifest does not record pool ordering, only `selected_roots` and `slice_nodes`. Consider adding `"candidate_root_traversal_order": list(candidates)` to the manifest for full reproducibility.

**C-3. `attempts` list in manifests can grow very large for wide scope roots**

If a scope root has 10 000 direct children and none qualify, the manifest will contain a 10 000-element `candidate_root_attempts` array. For Wikipedia-scale graphs this is plausible. Consider capping the attempts list at a configurable max (e.g. `--max-attempts-logged 100`) or only logging attempts that fail the `min_descendants` gate vs. those that fail `can_satisfy_targets`.

---

### D. API Surface / Interface Design

**D-1. `LmdbTitleGraph.node_id` silently falls back to `int(title_or_id)` for unresolved numeric strings**

```python
if looks_int(title_or_id):
    return int(title_or_id)
```
A caller passing `"12345"` that is *not* in the title_s2i db will get `12345` back without error. This bypasses the title-layer requirement for score-input production and could allow numeric IDs to leak into the output TSV. The guard in `LmdbTitleGraph.title()` (checking `is_identity_numeric_title`) would catch this *if* that node_id exists in `title_i2s`, but if it does not (because it was never materialized), `title()` raises `FreshCorpusError("title layer lacks node id ...")` — which is a different, less clear error than the identity-numeric guard. The intent seems to be to allow users to supply raw IDs on the command line; if so, this should be explicitly documented and tested.

**D-2. `_open_title_db` fallback to `"i2s"` / `"s2i"` is undocumented**

```python
def _open_title_db(self, preferred, fallback):
    db = self._open_optional_db(preferred)
    if db is not None:
        return preferred, db
    db = self._open_optional_db(fallback)
```
When called as `self._open_title_db(title_i2s_db, "i2s")`, the fallback is the bare string `"i2s"`. This appears to be a legacy sub-db name. There is no documentation of when a Phase-1 LMDB would have `"i2s"` instead of `"title_i2s"`. If this legacy name was never actually used in production, the fallback is dead code that adds confusion. If it was used, it should be tested.

**D-3. `sample_from_lmdb` passes `targets` computed locally, while `sample_from_tsv` does not compute its own `targets`**

```python
# sample_from_lmdb:
targets = hop_targets(args.pairs, args.hmax)
...
first_eligible_lmdb_roots(..., targets, ...)

# sample_from_tsv:
# no local targets — first_eligible_roots computes its own internally
```
The TSV path passes `targets` as a computed argument inside `first_eligible_roots`, while the LMDB path computes `targets` externally and passes it in. This asymmetry means that if `hop_targets` ever changes its signature, one call site could be missed. Unifying to always pass `targets` explicitly (or always let the callee compute it) would be cleaner.

---

### E. Minor Style / Hygiene

**E-1. Blank line duplication** — `test_sigma_hop_fresh_corpus.py` has a double blank line between `test_load_edges_skips_header` and `_build_lmdb`, which is a PEP 8 violation (two blank lines between top-level definitions is standard, but three appears in the diff).

**E-2. `import sys` inside test functions** — Both new CLI tests import `sys` inside the function body rather than at module level. This is not wrong but inconsistent with the module's style.

**E-3. `scanned_category_titles` counts only namespace-14 rows** — The variable name implies all SQL rows are counted, but `iter_page_titles` already filters for `ns == CATEGORY_NS` before yielding. So `scanned_category_titles` in the manifest actually means *"namespace-14 SQL rows processed"*, not *"total SQL rows scanned"*. The name is misleading; `matched_category_sql_rows` or `candidate_category_rows` would be more precise.

---

### Priority Matrix

| ID | Severity | Description |
|----|----------|-------------|
| B-1 | **High** | `graph_node_ids` misses non-first dup parents → under-populated title layer |
| A-1 | Medium | `can_satisfy_targets` called before `min_descendants` gate → unnecessary O(n) work |
| B-3 | Medium | `lock=False` concurrent-write hazard undocumented |
| C-2 | Medium | TSV vs LMDB pool ordering differs with same seed — reproducibility gap undocumented |
| D-1 | Medium | Numeric fallback in `node_id()` bypasses title-layer requirement silently |
| A-4 | Low | Dead `targets` recomputation in `main()` |
| B-2 | Low | `os.path.exists` race on NFS in manifest writer |
| B-4 | Low | `int()` truncation of float GiB → use `round()` |
| C-3 | Low | Unbounded `attempts` list in manifest for wide scope roots |
| A-2 | Low | `set` vs `list` in parents/children dicts — confirm downstream compat |
| A-3 | Low | `selection_rule` string accuracy depends on `sort_key` semantics |
| C-1 | Low | Comment needed: `--allow-small-sample` does not relax `--hmax` |
| D-2 | Low | Undocumented `"i2s"`/`"s2i"` legacy fallback |
| D-3 | Low | Asymmetric `targets` computation between TSV and LMDB paths |
| E-1–3 | Trivial | Style/naming hygiene |

**Recommended pre-merge fixes: B-1 and B-3.** B-1 is a correctness bug that could produce an incomplete title layer on production Wikipedia dumps without any error signal.

s243a (commented):
## Third Academic Code Review — PR #3521: Add LMDB title-layer support for sigma-hop fresh sampling

This third pass focuses on error-handling completeness, Unicode/encoding correctness, title-namespace collision hazards, CLI usability, and operational concerns not addressed in the first two reviews.

---

### F. Unicode and Encoding Correctness

**F-1. `errors="replace"` silently corrupts titles in `LmdbTitleGraph.title()` and `iter_page_titles`**

```python
# LmdbTitleGraph.title():
title = bytes(raw).decode("utf-8", "replace")

# iter_page_titles:
with gzip.open(page_dump, "rt", encoding="utf-8", errors="replace") as f:
```

Using `errors="replace"` substitutes U+FFFD (`�`) for undecodable bytes. If a MediaWiki page dump contains a title with a UTF-8 encoding error (rare but possible for older dumps), the materializer writes a corrupted title to `title_i2s`, and `LmdbTitleGraph.title()` returns the corrupted form. The two corruption events are independent: the materializer writes `"Caf�é"` and the sampler reads `"Caf�é"` — these match bidirectionally, so the pipeline does not crash. However, the `title_s2i` reverse lookup uses the corrupted key, meaning any user-supplied `--root "Café"` will fail to resolve while `--root "Caf�é"` would work. For a correctness-critical preregistered pipeline, `errors="strict"` with a fallback to explicit logging is preferable so that encoding problems are surfaced rather than silently mangled.

**F-2. `iter_page_titles` strips the `Category:` namespace prefix implicitly**

MediaWiki `page` SQL dumps store titles *without* the namespace prefix for namespace-14 rows (i.e., the title for category page "Category:Mammals" is stored as `"Mammals"`, not `"Category:Mammals"`). The code correctly uses `row[2]` directly, but this is not documented. Meanwhile, `LmdbTitleGraph.node_id()` explicitly strips `"Category:"` prefixes from user-supplied roots:
```python
if title_or_id.startswith("Category:"):
    candidates.append(title_or_id[len("Category:"):])
```
This means users *can* supply `--root "Category:Mammals"` and it will resolve correctly. But the `title_i2s` / `title_s2i` dbs store bare `"Mammals"`, so the manifest's `selected_roots` will list `"Mammals"` while the user supplied `"Category:Mammals"`. This mismatch between user input and manifest output could cause confusion in downstream consumers that reconstruct Wikipedia URLs. A normalization note in the module docstring, or storing both the bare and prefixed forms in the manifest, would prevent ambiguity.

**F-3. MediaWiki SQL title escaping includes `\_` and `\%` for LIKE patterns**

`parse_values` handles `\\` (backslash escape) and `''` (SQL doubled-quote) but does not handle `\_` or `\%`. In standard MediaWiki SQL dumps these are used in title fields with underscores or percent signs (since underscores double as spaces in MediaWiki titles). For example, a category title `"Science_and_technology"` may appear as `'Science\_and\_technology'` in some dump formats. The current parser would emit `"Science_and_technology"` correctly (since `\_` would be parsed as the escape-char handler consuming `\_` → `_` via `cur.append(sql_values[i + 1])`), but only because `_` is a single-character escape that happens to be its own literal. This is *accidentally correct* for `\_` but worth a targeted test to confirm, since a future change to the escape handler could break it silently.

---

### G. Title-to-ID Collision and Bijection Assumptions

**G-1. `title_s2i` does not handle title collisions (non-injective mapping)**

`txn.put(title_bytes, page_bytes, db=title_s2i)` uses the default `overwrite=True`, meaning if two different page IDs share the same title string (possible due to historical redirects or capitalisation normalization in the dump), the later write silently overwrites the earlier one. MediaWiki enforces title uniqueness within a namespace, so this should not occur in a well-formed dump — but malformed or concatenated dumps (common in pipeline environments) can violate this. An assertion or count of skipped duplicates written to the manifest would make violations visible.

**G-2. `title_i2s` and `title_s2i` are not opened with `dupsort=True`, so they cannot represent one-to-many or many-to-one relationships**

This is a design choice, not a bug — but it deserves explicit documentation. If the pipeline is ever extended to handle redirect chains (where one title maps to multiple page IDs across dump versions), the current schema silently loses all but one entry. The module docstring's schema description should note that these are bijective sub-dbs.

**G-3. `is_identity_numeric_title` check is performed on every cache miss in `LmdbTitleGraph.title()`**

```python
if is_identity_numeric_title(node_id, title):
    raise FreshCorpusError(...)
```

This raises on the *first* such node encountered and aborts the entire run. If a large graph has 1 000 nodes with identity-numeric titles (e.g. from a partial materialization), the error message names only one node ID, making diagnosis slow. Consider collecting a sample of violating nodes (up to 5–10) before raising, or reporting the total count in the error message.

---

### H. Error Handling Completeness

**H-1. `sample_from_lmdb` catches `FreshCorpusError` implicitly via the `finally: graph.close()` pattern, but non-`FreshCorpusError` exceptions bypass `graph.close()` in one path**

```python
def sample_from_lmdb(args, exploratory_nodes):
    ...
    graph = LmdbTitleGraph(...)
    try:
        root_ids, selection_rule, scope_info = lmdb_candidate_root_ids(graph, ...)
        selected_roots, ... = first_eligible_lmdb_roots(...)
        source_manifest = {...}
        ...
        return selected_roots, slice_nodes, pool, source_manifest
    finally:
        graph.close()
```

The `finally: graph.close()` block correctly runs on all exit paths, including exceptions — this is actually correct Python. **However**, `graph.close()` calls `self.txn.abort()` followed by `self.env.close()`. If `self.txn` was already aborted or committed (e.g. due to an LMDB internal exception that auto-aborts), calling `txn.abort()` again raises `lmdb.Error: Attempt to operate on closed/deleted/dropped object`. This would mask the original exception. Guarding with a `try/except lmdb.Error: pass` inside `close()`, or tracking transaction state with a `self._txn_open` flag, prevents exception masking.

**H-2. `lmdb_candidate_root_ids` evaluates `graph.meta_int("scoped_root") or graph.meta_int("metric_min_dist_to_root.root")`**

```python
scope_id = graph.meta_int("scoped_root") or graph.meta_int("metric_min_dist_to_root.root")
```

Python's `or` short-circuits, so `meta_int("metric_min_dist_to_root.root")` is only called if `meta_int("scoped_root")` returns `None` or `0`. A `scoped_root` of `0` is a valid page ID (the root of the category tree in some MediaWiki instances uses page_id=0 as a sentinel). This means a LMDB with `meta.scoped_root = 0` would silently fall through to check `metric_min_dist_to_root.root` instead. Use `if scope_id is None` rather than relying on truthiness.

**H-3. `validate_args` does not validate `--pairs > 0` or `--hmax > 0`**

Both `hop_targets` and `sample_balanced_pairs` implicitly require positive values. `--pairs 0` with `--hmax 5` would produce an empty `targets` dict (`base = 0, rem = 0`) and `sample_balanced_pairs` would return zero rows, writing an empty output TSV with no error. Since this is a preregistered confirmatory pipeline, an explicit guard `if args.pairs <= 0` with a clear `FreshCorpusError` would prevent silent empty-output runs.

---

### I. Operational / Pipeline Concerns

**I-1. `materialize_lmdb_title_layer.py` does not check whether the LMDB map size is sufficient before writing**

LMDB raises `lmdb.MapFullError` when the map is exhausted. This is caught by the outer `except Exception: txn.abort()` in `main()`, which correctly aborts the current transaction. However, after a `MapFullError`, partial data from *already-committed* batch transactions (written before the map-full event) remains in the LMDB. A subsequent re-run with `--map-size-gib` increased will call `txn.drop(title_i2s, delete=False)` first, clearing the partial data — so recovery is possible. But the error message from `lmdb.MapFullError` is opaque (`lmdb.MapFullError: mdb_put: MDB_MAP_FULL: Environment mapsize limit reached`). Wrapping the exception to print `"Hint: re-run with --map-size-gib <larger value>"` would significantly improve operator experience.

**I-2. The `--page-dump` argument does not validate that the file is actually gzip-compressed**

`gzip.open` with mode `"rt"` will raise `gzip.BadGzipFile` (Python ≥ 3.8) or `OSError` if the file is not gzip-compressed. This is a common user error (supplying an uncompressed `.sql` file). Wrapping the `gzip.open` call with a clear `FreshCorpusError` or `SystemExit` mentioning gzip would improve the user experience over a raw traceback.

**I-3. `materialize_lmdb_title_layer.py` has no `--dry-run` or `--check` mode**

For a potentially multi-hour operation on a full Wikipedia dump, operators typically want a dry-run that scans the dump and counts candidates without writing to the LMDB. Adding `--dry-run` (skip all `txn.put` calls, print counts, write manifest with `"refreshed": False`) would make the tool significantly more operationally friendly, especially for map-size estimation before committing a full run.

**I-4. Progress reporting uses a fixed 500 000 batch interval, but startup I/O (`graph_node_ids` + `iter_page_titles` scanning) has no progress output**

On a full English Wikipedia dump (~100 M rows in `page.sql.gz`), the initial `graph_node_ids` scan and the dump scan can take several minutes each with no output. Adding a `--verbose` flag or a simple `print(f"scanning dump... {scanned} rows", end="\r")` every million rows would help operators distinguish a stalled process from a running one.

---

### J. Test Infrastructure

**J-1. Both new test files use `sys.argv` patching to invoke `main()` directly**

```python
old_argv = sys.argv
try:
    sys.argv = ["sample_sigma_hop_fresh_corpus.py", ...]
    main()
finally:
    sys.argv = old_argv
```

This pattern is not thread-safe: if pytest ever runs tests in parallel (e.g. via `pytest-xdist`), two tests patching `sys.argv` concurrently will interfere. The standard solution is `unittest.mock.patch("sys.argv", [...])` or refactoring `main()` to accept an `argv=None` parameter (defaulting to `sys.argv[1:]`) so tests can call `main(argv=[...])` directly.

**J-2. Test fixtures do not exercise the `--scope-root` → LMDB meta fallback path**

The path `scope_id = graph.meta_int("scoped_root") or graph.meta_int("metric_min_dist_to_root.root")` is exercised only when neither `--root` nor `--scope-root` is supplied and the LMDB has a `meta.scoped_root` key. None of the new tests cover this code path, meaning it is untested. This is one of the more complex branching paths (with the `or`-truthiness bug noted in H-2), and a targeted unit test would be valuable.

**J-3. `test_cli_writes_from_lmdb_real_title_layer` asserts `hop_counts == {"1": 2, ..., "5": 2}` but the fixture topology does not guarantee this**

The `_lmdb_chain_fixture` builds 8 independent depth-5 chains rooted at node `1`. With `--pairs 10` and `--hmax 5`, `hop_targets(10, 5)` gives `{1: 2, 2: 2, 3: 2, 4: 2, 5: 2}`. This *happens* to work because there are exactly 8 chains × 1 pair/hop = 8 pairs available per hop, well above the target of 2. However, the test hardcodes `{"1": 2, ..., "5": 2}` without documenting the arithmetic. If `hop_targets` or `sample_balanced_pairs` logic changes (e.g. to handle odd `pairs`), this assertion will fail with a non-obvious message. Deriving the expected value programmatically inside the test would be more maintainable:
```python
expected = {str(h): 2 for h in range(1, 6)}
assert m["hop_counts"] == expected
```

---

### K. Documentation and Module Contracts

**K-1. Neither new module has a `__all__` declaration**

`materialize_lmdb_title_layer.py` exposes `parse_values`, `iter_page_titles`, `graph_node_ids`, `write_manifest`, and `main` as importable names. Without `__all__`, `from materialize_lmdb_title_layer import *` would pull in `I32`, `CATEGORY_NS`, `enc_i32`, and `dec_i32` as well. For a prototype script this is acceptable, but since the test file already imports `main` directly (`from materialize_lmdb_title_layer import main`), an `__all__` would document the intended public API.

**K-2. `LmdbTitleGraph` has no `__repr__` or `__str__`**

For debugging long pipeline runs, a `__repr__` showing `lmdb_dir`, `title_layer_kind` from meta, and the count of cached titles would be useful in tracebacks and interactive sessions.

**K-3. The PR description test commands reference relative paths without context**

```
python3 prototypes/mu_cosine/test_sigma_hop_fresh_corpus.py
```

These commands assume the working directory is the repository root and that `prototypes/mu_cosine/` is on `sys.path`. There is no `conftest.py` or `__init__.py` that ensures this. A note in the PR description or a `Makefile` target would clarify the expected invocation context.

---

### Priority Matrix (new findings only)

| ID | Severity | Description |
|----|----------|-------------|
| H-2 | **High** | `or`-truthiness on `meta_int` masks `scoped_root = 0` |
| F-1 | Medium | `errors="replace"` silently corrupts titles; use `errors="strict"` with logging |
| G-1 | Medium | `title_s2i` silent overwrite on title collision |
| H-1 | Medium | Double-abort in `graph.close()` can mask original exception |
| H-3 | Medium | No validation of `--pairs > 0` / `--hmax > 0` in `validate_args` |
| J-1 | Medium | `sys.argv` patching not thread-safe under `pytest-xdist` |
| I-1 | Low | `MapFullError` recovery path needs user-friendly hint message |
| I-2 | Low | Non-gzip `--page-dump` gives opaque traceback |
| F-2 | Low | `Category:` prefix stripping asymmetry between CLI input and manifest output |
| G-3 | Low | Identity-numeric error names only one node; sample multiple violators |
| J-2 | Low | `meta.scoped_root` fallback path untested |
| J-3 | Low | Hardcoded `hop_counts` assertion should be derived programmatically |
| F-3 | Informational | `\_` escape in SQL parser accidentally correct; add targeted test |
| G-2 | Informational | `title_i2s`/`title_s2i` bijection assumption undocumented |
| I-3 | Enhancement | `--dry-run` mode for map-size estimation |
| I-4 | Enhancement | Progress output during dump scan |
| K-1–3 | Trivial | `__all__`, `__repr__`, PR description test invocation context |

**Recommended pre-merge fixes: H-2** (the `or`-truthiness bug is a silent semantic error that will produce wrong scope roots on any LMDB with `scoped_root = 0`), **F-1** (encoding corruption in a preregistered pipeline is a reproducibility risk), and **J-1** (unsafe `sys.argv` patching if CI ever enables parallel test execution).

s243a (commented):
## Fourth Academic Code Review — PR #3521: Add LMDB title-layer support for sigma-hop fresh sampling

This fourth pass focuses on concurrency and re-entrancy hazards, algorithmic complexity and scalability bounds, boundary-condition arithmetic, CLI contract completeness, and cross-module invariant consistency — areas not substantively addressed in the first three reviews.

---

### L. Concurrency, Re-entrancy, and Signal Safety

**L-1. `materialize_lmdb_title_layer.py` `main()` is not signal-safe during the batch-commit rotation**

The batch-commit loop does:
```python
txn.commit()
txn = env.begin(write=True)
print(f"materialized {matched} category titles")
```
If SIGINT (Ctrl-C) arrives between `txn.commit()` and `txn = env.begin(write=True)`, the `except Exception: txn.abort()` handler will call `abort()` on the *newly assigned* variable, which may be unassigned or partially constructed. In CPython this is unlikely to cause corruption due to the GIL, but on POSIX the OS can deliver the signal between any two bytecode instructions. More critically, if SIGINT arrives *after* `txn.commit()` but *before* the new `txn = env.begin()` assignment, the `finally: env.sync(); env.close()` block runs without an active transaction — which is correct, but the `except Exception: txn.abort()` block may try to abort an already-committed transaction if the exception path is taken. A `signal.signal(signal.SIGINT, ...)` handler that sets a flag checked at the top of the loop, or a `contextlib.contextmanager` for transaction rotation, would make this safe.

**L-2. `LmdbTitleGraph._parents` and `_children` caches are populated lazily from a shared transaction but are plain dicts with no thread safety**

If `LmdbTitleGraph` is ever used from multiple threads (e.g., future parallelization of root candidate evaluation), two threads calling `graph.parents(node_id)` simultaneously could both miss the cache and both call `iter_dups`, then both write `self._parents[node_id]` with the result. In CPython this is safe due to the GIL (dict assignment is atomic at the bytecode level), but the `iter_dups` call itself uses a cursor that is not documented as thread-safe in python-lmdb. A docstring noting the single-threaded assumption is the minimum fix.

**L-3. `write_manifest` in both modules uses `tempfile.mkstemp` with `dir=str(out_dir)`, but `out_dir.mkdir(parents=True, exist_ok=True)` is called immediately before**

If two processes call `write_manifest` for the same path concurrently (possible in a `-j2` make pipeline), both will successfully create temp files and both will call `os.replace`. POSIX guarantees `os.replace` is atomic, so the last writer wins without corruption — this is correct behavior. However, the `finally: if os.path.exists(tmp): os.unlink(tmp)` cleanup in the *losing* process will find that `tmp` was already replaced (the file still exists with the new content), and will silently delete the winner's manifest. This is a race: Process A writes tmp_A, Process B writes tmp_B, A does `os.replace(tmp_A, path)`, B does `os.replace(tmp_B, path)`, then A's `finally` checks `os.path.exists(tmp_A)` — since `tmp_A` was successfully replaced (no longer exists), it returns `False` and A's finally is a no-op. This is actually **correct** — the concern is unfounded — but the code has a subtle correctness argument that is worth a comment.

---

### M. Algorithmic Complexity and Scalability

**M-1. `load_lmdb_slice_maps` builds `id_to_title` as a local dict that shadows `LmdbTitleGraph._title_cache`**

```python
def load_lmdb_slice_maps(graph, root_id, exploratory_nodes, slice_depth=None):
    id_to_title = {}
    def title_for(node_id):
        if node_id not in id_to_title:
            id_to_title[node_id] = graph.title(node_id)  # populates graph._title_cache too
            ...
        return id_to_title[node_id]
```

`graph.title()` populates `graph._title_cache` on every call, so `id_to_title` and `_title_cache` hold the same data. For a slice of N nodes this doubles the memory footprint. If `first_eligible_lmdb_roots` iterates over K candidate roots before finding an eligible one, and each slice has N nodes, the total memory is O(K × N) for `id_to_title` (a new dict per call) plus O(total unique nodes) for `_title_cache`. Since `id_to_title` is local and goes out of scope after each `load_lmdb_slice_maps` call, the K×N growth is bounded to a single call — but for a scope root with 500 children each with 5 000-node slices, the peak usage is 500 × 5 000 = 2.5M cache entries. Consider whether `id_to_title` is necessary at all, or whether `title_for` could simply delegate entirely to `graph.title()` (which already caches).

**M-2. `first_eligible_lmdb_roots` calls `graph.title(nid)` for sorting the initial candidate list, then calls it again inside `load_lmdb_slice_maps` via `title_for`**

```python
candidates = sorted(root_ids, key=lambda nid: sort_key(graph.title(nid)))
for root_id in candidates:
    root_title = graph.title(root_id)   # second call — cache hit, but still two call sites
    ...
    root_title, slice_nodes, ... = load_lmdb_slice_maps(graph, root_id, ...)
    # load_lmdb_slice_maps calls title_for(root_id) — third logical title lookup
```

Because `graph.title()` caches, these are all O(1) after the first. The redundancy is harmless at runtime but increases cognitive load when tracing the call graph. Storing `candidates` as `[(sort_key, nid, title), ...]` would make the flow explicit.

**M-3. `pair_pool_for_roots` is called with the materialized title-keyed dicts from `load_lmdb_slice_maps`, but its inner `ancestors_by_hop` and `descendants_within` use `sorted(..., key=sort_key)` which is O(d log d) per node (d = degree)**

For a Wikipedia category graph with high-degree hubs (e.g., "Hidden categories" can have tens of thousands of children), this sort dominates runtime. The TSV path has the same issue. This is a pre-existing concern that this PR does not worsen, but since the LMDB path now enables much larger graphs (full Wikipedia rather than a 100k-category slice), the scalability concern is materially more relevant after this PR. A note in the module docstring about the expected graph size bounds would set appropriate operational expectations.

---

### N. Boundary Condition Arithmetic

**N-1. `hop_targets` distributes remainder to the *lowest* hops, which biases the preregistered dataset**

```python
def hop_targets(total_pairs, hmax):
    base = total_pairs // hmax
    rem = total_pairs % hmax
    return {hop: base + (1 if hop <= rem else 0) for hop in range(1, hmax + 1)}
```

For `total_pairs=251, hmax=5`: `base=50, rem=1`, so hop-1 gets 51 pairs and hops 2–5 get 50. The extra pair always goes to the *lowest* hop (shortest-distance pairs). For a confirmatory study that relies on balanced transitive-hop representation, this asymmetry should be documented as a deliberate design choice. If it was unintentional, distributing the remainder to the *highest* hop (most informative for long-range transitivity) or using round-robin assignment would be preferable.

**N-2. `sample_balanced_pairs` uses `random.shuffle` seeded via `random.seed(seed ^ 0x5EED5EED)`, but `random` is a module-level singleton**

```python
random.seed(seed ^ 0x5EED5EED)
...
random.shuffle(rows)
```

If any other code in the process calls `random.random()` or `random.shuffle()` between the `random.seed()` call and the `random.shuffle(rows)` call (e.g., in a future refactor that adds randomized retry logic), the shuffle order will change silently, breaking reproducibility. Using `rng = random.Random(seed ^ 0x5EED5EED); rng.shuffle(rows)` creates an isolated RNG instance that is immune to external state mutation.

**N-3. `descendants_within` uses a BFS `depth` dict but does not handle DAG diamonds correctly for `max_depth`**

```python
depth[child] = depth[node] + 1
```

In a DAG (which the Wikipedia category graph is), a node can be reachable via multiple paths of different lengths. BFS assigns the *shortest-path* depth. If `max_depth=3` and a node is reachable at depth 2 via one path and depth 4 via another, BFS correctly assigns depth 2. However, `descendants_within` uses a `kept` set to avoid re-enqueuing, so a node first reached at depth 3 (= max_depth) is added to `kept` but *not* enqueued for further expansion. If the same node is also reachable at depth 1 via a different path, it would have been enqueued at depth 1 and its children explored. This is correct BFS behavior. However, the `max_depth` semantics are therefore *minimum-depth based*, not *maximum-depth based* — a node at minimum BFS depth 2 will have its children explored even if it is also reachable at depth 5. This is documented as intentional (`hop` means minimum upward graph distance), but the `slice_depth` parameter in `descendants_within` has the same minimum-depth semantics, which may surprise users who expect it to cap the *tree depth* of the output.

---

### O. CLI Contract and Usability

**O-1. `--scope-root` is accepted by `validate_args` without checking for mutual incompatibility with `--root` in the LMDB path**

```python
# lmdb_candidate_root_ids:
if roots:   # args.root is non-empty
    return [graph.node_id(root) for root in roots], "user-supplied root validation", None
# scope_root is silently ignored if roots is non-empty
```

If a user passes both `--root CategoryA` and `--scope-root BroadCategory`, `--scope-root` is silently ignored. This is the same behavior as the TSV path, but is not documented in `--scope-root`'s help string (`"broad LMDB/TSV root whose direct children are candidate root slices"`). Adding `"(ignored if --root is also supplied)"` to the help text, or raising a `FreshCorpusError` on the combination, would prevent silent no-ops.

**O-2. `--title-i2s-db` and `--title-s2i-db` are CLI flags but are also written to the manifest, creating a round-trip ambiguity**

If a user runs the materializer with non-default `--title-i2s-db my_i2s` and then runs the sampler without specifying `--title-i2s-db`, the sampler defaults to `"title_i2s"` and fails to open the correct sub-db. The manifest from the materializer records the actual db name, but there is no mechanism to feed that back into the sampler automatically. A `--title-layer-manifest` flag that reads the materializer manifest and auto-populates `--title-i2s-db` / `--title-s2i-db` would close this workflow gap.

**O-3. Error messages for missing LMDB sub-dbs expose internal implementation names**

```python
raise FreshCorpusError(
    f"candidate LMDB has no real title layer: missing `{preferred}` and fallback `{fallback}` sub-dbs"
)
```

The fallback name (`"i2s"` / `"s2i"`) is an internal detail. Exposing it in the error message implies to users that creating a sub-db named `"i2s"` is a supported workflow, which it may not be. The error message should guide the user toward the correct fix: `"Run materialize_lmdb_title_layer.py --lmdb-dir <path> --page-dump <dump> to add the required title layer."`

---

### P. Cross-Module Invariant Consistency

**P-1. `materialize_lmdb_title_layer.py` and `sample_sigma_hop_fresh_corpus.py` define `I32 = struct.Struct("<i")` independently**

Already flagged in prior reviews as a DRY concern. The new angle: `"<i"` is a *signed* 32-bit little-endian integer. MediaWiki page IDs are unsigned 32-bit integers in the database schema (`unsigned int(8)`). For IDs ≥ 2³¹ = 2 147 483 648 (possible in the English Wikipedia as of 2026, which has page IDs exceeding 70M — well under 2³¹, but the schema does not bound them), `I32.pack(int(value))` would raise `struct.error: byte format requires -2147483648 <= number <= 2147483647`. Using `struct.Struct("<I")` (unsigned) would be semantically correct and extend the safe range to 4 294 967 295. This is not a current issue but is a latent correctness hazard as Wikipedia page IDs grow.

**P-2. `filter_candidate_edges` in the TSV path removes edges involving `exploratory_nodes` before building `parents`/`children` maps; the LMDB path does this at slice-build time via `title_block_reason`**

The TSV path removes *all* edges touching an exploratory node (including edges between two non-exploratory nodes that pass through an exploratory intermediate). The LMDB path removes only nodes from the BFS frontier when their title matches an exploratory node. These are semantically different: the TSV path is more aggressive (removes edges; LMDB path only prunes subtree traversal). For a strictly preregistered pipeline these behaviors should be either aligned or explicitly justified in a comment. The difference could produce different `slice_nodes` sets for the same graph structure.

**P-3. The `ADMIN` regex (imported/defined elsewhere in the file) is applied to string titles in both TSV and LMDB paths**

The LMDB path applies `ADMIN.search(root_title)` in `first_eligible_lmdb_roots` and `ADMIN.search(title)` in `title_block_reason`. The TSV path applies `ADMIN.search(root)` in `first_eligible_roots`. Both paths use string titles here, which is consistent. However, `title_block_reason` is called with `child_title` *before* checking `child_title in exploratory_nodes`, meaning the admin check runs even for exploratory nodes. The order should arguably be: check exploratory overlap first (fast dict lookup), then run the regex (slower). This is a minor performance concern but also a logical ordering issue: an exploratory admin node should be classified as `"overlap_node"`, not `"admin_node"`.

---

### Priority Matrix (new findings only)

| ID | Severity | Description |
|----|----------|-------------|
| P-1 | **High** | Signed `struct.Struct("<i")` will raise for page IDs ≥ 2³¹; should be `"<I"` (unsigned) |
| P-2 | **High** | TSV vs LMDB exploratory-node filtering semantics differ — non-equivalent slice construction |
| N-2 | Medium | Module-level `random.seed()` is not isolated; use `random.Random(seed)` instance |
| L-1 | Medium | SIGINT between batch-commit and re-open can leave inconsistent exception state |
| O-1 | Medium | `--scope-root` silently ignored when `--root` also supplied; undocumented |
| M-1 | Medium | `id_to_title` local dict duplicates `_title_cache` memory unnecessarily |
| N-1 | Low | `hop_targets` remainder always goes to lowest hop — deliberate? Document it |
| M-2 | Low | Triple logical title-lookup per candidate root — minor clarity issue |
| O-2 | Low | No round-trip mechanism from materializer manifest to sampler `--title-*-db` flags |
| O-3 | Low | Error message exposes internal fallback db names instead of guiding to fix |
| P-3 | Low | `title_block_reason` checks admin before exploratory overlap (suboptimal order) |
| N-3 | Informational | `slice_depth` semantics are minimum-BFS-depth, not tree-depth; worth documenting |
| L-2 | Informational | `LmdbTitleGraph` caches not thread-safe; single-threaded assumption undocumented |
| M-3 | Informational | LMDB path enables full-Wikipedia graphs; O(d log d) sort on high-degree hubs may be slow |

**Recommended pre-merge fixes: P-1** (the signed/unsigned struct bug will raise on real-world Wikipedia data if page IDs ever exceed 2.1B, and the fix is one character) and **P-2** (the TSV vs LMDB exploratory filtering semantics difference could produce non-equivalent samples, undermining the preregistration guarantee).

PR #3525, opened 2026-07-07
Codex/sigma hop confirmatory run

s243a:
# Report confirmatory Sigma(hop) result

## Summary
- add LMDB feature-graph support to the Sigma(hop) confirmatory runner
- run the preregistered fresh-corpus test on the `Behavior` LMDB slice
- report the confirmed result: mean gain `+0.059799`, `K=1000`, permutation `p=0.000999`

## Tests
- `python3 prototypes/mu_cosine/test_sigma_hop_confirmatory.py`
- `python3 -m py_compile prototypes/mu_cosine/sigma_hop_confirmatory.py prototypes/mu_cosine/sample_sigma_hop_fresh_corpus.py`
- full preregistered confirmatory run, `K=1000`

### DIFF
--- a/prototypes/mu_cosine/REPORT_sigma_hop_confirmatory.md
+++ b/prototypes/mu_cosine/REPORT_sigma_hop_confirmatory.md
@@ -44,6 +44,9 @@ rendered run report: /tmp/mu_data/sigma_hop_confirmatory_REPORT.md
 runner commit: 98efc2fdc Add LMDB graph support to sigma-hop confirmatory runner
 ```
 
+The `/tmp/mu_data/...` paths above are local run artifacts, not durable repository records. The durable record in this
+PR is this committed report plus the runner change needed to consume the LMDB feature graph.
+
 The selected root was `Behavior`. The retained LMDB slice had 75,901 nodes after the preregistered no-overlap and
 admin-title filtering. Sampling produced exactly 50 pairs at each shortest upward hop 1..5.
 
--- a/prototypes/mu_cosine/REPORT_sigma_hop_confirmatory.md
+++ b/prototypes/mu_cosine/REPORT_sigma_hop_confirmatory.md
@@ -1,0 +1,71 @@
+# Confirmatory Sigma(hop) run: confirmed on a fresh Behavior slice
+
+*Run 2026-07-07 against `PREREG_sigma_hop_confirmatory.md`, after sampling and scoring a fresh Wikipedia category
+slice that shares no category nodes with the exploratory `100k_cats` graph.*
+
+## Decision
+
+The preregistered confirmatory decision is **confirmed**: smooth `Sigma(hop)` beat constant-Sigma on held-out
+continuous-mu residual likelihood, and the one-sided hop-shuffle permutation test passed the frozen `p < 0.01`
+threshold.
+
+```text
+fresh corpus dump/root: enwiki_cats_correct scoped LMDB; title layer mediawiki_page_titles; scope Main_topic_classifications; selected root Behavior
+node-overlap check with 100k_cats: passed, 0 overlapping category titles
+n scored pairs: 250
+hop counts: {1: 50, 2: 50, 3: 50, 4: 50, 5: 50}
+judge/prompt/model: gpt-5.5 with model_reasoning_effort=low via score_with_codex.py / score_inferred_tail.py prompt
+valid descendant-disjoint splits: 40
+mean held pairs/split: 75.0
+observed mean gain: +0.059799
+constant-Sigma NLL: -0.604010
+Sigma(hop) NLL: -0.663809
+hop-shuffle null mean: -0.009487
+hop-shuffle null 95%ile: +0.000456
+K: 1000
+permutation p: 0.000999
+decision: confirmed
+skipped splits: 0
+```
+
+At `K=1000`, `p=0.000999` is the finite-permutation floor `(1 + 0) / (1000 + 1)`: no shuffled-hop null run reached
+the observed mean gain.
+
+## Artifacts
+
+```text
+score input: /tmp/mu_data/sigma_hop_fresh_pairs.tsv
+sampling manifest: /tmp/mu_data/sigma_hop_fresh_manifest.json
+raw judge responses: /tmp/mu_data/sigma_hop_fresh_responses_gpt55low.txt
+ingested judge scores: /tmp/mu_data/sigma_hop_fresh_scored_gpt55low.tsv
+retained-slice e5 cache: /tmp/mu_data/sigma_hop_behavior_slice_e5.pt
+result JSON: /tmp/mu_data/sigma_hop_confirmatory_result.json
+rendered run report: /tmp/mu_data/sigma_hop_confirmatory_REPORT.md
+runner commit: 98efc2fdc Add LMDB graph support to sigma-hop confirmatory runner
+```
+
+The selected root was `Behavior`. The retained LMDB slice had 75,901 nodes after the preregistered no-overlap and
+admin-title filtering. Sampling produced exactly 50 pairs at each shortest upward hop 1..5.
+
+## Implementation Notes
+
+Two mechanical issues were fixed before any gain/permutation result was computed:
+
+- The confirmatory runner was originally TSV-only for feature-graph loading, while the fresh corpus came from the
+  LMDB title graph. Commit `98efc2fdc` added an LMDB feature-graph adapter that reuses the sampler's retained-slice
+  traversal. This changed no labels, model family, split protocol, permutation null, threshold, or decision rule.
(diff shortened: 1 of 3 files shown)


prototypes/mu_cosine/sigma_hop_confirmatory.py
```diff
@@ -260,6 +261,63 @@ def assert_no_node_overlap(pairs, exploratory_graph):
...
+    title_s2i_db="title_s2i",
+    lmdb_no_lock=False,
+):
+    """Return parents/children/degree maps for confirmatory feature construction.
+
+    TSV keeps backward compatibility with the exploratory runner. LMDB reuses the sampler's retained-slice
+    traversal so `hit_prob` is computed on the same no-overlap/admin-filtered graph that produced the fresh pairs.
+    """
+    if graph:
+        parents, children, deg = load_dag(graph)
```
s243a: **`load_feature_graph` — resource safety in the LMDB branch**

The `lmdb_graph.close()` is inside a `finally` block, which is correct, but `LmdbTitleGraph` is not used as a context manager here. If the class supports `__enter__`/`__exit__` (or can be cheaply extended to), prefer:

```python
with LmdbTitleGraph(candidate_lmdb, title_i2s_db, title_s2i_db, lock=not lmdb_no_lock) as lmdb_graph:
    root_id = lmdb_graph.node_id(lmdb_root)
    root_title, slice_nodes, parents, children, stats = load_lmdb_slice_maps(
        lmdb_graph, root_id, exploratory_nodes
    )
```

This is stylistically cleaner and eliminates the `try/finally` boilerplate. If `LmdbTitleGraph` is a thin wrapper around `lmdb.open()`, adding `__enter__`/`__exit__` is trivial.

prototypes/mu_cosine/sigma_hop_confirmatory.py
```diff
@@ -260,6 +261,63 @@ def assert_no_node_overlap(pairs, exploratory_graph):
     return overlap
 
 
+def degree_from_maps(parents, children):
+    nodes = set(parents) | set(children)
```
s243a, line 278: **`degree_from_maps` — small correctness note**

The current implementation builds degree as `len(parents.get(node, ())) + len(children.get(node, ()))`. This is a pure in-degree + out-degree sum for the stored maps, which matches the contract used in `load_dag`. However, if a node appears as a *value* inside a parent/children list without being a *key* in either map (i.e. a terminal leaf or isolated root), it would be missing from the `nodes` set and therefore missing from `deg`.

Consider whether `hit_prob` / the `Tokenizer` ever queries `deg` for such nodes. If so, add an explicit default:

```python
def degree_from_maps(parents, children):
    nodes = set(parents) | set(children)
    # also include pure leaf/root nodes referenced as values
    for v_list in parents.values():
        nodes.update(v_list)
    for v_list in children.values():
        nodes.update(v_list)
    return {
        node: len(parents.get(node, ())) + len(children.get(node, ()))
        for node in nodes
    }
```

At minimum, add a comment documenting the assumption that all queried nodes are keys.

prototypes/mu_cosine/sigma_hop_confirmatory.py
```diff
@@ -281,8 +339,8 @@ def load_e5_cache_and_filter(pairs, hop, D, S, e5_cache):
     return cache, idx, filtered_pairs, hop[keep], D[keep], S[keep]
 
 
-def build_confirmatory_features(pairs, cache, idx, graph, model_path, device):
-    parents, _, deg = load_dag(graph)
+def build_confirmatory_features(pairs, cache, idx, graph, model_path, device, **feature_graph_kwargs):
+    parents, _, deg, _ = load_feature_graph(graph=graph, **feature_graph_kwargs)
     tokenizer = Tokenizer(cache["query"], cache["passage"], idx, parents, deg)
     model = build_model(model_path, device)
```
s243a, line 358: **`build_confirmatory_features` / `build_confirmatory_data*` — `**feature_graph_kwargs` threading**

Propagating kwargs through three layers (`build_confirmatory_data` → `build_confirmatory_data_from_labels` → `build_confirmatory_features` → `load_feature_graph`) makes call-site readability harder and can silently swallow unexpected keyword arguments. A lightweight named dataclass or `typing.TypedDict` would make the contract explicit and provide IDE completion:

```python
@dataclasses.dataclass
class FeatureGraphConfig:
    graph: str | None = None
    candidate_lmdb: str | None = None
    lmdb_root: str | None = None
    exploratory_graph: str | None = None
    title_i2s_db: str = "title_i2s"
    title_s2i_db: str = "title_s2i"
    lmdb_no_lock: bool = False
```

This is a non-blocking suggestion — the current approach works — but worth considering before the runner grows more graph backends.

prototypes/mu_cosine/sigma_hop_confirmatory.py
```diff
@@ -372,7 +434,13 @@ def main():
...
     ap.add_argument("--e5-cache", required=True)
-    ap.add_argument("--graph", required=True)
+    graph_source = ap.add_mutually_exclusive_group(required=True)
+    graph_source.add_argument("--graph", help="child<TAB>parent TSV graph for feature construction")
+    graph_source.add_argument("--candidate-lmdb", help="Phase-1 category LMDB graph for feature construction")
+    ap.add_argument("--lmdb-root", help="selected retained LMDB slice root, e.g. the sampled root recorded in the manifest")
+    ap.add_argument("--title-i2s-db", default="title_i2s", help="LMDB uint32 id -> real category title sub-db")
+    ap.add_argument("--title-s2i-db", default="title_s2i", help="LMDB real category title -> uint32 id sub-db")
+    ap.add_argument("--lmdb-no-lock", action="store_true", help="open candidate LMDB with lock=False; use only for immutable fixtures")
     ap.add_argument("--model", default="model_prod.pt")
```
s243a, line 456: **CLI argument group: `--lmdb-root` should arguably be `required` within the LMDB sub-group**

Right now `--lmdb-root` is an independent optional argument. If a caller passes `--candidate-lmdb` but omits `--lmdb-root`, the error surfaces deep inside `load_feature_graph` as a `ConfirmatoryInputError` rather than at argparse validation time. Since `--lmdb-root` is always required when `--candidate-lmdb` is supplied, consider a custom validation step right after `ap.parse_args()`:

```python
if args.candidate_lmdb and not args.lmdb_root:
    ap.error("--lmdb-root is required when --candidate-lmdb is given")
```

This gives the user a clean usage message instead of a traceback.

prototypes/mu_cosine/REPORT_sigma_hop_confirmatory.md
```diff
@@ -0,0 +1,71 @@
...
+
+## Decision
+
+The preregistered confirmatory decision is **confirmed**: smooth `Sigma(hop)` beat constant-Sigma on held-out
+continuous-mu residual likelihood, and the one-sided hop-shuffle permutation test passed the frozen `p < 0.01`
+threshold.
+
+```text
+fresh corpus dump/root: enwiki_cats_correct scoped LMDB; title layer mediawiki_page_titles; scope Main_topic_classifications; selected root Behavior
+node-overlap check with 100k_cats: passed, 0 overlapping category titles
```
s243a, line 14: **Artifact paths are `/tmp` — consider noting reproducibility caveat**

All artifact paths (`score input`, `raw judge responses`, `e5 cache`, etc.) point to `/tmp/mu_data/`. These are ephemeral and will not survive a reboot or CI runner teardown. The report reads as a permanent record, so a brief callout like:

> *Artifact paths above are local run-time locations; see the sampling manifest and result JSON committed under `prototypes/mu_cosine/` for durable records.*

…would help future readers distinguish the live-run artifacts from the committed reproducibility artifacts.

s243a (commented):
## Overall

Solid confirmatory run — the result is clearly documented, the preregistered decision criteria are satisfied, and the LMDB adapter is a clean mechanical extension with no changes to labels, model, split protocol, or permutation null. The two-commit history (infrastructure first, report second) makes the audit trail easy to follow.

Five focused comments inline:

1. **`load_feature_graph`** — prefer a `with`-statement / context manager for `LmdbTitleGraph` over manual `try/finally`.
2. **`degree_from_maps`** — potential missing-node edge case for pure leaf/root nodes that are values but not keys in the parent/children maps.
3. **`**feature_graph_kwargs` threading** — consider a `FeatureGraphConfig` dataclass to make the contract explicit across the three-layer call chain.
4. **CLI validation** — add an explicit `ap.error()` check so `--lmdb-root` is validated at argument-parse time when `--candidate-lmdb` is supplied.
5. **Report artifact paths** — note that `/tmp/mu_data/` paths are ephemeral; point readers to committed durable records.

None of these are blockers for the scientific result. Items 1, 4 are quick fixes; items 2, 3, 5 are polish/documentation.

PR #3527, opened 2026-07-07
Document Product-Kalman PoE design

s243a:
## Summary

Adds a detailed design note for the Product-Kalman / correlated-PoE theory that grew out of the Sigma(hop) confirmatory manuscript discussion.

The note clarifies that:
- PoE can provide a lower-confidence / consensus point estimate of `mu`
- `Sigma(hop)` supplies the predicted-error covariance `V(hop)` needed by calibrated NLL
- a naive weighted blend of PoE and joint-covariance losses conflates mean estimation with error-geometry modeling
- a Product-Kalman update should be trained as correlated PoE, not assumed independent PoE

## Contents

- additive predicted-error / Kalman formulation
- PoE as a mean/prior mechanism
- product and geometric-mean error propagation
- log-space Product-Kalman update
- explanation of why PoE and joint covariance are complements rather than substitutes
- evaluation plan and guardrails
- cross-reference from the Sigma(hop) manuscript scaffold

## Validation

Docs-only change. Ran `git diff --cached --check`.

### DIFF
--- a/prototypes/mu_cosine/DESIGN_product_kalman_poe.md
+++ b/prototypes/mu_cosine/DESIGN_product_kalman_poe.md
@@ -240,6 +240,11 @@ where `C` projects the vector residual covariance into the scalar or vector prod
 measurement noise and linearization error. Treating `R_ell` as a free scalar while citing `V(hop)` would lose the
 bridge from the confirmatory result.
 
+Here `J` is the per-channel link Jacobian, such as `d log(mu) / d mu = 1 / mu` or
+`d logit(mu) / d mu = 1 / (mu * (1 - mu))`, so the clipping convention in the preconditions is part of the
+definition rather than cosmetic numerical hygiene. `C` may be the identity when `R_ell` is kept per-channel; otherwise
+it is the projection from linked residual channels into the fused product channel.
+
 This is a correlated PoE, not an independent PoE. The exponents, source weights, and covariance terms should be
 learned or calibrated on held-out node-disjoint data.
 
--- a/prototypes/mu_cosine/DESIGN_product_kalman_poe.md
+++ b/prototypes/mu_cosine/DESIGN_product_kalman_poe.md
@@ -58,6 +58,31 @@ mu_T = H mu_z
 
 where `H` may be a scalar weight vector for one observed channel or a matrix for multiple observed/judge channels.
 
+## Preconditions and implementation boundaries
+
+This is a design note, not an implementation recipe. The closed-form Kalman and Gaussian-PoE formulas below assume
+experts are Gaussian, or are being approximated as Gaussian in the chosen coordinates, and that shared evidence is
+represented in a covariance matrix rather than counted twice.
+
+For independent Gaussian expert densities over the same latent state, PoE reduces to information-form fusion:
+
+```text
+Lambda_fused = sum_j Lambda_j
+mu_fused     = Lambda_fused^-1 * sum_j Lambda_j mu_j
+Sigma_fused  = Lambda_fused^-1
+```
+
+where `Lambda_j = Sigma_j^-1`. This precision-summation identity is exact only under the Gaussian approximation and
+conditional independence of expert noise given the state. Correlated experts require a full joint covariance, a learned
+joint-channel model, or a conservative unknown-correlation method such as generalized/weighted PoE or covariance
+intersection. This note does not implement those mitigations; it points to the joint-channel Kalman variant as the
+preferred direction for this project.
+
+All covariance and precision matrices are assumed symmetric positive-definite after regularization. Implementations
+should clamp near-zero variances to avoid infinite-precision experts, use a finite-evidence convention for `log` and
+`logit` transforms such as `mu <- clip(mu, eps, 1 - eps)`, and preserve covariance validity with Joseph-form updates
+or an SPD projection when floating-point arithmetic nudges a matrix out of range.
+
 ## Additive predicted-error model
 
 The Gaussian/Kalman version treats the observed fuzzy label as:
@@ -97,10 +122,12 @@ The Kalman update follows from the same objects:
 ```text
 K       = P H^T (H P H^T + R)^-1
 mu_post = mu_prior + K (y - H mu_prior)
+P_post  = (I - K H) P (I - K H)^T + K R K^T
 ```
 
-The pseudoinverse appears as a limiting case with weak prior structure and simple noise. With correlated operators or
-correlated judge channels, the covariance-aware update is the relevant generalization.
+The `P_post` line is the Joseph-form covariance update. The pseudoinverse appears as a limiting case with weak prior
+structure and simple noise. With correlated operators or correlated judge channels, the covariance-aware update is the
+relevant generalization.
 
 ## Product-of-experts as a mean/prior mechanism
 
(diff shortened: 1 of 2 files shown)


prototypes/mu_cosine/DESIGN_product_kalman_poe.md
```diff
@@ -0,0 +1,316 @@
...
+The product version treats model and graph superpositions as two experts:
+
+```text
+g_prior = prod_i mu_model,i ^ alpha_i * prod_j mu_graph,j ^ beta_j
+```
+
+Replacing the model expert with a judge expert gives a measurement-like channel:
+
+```text
+g_meas = prod_i mu_judge,i ^ alpha_i * prod_j mu_graph,j ^ beta_j
```
s243a, line 143: **[non-blocking, design clarity]** `mu_graph` appears in both `g_prior` and `g_meas`, but the later Kalman update is written in the standard prior/measurement form. If these two channels share graph evidence, the prior and measurement errors are not independent, so a literal implementation would double-count that evidence.

**Suggested fix:** either remove the shared graph term from one side for the simple variant, or explicitly say this variant requires carrying prior-measurement cross-covariance / using the joint-channel formulation below.

prototypes/mu_cosine/DESIGN_product_kalman_poe.md
```diff
@@ -0,0 +1,316 @@
...
+```
+
+This can be useful, but it is a different object from `V_y`. In membership space `[0, 1]`, the product is an AND-like
+quantity and is at most 1. Values above 1 only make sense after moving to odds, likelihood ratios, precisions, or
+another positive evidence scale.
+
+PoE can therefore give a lower-confidence bound or consensus point estimate of membership, but it does not by itself
```
s243a, line 154: **[non-blocking, terminology/precondition]** This statement is true for the membership-space product as a heuristic `mu` aggregator, but formal Gaussian PoE / information-filter fusion does produce a fused covariance through precision addition. Since this document later introduces log-space Kalman objects (`P_ell`, `R_ell`), readers may conflate the two meanings of “PoE.”

**Suggested fix:** add a short note distinguishing “product of membership scores as a point estimate” from “product of expert densities.” If the Gaussian case is intended, state the precision-summation identity and the Gaussian/independence preconditions explicitly.

prototypes/mu_cosine/DESIGN_product_kalman_poe.md
```diff
@@ -0,0 +1,316 @@
...
+ell_prior = log g_prior
+ell_meas  = log g_meas
+```
+
+With scalar prior variance `P_ell` and measurement variance `R_ell`:
+
+```text
+K_ell    = P_ell / (P_ell + R_ell)
+ell_post = ell_prior + K_ell * (ell_meas - ell_prior)
+g_post   = exp(ell_post)
```
s243a, line 219: **[non-blocking, implementation gap]** The log-space update needs an explicit finite-domain convention. Membership products can be 0 or 1, so `log g`, `log x`, and later `logit(mu)` can become infinite or undefined; near-boundary values can also dominate the update and variance estimates.

**Suggested fix:** document the intended clipping epsilon, likelihood-ratio/evidence transform, or other finite-evidence convention before this becomes implementation guidance.

prototypes/mu_cosine/DESIGN_product_kalman_poe.md
```diff
@@ -0,0 +1,316 @@
...
+```
+
+In vector form:
+
+```text
+K       = P_ell H^T (H P_ell H^T + R_ell)^-1
+ell_post = ell_prior + K (ell_meas - H ell_prior)
```
s243a, line 226: **[non-blocking, missing covariance mapping]** The design introduces `P_ell` and `R_ell` for the product/log-evidence Kalman update, but it does not yet explain how the residual-space `V(hop)` from Sigma(hop) maps into `R_ell`. That mapping is the key bridge between the Sigma(hop) result and Product-Kalman variants 2/5.

**Suggested fix:** add a sentence or small equation saying whether `R_ell` is learned directly in log-evidence space, obtained by a delta-method/Jacobian transform from `V(hop)`, or treated as a separate covariance head.

prototypes/mu_cosine/DESIGN_product_kalman_poe.md
```diff
@@ -0,0 +1,316 @@
...
+   product supplies a measurement-like channel. Sigma(hop) or another covariance head supplies the predicted-error
+   terms that decide the update rate.
+3. **Joint direct-mu + PoE Kalman.** The observation vector includes the direct `mu` readout, the lower-product proxy,
+   the upper noisy-OR proxy, and any graph/judge channels. The covariance matrix `R` carries their correlations, so
+   the update can use PoE information without double-counting the direct model signal.
+
+A schematic joint-channel form is:
+
+```text
+s = [logit(mu_direct), log(mu_lower), logit(mu_upper), graph_features, judge_features]^T
```
s243a, line 268: **[non-blocking, coordinate consistency]** This schematic mixes `logit(mu_direct)`, `log(mu_lower)`, `logit(mu_upper)`, and raw graph/judge features under one linear-Gaussian observation equation. A constant linear `H` is only well-defined after the channels are put in compatible coordinates or explicitly modeled through per-channel transforms/Jacobians.

**Suggested fix:** clarify which channels are modeled as approximately Gaussian measurements of `ell`, how heterogeneous feature channels are scaled/embedded, and how `R` captures cross-channel covariance without double-counting shared upstream evidence.

prototypes/mu_cosine/DESIGN_product_kalman_poe.md
```diff
@@ -0,0 +1,316 @@
...
+## How Sigma(hop) fits
+
+Sigma(hop) is not a competing mean estimator. It is a conditional predicted-error model:
+
+```text
+V(hop) = [[sigma_D(hop)^2, rho(hop) sigma_D(hop) sigma_S(hop)],
+          [rho(hop) sigma_D(hop) sigma_S(hop), sigma_S(hop)^2]]
+```
+
+The confirmatory run showed that this `V(hop)` improves held-out residual NLL over a constant covariance baseline on
```
s243a, line 315: **[nit, readability]** The variables `x` and `y` are overloaded: earlier `(x, y)` names the entity pair, `y` names the observed fuzzy label, and here `x, y` are two generic random estimates. This makes the product-error formula easier to misread.

**Suggested fix:** rename the generic estimates here, e.g. `a` and `b`, or `u` and `v`, to avoid colliding with the entity pair and observation notation.

prototypes/mu_cosine/DESIGN_product_kalman_poe.md
s243a, line 226: **[non-blocking, numerical validity]** The Kalman examples show the gain and posterior mean, but not the posterior covariance/precision update or numerical validity requirements. For an implementation-facing design, readers need to know whether covariance matrices must be SPD/PSD, how near-singular matrices are handled, and what happens with zero-variance / infinite-precision experts.

**Suggested fix:** either state that the examples are mean-update-only pseudocode, or include the covariance/precision update and numerical guardrails for symmetric positive-definite covariance/precision matrices.

s243a (commented):
## Model-council review: Product-Kalman PoE design

Recommendation: needs discussion before relying on this as an implementation recipe, but not necessarily blocking as a docs-only design note.

Council summary:
- Claude Opus 4.8: approve with one substantive non-blocking concern about shared graph evidence appearing in both prior and measurement channels, which can double-count evidence unless cross-covariance is modeled.
- Claude Sonnet 5: approve with non-blocking documentation clarifications; strongest concerns were distinguishing membership-product PoE from Gaussian/log-space PoE, mixed coordinates under one linear measurement model, overloaded variables, and missing mapping from V(hop) to R_ell.
- GPT-5.5: request changes; considered the implicit Gaussian/independence/domain/numerical preconditions important enough that future implementers could misread the note.
- Deep research grounding: Product-of-Experts needs renormalization in general; Gaussian PoE is the tractable case where precision matrices add; ignoring correlation/conditional-dependence risks overconfident fused uncertainty.

My synthesis: the document’s core thesis is directionally sound: PoE can be treated as a mean/prior mechanism while Sigma(hop) or another covariance head supplies error geometry. I would not block the PR as a design note, because it repeatedly warns against independent PoE and double-counting. However, I recommend adding a few precise guardrails now, because the title and Kalman formulas could otherwise be read as a direct implementation recipe.

PR #3529, opened 2026-07-08
Add product-space transform diagnostics

s243a:
Adds the finite product-space transform checkpoint for the Product-Kalman / correlated-PoE design path.

Changes:
- Add `product_space.py` with clipped log/logit coordinates, lower/upper product proxies, interval width, and first-order Jacobians.
- Add `test_product_space.py` for closed-form transform and Jacobian checks.
- Add `test_product_space_monte_carlo.py`, a CPU-default diagnostic with `--device cuda` / `--device auto` support for larger Monte Carlo checks.
- Update `DESIGN_product_kalman_poe.md` to record this as the nonlinear-statistics sanity-check step.

Validation:
- `python3 prototypes/mu_cosine/test_product_space_monte_carlo.py`
- `python3 prototypes/mu_cosine/test_product_space_monte_carlo.py --device auto --samples 5000`
- `python3 -m py_compile prototypes/mu_cosine/product_space.py prototypes/mu_cosine/test_product_space.py prototypes/mu_cosine/test_product_space_monte_carlo.py`
- `git diff --check`

### DIFF
--- a/prototypes/mu_cosine/DESIGN_product_kalman_poe.md
+++ b/prototypes/mu_cosine/DESIGN_product_kalman_poe.md
@@ -356,8 +356,10 @@ calibration against both the naive-PoE controls and the additive/joint covarianc
 
 1. Keep the synthetic overconfidence check in `test_product_kalman_poe_synthetic.py` as the first guardrail:
    shared expert noise should make naive independent PoE report variance below empirical error.
-2. Implement a product-space transform that exposes `log_mu`, `logit_mu`, or likelihood-ratio coordinates explicitly.
-3. Add the noisy-OR upper proxy and track the width `mu_upper - mu_lower` as a disagreement diagnostic.
+2. Keep finite product-space transforms in `product_space.py`: clipped `log_mu`/`logit_mu`, lower/upper
+   product proxies, interval width, and first-order Jacobians for covariance propagation.
+3. Use `test_product_space_monte_carlo.py` as the local nonlinear-statistics sanity check: CPU by default,
+   optional CUDA for larger Monte Carlo runs.
 4. Fit scalar/vector product-Kalman updates with learned `P_ell` and `R_ell`.
 5. Compare against `JointPosterior` and Sigma-conditioned covariance on held-out node-disjoint splits.
 6. Only after the held-out comparison, decide whether this belongs in the training objective.
@@ -373,5 +375,7 @@ calibration against both the naive-PoE controls and the additive/joint covarianc
 - `DESIGN_transitive_relations.md` — predicted-distribution loss for transitive relations and covariance caveats.
 - `test_product_kalman_poe_synthetic.py` — runnable synthetic check that shared expert noise makes independent
   Gaussian PoE overconfident unless the full covariance is modeled.
+- `product_space.py`, `test_product_space.py`, and `test_product_space_monte_carlo.py` — finite product-space
+  transform helpers, closed-form Jacobian tests, and CPU/CUDA Monte Carlo checks for nonlinear covariance propagation.
 - `REPORT_sigma_hop_confirmatory.md` and `PAPER_sigma_hop_confirmatory.md` — confirmatory Sigma(hop) result and
   publication scaffold.
--- a/prototypes/mu_cosine/DESIGN_product_kalman_poe.md
+++ b/prototypes/mu_cosine/DESIGN_product_kalman_poe.md
@@ -357,9 +357,10 @@ calibration against both the naive-PoE controls and the additive/joint covarianc
 1. Keep the synthetic overconfidence check in `test_product_kalman_poe_synthetic.py` as the first guardrail:
    shared expert noise should make naive independent PoE report variance below empirical error.
 2. Keep finite product-space transforms in `product_space.py`: clipped `log_mu`/`logit_mu`, lower/upper
-   product proxies, interval width, and first-order Jacobians for covariance propagation.
+   product proxies, interval width, and first-order Jacobians for covariance propagation. *(Completed in PR #3529.)*
 3. Use `test_product_space_monte_carlo.py` as the local nonlinear-statistics sanity check: CPU by default,
-   optional CUDA for larger Monte Carlo runs.
+   optional CUDA for larger Monte Carlo runs. *(Completed in PR #3529; keep as a prototype diagnostic until a
+   Product-Kalman implementation depends on it enough to promote into CI.)*
 4. Fit scalar/vector product-Kalman updates with learned `P_ell` and `R_ell`.
 5. Compare against `JointPosterior` and Sigma-conditioned covariance on held-out node-disjoint splits.
 6. Only after the held-out comparison, decide whether this belongs in the training objective.
--- a/prototypes/mu_cosine/product_space.py
+++ b/prototypes/mu_cosine/product_space.py
@@ -1,0 +1,107 @@
+#!/usr/bin/env python3
+"""Finite product-space transforms for Product-Kalman / correlated-PoE prototypes.
+
+These helpers implement the clipping convention from `DESIGN_product_kalman_poe.md`.
+They are intentionally small and numpy-only: they expose stable log/logit coordinates,
+lower/upper product proxies, and link Jacobians without committing to a Kalman model.
+"""
+
+import numpy as np
+
+
+DEFAULT_EPS = 1e-6
+
+
+def clip_mu(mu, eps=DEFAULT_EPS):
+    """Clip membership values into the finite open interval used by log/logit links."""
+    if not 0.0 < eps < 0.5:
+        raise ValueError("eps must be in (0, 0.5)")
+    return np.clip(np.asarray(mu, dtype=float), eps, 1.0 - eps)
+
+
+def log_mu(mu, eps=DEFAULT_EPS):
+    """Finite log-membership evidence, `log(clip(mu))`."""
+    return np.log(clip_mu(mu, eps))
+
+
+def logit_mu(mu, eps=DEFAULT_EPS):
+    """Finite logit-membership evidence, `log(clip(mu)/(1-clip(mu)))`."""
+    m = clip_mu(mu, eps)
+    return np.log(m) - np.log1p(-m)
+
+
+def link_jacobian(mu, link="logit", eps=DEFAULT_EPS):
+    """Derivative d link(mu) / d mu at clipped `mu`.
+
+    Supported links:
+    - `log`:   d log(mu) / d mu = 1 / mu
+    - `logit`: d logit(mu) / d mu = 1 / (mu * (1 - mu))
+    """
+    m = clip_mu(mu, eps)
+    if link == "log":
+        return 1.0 / m
+    if link == "logit":
+        return 1.0 / (m * (1.0 - m))
+    raise ValueError(f"unknown link {link!r}; expected 'log' or 'logit'")
+
+
+def _weights_like(mu, weights):
+    m = np.asarray(mu, dtype=float)
+    if weights is None:
+        return np.ones_like(m, dtype=float)
+    w = np.asarray(weights, dtype=float)
+    if w.shape != m.shape:
+        raise ValueError(f"weights shape {w.shape} must match mu shape {m.shape}")
+    if np.any(w < 0):
+        raise ValueError("weights must be nonnegative")
+    return w
+
+
+def product_lower(mu, weights=None, eps=DEFAULT_EPS):
+    """AND-style lower-support proxy, `prod_i clip(mu_i)^w_i`."""
+    m = clip_mu(mu, eps)
+    w = _weights_like(m, weights)
+    return float(np.exp(np.sum(w * np.log(m))))
+
+
+def product_upper(mu, weights=None, eps=DEFAULT_EPS):
+    """Noisy-OR upper-support proxy, `1 - prod_i (1 - clip(mu_i))^w_i`."""
+    m = clip_mu(mu, eps)
+    w = _weights_like(m, weights)
+    return float(1.0 - np.exp(np.sum(w * np.log1p(-m))))
+
+
+def product_interval(mu, weights=None, eps=DEFAULT_EPS):
+    """Return `(lower, upper, width)` product proxies for a source vector."""
+    lo = product_lower(mu, weights, eps)
+    hi = product_upper(mu, weights, eps)
+    return lo, hi, hi - lo
+
+
+def product_lower_jacobian(mu, weights=None, eps=DEFAULT_EPS):
+    """Derivative of the lower product proxy with respect to clipped `mu`."""
+    m = clip_mu(mu, eps)
+    w = _weights_like(m, weights)
+    return product_lower(m, w, eps) * w / m
+
+
+def product_upper_jacobian(mu, weights=None, eps=DEFAULT_EPS):
+    """Derivative of the upper noisy-OR proxy with respect to clipped `mu`."""
+    m = clip_mu(mu, eps)
+    w = _weights_like(m, weights)
+    complement = np.exp(np.sum(w * np.log1p(-m)))
+    return complement * w / (1.0 - m)
+
+
+def product_interval_jacobian(mu, weights=None, eps=DEFAULT_EPS):
+    """Return Jacobian rows for `(lower, upper, width)` in clipped product space."""
+    lo_j = product_lower_jacobian(mu, weights, eps)
+    hi_j = product_upper_jacobian(mu, weights, eps)
+    return np.vstack([lo_j, hi_j, hi_j - lo_j])
+
+
+def normalized_geometric_weights(n):
+    """Equal exponents summing to one for softened geometric product proxies."""
+    if n <= 0:
+        raise ValueError("n must be positive")
+    return np.full(int(n), 1.0 / float(n))
--- a/prototypes/mu_cosine/product_space.py
+++ b/prototypes/mu_cosine/product_space.py
@@ -4,12 +4,40 @@
 These helpers implement the clipping convention from `DESIGN_product_kalman_poe.md`.
 They are intentionally small and numpy-only: they expose stable log/logit coordinates,
 lower/upper product proxies, and link Jacobians without committing to a Kalman model.
+
+Example:
+
+    mu = [0.25, 0.8]
+    weights = normalized_geometric_weights(len(mu))
+    lower, upper, width = product_interval(mu, weights)
+    J = product_interval_jacobian(mu, weights)
+
+The product proxies are AND/noisy-OR style diagnostics, not calibrated confidence
+intervals. Under correlated experts the true posterior can fall outside the proxy
+interval unless a downstream model calibrates the dependence structure.
 """
 
 import numpy as np
 
 
 DEFAULT_EPS = 1e-6
+
+__all__ = [
+    "DEFAULT_EPS",
+    "clip_mu",
+    "link_jacobian",
+    "log_mu",
+    "logit_mu",
+    "normalized_geometric_weights",
+    "product_interval",
+    "product_interval_jacobian",
+    "product_lower",
+    "product_lower_jacobian",
+    "product_lower_log",
+    "product_upper",
+    "product_upper_complement_log",
+    "product_upper_jacobian",
+]
 
 
 def clip_mu(mu, eps=DEFAULT_EPS):
@@ -45,8 +73,17 @@ def link_jacobian(mu, link="logit", eps=DEFAULT_EPS):
     raise ValueError(f"unknown link {link!r}; expected 'log' or 'logit'")
 
 
-def _weights_like(mu, weights):
+def _source_vector(mu):
     m = np.asarray(mu, dtype=float)
+    if m.ndim != 1:
+        raise ValueError("product proxies expect a 1-D source vector; batch matrices are not supported")
+    if m.size == 0:
+        raise ValueError("product proxies require at least one source")
+    return m
+
+
+def _weights_like(mu, weights):
+    m = _source_vector(mu)
     if weights is None:
         return np.ones_like(m, dtype=float)
     w = np.asarray(weights, dtype=float)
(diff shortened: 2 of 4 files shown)


s243a (commented):
## Academic Code Review — PR #3529: Add product-space transform diagnostics

**Summary**

This PR introduces a self-contained finite product-space transform layer (`product_space.py`) along with two test modules — a closed-form unit test suite (`test_product_space.py`) and a Monte Carlo covariance propagation diagnostic (`test_product_space_monte_carlo.py`) — as an intermediate checkpoint on the Product-Kalman / correlated-PoE design path. The design is intentionally narrow: numpy-only, no model commitment, minimal surface area. Overall the contribution is coherent and of high quality. The review below is structured into **(1) Correctness & Mathematics**, **(2) API & Interface Design**, **(3) Testing & Validation**, **(4) Documentation**, and **(5) Minor / Nits**.

---

### 1. Correctness & Mathematics

#### 1.1 Double-clipping in `product_lower_jacobian` and `product_upper_jacobian`

`product_lower_jacobian` calls `clip_mu` on `mu`, then passes the already-clipped array `m` to `product_lower(m, w, eps)`. `product_lower` internally calls `clip_mu` again — harmless numerically (idempotent), but it does allocate a second copy.

More importantly: the Jacobian formula is derived as
```
d/d mu_i  prod_j mu_j^{w_j}  =  (prod_j mu_j^{w_j}) * w_i / mu_i
```
The denominator `mu_i` here must be the *clipped* value used in the product, not the raw input. The current code uses `m` (clipped) in the denominator — **correct**. However, `product_lower(m, w, eps)` re-clips `m` through `clip_mu`, so the numerator product and the denominator are evaluated at the *same* clipped point — also consistent. No bug, but the redundant clipping is worth flagging for clarity.

**Recommendation:** Compute `product_lower(m, w, eps=0.0)` or accept a private `_no_clip` path to make it explicit that you are intentionally using pre-clipped inputs, and add a comment to that effect.

#### 1.2 `product_upper_jacobian` sign convention

The noisy-OR upper proxy is:
```
U(mu) = 1 - prod_i (1 - mu_i)^{w_i}
```
Its partial derivative is:
```
dU/d mu_i = w_i * (1-mu_i)^{w_i - 1} * prod_{j≠i} (1-mu_j)^{w_j}
           = prod_j (1-mu_j)^{w_j} * w_i / (1 - mu_i)
           = complement * w_i / (1 - mu_i)
```
The implementation matches. One subtlety: when `w_i = 0` and `mu_i = 1 - eps`, the denominator `(1 - m_i)` is `eps`, which is fine because `w_i = 0` zeroes the term. But if a caller passes `weights` with some zeros paired with `mu` values very close to 1, the result is `0/eps` which is `0.0` — correct but potentially confusing. Worth a docstring note.

#### 1.3 `product_interval_jacobian` width row

The width Jacobian row is `hi_j - lo_j`, which is the Jacobian of `upper - lower` with respect to `mu`. This is mathematically correct by linearity of differentiation. However, calling `product_lower_jacobian` and `product_upper_jacobian` independently incurs two separate `clip_mu` calls and two separate evaluations of `product_lower`/`product_upper`. A small efficiency improvement is possible by computing them jointly, but for a prototype diagnostic this is acceptable.

#### 1.4 Monte Carlo tolerance `rel_tol=0.18` is generous

With 30,000 samples, float64 arithmetic, and sigmas in the range [0.012, 0.018], the delta-method approximation should be quite tight (well under 5% relative error for smooth nonlinear transforms in the small-noise regime). The default tolerance of **18%** is much wider than the empirical accuracy. This is not wrong, but it means the test will pass even with a noticeably incorrect Jacobian.

**Recommendation:** Consider tightening to `rel_tol=0.05` at 30k samples, or documenting why 18% is the appropriate bar (e.g., to account for boundary effects when `mu` gets clamped).

#### 1.5 `test_weighted_product_proxies_support_geometric_mean` — upper-proxy tolerance

```python
assert abs(product_upper(mu, weights, eps=1e-9) - 1.0) < 5e-5
```
`mu = [0.25, 1.0]`, but with `eps=1e-9`, the second value clips to `1 - 1e-9`. The exact upper proxy value is:
```
1 - (0.75)^0.5 * (1e-9)^0.5  ≈  1 - 8.66e-5
```
So `|upper - 1.0| ≈ 8.66e-5`, which **exceeds** the asserted `5e-5`. This test may be relying on a slightly different arithmetic path. You should verify this assertion analytically or tighten `eps` further.

---

### 2. API & Interface Design

#### 2.1 `clip_mu` returns a numpy array from scalar input silently

```python
clip_mu(0.5)  # returns np.float64, not float
```
Not a bug, but users who pass a scalar `mu` to `log_mu` or `logit_mu` will get a 0-d numpy array. For downstream callers building Kalman update matrices this is harmless, but worth noting in the module docstring.

#### 2.2 `product_lower` / `product_upper` return `float`, but Jacobians return `np.ndarray`

The scalar-return convention for the proxy values is asymmetric with the array-return for Jacobians. This is reasonable (scalars are natural for scalar aggregates), but it means `product_lower_jacobian` calls `product_lower(m, w, eps)` and wraps the result in a `float(...)` internally — fine, but a type annotation would help downstream static analysis.

#### 2.3 No `__all__` export list

Since this module is intentionally small and used as a library by the test files, an `__all__` list would make the public API explicit and prevent future accidental exposure of private helpers like `_weights_like`.

#### 2.4 `normalized_geometric_weights` does not validate float `n`

```python
normalized_geometric_weights(2.7)  # returns array of length 2, silently truncates
```
The `int(n)` cast swallows non-integer inputs. A `isinstance(n, int)` check or an explicit `if n != int(n): raise ValueError(...)` guard would be more robust.

---

### 3. Testing & Validation

#### 3.1 Test runner is not pytest-compatible as written

The test files use a manual `if __name__ == "__main__"` runner. They will also work fine with `pytest` (each `test_*` function is independently discoverable), which is good. However, the `test_invalid_weights_and_links_fail_fast` tests use bare `try/except/else` patterns. Consider switching to `pytest.raises` for clearer intent and error messages, or at minimum add `assert isinstance(exc, ValueError)` inside the `except` blocks to prevent accidentally catching other exception types.

#### 3.2 No test for `normalized_geometric_weights(1)`

The edge case `n=1` should return `[1.0]`. This is covered implicitly by `normalized_geometric_weights(2)` in `test_weighted_product_proxies_support_geometric_mean` but not the `n=1` boundary. Worth adding, given the `n <= 0` guard.

#### 3.3 Monte Carlo test: `upper_complement` variable computed but not used for interval check

In `test_product_space_monte_carlo.py`:
```python
upper_complement0 = (1.0 - mean).prod()
upper_j = upper_complement0 / (1.0 - mean)
```
The variable `upper_complement0` is used only in computing `upper_j`, and is not independently validated. A brief assertion (e.g., that `lower0 + upper_complement0 <= 1 + eps`) could serve as a useful sanity check on the geometry of the proxy bounds.

#### 3.4 No integration test with `test_product_kalman_poe_synthetic.py`

The design document records this PR as a prerequisite step before the Product-Kalman fitting. It would be valuable to add at least a smoke test verifying that `product_space.py` functions can be imported and called from the synthetic test context, confirming no module path or namespace issues.

---

### 4. Documentation

#### 4.1 Design doc update is accurate and helpful

The rewording of steps 2–3 from forward-looking imperatives to present-tense descriptions of what now exists is exactly the right approach for a living design document. The cross-reference list addition is complete. No issues here.

#### 4.2 Module docstring could name the PoE connection more explicitly

The `product_space.py` module docstring says "without committing to a Kalman model" — excellent framing. However, it doesn't mention the connection to correlated Product-of-Experts (PoE) or the broader `mu_cosine` graph-scoring context. A reader who discovers this file in isolation will lack orientation. One additional sentence would help:
```
# These transforms appear in the covariance propagation step of the correlated-PoE
# path described in DESIGN_product_kalman_poe.md.
```

#### 4.3 `test_product_space_monte_carlo.py` tolerance flag underdocumented

The `--rel-tol` argument description reads only `"default: 0.18"`. Because this flag controls the pass/fail threshold for a statistical validation, it deserves a brief explanation of what relative tolerance means in this context, e.g., `"Maximum allowed relative error between empirical and delta-method covariance (default: 0.18 = 18%)."`

---

### 5. Minor / Nits

- **`logit_mu` uses `np.log(m) - np.log1p(-m)` instead of `np.log(m / (1 - m))`** — the former is numerically superior (avoids catastrophic cancellation near `m=1`), which is correct and worth a brief inline comment to prevent future "simplification".
- **`DEFAULT_EPS = 1e-6` vs. `eps=1e-9` in tests** — using a stricter `eps` in tests than the module default is fine, but the mismatch in `test_weighted_product_proxies_support_geometric_mean` (see §1.5) suggests the interplay between `eps` and tolerance should be documented.
- **`resolve_device` in Monte Carlo test uses `name.startswith("cuda")`** — this would match hypothetical future device strings like `"cuda_fp8"`. Consider `name == "cuda" or name.startswith("cuda:")` for future-safety.
- **PR description validation steps** include `python3 -m py_compile` — good hygiene. Consider adding it as a pre-commit hook or CI step to avoid regressing.

---

### Overall Assessment

The implementation is mathematically correct (with the minor concern in §1.5 worth verifying), well-scoped, and represents a clean checkpoint in the Product-Kalman design path. The primary actionable items before merging are:

1. **Verify the upper-proxy tolerance in `test_weighted_product_proxies_support_geometric_mean`** (§1.5) — this may be a latent test failure.
2. **Tighten or document `rel_tol=0.18`** (§1.4) to ensure the Monte Carlo check is meaningful.
3. **Add `__all__`** (§2.3) to make the public API explicit.
4. **Guard `normalized_geometric_weights` against non-integer input** (§2.4).

All other points are suggestions for improvement, not blockers.

s243a (commented):
## Second Academic Code Review — PR #3529: Add product-space transform diagnostics

*This is an independent second-pass review focusing on aspects not covered or deserving deeper treatment beyond the first review.*

---

### A. Numerical Stability & Floating-Point Analysis

#### A.1 `logit_mu` implementation is superior but undocumented

```python
def logit_mu(mu, eps=DEFAULT_EPS):
    m = clip_mu(mu, eps)
    return np.log(m) - np.log1p(-m)
```

This is strictly better than `np.log(m / (1 - m))` because `np.log1p(-m)` avoids cancellation when `m` is close to 1. However, there is a subtlety: after clipping, `m` is at most `1 - eps`, so `1 - m >= eps = 1e-6`. The `log1p` trick has maximum benefit when the argument is close to -1 (i.e., `m` close to 0), specifically because `log1p(x)` is more accurate than `log(1+x)` for small `x`. For the clipped domain `[eps, 1-eps]`, `log(m)` is the dominant precision concern at the lower boundary, not `log1p(-m)` at the upper boundary. The current implementation is still correct and not worse — but the code comment rationale, if added, should note the lower boundary is the actual precision-sensitive case.

#### A.2 `product_lower` precision degrades silently with large `n`

The log-sum-exp pattern used:
```python
np.exp(np.sum(w * np.log(m)))
```
is numerically sound for moderate `n`. But as `n` grows and `mu_i` values are small (say, all near `eps = 1e-6`), the sum `w * log(m)` accumulates toward `-infinity`, and the result underflows to `0.0` silently. This is mathematically correct (the product of many small numbers is essentially 0), but the Jacobian computed afterward divides by `m`, which remains `eps`. The ratio `product_lower / eps` can produce a very large gradient signal where the true gradient should be near-zero in practical terms.

**Recommendation:** Add a note in the docstring that the Jacobian is only interpretable when `product_lower` is not numerically zero. Optionally, return `np.nan` or raise if the result underflows, rather than silently propagating a potentially misleading gradient.

#### A.3 Monte Carlo: `mu.clamp(1e-6, 1.0 - 1e-6)` uses a different `eps` than `DEFAULT_EPS`

In `test_product_space_monte_carlo.py`:
```python
mu = mu.clamp(1e-6, 1.0 - 1e-6)
```
This matches `DEFAULT_EPS = 1e-6` — consistent. However, the Jacobian predictions are computed at `mean`, not at `clipped_mean`. If any component of `mean` falls outside `[1e-6, 1-1e-6]`, the predicted covariance would be computed at an unclipped point while the empirical covariance is computed on clipped samples. With the current hardcoded `mean = [0.31, 0.55, 0.74]` this is fine, but it is a latent correctness hazard for future parameter changes.

**Recommendation:** Compute Jacobians at `clamp(mean, 1e-6, 1-1e-6)` rather than raw `mean` in the Monte Carlo test.

---

### B. Statistical Design of the Monte Carlo Diagnostic

#### B.1 Single fixed seed makes the test deterministic but non-robust

`--seed 17` is hardcoded as the default. A single seed means the test can pass for seed 17 but fail for seeds that produce more extreme samples. For a prototype checkpoint this is acceptable, but worth noting explicitly in the module docstring: "Results are seed-dependent; this is not a coverage guarantee."

#### B.2 The delta-method approximation is first-order only

The Monte Carlo test validates the first-order (delta-method) covariance propagation:
```
Cov[f(mu)] ≈ J * Cov[mu] * J^T
```
This approximation is valid in the small-noise regime where `sigma << mean * (1 - mean)`. With `sigmas = [0.018, 0.015, 0.012]` and means near 0.3–0.74, the signal-to-noise ratio is large and the approximation is excellent. The test therefore validates the Jacobian implementation, not a general nonlinear covariance claim. This is correctly noted in the docstring (`"small-noise regime"`), but the chosen `mean = [0.31, 0.55, 0.74]` with moderate `rho = 0.35` gives a specific correlated structure. A second test case with higher correlation (`rho = 0.9`) or with means closer to 0 or 1 would probe the delta-method break-down regime.

#### B.3 `empirical_cov` divides by `n-1` (Bessel-corrected) but predicted uses population formula

```python
return centered.T @ centered / float(x.shape[0] - 1)  # sample covariance
```

The predicted covariance `log_pred = log_j[:, None] * cov * log_j[None, :]` uses the population covariance `cov = source_covariance(sigmas, rho)` — not a sample estimate. Comparing a sample covariance estimator against a population covariance target introduces a bias of `1/n` which is negligible at 30k samples but is worth acknowledging. At 1000 samples (the minimum allowed), the bias is `0.1%` — still within the 18% tolerance, but the argument is weak if tolerance is ever tightened.

---

### C. Architecture & Dependency Concerns

#### C.1 `product_space.py` is numpy-only but `test_product_space_monte_carlo.py` requires torch

This is by design (the module docstring says "numpy-only") and the Monte Carlo test gracefully skips when torch is absent:
```python
except ModuleNotFoundError:
    print("skipped: torch is not installed")
    sys.exit(0)
```
The `sys.exit(0)` causes pytest to interpret the test file as having **passed** (exit code 0). This is intentional but means CI will not flag a missing torch installation as a problem. If torch is a required dependency for the full test suite, consider `sys.exit(77)` (the POSIX "skipped" convention) or `pytest.skip()` to make the skip semantics explicit.

#### C.2 `test_product_space.py` imports from `product_space` without `sys.path` manipulation

Both test files import `from product_space import ...` using a bare module name. This works only when executed from the `prototypes/mu_cosine/` directory or when that directory is on `sys.path`. Running `pytest` from the repo root will fail unless a `conftest.py` or `pyproject.toml` `pythonpath` setting is present.

**Recommendation:** Add a `conftest.py` in `prototypes/mu_cosine/` with:
```python
import sys, pathlib
sys.path.insert(0, str(pathlib.Path(__file__).parent))
```
or use a relative import structure (`from .product_space import ...`) if a package `__init__.py` is added.

#### C.3 No `requirements.txt` or dependency declaration for the `prototypes/` subtree

The prototype uses numpy (always present) and torch (optional, gracefully skipped). There is no `requirements.txt` or `pyproject.toml` in `prototypes/mu_cosine/`. Downstream contributors won't know the numpy version constraint. Given that `np.log1p` and `np.clip` have been stable since numpy 1.x, this is not an urgent concern, but a `requirements-prototype.txt` with `numpy>=1.20` would be a low-cost improvement.

---

### D. Gaps in Test Coverage

| Scenario | Covered? | Notes |
|---|---|---|
| `clip_mu` scalar input | ✅ | Implicitly via `log_mu`/`logit_mu` tests |
| `clip_mu` out-of-range eps | ❌ | `eps=0.6` should raise `ValueError`; not tested |
| `product_lower` with all-unity weights | ✅ | Default weights path |
| `product_lower` large `n` underflow | ❌ | Not tested; see §A.2 |
| `normalized_geometric_weights(1)` | ❌ | Edge case `n=1` |
| `normalized_geometric_weights` float input | ❌ | Silently truncates; see prior review §2.4 |
| `product_interval_jacobian` weight path | ❌ | Only unweighted case tested |
| `link_jacobian` with `link="log"` edge values | ❌ | `mu=0.0` with `log` link only tested via `logit` |
| `resolve_device("auto")` with no CUDA | ❌ | Only testable in a CPU-only environment |

**Priority additions:** `eps` validation test for `clip_mu`, and the weighted `product_interval_jacobian` path.

---

### E. Design-Level Observations

#### E.1 Width Jacobian encodes disagreement, not uncertainty

The width `mu_upper - mu_lower` is framed as a "disagreement diagnostic" in the design doc. This is a meaningful interval, but the Jacobian of the width has an interesting property: `d(width)/d(mu_i)` can be negative for some `mu_i`. Specifically:
```
d(U - L)/d mu_i = [upper_complement / (1-mu_i) - product_lower / mu_i] * w_i
```
For large `mu_i` close to 1, the `product_lower / mu_i` term dominates and the width *decreases* as `mu_i` increases. This means "increasing confidence in source i" can paradoxically narrow the disagreement interval. This is mathematically expected (higher lower bound closes the gap faster than upper bound rises) but is non-obvious to a reader using width as a diagnostic. A note in `product_interval_jacobian`'s docstring would prevent misinterpretation.

#### E.2 No inverse transforms

The module provides forward transforms (mu → log/logit/product space) and their Jacobians, but no inverse transforms. For a Kalman update, measurements in log/logit space need to be mapped back to mu space. This is clearly out of scope for this checkpoint PR, but a `# TODO: inverse transforms (logistic, exp) for Kalman measurement update` comment would help future contributors understand the intended completion.

---

### Overall Assessment (Second Pass)

The implementation holds up under deeper scrutiny. The priority concerns from this pass are:

1. **`sys.exit(0)` on missing torch masks skipped tests in CI** (§C.1) — low effort to fix, high CI hygiene value.
2. **Bare import path for `product_space`** (§C.2) — will cause `pytest` failures from repo root without a `conftest.py`.
3. **Jacobians computed at unclipped `mean` in Monte Carlo test** (§A.3) — latent correctness hazard for future parameter changes.
4. **Width Jacobian sign behavior** (§E.1) — worth documenting to prevent diagnostic misinterpretation.

The PR is appropriate to merge as a prototype checkpoint with these issues tracked.

s243a (commented):
## Third Academic Code Review — PR #3529: Add product-space transform diagnostics
*(Sonnet 5 pass — independent review focusing on aspects not yet covered)*

---

### F. Semantic Correctness of the Product Proxy Formulations

#### F.1 The AND/noisy-OR framing is asymmetric in a potentially misleading way

`product_lower` is documented as an "AND-style lower-support proxy" and `product_upper` as a "Noisy-OR upper-support proxy." These are conceptually dual operations, but their weighted generalizations are not symmetric duals in the same sense:

- Lower: `prod_i mu_i^{w_i}` — a weighted geometric mean of memberships
- Upper: `1 - prod_i (1-mu_i)^{w_i}` — a weighted noisy-OR on complement space

For uniform weights summing to 1 (`w_i = 1/n`), the lower proxy is the geometric mean and the upper proxy is `1 - geometric_mean(1 - mu_i)`. These are **not** lower and upper bounds on the same quantity unless a specific probabilistic model (e.g., independent Bernoulli) is assumed. The docstrings present them as bounds without stating the model assumption. This could mislead a reader who uses `product_interval` as a credible interval.

**Recommendation:** Add a note such as: *"The lower/upper pair brackets the independent-Bernoulli PoE posterior, not a general credible interval. Under correlated experts, the true posterior may lie outside this interval."*

#### F.2 `normalized_geometric_weights` name implies normalization but the weights sum to 1 only for the exponent convention

In the standard geometric mean `(prod_i x_i)^{1/n}`, weights sum to 1 and the result is in `[min, max]`. In `product_lower`, the weight enters as the exponent: `prod_i mu_i^{w_i}`. When `sum(w_i) = 1`, this is the geometric mean. When `sum(w_i) != 1`, the proxy has a different scale. The function `normalized_geometric_weights` always makes `sum(w_i) = 1`, but the name "normalized" doesn't clarify *what* is normalized or *why* that normalization matters. A docstring explaining the geometric-mean interpretation would help.

---

### G. Interface Consistency and Composability

#### G.1 Inconsistent handling of 0-d arrays vs scalars in proxy functions

`product_lower` and `product_upper` both end with `return float(...)`, coercing the result to a Python scalar. But `product_lower_jacobian` returns a numpy array. If a caller passes a length-1 array `mu = np.array([0.5])`, they get:

```python
product_lower(np.array([0.5]))   # → float: 0.5
product_lower_jacobian(np.array([0.5]))  # → np.array([2.0])
```

This asymmetry is benign for scalars but will surface if these functions are embedded in a vectorized pipeline (e.g., batched Kalman updates over graph nodes). The interval functions return tuples of floats, while the interval Jacobian returns a `(3, d)` matrix. The API would be more composable if the proxy functions returned numpy scalars (`np.float64`) rather than Python `float`, keeping everything in the numpy type system.

#### G.2 No batch/vectorized path for multiple source vectors

All functions operate on a single source vector `mu` of shape `(d,)`. In the Product-Kalman context, you will likely want to compute product proxies over a batch of `(n, d)` membership matrices — one row per node or time step. The current API requires a Python loop, which is inefficient and non-idiomatic for numpy.

This is explicitly out of scope for a prototype checkpoint, but a `# Note: functions accept 1-D mu only; batch support is future work` comment in the module docstring would prevent a future contributor from accidentally passing a 2-D array and getting silently wrong results (numpy broadcasting would produce an incorrect scalar from a 2-D `np.sum`).

#### G.3 `_weights_like` is called twice in `product_lower_jacobian`

```python
def product_lower_jacobian(mu, weights=None, eps=DEFAULT_EPS):
    m = clip_mu(mu, eps)
    w = _weights_like(m, weights)
    return product_lower(m, w, eps) * w / m
```

`product_lower(m, w, eps)` internally calls `_weights_like` again on `m` and the already-validated `w`. Since `w` is already a numpy array of correct shape and nonnegative values, the second `_weights_like` call just re-validates. The same applies to `product_upper_jacobian`. Consider a private `_product_lower_raw(m, w)` that skips validation, or simply inline the scalar computation to avoid redundant validation.

---

### H. Test Isolation and Hermetic Design

#### H.1 Tests are not hermetic — they depend on execution directory

As noted in the prior review (§C.2), bare imports require running from `prototypes/mu_cosine/`. This also means the test files cannot be discovered by a standard `pytest` invocation from the repo root without configuration. However, there is an additional issue: if `product_space.py` is ever installed as a package (e.g., via a future `pip install -e .`), the local relative import would shadow the installed version. Hermetic test design using `importlib` or package-relative imports would prevent this class of confusion.

#### H.2 `test_product_space.py` has a test ordering dependency risk

The manual `__main__` runner iterates tests in **sorted name order**. Pytest discovers them in **definition order** by default. This divergence means a future test that passes when run in sorted order but fails in definition order (due to shared mutable state) would go undetected in CI if pytest is used. The current tests are all stateless, so there is no present bug — but a comment noting that tests must remain stateless would be a useful guard.

#### H.3 `test_product_lower_and_upper_match_manual_formulas` uses exact `mu = [0.2, 0.5, 0.8]`

The test asserts:
```python
assert abs(product_lower(mu) - (0.2 * 0.5 * 0.8)) < 1e-12
```
With `DEFAULT_EPS = 1e-6`, none of these values are clipped, so this is an exact comparison of the log-sum-exp path against direct multiplication. The tolerance `1e-12` is appropriate for float64, but the comment should note that clipping is not active here. A second case with boundary-near values (e.g., `mu = [1e-5, 0.5, 1 - 1e-5]`) would give better coverage of the clip-active path.

---

### I. Documentation and Knowledge-Transfer Quality

#### I.1 The DESIGN doc update converts future imperatives to present tense correctly — but loses traceability

The original steps 2–3 in `DESIGN_product_kalman_poe.md` were written as work-to-be-done:
> 2. Implement a product-space transform...
> 3. Add the noisy-OR upper proxy...

The PR rewrites them as accomplished facts. This is correct for a living design document, but it erases the record of *when* and *why* the design decision was made. Consider adding a completion date or a cross-reference to the PR number:
> 2. *(Completed in PR #3529)* Keep finite product-space transforms in `product_space.py`...

This is especially valuable in a prototype that may be revisited months later.

#### I.2 No module-level example in `product_space.py`

For a helper library intended to be used by future Kalman update code, a `Examples` section in the module docstring (or a simple `if __name__ == "__main__"` demo block) would significantly reduce the ramp-up time for new contributors. Even a three-line example:

```python
# mu = np.array([0.3, 0.6, 0.8])
# lo, hi, w = product_interval(mu)
# J = product_interval_jacobian(mu)  # shape (3, 3)
```

would make the intended usage immediately clear.

#### I.3 `--device auto` is underdocumented in the CLI help

The `--device` help string reads: `"cpu, cuda, cuda:0, or auto; default: cpu"`. The behavior of `auto` (fall back to CPU if CUDA unavailable) is only described in the module docstring, not in the argparse help. A user running `--help` in a CUDA environment won't know that `auto` is safe to use without CUDA.

---

### J. Structural Risk: Prototype vs. Production Boundary

#### J.1 `prototypes/` path implies experimental status, but tests are rigorous enough to be production-grade

The closed-form Jacobian tests and the Monte Carlo diagnostics are carefully written and would be appropriate in a production test suite. Placing them under `prototypes/` creates ambiguity: will these be promoted to `src/` eventually, or abandoned? A brief note in the PR description or in the design doc about the intended promotion path (or explicit abandonment criteria) would clarify the lifecycle expectation.

#### J.2 No CI configuration references these test files

The PR adds three runnable files but does not update any CI configuration (`.github/workflows/`, `Makefile`, `tox.ini`, etc.) to run them automatically. The validation section in the PR description lists manual `python3` commands. For a prototype checkpoint this is acceptable, but the risk is that future changes silently break these tests without anyone noticing. Even a comment in the PR body like *"CI integration deferred to promotion PR"* would set clear expectations.

---

### Priority Items from This Pass

1. **Document the independent-Bernoulli assumption** behind the lower/upper bound framing (§F.1) — this is a correctness-of-interpretation issue, not just a style point.
2. **Add a `# batch not supported` note** to the module docstring (§G.2) to prevent silent errors from 2-D array inputs.
3. **Add boundary-near test case** in `test_product_lower_and_upper_match_manual_formulas` to exercise the clip-active path (§H.3).
4. **Document the `auto` device fallback behavior** in the argparse help string (§I.3).
5. **Record PR number in design doc** step completions (§I.1) for future traceability.

PR #3530, opened 2026-07-08
Add Product-Kalman update core

s243a:
Adds the scalar/vector Gaussian-conditioning core for Product-Kalman experiments.

Changes:
- Add `product_kalman.py` with:
  - scalar Product-Kalman update in linked evidence coordinates
  - vector Gaussian conditioning with optional prior-measurement cross-covariance
  - residual/error covariance fitting from calibration residuals
  - Gaussian NLL helper for later held-out comparisons
- Add `test_product_kalman.py` covering scalar/vector gains, cross-covariance behavior, residual covariance fitting, invalid covariance rejection, and NLL sanity checks.
- Update `DESIGN_product_kalman_poe.md` to make this the core update primitive before empirical held-out comparisons.

Validation:
- `python3 prototypes/mu_cosine/test_product_kalman.py`
- `python3 -m pytest -s prototypes/mu_cosine/test_product_kalman.py`
- `python3 -m py_compile prototypes/mu_cosine/product_kalman.py prototypes/mu_cosine/test_product_kalman.py`
- `git diff --check`

### DIFF
--- a/prototypes/mu_cosine/product_kalman.py
+++ b/prototypes/mu_cosine/product_kalman.py
@@ -1,0 +1,205 @@
+#!/usr/bin/env python3
+"""Gaussian conditioning core for Product-Kalman / correlated-PoE prototypes.
+
+The functions here operate in an already chosen evidence coordinate, such as
+`log(mu_lower)`, `logit(mu_direct)`, or another finite product-space link. They do
+not construct product proxies themselves; use `product_space.py` for that layer.
+
+This module deliberately exposes cross-covariance between the prior state error and
+measurement noise. When prior and measurement channels share evidence, treating that
+cross term as zero is an assumption, not a default fact.
+"""
+
+from dataclasses import dataclass
+
+import numpy as np
+
+
+__all__ = [
+    "GaussianUpdate",
+    "fit_error_covariance",
+    "fit_residual_covariance",
+    "gaussian_condition_update",
+    "gaussian_nll",
+    "regularize_covariance",
+    "scalar_product_kalman_update",
+]
+
+
+@dataclass(frozen=True)
+class GaussianUpdate:
+    """Result of one Gaussian conditioning / Kalman-style update."""
+
+    mean: np.ndarray
+    covariance: np.ndarray
+    gain: np.ndarray
+    innovation: np.ndarray
+    innovation_covariance: np.ndarray
+    cross_covariance: np.ndarray
+
+
+def _as_vector(name, value):
+    arr = np.asarray(value, dtype=float)
+    if arr.ndim == 0:
+        arr = arr.reshape(1)
+    if arr.ndim != 1:
+        raise ValueError(f"{name} must be a 1-D vector or scalar")
+    if arr.size == 0:
+        raise ValueError(f"{name} must be nonempty")
+    if not np.isfinite(arr).all():
+        raise ValueError(f"{name} must be finite")
+    return arr
+
+
+def _as_matrix(name, value, rows=None, cols=None):
+    arr = np.asarray(value, dtype=float)
+    if arr.ndim == 0:
+        arr = arr.reshape(1, 1)
+    if arr.ndim != 2:
+        raise ValueError(f"{name} must be a 2-D matrix or scalar")
+    if rows is not None and arr.shape[0] != rows:
+        raise ValueError(f"{name} has {arr.shape[0]} rows, expected {rows}")
+    if cols is not None and arr.shape[1] != cols:
+        raise ValueError(f"{name} has {arr.shape[1]} columns, expected {cols}")
+    if not np.isfinite(arr).all():
+        raise ValueError(f"{name} must be finite")
+    return arr
+
+
+def _square_covariance(name, value, dim):
+    cov = _as_matrix(name, value, dim, dim)
+    return 0.5 * (cov + cov.T)
+
+
+def regularize_covariance(covariance, jitter=1e-9, name="covariance"):
+    """Return a symmetric positive-definite covariance with diagonal jitter.
+
+    Small negative eigenvalues from floating-point symmetry error are lifted. A
+    materially negative eigenvalue raises instead of silently accepting an invalid
+    covariance model.
+    """
+    if jitter < 0:
+        raise ValueError("jitter must be nonnegative")
+    cov = _as_matrix(name, covariance)
+    if cov.shape[0] != cov.shape[1]:
+        raise ValueError(f"{name} must be square")
+    cov = 0.5 * (cov + cov.T)
+    eig = np.linalg.eigvalsh(cov)
+    min_eig = float(eig[0])
+    floor = max(float(jitter), 1e-12)
+    if min_eig < -100.0 * floor:
+        raise ValueError(f"{name} is not positive semidefinite; min eigenvalue={min_eig:g}")
+    lift = max(float(jitter) - min_eig, 0.0)
+    if lift:
+        cov = cov + np.eye(cov.shape[0]) * lift
+    return cov
+
+
+def fit_residual_covariance(residuals, shrinkage=0.0, jitter=1e-9):
+    """Estimate residual covariance from rows of calibration residuals.
+
+    `shrinkage=1` keeps only the diagonal; `shrinkage=0` keeps the empirical full
+    covariance. Use residuals from an appropriate calibration split, not the data
+    used for the claimed evaluation.
+    """
+    if not 0.0 <= shrinkage <= 1.0:
+        raise ValueError("shrinkage must be in [0, 1]")
+    r = np.asarray(residuals, dtype=float)
+    if r.ndim == 1:
+        r = r.reshape(-1, 1)
+    if r.ndim != 2:
+        raise ValueError("residuals must be a 1-D or 2-D array")
+    if r.shape[0] < 2:
+        raise ValueError("at least two residual rows are required")
+    if not np.isfinite(r).all():
+        raise ValueError("residuals must be finite")
+    centered = r - r.mean(axis=0, keepdims=True)
+    cov = centered.T @ centered / float(r.shape[0] - 1)
+    if shrinkage:
+        diag = np.diag(np.diag(cov))
+        cov = (1.0 - shrinkage) * cov + shrinkage * diag
+    return regularize_covariance(cov, jitter=jitter, name="residual covariance")
+
+
+def fit_error_covariance(predicted, observed, shrinkage=0.0, jitter=1e-9):
+    """Estimate covariance of `observed - predicted` calibration errors."""
+    pred = np.asarray(predicted, dtype=float)
+    obs = np.asarray(observed, dtype=float)
+    if pred.shape != obs.shape:
+        raise ValueError(f"predicted shape {pred.shape} must match observed shape {obs.shape}")
+    return fit_residual_covariance(obs - pred, shrinkage=shrinkage, jitter=jitter)
+
+
+def gaussian_condition_update(mean, covariance, observation, observation_covariance, H=None, cross_covariance=None,
+                              jitter=1e-9):
+    """Condition a Gaussian state on a linear noisy observation.
+
+    State prior:
+        x ~ N(mean, covariance)
+    Observation:
+        y = H x + v
+
+    `observation_covariance` is Cov(v). `cross_covariance` is Cov(x - mean, v),
+    with shape `(state_dim, obs_dim)`. If omitted, it is treated as zero; callers
+    should pass it explicitly when prior and measurement channels share evidence.
+    """
+    m = _as_vector("mean", mean)
+    y = _as_vector("observation", observation)
+    n = m.size
+    k = y.size
+    P = regularize_covariance(_square_covariance("covariance", covariance, n), jitter=jitter, name="covariance")
+    R = regularize_covariance(
+        _square_covariance("observation_covariance", observation_covariance, k),
+        jitter=jitter,
+        name="observation covariance",
+    )
+    Hm = np.eye(n) if H is None else _as_matrix("H", H, k, n)
+    C = np.zeros((n, k)) if cross_covariance is None else _as_matrix("cross_covariance", cross_covariance, n, k)
+
+    innovation = y - Hm @ m
+    innovation_cov = Hm @ P @ Hm.T + Hm @ C + C.T @ Hm.T + R
+    innovation_cov = regularize_covariance(innovation_cov, jitter=jitter, name="innovation covariance")
+    state_innovation_cov = P @ Hm.T + C
+    gain = np.linalg.solve(innovation_cov.T, state_innovation_cov.T).T
+    posterior_mean = m + gain @ innovation
+    posterior_cov = P - gain @ innovation_cov @ gain.T
+    posterior_cov = regularize_covariance(posterior_cov, jitter=jitter, name="posterior covariance")
+
+    return GaussianUpdate(
+        mean=posterior_mean,
+        covariance=posterior_cov,
+        gain=gain,
+        innovation=innovation,
+        innovation_covariance=innovation_cov,
+        cross_covariance=C,
+    )
+
+
+def scalar_product_kalman_update(ell_prior, ell_measurement, prior_var, measurement_var, cross_covariance=0.0,
+                                 jitter=1e-9):
+    """Scalar Product-Kalman update in one log/link evidence coordinate."""
+    return gaussian_condition_update(
+        [ell_prior],
+        [[prior_var]],
+        [ell_measurement],
+        [[measurement_var]],
+        H=[[1.0]],
+        cross_covariance=[[cross_covariance]],
+        jitter=jitter,
+    )
+
+
+def gaussian_nll(observed, mean, covariance, jitter=1e-9, include_constant=True):
+    """Negative log likelihood of one Gaussian observation."""
+    y = _as_vector("observed", observed)
+    m = _as_vector("mean", mean)
+    if y.shape != m.shape:
+        raise ValueError(f"observed shape {y.shape} must match mean shape {m.shape}")
+    cov = regularize_covariance(_square_covariance("covariance", covariance, y.size), jitter=jitter, name="covariance")
+    residual = y - m
+    solved = np.linalg.solve(cov, residual)
+    sign, logdet = np.linalg.slogdet(cov)
+    if sign <= 0:
+        raise ValueError("covariance must have positive determinant")
+    constant = y.size * np.log(2.0 * np.pi) if include_constant else 0.0
+    return float(0.5 * (residual @ solved + logdet + constant))
--- a/prototypes/mu_cosine/product_kalman.py
+++ b/prototypes/mu_cosine/product_kalman.py
@@ -8,15 +8,25 @@ not construct product proxies themselves; use `product_space.py` for that layer.
 This module deliberately exposes cross-covariance between the prior state error and
 measurement noise. When prior and measurement channels share evidence, treating that
 cross term as zero is an assumption, not a default fact.
+
+Prototype note: this module prioritizes correctness and auditability over speed. A
+full update regularizes several covariance matrices with eigendecompositions; a
+performance-sensitive path should use a Cholesky-specialized implementation.
 """
 
 from dataclasses import dataclass
+import warnings
 
 import numpy as np
 
 
+NUMERIC_EIGEN_FLOOR = 1e-12
+SEMIDEFINITE_TOL_MULTIPLIER = 100.0
+
 __all__ = [
     "GaussianUpdate",
+    "NUMERIC_EIGEN_FLOOR",
+    "SEMIDEFINITE_TOL_MULTIPLIER",
     "fit_error_covariance",
     "fit_residual_covariance",
     "gaussian_condition_update",
@@ -26,9 +36,19 @@ __all__ = [
 ]
 
 
+def _readonly_array(value):
+    arr = np.array(value, dtype=float, copy=True)
+    arr.setflags(write=False)
+    return arr
+
+
 @dataclass(frozen=True)
 class GaussianUpdate:
-    """Result of one Gaussian conditioning / Kalman-style update."""
+    """Result of one Gaussian conditioning / Kalman-style update.
+
+    Array fields are copied and marked read-only in `__post_init__`; `frozen=True`
+    alone would not prevent in-place mutation of numpy arrays.
+    """
 
     mean: np.ndarray
     covariance: np.ndarray
@@ -37,13 +57,19 @@ class GaussianUpdate:
     innovation_covariance: np.ndarray
     cross_covariance: np.ndarray
 
+    def __post_init__(self):
+        for name in ("mean", "covariance", "gain", "innovation", "innovation_covariance", "cross_covariance"):
+            object.__setattr__(self, name, _readonly_array(getattr(self, name)))
+
 
 def _as_vector(name, value):
     arr = np.asarray(value, dtype=float)
     if arr.ndim == 0:
         arr = arr.reshape(1)
+    elif arr.ndim == 2 and 1 in arr.shape:
+        arr = arr.reshape(-1)
     if arr.ndim != 1:
-        raise ValueError(f"{name} must be a 1-D vector or scalar")
+        raise ValueError(f"{name} must be a 1-D vector, column vector, row vector, or scalar")
     if arr.size == 0:
         raise ValueError(f"{name} must be nonempty")
     if not np.isfinite(arr).all():
@@ -76,7 +102,9 @@ def regularize_covariance(covariance, jitter=1e-9, name="covariance"):
 
     Small negative eigenvalues from floating-point symmetry error are lifted. A
     materially negative eigenvalue raises instead of silently accepting an invalid
-    covariance model.
+    covariance model. Even when `jitter=0`, exact semidefinite covariances are
+    lifted to `NUMERIC_EIGEN_FLOOR` so downstream solves see positive-definite
+    matrices rather than singular ones.
     """
     if jitter < 0:
         raise ValueError("jitter must be nonnegative")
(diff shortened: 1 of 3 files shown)


prototypes/mu_cosine/product_kalman.py
```diff
@@ -0,0 +1,205 @@
...
+    return regularize_covariance(cov, jitter=jitter, name="residual covariance")
+
+
+def fit_error_covariance(predicted, observed, shrinkage=0.0, jitter=1e-9):
+    """Estimate covariance of `observed - predicted` calibration errors."""
+    pred = np.asarray(predicted, dtype=float)
+    obs = np.asarray(observed, dtype=float)
+    if pred.shape != obs.shape:
+        raise ValueError(f"predicted shape {pred.shape} must match observed shape {obs.shape}")
+    return fit_residual_covariance(obs - pred, shrinkage=shrinkage, jitter=jitter)
```
s243a: **[Math correctness — innovation covariance with cross-covariance]**

The current formula:
```python
innovation_cov = Hm @ P @ Hm.T + Hm @ C + C.T @ Hm.T + R
```
This implements S = H P Hᵀ + H C + Cᵀ Hᵀ + R, but the standard cross-covariance-aware Kalman innovation covariance (see Bar-Shalom et al. §5.2 or Solin & Svensson 2018) for the model x̃ ~ N(0, P), ṽ ~ N(0, R), Cov(x̃, ṽ) = C is:

  **S = H P Hᵀ + H C + Cᵀ Hᵀ + R**

This is indeed correct. However, note the document says "Joseph-equivalent conditioning" — the Joseph form P⁺ = (I − KH)P(I − KH)ᵀ + KRKᵀ is *not* used here; instead the standard P⁺ = P − K S Kᵀ is applied. For this prototype that is fine (they are algebraically equivalent when S is exact), but since the design doc explicitly mentions Joseph-equivalent conditioning, it is worth either:
1. Actually using the Joseph form for numerical stability with larger state dimensions, or
2. Updating the design doc to say "standard Riccati reduction" rather than "Joseph-equivalent".

The risk is that `regularize_covariance` with a small jitter compensates for ill-conditioning that the Joseph form would handle more gracefully.

prototypes/mu_cosine/product_kalman.py
```diff
@@ -0,0 +1,205 @@
...
+
+
+def gaussian_condition_update(mean, covariance, observation, observation_covariance, H=None, cross_covariance=None,
+                              jitter=1e-9):
+    """Condition a Gaussian state on a linear noisy observation.
```
s243a, line 185: **[Numerical correctness — posterior covariance]**

```python
posterior_cov = P - gain @ innovation_cov @ gain.T
```
With the cross-covariance term C ≠ 0, the correct posterior covariance is:

  P⁺ = P − K S Kᵀ  **only when** K = (P Hᵀ + C) S⁻¹  (which is what is computed above)

This is self-consistent, so algebraically this is fine. However, it can be worth noting that expanding this gives:

  P⁺ = P − (PHᵀ + C)(HPHᵀ + HC + CᵀHᵀ + R)⁻¹(PHᵀ + C)ᵀ

which is *not* the same as the standard Joseph form and may lose symmetry faster in higher dimensions. Consider adding a unit test that explicitly checks `posterior_cov` symmetry for a 3×3+ system with nonzero C.

prototypes/mu_cosine/product_kalman.py
```diff
@@ -0,0 +1,205 @@
...
+    cov = _as_matrix(name, covariance)
+    if cov.shape[0] != cov.shape[1]:
+        raise ValueError(f"{name} must be square")
+    cov = 0.5 * (cov + cov.T)
+    eig = np.linalg.eigvalsh(cov)
+    min_eig = float(eig[0])
+    floor = max(float(jitter), 1e-12)
+    if min_eig < -100.0 * floor:
+        raise ValueError(f"{name} is not positive semidefinite; min eigenvalue={min_eig:g}")
+    lift = max(float(jitter) - min_eig, 0.0)
```
s243a: **[Redundant `eigvalsh` call]**

`regularize_covariance` already calls `_as_matrix` and then does `eigvalsh`. In `gaussian_condition_update`, `P` and `R` are symmetrized via `_square_covariance` and then passed to `regularize_covariance`, which calls `_as_matrix` again internally (so `_as_matrix` runs twice per covariance). For a prototype this is harmless, but if this ever moves toward higher-dimensional use (e.g. larger `n`), the double validation and symmetrization will double the overhead for each covariance argument. Consider either:
- Having `regularize_covariance` accept a pre-validated matrix and skip the inner `_as_matrix` call, or
- Adding a `_validated=False` bypass flag for internal call chains.

prototypes/mu_cosine/product_kalman.py
```diff
@@ -0,0 +1,205 @@
...
+    cov = _as_matrix(name, value, dim, dim)
+    return 0.5 * (cov + cov.T)
+
+
+def regularize_covariance(covariance, jitter=1e-9, name="covariance"):
+    """Return a symmetric positive-definite covariance with diagonal jitter.
+
+    Small negative eigenvalues from floating-point symmetry error are lifted. A
+    materially negative eigenvalue raises instead of silently accepting an invalid
+    covariance model.
```
s243a: **[Boundary condition in `regularize_covariance`: `jitter=0.0` + PSD border case]**

```python
floor = max(float(jitter), 1e-12)
```
When `jitter=0.0` (as tests pass), `floor` becomes `1e-12`. This means even with `jitter=0.0`, the rejection threshold is `min_eig < -100 * 1e-12 = -1e-10`. So a matrix with `min_eig = -5e-11` (barely negative due to fp rounding) is **silently accepted and lifted** to 1e-12 even when the caller explicitly opts out of regularization. This is a semantic mismatch: `jitter=0.0` semantically means "do not regularize", but the function still lifts. The doc string says "small negative eigenvalues from floating-point symmetry error are lifted" which justifies this, but the `jitter=0.0` contract should be explicitly documented.

Suggestion: add a line to the docstring: "Passing `jitter=0` still lifts floating-point near-zero eigenvalues up to `1e-12`; pass `jitter=None` (not yet supported) if you need a strict non-modifying assertion."

prototypes/mu_cosine/product_kalman.py
```diff
@@ -0,0 +1,205 @@
...
+    "scalar_product_kalman_update",
+]
+
+
+@dataclass(frozen=True)
+class GaussianUpdate:
+    """Result of one Gaussian conditioning / Kalman-style update."""
+
+    mean: np.ndarray
+    covariance: np.ndarray
```
s243a, line 54: **[`frozen=True` dataclass with `np.ndarray` fields]**

```python
@dataclass(frozen=True)
class GaussianUpdate:
    mean: np.ndarray
    ...
```
`frozen=True` prevents field reassignment, but `np.ndarray` is mutable; the caller can still mutate `result.mean[0] = 99.0` in-place without triggering any error. This is a documentation/API contract issue: `frozen` gives a false sense of immutability. Consider either:
- Adding a `__post_init__` that calls `.copy()` on each array field and sets `writeable=False` on the copies, or
- Noting in the docstring that outputs should be treated as read-only by convention.

prototypes/mu_cosine/product_kalman.py
```diff
@@ -0,0 +1,205 @@
...
+    if shrinkage:
+        diag = np.diag(np.diag(cov))
+        cov = (1.0 - shrinkage) * cov + shrinkage * diag
+    return regularize_covariance(cov, jitter=jitter, name="residual covariance")
+
+
+def fit_error_covariance(predicted, observed, shrinkage=0.0, jitter=1e-9):
+    """Estimate covariance of `observed - predicted` calibration errors."""
+    pred = np.asarray(predicted, dtype=float)
+    obs = np.asarray(observed, dtype=float)
```
s243a, line 167: **[H=None default assumes H=I, but state_dim can differ from obs_dim]**

```python
Hm = np.eye(n) if H is None else _as_matrix("H", H, k, n)
```
When `H is None`, `Hm = np.eye(n)` requires `k == n` (square observation). If `n != k` and H is not supplied, the subsequent `Hm @ P @ Hm.T` will raise a shape error from numpy rather than a clear domain error. A guard would make the failure mode explicit:

```python
if H is None:
    if n != k:
        raise ValueError(f"H must be supplied when state_dim ({n}) != obs_dim ({k})")
    Hm = np.eye(n)
```
This is a defensive improvement, especially for the vector update path.

prototypes/mu_cosine/product_kalman.py
```diff
@@ -0,0 +1,205 @@
...
+        cross_covariance=[[cross_covariance]],
+        jitter=jitter,
+    )
+
+
+def gaussian_nll(observed, mean, covariance, jitter=1e-9, include_constant=True):
+    """Negative log likelihood of one Gaussian observation."""
+    y = _as_vector("observed", observed)
+    m = _as_vector("mean", mean)
+    if y.shape != m.shape:
```
s243a, line 277: **[`gaussian_nll`: sign check after `regularize_covariance` is redundant but potentially confusing]**

```python
sign, logdet = np.linalg.slogdet(cov)
if sign <= 0:
    raise ValueError("covariance must have positive determinant")
```
`regularize_covariance` already lifts the covariance to be PD (all eigenvalues ≥ jitter ≥ 1e-12 > 0), so `slogdet` will always return `sign=1` here after a successful call to `regularize_covariance`. The guard is unreachable under normal use. It becomes reachable only if someone bypasses `regularize_covariance`, which can't happen in this function. Either document why it's kept as a belt-and-suspenders safeguard, or remove it and rely on `regularize_covariance`.

prototypes/mu_cosine/test_product_kalman.py
```diff
@@ -0,0 +1,137 @@
...
+    weak_measurement = scalar_product_kalman_update(0.0, 10.0, prior_var=1.0, measurement_var=99.0, jitter=0.0)
+    strong_measurement = scalar_product_kalman_update(0.0, 10.0, prior_var=99.0, measurement_var=1.0, jitter=0.0)
+    assert weak_measurement.mean[0] < 0.2
+    assert strong_measurement.mean[0] > 9.8
+    assert weak_measurement.gain[0, 0] < strong_measurement.gain[0, 0]
+
+
+def test_vector_update_matches_independent_coordinate_formula():
+    res = gaussian_condition_update(
+        mean=[0.0, 0.0],
```
s243a, line 94: **[Test: `test_cross_covariance_matches_gaussian_conditioning_formula` — manual formula has a sign issue]**

```python
innovation_cov = 2.0 + 3.0 + 2.0 * 0.5
```
This computes S = P + R + 2C (scalar, H=1). The derivation from the code formula is:
  S = H P Hᵀ + H C + Cᵀ Hᵀ + R = P + C + C + R = P + 2C + R

For scalar C=0.5: S = 2 + 2*0.5 + 3 = 6.0, and `state_innovation_cov = P + C = 2.5`, so `gain = 2.5/6.0 ≈ 0.4167`. The test is arithmetically self-consistent with the implementation.

However, this test does not verify the **sign** of the cross-covariance contribution. The test passes with `C = +0.5` only; there is no complementary test verifying that `C = -0.5` (negative cross-covariance) gives a **smaller** innovation covariance and **smaller** gain. Consider adding:
```python
res_neg = gaussian_condition_update(mean=[1.0], covariance=[[2.0]], observation=[4.0],
    observation_covariance=[[3.0]], H=[[1.0]], cross_covariance=[[-0.5]], jitter=0.0)
assert res_neg.innovation_covariance[0,0] < res.innovation_covariance[0,0]
```

prototypes/mu_cosine/test_product_kalman.py
```diff
@@ -0,0 +1,137 @@
...
+
+def assert_raises(fn, *args, **kwargs):
+    try:
+        fn(*args, **kwargs)
+    except ValueError:
+        return
+    raise AssertionError(f"{fn.__name__} should have raised ValueError")
+
+
+def test_scalar_product_update_matches_closed_form():
```
s243a, line 34: **[Missing test: Kalman identity / information-limit checks]**

There is no test verifying the asymptotic behavior in limiting regimes:
- **Infinite prior variance** (`prior_var → ∞`): gain should approach 1 and posterior mean should approach the measurement.
- **Zero measurement noise** (`measurement_var → 0`): posterior mean should approach the observation exactly.
- **Identical prior and measurement** (`ell_prior == ell_measurement`): mean should be unchanged regardless of variance ratio.

These are standard sanity checks from Kalman filter theory (see Maybeck Vol. 1, Ch. 5) and would catch off-by-one or inverted gain bugs that the current closed-form tests might not surface if the formula is internally consistent but wrong in a sign or inverse.

s243a (commented):
## Academic Review: Add Product-Kalman Update Core

**Overall assessment:** The implementation is mathematically well-grounded and the module structure is clean. The separation of concerns between `product_space.py` (product proxy construction) and `product_kalman.py` (Gaussian conditioning) is sound. The explicit exposure of `cross_covariance` is the right design choice for product-evidence channels and the docstring reasoning is clear. Input validation is thorough. Test coverage hits the main happy paths and key error paths.

I've left inline comments on the following concerns, ordered by severity:

### Correctness / Math
1. **Innovation covariance formula vs. Joseph form** (line 130): The implementation uses the standard Riccati reduction `P⁺ = P - KSKᵀ`, but the design doc says “Joseph-equivalent conditioning”. These are only numerically equivalent when S is exact; for higher-dim use the Joseph form should be considered.
2. **Posterior covariance symmetry** (line 135): Recommend adding a test for symmetry of the posterior covariance with nonzero C in a 3×3+ system.
3. **Negative cross-covariance sign test missing** (test line 53): The existing test only covers positive C; a test verifying C < 0 gives smaller innovation covariance would close the coverage gap.

### API / Safety
4. **`frozen=True` gives false immutability** (line 34): `np.ndarray` fields are still mutably addressable in-place; either document or enforce read-only.
5. **`H=None` with n≠k gives an opaque numpy error** (line 127): A pre-check with a clear `ValueError` is better.
6. **`jitter=0.0` still regularizes silently** (line 79): The `1e-12` floor is undocumented relative to the `jitter=0` contract.

### Efficiency
7. **Double validation path** (line 92): `_as_matrix` is called twice per covariance in the main update path; worth noting for future performance work.

### Test Gaps
8. **No asymptotic/limit tests** (test line 33): Kalman theory limit regimes (infinite prior variance, zero measurement noise) are missing.
9. **`gaussian_nll` sign check is unreachable** (line 196): Dead code after `regularize_covariance`; should be documented or removed.

None of the above block merging for a prototype, but items 1–3 in particular are worth resolving before this module is promoted to CI or used in empirical comparison.

s243a:
## Second Academic Review — Add Product-Kalman Update Core

*Independent second pass. Items already covered in the first review (Joseph form, posterior symmetry, jitter=0 contract, frozen dataclass, H=None dimension guard, slogdet dead code, negative C test, Kalman limits, eigvalsh double-validation) are not repeated.*

---

### 1. Gain computation — `np.linalg.solve` with redundant transpose juggling

```python
gain = np.linalg.solve(innovation_cov.T, state_innovation_cov.T).T
```

`innovation_cov` is already symmetrized by `regularize_covariance`, so `innovation_cov.T == innovation_cov` and the transpose juggling is unnecessary. More importantly, since `regularize_covariance` guarantees PD, a Cholesky solve is always valid here and is O(n³/3) vs O(n³) for the general solver. Suggested alternative:

```python
L = np.linalg.cholesky(innovation_cov)
gain = np.linalg.solve(L.T, np.linalg.solve(L, state_innovation_cov.T)).T
```

---

### 2. `fit_residual_covariance` — shrinkage target is sample diagonal, not scaled identity

```python
diag = np.diag(np.diag(cov))
cov = (1.0 - shrinkage) * cov + shrinkage * diag
```

Standard Ledoit-Wolf and OAS estimators shrink toward `(tr(S)/p) * I` (scaled identity), not `diag(S)`, because the scaled identity minimizes the Frobenius loss under an isotropic model. For small calibration sets (few residual rows relative to dimension), shrinking toward the sample diagonal can still yield a rank-deficient result. Consider documenting this limitation or offering a `target='diagonal' | 'scaled_identity'` parameter.

---

### 3. `fit_error_covariance` — missing finite-value guard on raw inputs

```python
pred = np.asarray(predicted, dtype=float)
obs  = np.asarray(observed,  dtype=float)
if pred.shape != obs.shape:
    raise ValueError(...)
return fit_residual_covariance(obs - pred, ...)
```

There is no `np.isfinite` check on `predicted` or `observed` before subtraction. If either contains `nan`, the error bubbles up from `fit_residual_covariance` as *"residuals must be finite"* — misleading when the inputs themselves are the source. Add:

```python
if not np.isfinite(pred).all():
    raise ValueError("predicted must be finite")
if not np.isfinite(obs).all():
    raise ValueError("observed must be finite")
```

---

### 4. `GaussianUpdate.innovation_covariance` is the post-jitter S — document the mismatch risk

The returned field is set **after** `regularize_covariance` applies jitter, so `result.innovation_covariance` may differ from `H @ result.covariance @ H.T + R` recomputed from scratch. A downstream caller using this field for Mahalanobis distances or NLL computation will get a subtly different result than one who recomputes S. Recommend adding to the docstring: *"The returned `innovation_covariance` is the regularized S used for the gain computation; it may include a small jitter lift not reflected in the prior covariance."*

---

### 5. `scalar_product_kalman_update` — no positivity guard on variance arguments

Negative values for `prior_var` or `measurement_var` propagate into `regularize_covariance` where they may be silently lifted if the magnitude is small. A scalar variance should always be explicitly positive, and the error message from the eigenvalue path is indirect. Add:

```python
if prior_var <= 0:
    raise ValueError(f"prior_var must be positive, got {prior_var}")
if measurement_var <= 0:
    raise ValueError(f"measurement_var must be positive, got {measurement_var}")
```

---

### 6. Performance profile: 4× `eigvalsh` per `gaussian_condition_update` — undocumented

One update call triggers `regularize_covariance` (and therefore `eigvalsh`) four times: on P, R, `innovation_cov`, and `posterior_cov`. For the scalar case this is trivially cheap, but at state dimension n=64 (plausible for cosine feature vectors) that is four O(n³) eigendecompositions per update. Worth noting in the module docstring: *"This module prioritises correctness over speed. Each update performs up to four O(n³) eigendecompositions; for performance-sensitive paths use a Cholesky-based implementation."*

---

### 7. `test_vector_update_matches_independent_coordinate_formula` — only exercises diagonal covariances

```python
covariance=[[4.0, 0.0], [0.0, 1.0]],
observation_covariance=[[1.0, 0.0], [0.0, 3.0]],
```

Both matrices are diagonal, so the test reduces to two independent scalar updates and never exercises off-diagonal coupling in `H P Hᵀ`. A test using a dense (non-diagonal) P and a non-identity H would provide meaningful vector-case coverage.

---

### 8. `DESIGN_product_kalman_poe.md` — Steps 5 and 6 both say "held-out splits"; split disjointness is ambiguous

Step 5 fits `P_ell`, `R_ell` on "held-out node-disjoint calibration splits"; step 6 compares against `JointPosterior` on "held-out node-disjoint splits." If the same split is used for both fitting and comparison there is a data-leakage risk — the calibration residuals that determine `R_ell` must come from a split disjoint from the evaluation split. Suggest adding: *"The calibration split (step 5) and evaluation split (step 6) must be disjoint from each other and from training data."*

---

### 9. No `__init__.py` — module not importable as a package from repo root

The test uses `sys.path.insert(0, ...)` for portability, but this means `product_kalman` is not importable via `from prototypes.mu_cosine.product_kalman import ...` from the repo root (e.g., in notebooks or when `pytest` is invoked from a different CWD). Adding an empty `prototypes/mu_cosine/__init__.py` would make the directory a proper package without breaking any existing usage.

---

### New Findings Summary

| # | Location | Severity | Topic |
|---|---|---|---|
| 1 | `product_kalman.py:133` | Minor | Cholesky solve preferred over redundant-transpose `linalg.solve` |
| 2 | `product_kalman.py:103` | Advisory | Shrinkage target is sample diagonal, not scaled identity |
| 3 | `product_kalman.py:111` | Minor | `fit_error_covariance` missing finite-value guard on raw inputs |
| 4 | `product_kalman.py:138` | Advisory | Returned `innovation_covariance` is post-jitter; document mismatch risk |
| 5 | `product_kalman.py:155` | Minor | `scalar_product_kalman_update` missing positivity guard on variances |
| 6 | Module docstring | Advisory | 4× `eigvalsh` per update; document performance profile |
| 7 | `test_product_kalman.py:65` | Minor | Vector test uses only diagonal covariances |
| 8 | `DESIGN_product_kalman_poe.md` | Advisory | Calibration vs. evaluation split disjointness not stated |
| 9 | `prototypes/mu_cosine/` | Minor | Missing `__init__.py` for package importability |

None of these block merge at the prototype stage. Items 3, 5, and 8 carry the highest practical risk (silent bad inputs and potential data-leakage in the empirical comparison).

s243a:
## Third Academic Review — Add Product-Kalman Update Core

*Independent third pass (Sonnet 5). Items already covered in Review 1 (Joseph form, posterior symmetry, jitter=0 contract, frozen dataclass, H=None guard, slogdet dead code, negative-C test, Kalman limit tests, double eigvalsh) and Review 2 (Cholesky solve, shrinkage target, fit_error_covariance NaN guard, post-jitter innovation_covariance field, scalar variance positivity, 4× eigvalsh profile, diagonal-only vector test, split disjointness, __init__.py) are not repeated.*

---

### 1. `gaussian_condition_update` — cross-covariance symmetry assumption is undocumented but load-bearing

The docstring says `cross_covariance` has shape `(state_dim, obs_dim)` and represents `Cov(x - mean, v)`. This is inherently **not** a square symmetric matrix, so `regularize_covariance` is never called on it — correct. However, the innovation covariance formula:

```python
innovation_cov = Hm @ P @ Hm.T + Hm @ C + C.T @ Hm.T + R
```

expands to `S = HPHᵀ + HC + CᵀHᵀ + R`. The middle terms `HC + CᵀHᵀ` are the symmetrization of `HC`. This is correct **only if** the model assumes `Cov(x̃, ṽ) = C` and `Cov(ṽ, x̃) = Cᵀ` (i.e., the cross-covariance matrix is being treated as exact). If the caller accidentally passes a non-square C with transposed convention (obs_dim × state_dim instead of state_dim × obs_dim), the shape check `_as_matrix("cross_covariance", cross_covariance, n, k)` will catch it — good. But the docstring should explicitly state the convention: *"`cross_covariance[i, j]` is `Cov(x_i - mean_i, v_j)`"*, not just the shape.

---

### 2. `regularize_covariance` — the 100× tolerance multiplier is an undocumented magic constant

```python
floor = max(float(jitter), 1e-12)
if min_eig < -100.0 * floor:
    raise ValueError(...)
```

The factor `100.0` determines how much negative eigenvalue is "tolerable" before raising. This is an arbitrary engineering choice with no citation or derivation. For a covariance with trace ~1 (common in normalized evidence coordinates), `100 * 1e-9 = 1e-7` is a generous tolerance. For a covariance with trace ~1e6 (large-scale raw features), this same threshold might be too tight. A dimensionally adaptive threshold like `100 * floor * max(1.0, np.trace(cov) / cov.shape[0])` would be more principled. At minimum, the magic constant should be named and documented:

```python
_NEG_EIG_TOLERANCE_FACTOR = 100.0  # reject if min eigenvalue < -factor * jitter_floor
```

---

### 3. `fit_residual_covariance` — the unbiased estimator (n-1 denominator) is not always appropriate

```python
cov = centered.T @ centered / float(r.shape[0] - 1)
```

This uses Bessel's correction (unbiased sample covariance). For Kalman filter residual covariance estimation, the **maximum-likelihood** estimator (dividing by n, not n-1) is often preferred because: (a) it is the MLE of the true noise covariance under a Gaussian model, and (b) it never produces negative-semidefinite corrections when combined with the innovation update. Neither choice is universally correct, but the current selection of `n-1` is not documented or justified. Recommend exposing a `ddof=1` parameter (matching `np.cov` convention) and documenting the trade-off in the docstring.

---

### 4. `gaussian_nll` — argument order differs from `scipy.stats.multivariate_normal` convention

```python
def gaussian_nll(observed, mean, covariance, ...):
```

SciPy's `multivariate_normal.logpdf(x, mean, cov)` and most NLL utilities put `mean` before `observed`. Here `observed` comes first. This is not wrong, but it is an ergonomic trap for anyone familiar with the scipy API — a common source of argument-transposition bugs. At minimum, add a note to the docstring: *"Note: argument order is `(observed, mean, covariance)`, not `(mean, observed, covariance)`."*

---

### 5. `_as_vector` — 2-D single-column arrays are silently rejected

```python
def _as_vector(name, value):
    arr = np.asarray(value, dtype=float)
    if arr.ndim == 0:
        arr = arr.reshape(1)
    if arr.ndim != 1:
        raise ValueError(f"{name} must be a 1-D vector or scalar")
```

A caller passing a column vector `np.array([[1.0], [2.0]])` (shape `(2, 1)`) — which is idiomatic in many linear-algebra workflows — will get `"{name} must be a 1-D vector or scalar"`. The function could silently squeeze single-column/single-row 2-D arrays, or at minimum give an improved error message: *"got shape (2, 1); did you mean to pass a 1-D array?"* This is especially likely in the `observation` argument path, where a user may pass a column-vector output from `H @ x`.

---

### 6. `scalar_product_kalman_update` — `cross_covariance` default of `0.0` promotes the wrong default behavior for shared-evidence channels

```python
def scalar_product_kalman_update(ell_prior, ell_measurement, prior_var, measurement_var, cross_covariance=0.0, ...):
```

The module-level docstring explicitly warns: *"When prior and measurement channels share evidence, treating that cross term as zero is an assumption, not a default fact."* Yet `scalar_product_kalman_update` defaults `cross_covariance=0.0`, making the incorrect assumption the path of least resistance. Consider at minimum changing the default to `None` and raising if the caller doesn't explicitly pass a value:

```python
def scalar_product_kalman_update(..., cross_covariance=None, ...):
    if cross_covariance is None:
        raise ValueError(
            "cross_covariance must be supplied explicitly. "
            "Pass 0.0 only when you have verified channels do not share evidence."
        )
```

Or, less strictly, warn via `warnings.warn` when `cross_covariance=0.0` is used so the omission is visible in logs.

---

### 7. Test suite uses bare `assert` — incompatible with `pytest`'s assertion rewriting for dicts/arrays

```python
assert np.allclose(res.gain, [[0.8, 0.0], [0.0, 0.25]])
```

The PR's own validation instructions include running both `python3 test_product_kalman.py` and `pytest`. When a `np.allclose` assertion fails under pytest, the error message is just `AssertionError` with no diff. Replacing with `np.testing.assert_allclose` gives a structured failure message showing actual vs. expected values and tolerances, and is the standard for numerical test suites:

```python
np.testing.assert_allclose(res.gain, [[0.8, 0.0], [0.0, 0.25]], rtol=1e-10)
```

This also makes tolerances explicit rather than relying on `np.allclose`'s default `rtol=1e-5, atol=1e-8`.

---

### 8. No test for `regularize_covariance` with a non-square input

`regularize_covariance` checks:
```python
if cov.shape[0] != cov.shape[1]:
    raise ValueError(f"{name} must be square")
```

but there is no test that exercises this path. The existing test `test_regularize_covariance_lifts_tiny_semidefinite_floor_only` only covers the PD-lift and the rejection path for a materially negative eigenvalue. A test like:

```python
assert_raises(regularize_covariance, [[1.0, 2.0, 3.0], [4.0, 5.0, 6.0]])
```

would close this coverage gap.

---

### 9. Design doc: "Joseph-equivalent conditioning" claim needs a formal definition or citation

The updated step 4 in `DESIGN_product_kalman_poe.md` introduces the phrase *"Joseph-equivalent conditioning"* as a design goal without defining what it means or citing a reference. A reader unfamiliar with Kalman filter numerics will not understand why this matters vs. standard conditioning. Given that Review 1 already flagged the mismatch between this claim and the code (which uses `P - KSKᵀ`), the design doc should either:
- Cite the Joseph form explicitly: P⁺ = (I − KH)P(I − KH)ᵀ + KRKᵀ (Kalman 1960; Joseph's stabilized form, see Bierman 1977 *Factorization Methods for Discrete Sequential Estimation* §IV)
- Or replace the phrase with the accurate description of what the code actually does.

---

### Summary of New Findings

| # | Location | Severity | Topic |
|---|---|---|---|
| 1 | `product_kalman.py:128` | Advisory | Cross-covariance index convention not documented in docstring |
| 2 | `product_kalman.py:79` | Minor | Magic constant `100.0` tolerance multiplier undocumented/unexplained |
| 3 | `product_kalman.py:103` | Advisory | Unbiased (n-1) vs. MLE (n) denominator choice not justified |
| 4 | `product_kalman.py:183` | Minor | `gaussian_nll` argument order differs from scipy convention; trap for callers |
| 5 | `product_kalman.py:45` | Minor | `_as_vector` rejects idiomatic 2-D column vectors with unhelpful error |
| 6 | `product_kalman.py:155` | Advisory | `cross_covariance=0.0` default contradicts module-level warning; should require explicit opt-in |
| 7 | `test_product_kalman.py` (all) | Minor | Bare `assert` + `np.allclose` should be `np.testing.assert_allclose` for pytest compatibility |
| 8 | `test_product_kalman.py` | Minor | No test for `regularize_covariance` with non-square input |
| 9 | `DESIGN_product_kalman_poe.md` | Advisory | "Joseph-equivalent conditioning" lacks definition or citation |

None of these block merge at the prototype stage. Items 6 and 3 are the most semantically important: the `cross_covariance=0.0` default quietly undermines the module's own correctness philosophy, and the unbiased estimator choice is silent and consequential for small calibration sets.

PR #3531, opened 2026-07-08
Add Product-Kalman calibration helpers

s243a:
Adds the calibration layer for Product-Kalman experiments, between the core Gaussian update primitive and any real-data held-out comparison.

Changes:
- Add `product_kalman_calibration.py` to estimate `P`, `R`, and prior-measurement cross-covariance from calibration residuals.
- Add batch application of a fitted Product-Kalman calibration.
- Add `assert_disjoint_ids` to catch calibration/evaluation split leakage.
- Add synthetic tests for covariance block fitting, rowwise equivalence to the core update, read-only result arrays, shape/type guards, and held-out synthetic NLL improvement over the prior.
- Update `DESIGN_product_kalman_poe.md` to record the calibration/evaluation split boundary.

Validation:
- `python3 prototypes/mu_cosine/test_product_kalman_calibration.py`
- `python3 -m pytest -s prototypes/mu_cosine/test_product_kalman.py prototypes/mu_cosine/test_product_kalman_calibration.py`
- `python3 prototypes/mu_cosine/test_product_kalman.py`
- `python3 -m py_compile prototypes/mu_cosine/product_kalman.py prototypes/mu_cosine/product_kalman_calibration.py prototypes/mu_cosine/test_product_kalman.py prototypes/mu_cosine/test_product_kalman_calibration.py`
- `git diff --check`

### DIFF
--- a/prototypes/mu_cosine/product_kalman_calibration.py
+++ b/prototypes/mu_cosine/product_kalman_calibration.py
@@ -1,0 +1,198 @@
+#!/usr/bin/env python3
+"""Calibration helpers for Product-Kalman / correlated-PoE prototypes.
+
+This layer estimates the covariance blocks used by `product_kalman.py` from a
+calibration split with known target states in the chosen evidence coordinate:
+
+    prior_error       = target_state - prior_mean
+    measurement_error = measurement - H @ target_state
+
+The fitted blocks are:
+
+    P = Cov(prior_error)
+    R = Cov(measurement_error)
+    C = Cov(prior_error, measurement_error)
+
+Use calibration data that is node-disjoint from training data and from the final
+evaluation split. These helpers do not decide the split; they make leakage harder
+to miss once the split IDs are known.
+"""
+
+from dataclasses import dataclass
+
+import numpy as np
+
+try:
+    from .product_kalman import fit_residual_covariance, gaussian_condition_update, regularize_covariance
+except ImportError:  # direct script execution from prototypes/mu_cosine
+    from product_kalman import fit_residual_covariance, gaussian_condition_update, regularize_covariance
+
+
+__all__ = [
+    "ProductKalmanBatchUpdate",
+    "ProductKalmanCalibration",
+    "assert_disjoint_ids",
+    "apply_product_kalman_calibration",
+    "fit_product_kalman_calibration",
+]
+
+
+def _readonly_array(value):
+    arr = np.array(value, dtype=float, copy=True)
+    arr.setflags(write=False)
+    return arr
+
+
+@dataclass(frozen=True)
+class ProductKalmanCalibration:
+    """Covariance blocks fitted on a calibration split."""
+
+    state_covariance: np.ndarray
+    observation_covariance: np.ndarray
+    cross_covariance: np.ndarray
+    H: np.ndarray
+    n_samples: int
+    ddof: int
+    shrinkage: float
+    shrinkage_target: str
+
+    def __post_init__(self):
+        for name in ("state_covariance", "observation_covariance", "cross_covariance", "H"):
+            object.__setattr__(self, name, _readonly_array(getattr(self, name)))
+
+    @property
+    def state_dim(self):
+        return int(self.state_covariance.shape[0])
+
+    @property
+    def observation_dim(self):
+        return int(self.observation_covariance.shape[0])
+
+
+@dataclass(frozen=True)
+class ProductKalmanBatchUpdate:
+    """Batch application of one Product-Kalman calibration."""
+
+    mean: np.ndarray
+    covariance: np.ndarray
+    gain: np.ndarray
+    innovation: np.ndarray
+    innovation_covariance: np.ndarray
+
+    def __post_init__(self):
+        for name in ("mean", "covariance", "gain", "innovation", "innovation_covariance"):
+            object.__setattr__(self, name, _readonly_array(getattr(self, name)))
+
+
+def _as_2d(name, value):
+    arr = np.asarray(value, dtype=float)
+    if arr.ndim == 1:
+        arr = arr.reshape(1, -1)
+    if arr.ndim != 2:
+        raise ValueError(f"{name} must be a 1-D row or 2-D array")
+    if arr.shape[0] == 0 or arr.shape[1] == 0:
+        raise ValueError(f"{name} must be nonempty")
+    if not np.isfinite(arr).all():
+        raise ValueError(f"{name} must be finite")
+    return arr
+
+
+def _observation_matrix(H, state_dim, observation_dim):
+    if H is None:
+        if state_dim != observation_dim:
+            raise ValueError("H is required when target_state and measurement dimensions differ")
+        return np.eye(state_dim)
+    mat = np.asarray(H, dtype=float)
+    if mat.shape != (observation_dim, state_dim):
+        raise ValueError(f"H shape {mat.shape} must be ({observation_dim}, {state_dim})")
+    if not np.isfinite(mat).all():
+        raise ValueError("H must be finite")
+    return mat
+
+
+def assert_disjoint_ids(left_ids, right_ids, left_name="calibration", right_name="evaluation"):
+    """Raise if two split-ID collections overlap."""
+    left = set(left_ids)
+    right = set(right_ids)
+    overlap = sorted(left & right, key=lambda x: repr(x))
+    if overlap:
+        preview = ", ".join(repr(x) for x in overlap[:5])
+        suffix = "" if len(overlap) <= 5 else f", ... (+{len(overlap) - 5} more)"
+        raise ValueError(f"{left_name} and {right_name} IDs overlap: {preview}{suffix}")
+
+
+def fit_product_kalman_calibration(prior_mean, measurement, target_state, H=None, shrinkage=0.0, jitter=1e-9,
+                                   ddof=1, shrinkage_target="diagonal"):
+    """Fit Product-Kalman covariance blocks from calibration residuals.
+
+    `prior_mean` and `target_state` are rows in the state/evidence coordinate.
+    `measurement` rows are observations of `H @ target_state` in the same linked
+    coordinate system used by the update. `C` is estimated jointly with `P` and
+    `R`, so correlated prior/measurement errors are retained instead of silently
+    dropped.
+    """
+    prior = _as_2d("prior_mean", prior_mean)
+    target = _as_2d("target_state", target_state)
+    obs = _as_2d("measurement", measurement)
+    if prior.shape != target.shape:
+        raise ValueError(f"prior_mean shape {prior.shape} must match target_state shape {target.shape}")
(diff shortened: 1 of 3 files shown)


prototypes/mu_cosine/product_kalman_calibration.py
```diff
@@ -0,0 +1,198 @@
...
+    "apply_product_kalman_calibration",
+    "fit_product_kalman_calibration",
+]
+
+
+def _readonly_array(value):
+    arr = np.array(value, dtype=float, copy=True)
+    arr.setflags(write=False)
+    return arr
+
```
s243a, line 44: **[Design] `frozen=True` + `__post_init__` re-assignment pattern**

Using `object.__setattr__` inside `__post_init__` of a `frozen=True` dataclass is a well-known but slightly surprising idiom. Consider documenting _why_ you bypass the freeze here (to enforce the read-only flag on the numpy array in-place before the object is handed out). Alternatively, a `__init_subclass__` guard or a factory classmethod that returns a normally-mutable intermediate and then freezes it would make the intent clearer for future maintainers.

prototypes/mu_cosine/product_kalman_calibration.py
```diff
@@ -0,0 +1,198 @@
...
+        arr = arr.reshape(1, -1)
+    if arr.ndim != 2:
+        raise ValueError(f"{name} must be a 1-D row or 2-D array")
+    if arr.shape[0] == 0 or arr.shape[1] == 0:
+        raise ValueError(f"{name} must be nonempty")
+    if not np.isfinite(arr).all():
+        raise ValueError(f"{name} must be finite")
+    return arr
+
+
```
s243a, line 209: **[Correctness] Cross-covariance `C` is not re-regularized**

`P` and `R` are passed through `regularize_covariance` (which adds a jitter diagonal), but `C` is returned raw from the joint covariance block. This is mathematically correct only when jitter=0; with any positive `jitter`, the diagonal blocks of the joint covariance are inflated while the off-diagonal block `C` is not, which breaks the exact positive-semidefiniteness of the block-structured covariance matrix you constructed. Either apply the same jitter to the full joint covariance *before* slicing, or document explicitly that the inflated `P`/`R` and un-inflated `C` are intended.

prototypes/mu_cosine/product_kalman_calibration.py
```diff
@@ -0,0 +1,198 @@
...
+    prior = _as_2d("prior_mean", prior_mean)
+    target = _as_2d("target_state", target_state)
+    obs = _as_2d("measurement", measurement)
+    if prior.shape != target.shape:
+        raise ValueError(f"prior_mean shape {prior.shape} must match target_state shape {target.shape}")
+    if obs.shape[0] != target.shape[0]:
+        raise ValueError("measurement rows must match target_state rows")
+    Hm = _observation_matrix(H, target.shape[1], obs.shape[1])
+
+    prior_error = target - prior
```
s243a, line 282: **[Performance / Correctness] Template computed from zero-mean prior/observation**

`apply_product_kalman_calibration` computes the Kalman gain template by calling `gaussian_condition_update` with `np.zeros(...)` for both prior mean and measurement. This is valid for the gain/covariance outputs (which are mean-independent), but only if `gaussian_condition_update` doesn't internally use the means to compute innovation or gain. Verify (or assert) that the `template.gain` and `template.covariance` returned are identical regardless of the mean arguments passed. A brief inline comment would remove the reader's doubt.

prototypes/mu_cosine/product_kalman_calibration.py
```diff
@@ -0,0 +1,198 @@
...
+    measurement_error = obs - target @ Hm.T
+    joint_error = np.concatenate([prior_error, measurement_error], axis=1)
+    joint_cov = fit_residual_covariance(
+        joint_error,
+        shrinkage=shrinkage,
```
s243a, line 287: **[Correctness] Innovation computed against `prior`, not posterior mean**

Line: `innovation = obs - prior @ calibration.H.T`

This computes the *predicted* measurement residual from the prior, which is the standard Kalman innovation and is correct. Good. However note that the posterior mean `mean = prior + innovation @ template.gain.T` is then the classical Kalman correction formula, valid only when H is linear and the gain is computed from the *same* P/R/C used to build the template. Since the template is always recomputed from the calibration object, this invariant holds by construction — but it would be worth an explicit assertion or docstring note that `calibration` and the `prior` are expected to live in the same evidence coordinate space as specified in the module docstring.

prototypes/mu_cosine/product_kalman_calibration.py
```diff
@@ -0,0 +1,198 @@
...
+    if not np.isfinite(mat).all():
+        raise ValueError("H must be finite")
+    return mat
+
+
+def assert_disjoint_ids(left_ids, right_ids, left_name="calibration", right_name="evaluation"):
+    """Raise if two split-ID collections overlap."""
+    left = set(left_ids)
+    right = set(right_ids)
+    overlap = sorted(left & right, key=lambda x: repr(x))
```
s243a, line 248: **[Minor] Unused variable `obs_dim`**

`obs_dim = obs.shape[1]` is assigned but never referenced; `R` and `C` are already sliced correctly via `state_dim`. Remove to avoid reader confusion.

prototypes/mu_cosine/test_product_kalman_calibration.py
```diff
@@ -0,0 +1,148 @@
...
+def test_assert_disjoint_ids_flags_leakage():
+    assert_disjoint_ids(["a", "b"], ["c", "d"])
+    assert_raises(assert_disjoint_ids, ["a", "b"], ["b", "c"])
+
+
+def test_shape_and_type_errors_fail_fast():
+    prior, measurement, target, H, _, _ = toy_calibration_inputs()
+    assert_raises(fit_product_kalman_calibration, prior[:, :1], measurement, target, H=H)
+    assert_raises(fit_product_kalman_calibration, prior, measurement[:-1], target, H=H)
+    assert_raises(fit_product_kalman_calibration, prior, measurement, target, H=None)
```
s243a, line 192: **[Test robustness] Read-only test only attempts one write per array**

The test confirms `arr.flat[0] = 99.0` raises `ValueError`, which is correct. One edge case: if the array happens to be 0-dimensional or empty, `flat[0]` raises `IndexError` rather than `ValueError`, causing the test to mis-pass. Consider using `arr.flags.writeable` as the assertion directly (`assert not arr.flags.writeable`) instead of relying on a mutation side-effect.

prototypes/mu_cosine/test_product_kalman_calibration.py
```diff
@@ -0,0 +1,148 @@
...
+
+
+def test_synthetic_calibration_update_improves_eval_nll_over_prior():
+    rng = np.random.default_rng(11)
+    n_cal = 5000
+    n_eval = 2000
+    n = n_cal + n_eval
+    target = rng.normal(size=(n, 1))
+    error_cov = np.array([[0.9, 0.25], [0.25, 0.35]])
+    errors = rng.multivariate_normal([0.0, 0.0], error_cov, size=n)
```
s243a, line 261: **[Test gap] NLL improvement threshold (0.25 nats) is not statistically motivated**

`assert post_nll < prior_nll - 0.25` — the 0.25 nat margin is seed-fixed and empirically chosen. For a reproducible test suite this is fine, but consider adding a brief comment stating what the theoretical lower bound on NLL improvement is given the synthetic error covariance `[[0.9, 0.25],[0.25, 0.35]]`. This would also act as a regression sentinel: if the gap narrows after a refactor, the reader knows whether the threshold is conservative or tight.

prototypes/mu_cosine/test_product_kalman_calibration.py
```diff
@@ -0,0 +1,148 @@
...
+    prior_error = np.array([
+        [0.2, -0.1],
+        [-0.3, 0.2],
+        [0.1, 0.0],
+        [0.4, -0.2],
+        [-0.4, 0.1],
+    ])
+    measurement_error = np.array([
+        [0.1],
+        [-0.2],
```
s243a, line 48: **[Test coverage] `toy_calibration_inputs` uses only n=5 samples**

With n=5 and ddof=1, the sample covariance matrix has rank ≤ 4. For the 2-state + 1-observation joint error matrix (3 columns), this is sufficient for full rank, but only barely. The block-fitting test (`test_fit_calibration_splits_joint_covariance_into_blocks`) verifies exact match against `np.cov`, which would also pass with a rank-deficient matrix. Adding a case where n < state_dim + observation_dim would stress-test `regularize_covariance`'s jitter guard.

s243a (commented):
## Academic Review — PR #3531: Add Product-Kalman calibration helpers

**Summary**

This PR introduces a well-structured calibration layer between the core `gaussian_condition_update` primitive (PR #3530) and real-data held-out evaluation. The design decision to estimate `P`, `R`, and the prior-measurement cross-covariance `C` jointly from a single concatenated residual matrix is statistically sound: it avoids the independence assumption that would cause the standard PoE precision-summation formula to over-sharpen posteriors when prior and measurement errors are correlated. The leakage guard (`assert_disjoint_ids`) and the read-only result arrays are good defensive-programming practices for a prototype that will be compared against held-out evaluation data.

**Substantive issues (require resolution before merge)**

1. **Jitter inconsistency in block decomposition** — `P` and `R` receive a jitter diagonal via `regularize_covariance`, but `C` is taken raw from the joint sample covariance. This means the block-structured matrix `[[P, C], [C.T, R]]` is no longer exactly equal to the jitter-regularized joint covariance, potentially breaking PSD-ness of the full block. See inline comment at line 99.

2. **Template mean-independence assumption is silent** — `apply_product_kalman_calibration` correctly passes zero means to `gaussian_condition_update` to extract a mean-independent gain/covariance template, but this correctness depends on `gaussian_condition_update` not mixing mean values into the gain computation. This implicit contract should be documented or asserted. See inline comment at line 143.

**Minor issues**

3. Unused variable `obs_dim` (line 117).
4. `frozen=True` + `object.__setattr__` idiom should carry a short explanatory comment (line 44).
5. Innovation formula at line 148 is correct, but a docstring note linking the coordinate-space precondition to the module docstring would help readers.

**Test observations**

6. Read-only check pattern (line 113) could falsely pass on zero-size arrays; prefer `arr.flags.writeable` assertion.
7. NLL threshold of 0.25 nats (line 128) is empirically chosen; a brief comment on the theoretical gap would make regression intent clear.
8. `toy_calibration_inputs` with n=5 is sufficient but does not stress the jitter guard. Consider adding an underdetermined case.

**DESIGN_product_kalman_poe.md**

The updated steps 5–6 now correctly reflect the actual implementation boundary (calibration module fits blocks; evaluation is separate). The addition to `Related local artifacts` is accurate. No issues.

**Overall assessment:** Approve pending resolution of issue #1 (jitter inconsistency), which is a potential correctness bug. Issues #2–8 are non-blocking but recommended before promotion to a non-prototype module.

PR #3651, opened 2026-07-11
mu_cosine: post-#3648 validation and GPU square-root/QR conditioning

s243a:
## Summary

This follow-up validates and extends the cheap-judge pipeline merged in #3648.

- Fixes matched-cost model-selection leakage by nesting calibration, covariance fitting, pseudo-target generation, ridge fitting, and scaling inside the paid-label training fold.
- Reruns the exact-budget comparison on strict node-disjoint evaluation.
- Confirms graph_S across 40 node partitions and distinguishes the graph judge’s two roles:
  1. an immediate final-posterior measurement; and
  2. nearly free structural supervision that helps a model learn a dataset’s entities, vocabulary, topology, and recurring relations.
- Separates Luna head initialization from the campaign-independent prior and applies train-only global affine calibration.
- Adds a same-split JointPosterior comparison, calibrated uncertainty metrics, factored controls, and a within-judge hard-max/weighted/softmax pooling audit.
- Adds NumPy and PyTorch joint square-root information conditioners with nonzero prior/measurement cross-covariance, conditionally independent block streaming, compact Householder `geqrf/ormqr`, CUDA batching, and a matched design-cached dense-gain baseline.

## Corrected findings

- **The merged spending headline does not survive leakage-free evaluation.** No exact-budget cheap-label arm has a positive training-subsample interval excluding zero. At `n=160,k=2`, all ten S subsample deltas are negative on both corpora, but this fixed-partition diagnostic is not a population-level harm claim.
- **graph_S remains useful free supervision.** Its free-only S-NLL gain is positive across all 40 node partitions. Its incremental final-posterior value after debiased Luna is small and unconfirmed.
- **The graph result does not measure dataset familiarization.** The graph judge can still be valuable as broad structural supervision that teaches the model the entities, terminology, topology, and relation patterns of a dataset.
- **Debiased Luna remains valuable in the analytic Gaussian ladder**, although the leakage-free matched-cost proxy does not establish a general spending advantage.
- **JointPosterior and Gaussian+bridge are currently tied on AURC.** After both experiments use the same audited, stratum-aware node splitter, every paired interval includes zero across both corpora and both audited seeds:

| seed / corpus | JointPosterior − Gaussian AURC [95% CI] |
|---|---:|
| 0 / exploratory | +0.0013 [-0.0061,+0.0081] |
| 0 / fresh | +0.0058 [-0.0075,+0.0193] |
| 1 / exploratory | -0.0085 [-0.0288,+0.0033] |
| 1 / fresh | +0.0022 [-0.0228,+0.0293] |

This does not establish equivalence, and it is not a reason to retire JointPosterior. The learned joint combiner remains relevant when Gaussian assumptions or the linear D/S decision bridge are misspecified.

## Council-review follow-up

The post-implementation council review recommended approval and identified non-blocking improvements. The follow-up commit:

- replaces the duplicated plain node splitter with the shared, coverage-aware, stratum-aware `node_disjoint_pair_split`;
- reruns the JointPosterior comparison on that audited split;
- states explicitly that endpoint node-disjointness is not edge-disjointness or k-hop isolation from the ambient graph;
- relabels matched-cost intervals as subsample precision rather than inferential confidence intervals;
- adds endpoint-bootstrap block counts beside AURC comparisons;
- documents AURC’s early-error sensitivity and finite-sample limitations;
- adds distinct-design CUDA parity tests;
- adds sequential CUDA root-threading parity tests;
- tests compiled-conditioner safety against caller mutation of `H`, `R`, and `C`;
- removes duplicate conditional-covariance regularization;
- improves batched rank-failure diagnostics;
- corrects the 1,770 raw rows versus 1,700 Luna-matched rows distinction.

## Joint square-root/QR conditioner

The conditioner maintains an information root

`P^-1 = U.T @ U`

and converts the correlated residual model into a conditional likelihood:

`Rc = R - C.T @ P^-1 @ C`

`J = H + C.T @ P^-1`

After factoring `Rc = L @ L.T`, it triangularizes the information pre-array

`stack([U, solve(L, J)])`

while applying the same Householder transformations to the information right-hand side.

This avoids explicit posterior covariance inversion or subtraction-form downdates. It also allows independent measurement blocks to be streamed by carrying `(U_post, z_post)` into the next update.

The independence requirement is block diagonality of **conditional** covariance `Rc`, not raw `R`:

`Rc_ij = R_ij - C_i.T @ P^-1 @ C_j`

Raw measurement families can therefore be uncorrelated while remaining conditionally coupled through their shared prior. The state prior is inserted exactly once; repeating a complete `[[P,C_i],[C_i.T,R_i]]` block for each bundle would double-count prior information.

Potter and Carlson are historical antecedents. The implemented vector, nonzero-cross-covariance algorithm is more precisely a **joint square-root information/QR conditioner**, not a direct implementation of classical Potter or Carlson recursions.

## CPU/GPU result

GPU QR works and substantially accelerates larger factorizations, but a cached dense gain remains faster in every tested fixed-`P,H,C,R` cell.

Selected 4,096-row timings:

| state / measurements | CPU QR | CUDA QR | CPU dense | CUDA dense |
|---|---:|---:|---:|---:|
| `n=2,m=4` | 0.596 ms | 0.520 ms | **0.113 ms** | 0.132 ms |
| `n=32,m=32` | 2.106 ms | 0.543 ms | 0.363 ms | **0.129 ms** |
| `n=128,m=32` | 5.694 ms | 1.121 ms | 0.734 ms | **0.131 ms** |

These are single-run, on-device compute measurements. They exclude host/device transfer and do not benchmark the sequential regime where QR carries the updated root.

The resulting deployment rule is:

- use a design-cached dense gain for fixed, well-conditioned `P,H,C,R`;
- retain square-root/QR for changing, sequential, or ill-conditioned correlated measurement blocks;
- rebenchmark CUDA as the measurement dimension, state dimension, batch size, or augmented block size grows.

## Pooling and expert-selection scope

The inherited hard maximum pools relation-specific μ values **within one judge response**. It does not select the best expert.

Prior, graph, and Luna remain simultaneous measurements in the covariance-aware update. Probability-weighted and temperature-softmax pooling were audited as alternatives, but the present diagnostic does not establish a replacement. A fair production comparison must rebuild every target and source under the same pooling rule.

## Validation

- GitHub Actions passed on the current head.
- Independent focused run: `61 passed, 5 environment-gated skips`.
- The NumPy/PyTorch square-root suite contains 29 tests covering:
  - dense/QR posterior parity;
  - float32 and float64;
  - CPU and CUDA;
  - distinct batched designs;
  - nonzero prior/measurement cross-covariance;
  - block-diagonal conditional-covariance streaming;
  - fixed-coordinate and recentered sequential updates;
  - row-permutation invariance;
  - invalid Schur complements;
  - rank-deficient batches;
  - compiled-input mutation safety.
- Python compilation and diff checks passed.

## Selected references

- Dan Simon, [*Optimal State Estimation*](https://onlinelibrary.wiley.com/doi/book/10.1002/0470045345) — correlated-noise Kalman filtering.
- Toshio Michael Chin, [“Square-Root Formulas for Kalman Filter, Information Filter, and RTS Smoother”](https://ipnpr.jpl.nasa.gov/progress_report/42-233/42-233A.html) — modern square-root and SRIF background.
- Gerald J. Bierman, [*Sequential Least-Squares Using Orthogonal Transformations*](https://ntrs.nasa.gov/citations/19750023773) — Householder information arrays and sequential least squares.
- Neal A. Carlson, [“Fast Triangular Formulation of the Square Root Filter”](https://doi.org/10.2514/3.6907) — classical triangular covariance-root recursion.
- Catherine L. Thornton and Gerald J. Bierman, [*A Numerical Comparison of Discrete Kalman Filtering Algorithms*](https://ntrs.nasa.gov/api/citations/19760019842/downloads/19760019842.pdf) — numerical-stability and computational-cost comparisons.
- Gavin C. Cawley and Nicola L. C. Talbot, [“On Over-fitting in Model Selection and Subsequent Selection Bias in Performance Evaluation”](https://jmlr.org/papers/v11/cawley10a.html) — nested model selection and evaluation leakage.
- Traub et al., [“Overcoming Common Flaws in the Evaluation of Selective Classification Systems”](https://proceedings.neurips.cc/paper_files/paper/2024/file/047c84ec50bd8ea29349b996fc64af4b-Paper-Conference.pdf) — AURC limitations and AUGRC.
- Zhou et al., [“A Novel Characterization of Population AURC and Rates of Finite-Sample Estimators”](https://arxiv.org/html/2410.15361v4) — finite-sample AURC estimation.

## Follow-up

1. Repeat exact-budget matched-cost cells across preregistered node partitions.
2. Measure graph-supervision dataset adaptation with frozen initialization and matched training spend.
3. Add independent or human-verified targets.
4. Add AUGRC as a separately named robustness metric.
5. Benchmark genuinely sequential root-threading workloads with median/MAD timing, transfer, memory, and log-spaced scale/condition sweeps.
6. Sweep measurement count, state dimension, batch size, block-diagonal versus dense conditional correlation, and skinny-information versus full augmented QR to locate the CPU/CUDA crossover.

### DIFF
--- a/prototypes/mu_cosine/REPORT_cheap_judge_joint_posterior.md
+++ b/prototypes/mu_cosine/REPORT_cheap_judge_joint_posterior.md
@@ -6,47 +6,53 @@ Run 2026-07-10 from the post-#3648 validation branch. Protocol and non-claims:
 ## Bottom line
 
 The learned `JointPosterior` does **not** establish a win over the dense correlated-Gaussian conditioner plus
-its D/S LR bridge on this declared three-way task. Seed 0 is near-tied; fresh seed 1 significantly favours the
-Gaussian+bridge pipeline. The #3648 dense method therefore remains a strong interpretable baseline rather than
-a control that has already been superseded.
+its D/S LR bridge on this declared three-way task. After both experiments use the same audited, stratum-aware
+node splitter, seeds 0 and 1 are near-tied on both corpora and every paired AURC interval includes zero. The
+#3648 dense method therefore remains a strong interpretable baseline rather than a control that has already
+been superseded.
 
 This is fidelity to GPT-5.5's macro decision (directional / symmetric / other), not independent ground truth.
-The campaign's `cur_rel` cannot be used: it is `subcategory` on all 1,770 rows.
+The raw campaign's `cur_rel` cannot be used: it is `subcategory` on all 1,770 rows; 1,700 rows remain after
+matching the exploratory and fresh campaign views to Luna outputs.
 The reported ECE/log-loss audit raw cross-entropy outputs; no held labels were used for post-hoc calibration.
 
 ## Primary split and headline comparison
 
-Seed 0 partitions nodes before retaining within-side pairs. No held endpoint appears in this follow-up's
-combiner/calibration training. The campaign-independent upstream checkpoint is not audited as node-independent
-from every earlier training source, so this is not an end-to-end unseen-node claim.
+Seed 0 selects the best of 64 outcome-blind node assignments using retained coverage and macro-class strata,
+then retains within-side pairs. No held endpoint appears in this follow-up's combiner/calibration training.
+This endpoint split is not edge-disjoint or k-hop-isolated from the ambient parent graph used to construct graph
+features. The campaign-independent upstream checkpoint is not audited as node-independent from every earlier
+training source, so this is not an end-to-end unseen-node claim.
 
 | corpus | train / held / crossing rows dropped | method | accuracy | log-loss | ECE-10 | margin AURC [component-bootstrap 95% CI] |
 |---|---:|---|---:|---:|---:|---:|
-| exploratory | 329 / 173 / 498 | D/S hard-max linear reference | 0.931 | 0.161 | 0.020 | 0.006 [0.002, 0.013] |
-| | | dense Gaussian + D/S LR bridge, all | **0.879** | **0.331** | 0.071 | **0.025 [0.012, 0.042]** |
-| | | JointPosterior LR, all sources | 0.873 | 0.333 | **0.043** | 0.026 [0.013, 0.045] |
-| | | factored PoE, equal | 0.844 | 0.508 | 0.100 | 0.032 [0.017, 0.050] |
-| | | factored PoE, separability | 0.809 | 0.543 | 0.118 | 0.065 [0.038, 0.097] |
-| fresh | 263 / 108 / 329 | D/S hard-max linear reference | **0.935** | **0.217** | **0.028** | **0.012 [0.003, 0.025]** |
-| | | dense Gaussian + D/S LR bridge, all | **0.787** | **0.573** | **0.062** | **0.084 [0.046, 0.137]** |
-| | | JointPosterior LR, all sources | 0.778 | 0.591 | 0.092 | 0.092 [0.049, 0.146] |
-| | | factored PoE, equal | 0.722 | 0.753 | 0.166 | 0.084 [0.046, 0.128] |
-| | | factored PoE, separability | 0.796 | 0.853 | 0.341 | 0.077 [0.044, 0.119] |
+| exploratory | 365 / 161 / 474 | D/S hard-max linear reference | **0.925** | **0.170** | **0.018** | **0.008 [0.003, 0.016]** |
+| | | dense Gaussian + D/S LR bridge, all | 0.851 | 0.378 | **0.037** | **0.033 [0.018, 0.054]** |
+| | | JointPosterior LR, all sources | **0.870** | **0.377** | 0.071 | 0.035 [0.020, 0.059] |
+| | | factored PoE, equal | 0.795 | 0.596 | 0.143 | 0.043 [0.024, 0.070] |
+| | | factored PoE, separability | 0.770 | 0.584 | 0.109 | 0.077 [0.043, 0.118] |
+| fresh | 262 / 117 / 321 | D/S hard-max linear reference | **0.923** | **0.212** | **0.058** | **0.009 [0.002, 0.021]** |
+| | | dense Gaussian + D/S LR bridge, all | **0.718** | **0.598** | 0.121 | **0.108 [0.069, 0.165]** |
+| | | JointPosterior LR, all sources | 0.701 | 0.660 | **0.113** | 0.113 [0.072, 0.175] |
+| | | factored PoE, equal | 0.726 | 0.808 | 0.135 | 0.118 [0.073, 0.175] |
+| | | factored PoE, separability | 0.650 | 0.817 | 0.156 | 0.146 [0.093, 0.217] |
 
 Paired `AURC(Joint/all) - AURC(Gaussian/all)` (negative favours JointPosterior):
 
-- exploratory: **+0.0008 [-0.0053, +0.0065]**;
-- fresh: **+0.0074 [-0.0123, +0.0280]**.
+- exploratory: **+0.0013 [-0.0061, +0.0081]** (122 blocks; largest 5 rows);
+- fresh: **+0.0058 [-0.0075, +0.0193]** (72 blocks; largest 7 rows).
 
 Partition sensitivity matters. On seed 1, exploratory Gaussian+bridge versus Joint is
-`.897/.326/.070/.033` versus `.903/.272/.019/.021` (accuracy/log-loss/ECE/AURC), paired delta
-`-0.0127 [-0.0403,+0.0030]`; fresh is `.728/.618/.131/.111` versus `.737/.670/.088/.150`, paired delta
-**`+0.0384 [+0.0056,+0.0781]`**, favouring Gaussian+bridge. Seed 0 remains the declared primary split, but
-its near-tie is not a partition-robust equivalence result.
+`.854/.339/.050/.033` versus `.884/.322/.046/.025` (accuracy/log-loss/ECE/AURC), paired delta
+`-0.0085 [-0.0288,+0.0033]` (122 blocks; largest 4); fresh is `.750/.593/.071/.102` versus
+`.759/.618/.080/.104`, paired delta `+0.0022 [-0.0228,+0.0293]` (69 blocks; largest 6). These audited
+splits do not support a method-level AURC difference, but two seeds are not an equivalence study.
 
-The endpoint-component bootstrap had 126 blocks (largest 8 rows) for exploratory and 73 blocks (largest 4)
-for fresh. These blocks are sufficiently dispersed for the interval to be more informative than a row
-bootstrap, though it remains a one-split operating-judge result.
+The endpoint-component bootstrap respects shared endpoints, but AURC can over-emphasize the earliest
+high-confidence errors ([AUGRC, NeurIPS 2024](https://proceedings.neurips.cc/paper_files/paper/2024/file/047c84ec50bd8ea29349b996fc64af4b-Paper-Conference.pdf))
+and its empirical estimator is biased at small held-set sizes
+([population AURC](https://arxiv.org/html/2410.15361v4)). An AUGRC robustness check and more node partitions
+remain follow-up work; the present intervals are one-split operating-judge results.
 
 The D/S linear-reference gap is diagnostic, not a ceiling. It can reflect max-pooling information loss, but
 also finite training data, LR misspecification/optimisation, and incoherence between the judge's relation
@@ -57,39 +63,35 @@ representation ceiling.
 
 | corpus | method family | prior only | +graph | +Luna | all |
 |---|---|---:|---:|---:|---:|
-| exploratory | dense Gaussian + bridge AURC | 0.181 | 0.072 | 0.032 | **0.025** |
-| | JointPosterior AURC | 0.181 | 0.066 | 0.044 | **0.026** |
-| fresh | dense Gaussian + bridge AURC | 0.348 | 0.149 | 0.101 | **0.084** |
-| | JointPosterior AURC | 0.299 | 0.166 | 0.115 | **0.092** |
+| exploratory | dense Gaussian + bridge AURC | 0.202 | 0.071 | 0.040 | **0.033** |
+| | JointPosterior AURC | 0.208 | 0.058 | 0.042 | **0.035** |
+| fresh | dense Gaussian + bridge AURC | 0.388 | 0.197 | 0.143 | **0.108** |
+| | JointPosterior AURC | 0.347 | 0.232 | 0.120 | **0.113** |
 
 The seed-0 all-source point estimate is best for each family, but its paired ablation is conditional on that
 one fitted partition. Deltas are `AURC(all) - AURC(without source)`; negative favours adding the source:
 
 | corpus | family | add graph [95% CI] | add debiased Luna [95% CI] |
 |---|---|---:|---:|
-| exploratory | dense Gaussian + bridge | -0.0073 [-0.0183, +0.0003] | **-0.0469 [-0.0791, -0.0183]** |
-| | JointPosterior | **-0.0177 [-0.0423, -0.0025]** | **-0.0402 [-0.0768, -0.0099]** |
-| fresh | dense Gaussian + bridge | -0.0163 [-0.0471, +0.0076] | **-0.0644 [-0.1337, -0.0133]** |
-| | JointPosterior | -0.0237 [-0.0815, +0.0325] | **-0.0746 [-0.1380, -0.0307]** |
+| exploratory | dense Gaussian + bridge | -0.0068 [-0.0166, +0.0010] | **-0.0373 [-0.0659, -0.0114]** |
+| | JointPosterior | -0.0074 [-0.0215, +0.0040] | **-0.0231 [-0.0452, -0.0059]** |
+| fresh | dense Gaussian + bridge | **-0.0357 [-0.0778, -0.0055]** | **-0.0888 [-0.1592, -0.0324]** |
+| | JointPosterior | -0.0067 [-0.0336, +0.0215] | **-0.1188 [-0.1864, -0.0514]** |
 
-Conditional on the seed-0 fitted split, debiased Luna's interval favours inclusion in all four cells. This is
-not robustness across node partitions: a seed-1 sensitivity makes fresh Joint add-Luna inconclusive and fresh
-Joint add-graph harmful (`+0.0441 [+0.0104,+0.0806]`); fresh Joint add-Luna becomes
-`-0.0739 [-0.1573,+0.0007]`. Graph's seed-0 point estimate is beneficial in all four, but its interval excludes
-zero only for exploratory JointPosterior. The result is not evidence of source independence. On exploratory
-training
-rows, `corr(graph_S,luna_S)=+0.68`,
-`corr(graph_D,luna_D)=+0.60`, and `corr(prior_D,graph_D)=+0.58`. On fresh, the corresponding graph-D/Luna-D
-correlation is +0.56. The learned joint head and fitted cross-covariance can price this redundancy; a factored
+Debiased Luna's interval favours inclusion in all four seed-0 cells and all four audited seed-1 cells. Graph's
+seed-0 point estimate is beneficial in all four, but its interval excludes zero only for fresh Gaussian; the
+same pattern holds at seed 1. This two-seed stability remains descriptive and is not evidence of source
+independence. On exploratory training rows, `corr(graph_S,luna_S)=+0.60`,
+`corr(graph_D,luna_D)=+0.63`, and `corr(prior_D,graph_D)=+0.40`. On fresh, the corresponding graph-D/Luna-D
+correlation is +0.65. The learned joint head and fitted cross-covariance can price this redundancy; a factored
 control cannot.
 
-The separability-weighted factored control illustrates why one metric is insufficient: on fresh it has high
-accuracy and low AURC, but its log-loss 0.853 and ECE 0.341 reveal severe miscalibration. It is not a preferred
-combiner.
+The factored controls illustrate why one metric is insufficient: fresh equal-weight PoE has slightly higher
+accuracy than both joint methods, but worse log-loss and ECE than either. It is not a preferred combiner.
 
 The hard macro target resolves exact ties in fixed class order (directional, symmetric, other). Across the
-1,770-row campaign there are 7 exact top ties and 41 normalized top-two margins below 0.02. The result JSON now
-persists each comparison row's macro vector, margin, tie flag, and train/held/cross-dropped membership.
+1,700-row matched campaign there are 7 exact top ties and 41 normalized top-two margins below 0.02. The result
+JSON persists each comparison row's macro vector, margin, tie flag, and train/held/cross-dropped membership.
 
 ## Within-judge pooling diagnostic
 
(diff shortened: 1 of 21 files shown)


prototypes/mu_cosine/run_cheap_judge_joint_posterior.py
```diff
@@ -0,0 +1,632 @@
...
+                "prob_weighted": np.asarray(prob_weighted),
+                f"softmax_t{temperature:g}": np.asarray(softmax),
+            }
+            if pair in out and any(not np.allclose(out[pair][key], val) for key, val in value.items()):
+                raise ValueError(f"conflicting duplicate pooling state for {pair!r}")
+            out[pair] = value
+    return out
+
+
+def strict_node_disjoint_split(pairs, seed=0, held_frac=0.40):
```
s243a: **[important] Duplicated, un-stratified node-disjoint splitter** — *(Opus 4.8, Sonnet 5; Fable 5 as nit)*

This runner defines its own `strict_node_disjoint_split` (plain `rng.shuffle`, no candidate search, no stratification, post-hoc `--min-class` check) while `sim_matched_cost.py` and `run_sym_channel_fusion.py` use the audited, coverage-aware `node_disjoint_eval.node_disjoint_pair_split`. At nominally identical "seed 0 / held-frac 0.40" the two produce different partitions (377/168 vs 329/173 across the two reports), so the "same-split JointPosterior comparison" is on a different split, held-class balance is seed-fragile, and part of the seed-0-vs-seed-1 divergence highlighted in the PR may reflect the splitter rather than the method. Two near-identical splitters in one PR is also a maintenance hazard.

**Suggested fix:** call `node_disjoint_pair_split(pairs, seed, held_node_fraction=..., strata=hard_target, minimum_per_stratum=args.min_class)` and drop the local variant (keep the endpoint-leak assertion as a post-check) — or add a clause in REPORT_cheap_judge_joint_posterior.md noting the plain strict splitter is used and why.

prototypes/mu_cosine/joint_square_root_conditioner_torch.py
```diff
@@ -0,0 +1,767 @@
...
+        z = _matrix("prior_information_rhs", prior_information_rhs, rows=n)
+        b = _matrix("measurement_rhs", measurement_rhs, rows=m)
+        _same_backend(self.packed, prior_information_rhs=z, measurement_rhs=b)
+        if z.shape[-1] != b.shape[-1]:
+            raise ValueError("prior and measurement RHS counts must match")
+
+        batch_shape = torch.broadcast_shapes(
+            self.batch_shape, z.shape[:-2], b.shape[:-2]
+        )
+        packed = self.packed.expand(batch_shape + self.packed.shape[-2:]).contiguous()
```
s243a, line 208: **[important] Batched-design `geqrf`/`ormqr` path is never exercised on CUDA** — *(GPT 5.5)*

The distinct-design batched path expands `packed`/`tau` to a leading batch shape and calls batched `torch.ormqr`/`torch.geqrf` — the most fragile part of the backend, since cuSOLVER coverage for batched `geqrf`/`ormqr` has historically lagged the 2-D case and varies by PyTorch/CUDA build (the module docstring itself concedes this). Yet the only CUDA test folds a batch of observations into RHS columns of a *single* design (unbatched `ormqr`); both batched-distinct-design tests run CPU-only. The path most likely to raise `NotImplementedError` or misbehave on a real GPU is unverified.

**Suggested fix:** parametrize `test_batched_distinct_designs_match_unbatched_calls` and `test_full_correlated_conditioner_batches_distinct_designs` over `cuda` behind `skipif`, or document batched designs as CPU-only and guard the branch with a device check + clear error.

prototypes/mu_cosine/joint_square_root_conditioner_torch.py
```diff
@@ -0,0 +1,767 @@
...
+            covariance=self.posterior_covariance.expand(batch_shape + (n, n)),
+            gain=self.gain.expand(batch_shape + (n, m)),
+            innovation=innovation,
+            innovation_covariance=self.innovation_covariance.expand(
+                batch_shape + (m, m)
+            ),
+        )
+
+
+@dataclass(frozen=True)
```
s243a, line 752: **[important] Sequential root-threading — QR's core justification — is untested and unbenchmarked on GPU** — *(GPT 5.5)*

`SquareRootInformationStateTorch` sequential threading is only tested on CPU, and the benchmark times only the fixed-design `compiled.condition` loop. So "dense wins every tested cell" is measured entirely inside the regime the PR already concedes dense should win, while the regime where QR is claimed to earn its cost (changing/sequential/ill-conditioned blocks feeding `(U_post, z_post)` forward) is neither timed nor GPU-tested. The report acknowledges this and Follow-up #4 plans it, but the asymmetry deserves a line next to the PR-body table, not only in the report.

**Suggested fix:** add a CUDA sequential-state parity test, and either a minimal sequential timing cell or an explicit "not measured" annotation adjacent to the results table.

prototypes/mu_cosine/sim_matched_cost.py
```diff
@@ -45,41 +85,243 @@ def budget_plan(n, k, n_ov, avail):
...
+    """Paid-label-only convenience wrapper around :func:`nested_ridge_fit_predict`."""
+    X = np.asarray(X, dtype=float)
+    y = np.asarray(y, dtype=float)
+    return nested_ridge_fit_predict(
+        X, y, Xq, np.arange(len(X)), lams=lams, sample_ids=sample_ids,
+        split_seed=split_seed, return_info=return_info,
+    )
+
+
+def paired_delta_summary(baseline, candidate, *, n_boot=5000, confidence=0.95, seed=0):
```
s243a, line 288: **[important] Subsample intervals are Monte-Carlo precision, not inferential CIs — but two cells are bolded as "nominal S harm"** — *(Sonnet 5)*

`paired_delta_summary` percentile-bootstraps over 10 paired replicate deltas, but those replicates are training-subsample permutations on a single fixed held-node split (`rep_orders`), so the interval quantifies subsampling precision on one partition, not generalization uncertainty — and a percentile bootstrap over n=10 has poor coverage. REPORT_cheap_judge_post3648_validation.md discloses this clearly, yet its §1 table still bolds the "-0.050/-0.062" cells as nominal S harm, inviting inferential over-reading (conservative in direction, but still).

**Suggested fix:** de-bold those cells or relabel the column "subsample precision (not a CI)", and defer harm/benefit language to the preregistered node-partition repetition (Follow-up #1). Optionally raise `--reps` well above 10.

prototypes/mu_cosine/run_cheap_judge_joint_posterior.py
```diff
@@ -0,0 +1,632 @@
...
+
+    for a, b in pairs:
+        union(a, b)
+    by = {}
+    for i, (a, _b) in enumerate(pairs):
+        by.setdefault(find(a), []).append(i)
+    return [np.asarray(ix, int) for ix in by.values()]
+
+
+def component_bootstrap_aurc(proba, labels, pairs, B=500, seed=0):
```
s243a, line 237: **[nit] AURC significance: carry block counts with the claim, and consider AUGRC + small-n caveats** — *(Sonnet 5 + Deep Research)*

The paired AURC delta CI resamples endpoint-connected components of the held pair graph; a hub node can collapse many pairs into one component, shrinking effective resampling units. The joint report gives block counts inline ("126 blocks (largest 8)"), but the §4 summary table where the bolded seed-1 "+.0384 [+.0056,+.0781]" significance claim lives omits them — a block-bootstrap significance claim is only interpretable with its block count.

Two literature caveats: AURC over-weights early high-confidence failures and can favour a worse classifier — AUGRC changed top-3 rankings on 5/6 benchmarks ([NeurIPS 2024](https://proceedings.neurips.cc/paper_files/paper/2024/file/047c84ec50bd8ea29349b996fc64af4b-Paper-Conference.pdf)); and empirical AURC estimators carry finite-sample bias at small n ([arXiv:2410.15361](https://arxiv.org/html/2410.15361v4)).

**Suggested fix:** add block-count/largest-block columns to the §4 table; rerun the paired comparison with AUGRC using the same bootstrap machinery (or report whether a handful of highest-confidence errors drive the seed-1 result); note the small-n estimator bias next to the interval.

prototypes/mu_cosine/benchmark_joint_square_root_qr.py
```diff
@@ -0,0 +1,166 @@
...
+        raise SystemExit("no requested benchmark device is available")
+    return devices
+
+
+def _synchronise(device: torch.device) -> None:
+    if device.type == "cuda":
+        torch.cuda.synchronize(device)
+
+
+def _time_ms(function, repeats: int, warmups: int, device: torch.device) -> float:
```
s243a, line 61: **[nit] Single-run, mean-only, compute-only timings** — *(GPT 5.5, Sonnet 5)*

`_time_ms` returns a bare mean over `repeats` from one process run — no median/MAD/min — and `_problem` allocates tensors on-device with results never leaving the device, so no transfer cost is included. For ~0.13 ms dense cells, a single mean is easily perturbed by launch jitter, and near-ties (n=2,m=4: 0.113 vs 0.132) could flip within noise, which makes "faster in every tested cell" fragile as stated. The sync placement and warmups are correct, and the report is candid — but the PR-body table carries none of these caveats.

**Suggested fix:** report median + MAD (or min) over ≥5 independent runs, interleave algorithm order per cell, and annotate the PR-body table as compute-only, on-device point measurements (Follow-up #4 already plans most of this).

prototypes/mu_cosine/joint_square_root_conditioner_torch.py
```diff
@@ -0,0 +1,767 @@
...
+    NumPy prototype's default ``1e-9`` floor, while float32 automatically uses
+    a representable stability floor.
+    """
+    if jitter < 0:
+        raise ValueError("jitter must be nonnegative")
+    covariance = _matrix(name, covariance)
+    if covariance.shape[-2] != covariance.shape[-1]:
+        raise ValueError(f"{name} must be square")
+    covariance = 0.5 * (covariance + covariance.transpose(-2, -1))
+    eigenvalues = torch.linalg.eigvalsh(covariance)
```
s243a, line 342: **[nit] Functional path forces an eigendecomposition + host sync per call** — *(Opus 4.8, Fable 5)*

`_tensor` deliberately avoids `isfinite(...).item()` to keep the batched path sync-free, but `regularize_covariance_torch` calls `torch.linalg.eigvalsh(...)` + `.item()`, and `condition_correlated_gaussian_qr_torch` reaches it via `conditional_measurement_model_torch` on every invocation — so the functional entry point pays an eigendecomposition and a device→host sync per call, contradicting the module's stated no-sync hot-path design. The compiled classes correctly do this once at `compile`.

**Suggested fix:** document the functional entry point as compile-per-call (not the throughput path), or add a `skip_validation`/precompiled route so repeated functional calls avoid the sync.

prototypes/mu_cosine/joint_square_root_conditioner_torch.py
```diff
@@ -0,0 +1,767 @@
...
+def _normalised_root_from_packed(
+    packed: torch.Tensor, n: int
+) -> tuple[torch.Tensor, torch.Tensor]:
+    raw_root = torch.triu(packed[..., :n, :n])
+    diagonal = torch.diagonal(raw_root, dim1=-2, dim2=-1)
+    signs = torch.where(diagonal < 0, -torch.ones_like(diagonal), torch.ones_like(diagonal))
+    return raw_root * signs.unsqueeze(-1), signs
+
+
+def _check_root_rank(pre_array: torch.Tensor, root: torch.Tensor) -> None:
```
s243a, line 117: **[nit] `_check_root_rank` aborts the whole batch without identifying the offender** — *(GPT 5.5)*

`bool(torch.any(smallest <= tolerance).item())` forces a device→host sync and raises for the entire batch if any single element is rank-deficient, with no index of which design failed. Fine on the compile path, but coarse for large batches; the relative tolerance (`eps · max(shape) · ‖pre_array‖_F`) can also false-flag a legitimately weak-but-full-rank precision direction (very diffuse prior).

**Suggested fix:** include the offending batch indices (`torch.nonzero(smallest <= tolerance)`) in the exception message; optionally offer a non-raising validation entry point.

prototypes/mu_cosine/joint_square_root_conditioner_torch.py
```diff
@@ -0,0 +1,767 @@
...
+            ("H", H),
+            ("observation_covariance", observation_covariance),
+            ("cross_covariance", cross_covariance),
+        ):
+            if not isinstance(value, torch.Tensor) or value.ndim != 2:
+                raise ValueError(f"{name} must be one unbatched matrix")
+        # Compiled objects own a snapshot of coefficient inputs. Otherwise an
+        # in-place caller mutation of H would combine a new innovation with
+        # stale QR reflectors.
+        prior_precision_root = prior_precision_root.clone()
```
s243a, line 556: **[nit] `R`/`C` caller-mutation safety is correct-by-construction but untested** — *(GPT 5.5)*

`compile` clones `prior_precision_root` and `H`, and the mutation test only mutates `H`. `R` and `C` are not cloned — safe today because they are consumed into derived tensors and never stored, but that invariant is load-bearing: a future refactor that retains `observation_covariance`/`cross_covariance` on the dataclass would silently reintroduce a mutation bug.

**Suggested fix:** extend `test_compiled_conditioners_snapshot_H_against_caller_mutation` to also `R.add_(...)`/`C.add_(...)` after compile and assert the conditioned mean is unchanged.

s243a (commented):
## Council review — approve (non-blocking findings only)

Independent review by four models — **Claude Opus 4.8** (numerics/architecture), **Claude Sonnet 5** (methodology/leakage/statistics), **GPT 5.5** (PyTorch/CUDA & benchmark rigor), **Claude Fable 5** (holistic adversarial) — plus a **Deep Research** literature pass (web + academic). All four code reviewers independently recommend **approve**; findings below are non-blocking.

**What was verified (multiple seats, independently):**
- The Schur/sign math (`Rc = R - C^T P^-1 C`, `J = H + C^T P^-1`) is correct in both backends; Opus 4.8 confirmed posterior mean/covariance against the independent dense reference to ~1e-9, and GPT 5.5/Fable 5 re-derived the conditional-decorrelation identity. The formulation is algebraically identical to the cross-covariance Kalman gain in Simon, *Optimal State Estimation* ([notes, eq. for correlated noise](https://www.lukesy.net/docs/self-study/simon-ose/simon-ose-notes.pdf)) — worth citing in the design doc.
- The leakage fix is genuinely nested (calibration, covariance, pseudo-targets, ridge, scaler all confined to the paid inner-train fold; held rows isolated) and is directly unit-tested at the exact invariant. The retraction of the spending headline under leakage-free evaluation is exactly what [Cawley & Talbot (JMLR 2010)](https://www.jmlr.org/papers/volume11/cawley10a/cawley10a.pdf) predicts.
- No information double-counting; QR=dense parity holds with nonzero C; reports are internally consistent and candid.

**Literature caveats (Deep Research seat):**
- AURC has documented structural flaws — it can favour a worse classifier by over-weighting early high-confidence failures; switching to AUGRC changed top-3 rankings on 5/6 benchmarks ([NeurIPS 2024](https://proceedings.neurips.cc/paper_files/paper/2024/file/047c84ec50bd8ea29349b996fc64af4b-Paper-Conference.pdf)). Empirical AURC estimators also carry finite-sample bias at small n ([population-AURC characterization](https://arxiv.org/html/2410.15361v4)), relevant at n=160. An AUGRC robustness check reusing the existing paired-bootstrap machinery would strengthen the JointPosterior-vs-Gaussian+bridge conclusion.
- Node-disjointness governs pair endpoints/labels, not the ambient graph topology used for features (`sym_graph_features` walks the full parent graph); a one-line scope note would prevent readers over-reading "node-disjoint", and it is worth stating whether the split is edge-disjoint / k-hop-isolated at the hop depth the graph judge uses.
- The dense-vs-QR trade-off matches 50 years of square-root-filtering literature (Potter/Carlson/Bierman; a [NASA comparative study](https://ntrs.nasa.gov/api/citations/19760019842/downloads/19760019842.pdf) found square-root filters numerically reliable but the most expensive tested), and the small-matrix batched-GPU regime matches the [MAGMA/batched-LA literature](https://ieeexplore.ieee.org/document/8476161/) — the PR's hedged framing is the right level of caution.

**Body-only note (lines outside this diff):** `calibrate_luna` / `sym_graph_features` now exist in three/two copies across `run_sym_channel_fusion.py`, `run_cheap_judge_joint_posterior.py`, and `fine_tune_fused_head_luna.py`; divergence risk is real for leakage-sensitive code — consider hoisting into a shared module (Sonnet 5).

Inline comments below are attributed per seat; severities: 4 important, 8 nit. Full seat reports available on request.

*Posted by Perplexity Computer on behalf of @s243a after a multi-model council review.*

prototypes/mu_cosine/joint_square_root_conditioner.py
```diff
@@ -0,0 +1,301 @@
...
+    )
+    return effective_H, conditional_R
+
+
+def whiten_measurement_block(measurement_matrix, measurement_covariance, measurement_rhs,
+                             jitter=1e-9):
+    """Whiten one independent likelihood block with a Cholesky solve."""
+    J = _matrix("measurement_matrix", measurement_matrix)
+    m, _ = J.shape
+    Rc = regularize_covariance(
```
s243a: **[nit] Conditional covariance is regularized twice in the NumPy path** — *(Opus 4.8)*

`conditional_measurement_model` already returns a `regularize_covariance`-lifted `Rc`, then `whiten_measurement_block` regularizes it again before the Cholesky. Harmless but it double-applies the jitter floor and obscures which floor is actually in effect.

**Suggested fix:** pass the already-regularized `Rc` to a whitening helper that factors without re-regularizing, or make the second call a no-op for PD input. Related: the PSD-rejection tolerance (`min_eig < -100·floor`) accepts and lifts an `Rc` with eigenvalue ≈ -1.2e-5 in float32 — worth a boundary-pinning test so the tolerance is an explicit, reviewed choice.

prototypes/mu_cosine/node_disjoint_eval.py
```diff
@@ -0,0 +1,304 @@
...
+        raise RuntimeError(
+            f"only {len(replicates)}/{n_resamples} non-empty node-bootstrap draws "
+            f"after {attempts} attempts"
+        )
+
+    boot = np.asarray(replicates)
+    alpha = (1.0 - confidence) / 2.0
+    low, high = np.quantile(boot, [alpha, 1.0 - alpha])
+    return NodeBootstrapInterval(
+        estimate=float(values.mean()),
```
s243a, line 306: **[nit] Unweighted point estimate paired with a multiplicity-weighted bootstrap interval** — *(Fable 5)*

`estimate` is the plain row mean of `values`, but each bootstrap replicate is a multiplicity-weighted ratio mean with pigeonhole endpoint weights. Under the uniform multinomial the two estimands coincide in expectation (and `bootstrap_mean` is returned for auditing), but the headline graph_S intervals pair the unweighted point with the weighted CI — if high-degree nodes carry systematically different gains, the interval need not be centered on the reported estimate.

**Suggested fix:** report the weighted ratio mean as the point estimate for consistency, or add one docstring sentence stating the estimate is unweighted while the interval is multiplicity-weighted and they agree only in expectation.

prototypes/mu_cosine/REPORT_cheap_judge_joint_posterior.md
```diff
@@ -0,0 +1,131 @@
...
+`corr(graph_D,luna_D)=+0.60`, and `corr(prior_D,graph_D)=+0.58`. On fresh, the corresponding graph-D/Luna-D
+correlation is +0.56. The learned joint head and fitted cross-covariance can price this redundancy; a factored
+control cannot.
+
+The separability-weighted factored control illustrates why one metric is insufficient: on fresh it has high
+accuracy and low AURC, but its log-loss 0.853 and ECE 0.341 reveal severe miscalibration. It is not a preferred
+combiner.
+
+The hard macro target resolves exact ties in fixed class order (directional, symmetric, other). Across the
+1,770-row campaign there are 7 exact top ties and 41 normalized top-two margins below 0.02. The result JSON now
```
s243a: **[nit] Row-count mislabel: 1,770 vs 1,700** — *(Fable 5)*

The tie/margin audit counts (7 exact top ties, 41 margins < 0.02) come from `audit_counts` over `decision_rows` built from the luna-matched `keep` set — 1,700 rows (the split table in this report sums to 1000 + 700, and REPORT_luna_campaign.md independently confirms 1,700). The "1,770" here appears to be the raw `campaign_scored.tsv` count used only for the `cur_rel` degeneracy check.

**Suggested fix:** change to "1,700-row matched campaign" (and if 1,770 is the raw file count, say so explicitly at line 14 to avoid conflation).

PR #3666, opened 2026-07-12
mu_cosine: batched square-root/QR conditioning with correlation whitening

s243a:
## Summary

- add a batched CPU/CUDA noise whitener that accepts covariance or correlation plus marginal standard deviations
- convert measurement correlation to an implicit precision row root with Cholesky triangular solves; no explicit inverse is formed
- expose correlated measurement blocks through `SquareRootInformationStateTorch.update_noise_block()`
- support stacked and streamed Householder-QR information updates over explicit `n`, `m_block`, `T`, and `B` axes
- add scale-relative covariance loading with observable diagnostics and hard rejection of materially indefinite models
- use a Joseph-form dense posterior covariance and an independent direct least-squares/QR benchmark reference
- document the correlation-to-precision-root algebra, conditional-`Rc` block-independence requirement, and semantic/graph covariance follow-up

## Why

For `Rc = L L.T`, apply the measurement precision row root implicitly:

```
A = solve(L, J)
b = solve(L, r)
```

The posterior precision is a sum,

```
U_post.T U_post = U.T U + J.T Rc^-1 J,
```

so the correct operation is QR triangularization of `stack([U, A])`. Multiplying a Cholesky factor of `Rc^-1` by the previous state root is generally dimensionally and statistically wrong: the factors live in different spaces, and Bayesian assimilation adds information rather than multiplying precisions.

This is the classical square-root-information vertical-stack pattern described by [Tracy (2022)](https://arxiv.org/abs/2208.06452) and the [CERFACS SRIF presentation](https://cerfacs.fr/wp-content/uploads/2017/05/Alina_presentation_kf.pdf). The nonzero-cross-covariance Schur reduction in this PR is documented separately.

## Review follow-up

This revision addresses the Perplexity council review:

- removes the duplicate pre-array scale reduction/device synchronization
- restores conservative Frobenius-scale rank sensitivity without a unit clamp, while comparing normalized quantities so the threshold cannot overflow near the dtype maximum
- adds `n=128,m=32` rank-boundary tests plus global-scale, near-maximum, and subnormal-scale regressions
- replaces the float64 normal-equations “gold” solve with direct stacked `torch.linalg.lstsq`; a separate reduced QR supplies the positive-diagonal reference root
- couples the default negative-eigenvalue validity tolerance to the remaining loading budget, while preserving explicit overrides and separate rejection gates
- renames the explicit lower row-root helper to `materialize_row_root()`
- hardens expanded batched triangular solves and documents the hot-path NaN/Inf policy
- adds compute/transfer, CUDA-validation, CPU-memory, single-seed, and hardware-specific caveats to the report
- cites the square-root-information literature in the design document

## Benchmark result — local prototype evidence

The recorded timing tables are local, compute-only measurements on a GTX 1660 SUPER and predate the post-review rank-guard cleanup. Transfers are excluded, and the corrected pre-cast transfer benchmark has not been rerun, so no transfer-inclusive crossover is established.

At `n=128`, `m_block=32`, `T=16`, and shared-design `B=4096`, stacked QR measured **3.939 ms CUDA versus 10.392 ms CPU**. It did not cross over at `B=128`, and distinct-design batched QR was slower on this local GPU. This is not a general `B=4096` recommendation.

In one seeded synthetic stress instance with target prior/within-block covariance condition number `1e6`, the historical run showed a float32 dense Cholesky failure in one scale regime and `1.50e-4` solution error in another, while QR remained near `1e-6`. Those CUDA error rows predate the current independent float64 stacked least-squares/QR reference and were not rerun. Old and new float64 references agree to `2.51e-13` or better on the tested solutions, so the historical magnitudes remain representative, but they show one concrete failure mode—not its frequency.

## API behavior worth noting

- The Torch square-root and conditional-conditioning helpers now default `jitter` to `0.0` instead of the historical absolute `1e-9` floor. Scale-relative loading remains active and observable; a near-singular covariance previously rescued by absolute jitter can now be rejected by the validity or loading-budget gate.
- Preparation validates covariance/design finiteness. Repeated compiled RHS paths avoid device-synchronizing finiteness checks, so NaN/Inf observations propagate unless callers gate them at batch ingress.

## Validation

- `70 passed, 9 skipped` across Product-Kalman and NumPy/Torch square-root suites
- all 9 skips are CUDA-only correctness cases and are not currently verified by CI
- CPU shared/distinct benchmark smoke runs pass against the direct least-squares/QR reference
- Python compilation and `git diff --check` pass
- final independent audit found no remaining actionable correctness issue

## Follow-up

- rerun the corrected transfer benchmark on CUDA before making an end-to-end crossover claim
- add CUDA CI or cross-device correctness coverage and a multi-seed conditioning sweep
- fit PSD semantic/graph residual kernels on held-out node-disjoint calibration data, then compare full-`Rc`, structured, and block-diagonal approximations
- separately test the graph judge as dataset-adaptation supervision—helping the model learn a dataset's entities and topology—not only as an input to the final judgment

### DIFF
--- a/prototypes/mu_cosine/REPORT_streamed_block_qr_benchmark.md
+++ b/prototypes/mu_cosine/REPORT_streamed_block_qr_benchmark.md
@@ -1,0 +1,301 @@
+# Batched measurement blocks with the joint square-root / QR conditioner
+
+Run 2026-07-11. This report benchmarks the sequential regime left open by
+`REPORT_joint_square_root_qr_benchmark.md`.
+
+## Result in one paragraph
+
+Batching several available measurement blocks into **one stacked Householder QR** is substantially faster than
+launching one QR per block. CUDA becomes worthwhile for QR when a fixed design has thousands of independent
+right-hand sides: at `n=128`, 32 measurements per block, 16 blocks, and `B=4096`, full stacked QR fell from
+`10.392 ± 0.197 ms` on CPU to `3.939 ± 0.034 ms` on CUDA. It did **not** cross over at `B=128`, and batched QR
+of 128 distinct designs was slower on the local GTX 1660 SUPER than CPU QR. Dense normal-equation updates were
+usually faster, but in the conditioning stress test their float32 Cholesky failed in one regime and incurred
+`1.50e-4` relative solution error in another, while Householder QR stayed near `1e-6`. The resulting policy is:
+stack blocks that are already available, stream only genuinely online/adaptive blocks, retain QR for its
+stability, and use CUDA after measuring the actual shared-versus-distinct design shape.
+
+This square-root conditioner is not `JointPosterior`. `JointPosterior` is a learned nonlinear decision model;
+the conditioner is a numerical backend for a correlated linear-Gaussian posterior. They can be used together.
+
+## Four dimensions, not two
+
+The benchmark uses separate symbols for quantities that are easy to conflate:
+
+| symbol | meaning | values exercised |
+|---|---|---|
+| `n` | latent-state dimension | `2, 32, 128` |
+| `m_block` | measurements/judges in one block | `32` |
+| `T` | blocks assimilated into one posterior | selected `1,4,16`; CLI also exposes `2,8` |
+| `B` | independent posterior systems processed together | `128`, plus the `4096` crossover cell |
+| `M_total` | total likelihood rows, `T * m_block` | `32,128,512` |
+
+Thus the proposed `n=128,m=32` case becomes a `(n + Tm_block) × n` information pre-array when all blocks are
+stacked. `B` is a separate CUDA batch/RHS axis.
+
+Two design regimes matter:
+
+- **shared:** all `B` systems use the same block coefficients and differ only in their right-hand sides. One
+  factorization can serve all systems.
+- **distinct:** every system has its own block coefficients, as can happen when covariance or observation
+  geometry depends on the item. CUDA must perform `B` distinct factorizations.
+
+## Statistical construction and algorithms
+
+Each synthetic block starts with a dense within-block conditional covariance `Rc_t`. It is Cholesky-whitened to
+
+```text
+A_t = chol(Rc_t)^-1 J_t
+b_t = chol(Rc_t)^-1 r_t
+```
+
+and conditional cross-block covariance is exactly zero. Whitening and covariance estimation are outside the
+timed region. That isolates the conditioner; it also means this benchmark does **not** yet test semantic- or
+graph-distance cross-block correlation.
+
+All algorithms start with `P^-1 = U.T @ U` and zero information RHS:
+
+1. **`full_stacked_qr`:** one Householder QR of `stack([U,A_1,...,A_T])`.
+2. **`streamed_qr`:** `T` Householder QR calls, carrying `(U_t,z_t)` into the next block.
+3. **`dense_recompute`:** accumulate `Lambda += A_t.T @ A_t` and `eta += A_t.T @ b_t`, Cholesky-refactoring
+   the normal equations after every block. This is a speed baseline, not a square-root-stability method.
+4. **`cached_dense_full`:** for a shared, completely known design only, factor the final normal equations once
+   and apply new RHS batches. It does not produce the intermediate online roots.
+
+The stacked and streamed QR paths produce the same posterior in exact arithmetic. Stacking wins when all blocks
+are already present because it replaces `T` factorizations and launches with one larger operation. Streaming is
+still necessary when a block arrives later, depends on an earlier posterior, or an intermediate root is itself
+the required output.
+
+## Hardware and timing protocol
+
+- CPU: Intel Core i7-10700KF, eight PyTorch threads.
+- GPU: NVIDIA GeForce GTX 1660 SUPER, 6 GiB.
+- PyTorch 2.4.1+cu121; CUDA runtime 12.1; WSL2.
+- Nominal workload: float32, target prior/within-block covariance condition number 10, conditional-noise scale
+  1, seed `20260711`.
+- CUDA timing synchronizes around every trial. Tables report median ± median absolute deviation (MAD).
+- Input H2D and result D2H copies are timed separately and excluded from trajectory latency. The current harness
+  pre-casts on the host so H2D measures transfer rather than transfer plus a CUDA-only dtype conversion.
+- Peak CUDA memory is incremental temporary allocation, not the allocator's reserved pool; it excludes any
+  compiled object already resident at the start of a conditioning trial.
+- The final `n=128,B=128` CUDA cells use 11 trials and CPU cells use 7; the smaller-state table uses 11. The
+  `B=4096` crossover uses 11 trials. All have warmups. Exact commands appear below.
+
+### Authoritative-run policy
+
+The compute tables use the **final recorded runs**: compute receives the configured warmups and CUDA's caching
+allocator is not emptied between timed trials. Peak allocation is measured with `reset_peak_memory_stats`
+without changing that steady-state timing protocol.
+
+These runs supersede exploratory figures produced while the harness still called `empty_cache()` before each
+trial and before transfer timing had a warmup. In particular, do not mix the superseded
+`2.934 / 196.941 ms` `B=128` CUDA values or the `10.666 / 3.967 ms`, `H2D=4.375 ms` `B=4096` values with the
+tables below. The authoritative corresponding compute values are `3.424 / 193.247 ms` and
+`10.392 / 3.939 ms`.
+
+The recorded `H2D=4.855 ms` value also included float64-to-float32 conversion. A post-review harness correction
+now pre-casts identically for CPU and CUDA, so that value is retained below only as a conservative historical
+upper bound; the current command must be rerun on a CUDA host for a transfer-only number. Compute timings are
+unaffected by this correction.
+
+## `n=128`, `m_block=32`, `B=128`
+
+### Shared design
+
+Trajectory median ± MAD, milliseconds. Cached dense is shown as `setup + condition`.
+
+| `T` | CPU stacked QR | CUDA stacked QR | CPU streamed QR | CUDA streamed QR | CPU dense recompute | CUDA dense recompute | CPU cached dense | CUDA cached dense |
+|---:|---:|---:|---:|---:|---:|---:|---:|---:|
+| 1 | 0.719 ± .005 | 3.147 ± .088 | 0.749 ± .017 | 2.980 ± .090 | 0.409 ± .010 | 0.767 ± .007 | .136 + .165 | .415 + .311 |
+| 4 | 0.873 ± .006 | 2.743 ± .179 | 3.075 ± .110 | 10.474 ± .342 | 1.484 ± .036 | 2.382 ± .087 | .173 + .205 | .410 + .329 |
+| 16 | **1.273 ± .024** | 3.424 ± .095 | 11.033 ± .495 | 39.879 ± 1.281 | 5.731 ± .244 | 9.574 ± .637 | .177 + .218 | .406 + .335 |
+
+At this batch size the CPU wins every shared-design QR cell. At `T=16`, stacking is 8.7× faster than streaming
+on CPU and 11.6× on CUDA. The cached-dense row is a static lower bound, not a matched cache comparison: this
+sequential harness does not yet time cached full/streamed QR chains. Compare `setup + condition` for a one-shot
+call; use `REPORT_joint_square_root_qr_benchmark.md` for the matched repeated fixed-design comparison.
+
+### Distinct design per system
+
+| `T` | CPU stacked QR | CUDA stacked QR | CPU streamed QR | CUDA streamed QR | CPU dense recompute | CUDA dense recompute |
+|---:|---:|---:|---:|---:|---:|---:|
+| 1 | 24.356 ± 1.412 | 113.350 ± .362 | 24.472 ± .513 | 113.111 ± .261 | 22.207 ± 4.185 | **1.625 ± .038** |
+| 4 | 34.799 ± .750 | 119.602 ± .466 | 109.855 ± 11.635 | 454.403 ± 2.043 | 45.038 ± 1.640 | **6.152 ± .103** |
+| 16 | **66.145 ± .878** | 193.247 ± .788 | 422.071 ± 11.815 | 1804.537 ± 4.470 | 177.832 ± 4.051 | **23.828 ± .187** |
+
+The local GPU's batched `geqrf` path is a poor fit for 128 distinct tall-skinny factorizations: CPU
+Householder wins decisively. CUDA does accelerate the dense batched Cholesky baseline. This is a backend and
+hardware result, not evidence against QR's numerical role.
+
+For the distinct `T=16` CUDA cell, the historical cast-plus-H2D upper bound is `15.097 ± .239 ms`; D2H is about
+`1.6 ms`. Peak incremental
+allocation is 160.94 MiB for stacked QR, 46.61 MiB for streamed QR, and 32.19 MiB for dense recomputation.
+Streaming trades substantially more time for bounded memory as `T` grows.
+
+## The shared-design CUDA crossover: `B=4096`
+
+At `n=128,m_block=32,T=16`, increasing the shared RHS batch from 128 to 4096 produces the expected GPU
+crossover:
+
+| algorithm | CPU trajectory ms | CUDA trajectory ms | CUDA peak MiB |
+|---|---:|---:|---:|
+| full stacked QR | 10.392 ± .197 | **3.939 ± .034** | 34.95 |
+| streamed QR | 54.951 ± 2.336 | **39.146 ± .407** | 17.31 |
+| dense recompute | 54.164 ± 3.870 | **20.464 ± .066** | 10.19 |
+| cached dense full | .183 setup + 2.812 condition | **.492 setup + 1.220 condition** | 6.25 |
+
+The historical cast-plus-input-transfer upper bound is `4.855 ± .171 ms`; output transfer is `.74–.93 ms`.
+Even with that conservative surcharge, stacked QR barely wins end-to-end if every batch begins and ends on the
+host; device residency materially strengthens the CUDA case. The current pre-cast harness needs a CUDA rerun
+before quoting a transfer-only crossover. Cached dense loses its end-to-end CUDA advantage under the historical
+cast-plus-transfer protocol.
+
+## Smaller states
+
+The selected distinct-design grid shows the same qualitative result at `B=128,T=16`:
+
+| `n` | CPU stacked QR ms | CUDA stacked QR ms | CPU streamed QR ms | CUDA streamed QR ms | CPU / CUDA dense recompute ms |
+|---:|---:|---:|---:|---:|---:|
+| 2 | **.808** | 5.958 | **10.387** | 54.136 | **2.388** / 7.739 |
+| 32 | **14.749** | 46.922 | **64.750** | 420.016 | **9.604** / 10.287 |
+| 128 | **66.145** | 193.247 | **422.071** | 1804.537 | 177.832 / **23.828** |
+
+There is no general rule that a larger pre-array automatically makes CUDA QR faster. Shared reflector reuse,
+the number of independent RHSs, distinct versus common designs, GPU generation, and transfer all matter.
+
+## Numerical behavior
+
+Every output is compared with a float64 dense full-trajectory reference. The CSV reports:
+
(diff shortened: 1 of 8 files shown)


prototypes/mu_cosine/joint_square_root_conditioner_torch.py
```diff
@@ -295,6 +426,7 @@ def compile_information_update_torch(
     _same_backend(root, measurement_matrix=matrix)
     root, matrix = _broadcast_matrices(root, matrix)
     pre_array = torch.cat([root, matrix], dim=-2).contiguous()
+    _pre_array_scale(pre_array)
```
s243a: **[important] Redundant `_pre_array_scale` call doubles host syncs on every QR compile** — *(GPT 5.5; Opus 4.8 and Sonnet 5 concur)*

This bare `_pre_array_scale(pre_array)` discards its return value, and `_check_root_rank` two lines later re-invokes it. Each invocation performs three `bool(torch.any(...).item())` device→host syncs, so every `compile_information_update_torch` pays ~6 sync round-trips where 3 suffice — squarely in the streamed hot path (`SquareRootInformationStateTorch.update` → `householder_information_update_torch` runs this per block, and `_streamed_qr` in the benchmark calls it per block per trial). On CUDA each `.item()` serializes the stream, so this directly inflates the streamed timings the benchmark attributes to launch overhead. Same double-call at `:767` in `precision_root_from_covariance_torch`.

**Suggested fix:** delete this line (the finite/zero/subnormal guard already runs inside `_check_root_rank`), or have `_check_root_rank` accept a precomputed `scale`. Pure win, one line.

prototypes/mu_cosine/benchmark_streamed_block_qr.py
```diff
@@ -0,0 +1,745 @@
...
+        problem.root,
+        torch.zeros(batch, n, dtype=problem.root.dtype, device=problem.root.device),
+        blocks,
+        rhs,
+    )
+    root = result.precision_root[0] if problem.shared_design else result.precision_root
+    return Result(root, result.information_rhs, result.solution)
+
+
+def _streamed_qr(problem: Problem) -> Result:
```
s243a, line 356: **[important] Streamed timings conflate validation-guard host syncs with kernel-launch overhead** — *(GPT 5.5, Sonnet 5)*

`_streamed_qr` calls `compile_information_update_torch` per block, so each timed iteration bundles (a) genuine per-launch overhead, (b) ~7 blocking `.item()` syncs from `_pre_array_scale`/`_check_root_rank` (see the conditioner comment), and (c) a redundant scale recomputation. The stacked path amortizes all of this into one compile. A production streaming loop with a fixed design would hoist (b) and (c), so part of the reported 8.7x/11.6x stacking win — and of the small-batch CUDA losses at B=128 — is a harness artifact rather than intrinsic launch cost. The docstring's claim that the benchmark "measures the information-conditioner core" is only partially true on CUDA. CPU ratios are largely real since `.item()` is cheap relative to the n=128 geqrf work there.

**Suggested fix:** add a "streamed, guards hoisted" variant (or a `validate=False` fast path on compile) to isolate true launch overhead, or add a sentence to REPORT_streamed_block_qr_benchmark.md noting streamed timings include per-block validation syncs a fixed-design production loop would not repeat.

prototypes/mu_cosine/test_joint_square_root_conditioner_torch.py
```diff
@@ -42,6 +43,300 @@ def _as_torch(*arrays, device="cpu", dtype=torch.float64):
...
+            singular,
+            relative_eigenvalue_floor=1e-3,
+            maximum_relative_loading=1e-4,
+        )
+
+
+@pytest.mark.parametrize("device", [
+    "cpu",
+    pytest.param(
+        "cuda",
```
s243a, line 139: **[important] The #3651 CUDA gap is not closed — all 9 skips are exactly the CUDA correctness checks** — *(GPT 5.5)*

Every CUDA parametrization in this suite (here, `:210`, and the batched-distinct-design, full-correlated, sequential-threading, and `test_cuda_geqrf_ormqr_matches_cpu` cases) is skip-gated on `torch.cuda.is_available()`, and in the PR's validation environment CUDA is unavailable — so the reported "60 passed, 9 skipped" skips precisely the batched geqrf/ormqr and sequential root-threading checks that the #3651 review asked to close. The `expand(...)`-without-`.contiguous()` factors passed to `solve_triangular`/`ormqr` (e.g. `NoiseWhitenerTorch.apply_columns`) are exactly the stride-0 batched inputs that can diverge between the CPU backend and cuSOLVER, so CPU passes are not evidence for CUDA. The headline B=4096 crossover therefore rests on one unreproduced local GTX 1660 SUPER run.

**Suggested fix:** state explicitly in the PR body that CUDA is unverified in CI, and capture at least one CUDA run (even a single small geqrf/ormqr + sequential-threading case) in CI logs or a documented artifact before the crossover numbers are cited as load-bearing.

prototypes/mu_cosine/REPORT_streamed_block_qr_benchmark.md
```diff
@@ -0,0 +1,301 @@
...
+crossover:
+
+| algorithm | CPU trajectory ms | CUDA trajectory ms | CUDA peak MiB |
+|---|---:|---:|---:|
+| full stacked QR | 10.392 ± .197 | **3.939 ± .034** | 34.95 |
+| streamed QR | 54.951 ± 2.336 | **39.146 ± .407** | 17.31 |
+| dense recompute | 54.164 ± 3.870 | **20.464 ± .066** | 10.19 |
+| cached dense full | .183 setup + 2.812 condition | **.492 setup + 1.220 condition** | 6.25 |
+
+The historical cast-plus-input-transfer upper bound is `4.855 ± .171 ms`; output transfer is `.74–.93 ms`.
```
s243a, line 175: **[important] PR body quotes the compute-only crossover without this section's transfer caveat** — *(Fable 5, Sonnet 5, GPT 5.5; Deep Research concurs)*

This report is candid: with the conservative historical H2D surcharge (~4.855 ms, which included a float64→float32 cast) plus ~0.74–0.93 ms D2H, stacked QR "barely wins end-to-end" (~9.6 ms vs 10.392 ms CPU) — and a transfer-only H2D number still needs a rerun. But the PR body's "Benchmark result" section states "3.939 ms CUDA vs 10.392 ms CPU" with no qualifier, which a reader will take as an end-to-end 2.6x win and draw the wrong deployment conclusion from.

**Suggested fix:** label the PR-body numbers "compute-only" and carry over this section's one-line caveat (device residency is required for a material CUDA advantage); ideally rerun the documented CUDA command with the corrected pre-cast harness to quote a genuine transfer-only H2D.

prototypes/mu_cosine/joint_square_root_conditioner_torch.py
```diff
@@ -114,9 +118,34 @@ def _normalised_root_from_packed(
     return raw_root * signs.unsqueeze(-1), signs
 
 
+def _pre_array_scale(pre_array: torch.Tensor) -> torch.Tensor:
+    # Use a homogeneous scale: clamping it to one falsely labels a perfectly
+    # conditioned factor such as 1e-20 * I as rank deficient.  Max-entry scale
+    # also avoids squaring underflow in a Frobenius norm.
+    scale = torch.amax(torch.abs(pre_array), dim=(-2, -1))
```
s243a: **[important] Rank-tolerance scale switch silently lowers detection sensitivity for large well-conditioned designs** — *(Fable 5)*

Switching the scale from a Frobenius norm clamped ≥1 to max-abs-entry (with `tolerance = eps · max_dim · scale`) correctly fixes the tiny-but-full-rank misclassification, and homogeneity under uniform rescaling is real (verified at 1e±150). But for a normal-scale design the Frobenius norm is ~√(rows·cols) larger than the max entry, so the tolerance drops by ~12.6x for the n=128, T=1 pre-array — a marginally rank-deficient design that #3651 would have rejected can now pass silently and return an unstable solution. This is changed behavior in an existing path with no boundary test: the new tests only exercise uniform rescaling of full-rank inputs. Related nit: the single global scale means "scale-homogeneous" holds under global rescaling only — a design mixing very different per-direction scales can still trip or miss detection (DESIGN acceptance #7 sweeps scale but not condition number). Same change applies to the NumPy `_scaled_norm` counterpart.

**Suggested fix:** add a boundary regression pair (well-conditioned n≈128 design with smallest singular value just above / just below the new tolerance), note the sensitivity change in the PR body's "removes two scale bugs" bullet (it reads as pure improvement), and document that homogeneity is under global rescaling.

prototypes/mu_cosine/benchmark_streamed_block_qr.py
```diff
@@ -0,0 +1,745 @@
...
+    return CachedDenseFull(cholesky.mT, cholesky, problem.blocks)
+
+
+def _reference(problem: Problem) -> tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
+    """Float64 dense full-trajectory reference on CPU."""
+    batch, block_count, block_size = problem.rhs.shape
+    n = problem.root.shape[-1]
+    if problem.shared_design:
+        blocks = problem.blocks.reshape(block_count * block_size, n)
+        precision = problem.root.mT @ problem.root + blocks.mT @ blocks
```
s243a: **[important] The float64 "gold standard" is itself a normal-equations solve — the method family whose instability is the headline** — *(Sonnet 5; Deep Research concurs)*

`_reference` forms `precision = U^T U + blocks^T blocks` and then `torch.linalg.solve(precision, eta)` in float64. At extreme conditioning this reference shares the exact kappa^2 failure mode of the dense baseline under test, so the stability comparison is not fully independent. Empirically it holds up here — the report's float64 rows show float64 QR agreeing with this reference to ~1e-13 at condition 1e6, which cross-validates it at the tested settings — but for a benchmark whose thesis is normal-equation instability, an independent reference makes the claim airtight. Deep Research adds: report a median/MAD distribution over repeated random instances at the target condition number rather than the single 1.50e-4 figure, since the kappa^2 argument is an upper bound, not a point prediction ([conditioning notes](https://www.cs.cornell.edu/~bindel/class/cs3220-s12/notes/lec11.pdf)).

**Suggested fix:** compute the reference via `torch.linalg.lstsq`/QR of the stacked pre-array, or add a sentence noting the reference is a float64 normal-equations solve cross-validated by the float64-QR agreement.

prototypes/mu_cosine/joint_square_root_conditioner_torch.py
```diff
@@ -321,17 +453,274 @@ def householder_information_update_torch(
...
+    zero_scale = matrix_scale == 0
+    if absolute_floor == 0.0 and bool(torch.any(zero_scale).item()):
+        raise ValueError(
+            f"{name} has zero spectral scale; supply positive variances, or "
+            "provide absolute_eigenvalue_floor > 0 together with "
+            "maximum_relative_loading >= 1 to authorize a full-scale repair"
+            + _batch_failure_detail(zero_scale)
+        )
+
+    materially_indefinite = minimum < -negative_tolerance * matrix_scale
```
s243a, line 671: **[important] float32 / large-n: the indefiniteness gate and the max-loading budget can disagree — in the regime the benchmark itself runs** — *(Opus 4.8)*

The "materially indefinite" threshold uses `negative_tolerance = 64·n·eps` while the loading budget defaults to `maximum_relative_loading = 1e-3`. The loading required for a borderline-negative input approaches `relative_floor + negative_tolerance`, so in float32 once `64·n·eps + sqrt(eps) > 1e-3` — i.e. n ≳ 85, and the streamed benchmark runs n=128 — an input that passes the indefiniteness gate can then be rejected by the loading gate with a confusing "exceeds maximum_relative_loading" error on a matrix just judged not materially indefinite. (Fable 5 verified the ordering itself is sound — loading cannot mask true indefiniteness; this is a defaults-coupling refinement, not a correctness bug.)

**Suggested fix:** default-couple the gates (e.g. `negative_tolerance = min(64·n·eps, maximum_relative_loading/2)`), or document the ordering (indefiniteness = hard reject; loading = budget that may fire first) and name the float32/n≳85 regime in the docstring.

s243a (commented):
## Council review — needs discussion (approve-able with two small fixes)

Independent review by four models — **Claude Opus 4.8** (numerics/architecture, approve), **Claude Sonnet 5** (benchmark methodology, needs discussion), **GPT 5.5** (PyTorch/CUDA & performance engineering, request changes), **Claude Fable 5** (holistic adversarial, approve) — plus a **Deep Research** literature pass. The split is not about the math: no seat found an algebra defect. It is about benchmark measurement hygiene and how the PR body presents the numbers.

**What was verified (multiple seats, independently):**
- Triangular-solve whitening (`W = L^-1` implicit, no explicit inverse), `diag(s) rho diag(s)` correlation scaling, the Schur complement `Rc = R - C^T P^-1 C` / `J = H + C^T P^-1`, QR of `stack([U, WJ])` for `U_post^T U_post = U^T U + J^T Rc^-1 J`, the innovation-uses-physical-`H`/design-uses-`J` split, sign-normalization cancellation, and the Joseph-form posterior are all correct. The PR's central claim — that multiplying a Cholesky factor of `Rc^-1` into the prior root is wrong — is itself correct: it matches the stacked-QR measurement update in [Tracy 2022, arXiv:2208.06452](https://arxiv.org/abs/2208.06452) and the classical Bierman/[CERFACS SRIF](https://cerfacs.fr/wp-content/uploads/2017/05/Alina_presentation_kf.pdf) pattern exactly (worth citing in DESIGN_joint_square_root_qr_conditioner.md).
- The scale-relative rank tolerance matches the [NumPy/LAPACK `S.max()*max(m,n)*eps` / RCOND convention](https://numpy.org/devdocs/reference/generated/numpy.linalg.matrix_rank.html) precisely, and the float32 cond-1e6 results are directionally what kappa^2-vs-kappa theory predicts ([conditioning notes](https://www.cs.cornell.edu/~bindel/class/cs3220-s12/notes/lec11.pdf)).
- Tests are genuine (independent normal-equations references, mutation snapshots, rejection paths), and the "9 skipped" count was corroborated exactly by static count of CUDA-gated parametrizations.

**Environment caveat:** none of the four review sandboxes had torch installed, so the torch suite was verified statically + via the NumPy suite (8 passed everywhere); runtime confirmation of "60 passed, 9 skipped" rests on the author's environment.

**Body-only notes (lines outside or spanning the diff):**
- `precision_root_from_covariance_torch` and the `conditional_*` helpers changed default `jitter` from `1e-9` to `0.0` and now route through the scale-relative whitener with `maximum_relative_loading=1e-3` — a behavior change that can newly *reject* near-singular covariances that previously passed with absolute jitter. Intentional and defensible, but worth one line in the PR body (Sonnet 5).
- `_tensor` intentionally skips finiteness checks, so NaN/Inf propagate silently through the hot whitening/QR path; a one-sentence note on the `update_noise_block`/module docstring would set caller expectations (Fable 5).
- Deep Research recommends stating the round-off-vs-materially-indefinite threshold explicitly in the docs, since the reject-vs-repair policy's safety hinges on where that boundary sits (cf. Higham-style nearest-PSD repair, which targets round-off only).

Inline comments below are attributed per seat; severities: 7 important, 4 nit.

*Posted by Perplexity Computer on behalf of @s243a after a multi-model council review.*

prototypes/mu_cosine/REPORT_streamed_block_qr_benchmark.md
```diff
@@ -0,0 +1,301 @@
...
+Run 2026-07-11. This report benchmarks the sequential regime left open by
+`REPORT_joint_square_root_qr_benchmark.md`.
+
+## Result in one paragraph
+
+Batching several available measurement blocks into **one stacked Householder QR** is substantially faster than
+launching one QR per block. CUDA becomes worthwhile for QR when a fixed design has thousands of independent
+right-hand sides: at `n=128`, 32 measurements per block, 16 blocks, and `B=4096`, full stacked QR fell from
+`10.392 ± 0.197 ms` on CPU to `3.939 ± 0.034 ms` on CUDA. It did **not** cross over at `B=128`, and batched QR
+of 128 distinct designs was slower on the local GTX 1660 SUPER than CPU QR. Dense normal-equation updates were
```
s243a, line 12: **[nit] Hardware caveat: a tensor-core-less, 1:32-fp64 Turing card makes the crossover and fp64 columns device-specific** — *(GPT 5.5 + Deep Research)*

The GTX 1660 SUPER (TU116) has no tensor cores and 1:32 FP64:FP32 throughput (~157 GFLOPS fp64 vs ~5 TFLOPS fp32, [Turing whitepaper](https://www.nvidia.com/content/dam/en-zz/Solutions/design-visualization/technologies/turing-architecture/NVIDIA-Turing-Architecture-Whitepaper.pdf)). The batched-LA literature treats crossover thresholds as hardware/shape-specific by design ([autotuned batched LA](https://ieeexplore.ieee.org/document/8476161/)), and consumer-vs-datacenter comparisons show large fp64 asymmetries ([CERN HEPiX benchmarking](https://indico.cern.ch/event/671509/contributions/2889650/attachments/1598240/2532715/HEPIX_BWG_GPU_ML_benchmarking.pdf)). On an A100/H100-class part the B=4096 crossover likely shifts materially (probably to smaller B). The report discloses hardware well; the summary paragraph and PR body quote the numbers without this qualifier.

**Suggested fix:** one line in the hardware section: the crossover B and fp64 columns are conservative, consumer-Turing-specific data points, not a general GPU recommendation.

prototypes/mu_cosine/joint_square_root_conditioner_torch.py
```diff
@@ -165,6 +199,103 @@ class DenseGaussianUpdateTorch:
...
+            columns = value.reshape(count, self.dimension).transpose(0, 1)
+            whitened = self.apply_columns(columns)
+            return whitened.transpose(0, 1).reshape(
+                vector_batch + (self.dimension,)
+            )
+        batch_shape = torch.broadcast_shapes(self.batch_shape, vector_batch)
+        value = value.expand(batch_shape + (self.dimension,))
+        return self.apply_columns(value.unsqueeze(-1)).squeeze(-1)
+
+    def materialize_precision_root(self) -> torch.Tensor:
```
s243a: **[nit] `materialize_precision_root` returns a lower-triangular factor; the name invites misuse as an upper root** — *(Opus 4.8)*

The method correctly returns `U = L^-1` (lower row root, `cov^-1 = U^T U`) and the docstring warns it is not the canonical upper precision Cholesky — used correctly internally (`precision_root_from_covariance_torch` re-triangularizes it). But the name `precision_root` collides conceptually with the upper `precision_root` used everywhere else in the module; a caller wiring this directly into a path expecting an upper root silently gets a valid-but-lower factor.

**Suggested fix:** rename to `materialize_row_root()` (or similar), or assert orientation at consumption sites. Optional.

prototypes/mu_cosine/benchmark_streamed_block_qr.py
```diff
@@ -0,0 +1,745 @@
...
+    _synchronise(device)
+
+    elapsed_ms: list[float] = []
+    peak_bytes: list[int] = []
+    for _ in range(trials):
+        if device.type == "cuda":
+            _synchronise(device)
+            baseline = torch.cuda.memory_allocated(device)
+            torch.cuda.reset_peak_memory_stats(device)
+        else:
```
s243a, line 215: **[nit] CPU peak memory is never measured — the memory story is CUDA-only** — *(Sonnet 5)*

`baseline = 0` on the non-CUDA branch and `peak_bytes` is only populated for CUDA, so the "streaming trades time for bounded memory" narrative has no CPU counterpart. The report labels its memory numbers as CUDA correctly; worth one line noting CPU memory is untracked so the memory/latency trade-off is only characterized on CUDA.

**Suggested fix:** either note the limitation in the report, or add `tracemalloc`/RSS sampling for the CPU branch in a follow-up.

prototypes/mu_cosine/benchmark_streamed_block_qr.py
```diff
@@ -0,0 +1,745 @@
...
+        "whitening is excluded"
+    )
+    print(
+        f"# warmups={args.warmups}; trials={args.trials}; "
+        f"inner_repeats={args.inner_repeats}; seed={args.seed}"
+    )
+    _print_header()
+
+    with torch.inference_mode():
+        for n in state_dims:
```
s243a, line 633: **[nit] `main()` is an eight-level nested sweep wrapping a broad `try/except`** — *(Sonnet 5)*

Six nested `for` axes plus device/algorithm loops and a wide `except Exception` make per-cell logic hard to follow and impossible to unit-test. (The lambda closures capture loop variables but are invoked within the same iteration, so there is no late-binding bug.)

**Suggested fix:** extract a `run_cell(...)` helper so the sweep body is testable and the failure-row path is isolated. Non-blocking.

PR #3675, opened 2026-07-12
mu_cosine: calibrate covariance sensitivity + adjacency-aware batching design

s243a:
## Summary

This PR adds a calibrated covariance-sensitivity study for `mu_cosine` and turns the next graph-correlation question into an explicit adjacent-positive / matched-distant-negative experiment.

It:

- adds marginal-matched PSD correlation paths, nested selection, family-wise null calibration, posterior sensitivity metrics, and Gaussian/centered-μ kernel helpers;
- adds corrected synthetic v2 controls separating mechanism recovery from end-to-end covariance discovery;
- labels the real-data v1 runner as a legacy diagnostic and blocks accidental multi-seed inference;
- records the reproducible corrected results;
- designs adjacency-aware sampling and distance-separated square-root/QR batching.

## Corrected result

The original 216-candidate selector selected nonzero covariance under the block null 64.9% of the time even with an oracle mean, increasing to 84.5% with fitted regional KRR.

The corrected controls show:

- Good correlation information really can improve the estimate.
- At coupling 0.10, it recovers 67–70% of the available NLL improvement and 74–86% of posterior-risk improvement.
- At coupling 0.20, recovery reaches 71–84% for NLL and 89–96% for posterior risk.
- Detection remains only 31–46%, below the frozen 80% power gate.

Therefore this PR retains block covariance for deployment. It does not claim that cross-item correlation is useless; it shows that the current one-field estimator and selector cannot discover it reliably enough.

## Adjacent-positive / distant-negative follow-up

The graph judge serves two related purposes:

1. Help the model learn the local structure of a dataset.
2. Supply candidate geometry for residual correlation and safe batching.

The follow-up treats adjacent rows as smoothing positives and matched distant—including semantically similar hard-negative—rows as negatives. Cross-fitted conditional residuals then test whether same-judge error coupling remains after the learned mean is removed.

If validated, a conservative threshold on whitened cross-block coupling determines which distant items may use independent updates. Adjacent connected components remain joint square-root/QR blocks.

## Validation

- 56 relevant tests pass.
- All four new Python modules compile.
- All three CLIs parse `--help`.
- Deterministic scientific artifact SHA-256: `bb4b81c6bd182ebe89bee3360345491b8068aecd56e3ff59d49da9ceac56a9ca`.
- Runtime remains available on stdout but is deliberately excluded from the hashed JSON.
- Claude's sole reproducibility finding is addressed; all scientific values are unchanged.

### DIFF
--- a/prototypes/mu_cosine/run_covariance_sensitivity_synthetic_v2.py
+++ b/prototypes/mu_cosine/run_covariance_sensitivity_synthetic_v2.py
@@ -831,12 +831,17 @@ def _validate_args(args):
         raise ValueError("--shrinkage must be in [0,1]")
 
 
-def _write_payload(path, payload):
+def _write_scientific_payload(path, payload):
+    """Write only deterministic scientific results, never runtime metadata."""
+    if "wall_seconds" in payload:
+        raise ValueError("wall_seconds belongs in stdout, not the reproducible artifact")
+    serialized = json.dumps(
+        _jsonable(payload), indent=2, sort_keys=True, allow_nan=False
+    ) + "\n"
     path = os.path.abspath(path)
     temporary = path + ".tmp"
-    with open(temporary, "w", encoding="utf-8") as stream:
-        json.dump(_jsonable(payload), stream, indent=2, sort_keys=True)
-        stream.write("\n")
+    with open(temporary, "w", encoding="utf-8", newline="\n") as stream:
+        stream.write(serialized)
     os.replace(temporary, path)
 
 
@@ -1016,16 +1021,16 @@ def main():
             "end_to_end_krr": end_records,
             "oracle_mean_known_B_mechanism": mechanism_records,
         },
-        "wall_seconds": time.perf_counter() - started,
     }
-    _write_payload(args.out, payload)
+    wall_seconds = time.perf_counter() - started
+    _write_scientific_payload(args.out, payload)
     print(json.dumps({
         "output": os.path.abspath(args.out),
         "null_calibrations": null_report,
         "v1_failure_attribution_audit": audit,
         "end_to_end_krr": payload["end_to_end_krr"],
         "oracle_mean_known_B_mechanism": payload["oracle_mean_known_B_mechanism"],
-        "wall_seconds": payload["wall_seconds"],
+        "wall_seconds": wall_seconds,
     }, indent=2, sort_keys=True))
 
 
--- a/prototypes/mu_cosine/run_covariance_sensitivity_synthetic_v2.py
+++ b/prototypes/mu_cosine/run_covariance_sensitivity_synthetic_v2.py
@@ -1,0 +1,1033 @@
+#!/usr/bin/env python3
+"""Versioned v2 family-wise calibration for the synthetic covariance selector.
+
+This executable preserves the invalidated v1 output record, not an exact v1 runner: the shared synthetic
+runner was later amended to use the corrected deranged wrong geometry.  V2 compares a conditional fixed-path
+null against a full block-null procedure that reruns regional KRR and block fitting, audits mean/search
+multiplicity, and evaluates two exploratory capacities: v2A's complete grid and v2B's canonical-geometry
+alpha grid.
+"""
+from __future__ import annotations
+
+import argparse
+from collections import Counter
+import json
+import os
+import time
+
+import numpy as np
+
+from covariance_sensitivity import (
+    finite_null_maximum_threshold,
+    maximum_eligible_macro_gain,
+    select_nested_candidate,
+    select_nested_candidate_null_calibrated,
+)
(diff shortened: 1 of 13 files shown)


prototypes/mu_cosine/run_covariance_sensitivity_synthetic_v2.py
```diff
@@ -0,0 +1,1033 @@
...
+                    "INVALID_pre_krr_denominator.json"
+                ),
+                "reason": "pre-KRR R_HH was not the covariance of e_H-W e_T",
+            },
+        ],
+        "replicate_records": None if args.summary_only else {
+            "end_to_end_krr": end_records,
+            "oracle_mean_known_B_mechanism": mechanism_records,
+        },
+        "wall_seconds": time.perf_counter() - started,
```
s243a: **Reproducibility gap (non-blocking):** this embeds wall-clock timing into the serialized `--summary-only --out` JSON, so `sha256sum` of the artifact differs on every run. The PR/report claim `6bbe4fdd…`, but running the documented command yields a different hash each time (I got `9f873eee…`; a second reviewer got yet another). Every *scientific* value is deterministic — dropping `wall_seconds` makes repeated runs byte-identical — so only the timing field breaks the hash.

Suggested fix (any one):
- keep `wall_seconds` in stdout only, out of the serialized artifact; or
- hash a canonicalized copy with the timing field removed; or
- point the SHA-256 claim at the deterministic tracked `repro/covariance_sensitivity/synthetic_v2_summary.json` (which is a separate hand-curated file that currently just stores the claimed hash as a string, rather than being the file that hashes to it).

---

s243a (commented):
## Review — calibrate covariance sensitivity + adjacency-aware batching design

**Verdict: Approve with one minor, non-blocking fix.** This is unusually careful research code. I read the core library line-by-line, verified the math, reproduced the headline results end-to-end (three independent reproductions, all matching), and confirmed the safety framing. The only defect is a reproducibility gap in the SHA-256 artifact claim (inline note on `run_covariance_sensitivity_synthetic_v2.py:1019`).

### Scope
Reviewed the single tip commit `222be4ba` — 13 files, +5432 (design/report markdown, one numerical library, three CLI runners, and their tests).

### Verified
**Core library `covariance_sensitivity.py` — mathematically sound.** Checked each nontrivial routine:
- `expected_rbf_kernel_gaussian` — correct Gaussian-expected RBF (amplitude `(ℓ²/denom)^{d/2}`, `denom = ℓ²+σᵢ²+σⱼ²`); normalization gives a correlation kernel with `ℓ_eff² = ℓ²+2σ²`.
- `CorrelationPath` — marginal-matched path preserves each within-item block exactly while varying only cross-item correlation; the fast `nll_per_scalar` reproduces the full Cholesky NLL including the `item_count·logdet_block + Σlog(values)` term.
- `directional_posterior_sensitivity` — the analytic derivative `dx = −S⁻¹HᵀR⁻¹ D R⁻¹(y−Hx)` is correct and matches finite-difference in the tests.
- Conservative order-statistic null threshold and nested selector tie-break are correct; PSD-preservation and input validation are pervasive.

**Empirical claims — all confirmed:**
- **34 focused tests pass** (matches the report); coverage is strong (analytic-vs-finite-difference, exact marginals, RBF broadening, selector behavior).
- **All three CLIs parse `--help`.**
- **Numbers reproduce exactly.** Re-ran the documented v2 command; thresholds (`0.008437`/`0.017601`), v1 selection rates (`64.9%`/`84.5%`), audit contrasts (`40.1`/`22.5` pts), recovery (67–84% NLL, 74–96% posterior-risk), and detection (31–46%) all match. Determinism is real (fixed seeds, `sort_keys=True`).
- **The "legacy diagnostic blocks multi-seed inference" behavior holds.** `run_covariance_sensitivity.py` raises on `--seeds>1` unless `--allow-failed-v1-selector`, with defense-in-depth (aggregate gate hardcoded non-passing) and a resume-config guard — all tested. The conditional fixed-path null is present but explicitly labeled non-inferential and never wired into a passing gate, consistent with "no real-data v2 selector implemented here."

### Finding (MEDIUM — non-blocking)
The SHA-256 reproducibility claim is not reproducible as documented, because `wall_seconds` (wall-clock timing) is embedded in the serialized artifact. Details and suggested fixes are in the inline comment. All scientific content is deterministic; only the timing field breaks the byte-level hash.

### Minor / optional
- `_filter_search` builds `output` before raising on an unknown search name — wasted work, still correct.
- The design docs are exemplary: the v1 erratum (pre-KRR denominator), the two-track correction, and the family-wise multiplicity attribution are documented honestly and match the code.

_Automated review posted as a comment (not a formal approval vote); the recommendation is to merge after addressing the reproducibility note._

---

PR #3685, opened 2026-07-12
mu_cosine: calibrate adjacency-residual confidence with component-safe folds

s243a:
## Summary

- freeze an outcome-blind adjacent-residual pilot over the existing exploratory and fresh campaigns
- cross-fit calibration, conditionalization, regional mean, and within-item covariance while holding out whole endpoint components
- compare same-descendant/direct-edge residual cross-products with anchor-sharing nonadjacent controls
- add a low-capacity PSD adjacency Gram and topology-deranged held-likelihood diagnostic
- add a deterministic known-mean/known-`B` mechanism/power audit, compact provenance, tests, and a full report

## Result

The observed adjacency-plus-local-hop contrast is strong and directionally stable:

| corpus | component-macro trace/4 | positive assignments | eligible positive components |
|---|---:|---:|---:|
| exploratory | +0.279877 ± 0.007426 | 10/10 | 28 |
| fresh | +0.169740 ± 0.011065 | 10/10 | 40 |

Every leave-one-positive-component-out estimate is positive. At the largest frozen diagnostic coupling (`alpha=0.50`), the true adjacency Gram improves held NLL per scalar over the independent block by +0.014673 exploratory and +0.021167 fresh; the deranged comparator improves by +0.003216 and +0.010030.

However, the formal frozen outcome is **not resolved by existing data**:

- adjacency is confounded with local hop/tag position in the existing campaign
- there is only one residual field per corpus, so persistent item structure cannot be separated from stochastic judge covariance
- the known-mean/known-`B` mechanism reaches only 18.0% / 19.5% pointwise detection at coupling 0.10 for 28 / 40 components
- the 80% synthetic power gate fails
- no end-to-end nuisance-estimation null or repeated-judge confirmation is present

The runner therefore hard-codes both advancement and QR-covariance deployment gates false.

## Confidence and deployment policy

Keep the 95% simultaneous bound for any production covariance or batching decision. An 80–90% pointwise band may be used only as a labelled discovery/shadow-mode screen; it cannot modify `R`, QR block membership, or a production decision.

A favorable diagnostic sizing extension first reaches 80% pointwise mechanism detection for coupling 0.10 at about 320 independent components (81%), but simultaneous detection is only 54% there and 65% at 400. The confirmatory campaign should therefore use more than 400 independent components for a covariance-wide target, balance adjacent and distant/hard-negative triples by hop/tag strata, and obtain at least three independent calls per judge family.

The eventual covariance adapter should shrink on a PSD path and add a simultaneous spectral envelope; elementwise “lower covariance” bounds are not generally PSD-safe.

## Review follow-up: portable provenance

Claude Opus 4.8 reproduced the scientific results but found that whole-file hashes embedded machine-local absolute paths. Commit `5cbe638f8` addresses the full class, including locators beyond the cited line:

- payload schema v2 stores role-keyed `{size_bytes, sha256}` records and no checkout/input paths
- output and input-locator arguments are excluded from scientific configuration
- graph provenance retains content identity and the LMDB lock-file exclusion contract without repository/directory locators
- two complete synthetic runs written to different output paths are byte-identical
- the schema-v2 real rerun reproduces every previous scientific field exactly

Portable artifact hashes:

- real: `407d52d8fe64df3bf9220da95976bee5c046c5d806d676aa38aafa4a475e100a`
- synthetic: `62695568b1eae247ece2d3e37a18250c756b57536bdf6fe7e0ec7a91f8b8c18e`

The optional resume-RNG concern was also inspected: `fit_context` consumes no process-global RNG, while component assignment, multiplier bands, and derangement use explicit seeded generators.

## Validation

- focused adjacent-pilot tests: 13 passed
- adjacent + inherited structured-covariance and NumPy/Torch QR suites: 88 passed, 9 skipped
- inherited covariance-sensitivity suites: 35 passed
- combined: **123 passed, 9 skipped**
- two full synthetic outputs at different locations: byte-identical
- real schema-v2 rerun: all prior scientific fields identical
- `git diff --check`: clean

Full deterministic output hashes and reproduction commands are recorded in `REPORT_adjacent_residual_pilot.md` and `repro/adjacent_residual_pilot/summary.json`.

### DIFF
--- a/prototypes/mu_cosine/run_adjacent_residual_synthetic.py
+++ b/prototypes/mu_cosine/run_adjacent_residual_synthetic.py
@@ -1,0 +1,206 @@
+#!/usr/bin/env python3
+"""Mechanism power audit for the adjacent component-multiplier statistic.
+
+This runner uses known zero mean and identity within-row covariance.  It does
+not emulate the end-to-end calibration/KRR/B fitting pipeline and therefore
+cannot by itself unlock the real-data pilot.
+"""
+from __future__ import annotations
+
+import argparse
+import hashlib
+import json
+import os
+import time
+
+import numpy as np
+
+from adjacent_residual_pilot import component_multiplier_stability
+
+
+COMPONENT_COUNTS = (28, 40)
+COUPLINGS = (0.0, 0.04, 0.10, 0.20)
+
+
+def _file_record(path):
+    path = os.path.abspath(path)
+    digest = hashlib.sha256()
+    with open(path, "rb") as stream:
+        for chunk in iter(lambda: stream.read(1 << 20), b""):
+            digest.update(chunk)
+    return {"path": path, "size_bytes": os.path.getsize(path), "sha256": digest.hexdigest()}
+
+
+def simulate_component_contrasts(components, coupling, seed):
+    """Generate one anchor-matched 4x4 contrast per independent component."""
+    if components < 2:
+        raise ValueError("components must be at least two")
+    if not 0.0 <= coupling < 1.0:
+        raise ValueError("coupling must be in [0,1)")
+    rng = np.random.default_rng(seed)
+    output = []
+    for _ in range(components):
+        anchor = rng.standard_normal(4)
+        positive_partner = (
+            coupling * anchor
+            + np.sqrt(1.0 - coupling * coupling) * rng.standard_normal(4)
+        )
+        first_control = rng.standard_normal(4)
+        second_control = rng.standard_normal(4)
+        positive = 0.5 * (
+            np.outer(anchor, positive_partner)
+            + np.outer(positive_partner, anchor)
+        )
+        controls = 0.25 * (
+            np.outer(anchor, first_control)
+            + np.outer(first_control, anchor)
+            + np.outer(positive_partner, second_control)
+            + np.outer(second_control, positive_partner)
+        )
+        output.append(positive - controls)
+    return np.asarray(output)
+
+
+def _summary(values):
+    values = np.asarray(values, dtype=float)
+    return {
+        "mean": float(np.mean(values)),
+        "sd": float(np.std(values, ddof=1)) if len(values) > 1 else 0.0,
+        "q05": float(np.quantile(values, 0.05)),
+        "median": float(np.median(values)),
+        "q95": float(np.quantile(values, 0.95)),
+    }
+
+
+def run_scenario(components, coupling, args):
+    records = []
+    for replicate in range(args.replicates):
+        seed = args.seed + 100000 * components + 10000 * int(round(coupling * 100)) + replicate
+        matrices = simulate_component_contrasts(components, coupling, seed)
+        band = component_multiplier_stability(
+            matrices,
+            draws=args.multiplier_draws,
+            seed=seed + 500000,
+            confidence=args.confidence,
+        )
+        primary = float(band.estimate[0])
+        pointwise_low = float(band.pointwise_low[0])
+        simultaneous_low = float(band.simultaneous_low[0])
+        true_matrix = np.eye(4) * coupling
+        error = float(np.max(np.abs(np.linalg.eigvalsh(
+            np.mean(matrices, axis=0) - true_matrix
+        ))))
+        records.append({
+            "primary_trace_over_4": primary,
+            "pointwise_lower_positive": bool(pointwise_low > 0.0),
+            "simultaneous_lower_positive": bool(simultaneous_low > 0.0),
+            "pointwise_contains_truth": bool(
+                band.pointwise_low[0] <= coupling <= band.pointwise_high[0]
+            ),
+            "spectral_radius_contains_truth": bool(error <= band.spectral_error_radius),
+        })
+    return {
+        "components": int(components),
+        "coupling": float(coupling),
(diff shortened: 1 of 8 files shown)


prototypes/mu_cosine/run_adjacent_residual_synthetic.py
```diff
@@ -0,0 +1,206 @@
...
+        ),
+        "design": "DESIGN_adjacent_residual_pilot.md",
+        "implementation": {
+            "design": _file_record(os.path.join(
+                os.path.dirname(os.path.abspath(__file__)), "DESIGN_adjacent_residual_pilot.md"
+            )),
+            "core": _file_record(os.path.join(
+                os.path.dirname(os.path.abspath(__file__)), "adjacent_residual_pilot.py"
+            )),
+            "runner": _file_record(os.path.abspath(__file__)),
```
s243a, line 187: **Reproducibility nit (LOW, non-blocking):** `_file_record` stores `os.path.abspath(...)` as `path`, so the serialized artifact embeds machine-specific absolute paths for core/design/runner. This means the claimed artifact SHA-256 (`c736e25…` synthetic, `61b4b3c…` real) can't reproduce across checkouts even on an identical numerical stack — three independent runs (yours, plus two during review) produced three different file hashes with *identical* scientific content. The real runner has the same pattern at `run_adjacent_residual_pilot.py:524`, and additionally puts `--artifact-repo`/`--ckpt` absolute paths in config.

The `sha256`/`size_bytes` fields here are content-based and fine — only the `path` string breaks portability. Nice work removing `wall_seconds` from the payload (the defect from the sibling PR); this is the residual of the same class.

**Fix (any one):** store repo-relative paths, or drop `path` and keep just `sha256`+`size_bytes`; for the real pilot also relativize/omit the data-path config from the hashed payload. The durable tracked `repro/adjacent_residual_pilot/summary.json` is already clean.

---

s243a (commented):
## Review — adjacency-residual confidence with component-safe folds

**Verdict: strong, methodologically careful pilot — approve the direction. One low-severity reproducibility nit; no correctness defects.** Draft PR; the two deployment gates are correctly hard-`False` and the result is honestly framed as "not resolved by existing data." I read the core library line-by-line, reproduced the synthetic audit end-to-end, verified tests/CLI, and ran an independent second review of both runners. Both reviews agree.

### Scope
Single commit `8d1d7d17` — 8 files, +2026 (design, report, `adjacent_residual_pilot.py` library, two runners, two test files). Builds on the adjacency design from #3675.

### Verified
**Core library — sound and defensive.**
- **Component-safe folds (central claim):** whole endpoint components assigned to folds via union-find over exact endpoint sharing, with two runtime assertions — exact-partition and an explicit cross-fold node-intersection check (`adjacent_residual_pilot.py:188-195`) that directly enforces no leakage. Positives and their controls are always same-descendant → same component → same fold, so the cross-product statistic cannot straddle fit/held. `_oof_whitened_rows` fits on `setdiff1d(all, held)` and writes only `held`, so every row is whitened by a model that never saw it.
- **Matched controls** are anchor-sharing and explicitly non-adjacent (`_direct_edge(...): continue`), matching the stated estimand.
- **Regional mean cross-fit before covariance** (block from train LOO residuals), so a smooth mean is removed before any adjacency statistic — no relabelling.
- Rademacher sign-flip multiplier bands honestly labelled *conditional stability*, not population CIs; PSD adjacency Gram and the within-descendant derangement are correct.

**All 7 PR claims implemented** (independently confirmed), including the two deployment gates being literal `False` with no truthy code path, and `real_data_gate_unlocked: False` in the synthetic audit.

- **10 focused tests pass**; both CLIs parse `--help`.
- **Synthetic detection reproduces exactly:** coupling 0.10 → 18.0%/5.5% (28 comp), 19.5%/6.5% (40 comp); 0.20 → 45.0%/23.5%, 61.0%/29.5%; block-null false-positive ≤10%; the 80% power gate **fails**, as claimed.
- **The `wall_seconds` artifact-hash defect from #3675 is fixed here** — timing is stdout-only, absent from the serialized payload (verified by byte comparison; two same-machine runs are byte-identical).

### Finding (LOW — reproducibility nit, non-blocking)
The claimed artifact SHA-256 values can't reproduce across checkouts because the serialized JSON embeds machine-specific absolute paths in its provenance block (`os.path.abspath(__file__)`). See the inline note on `run_adjacent_residual_synthetic.py:180` (same at `run_adjacent_residual_pilot.py:524`). Scientific content is fully deterministic and reproduces exactly; only the `path` strings break portability, and the durable tracked `summary.json` is already clean.

- `--help` requires the full heavy import chain (torch etc.) — pre-existing prototype pattern, not introduced here.
- `--folds` is a knob that `_validate_args` pins to 5 (intentional "frozen pilot", with a clear error).
- Worth an author spot-check: confirm `fit_context` doesn't consume the process-global `np.random` stream mid-pipeline, so a `--resume`d run matches a fresh one (per-assignment RNGs are explicitly seeded, so this is only a theoretical concern).

_Automated review posted as a comment (not a formal approval vote). The methodology, honesty about limitations, and hard-false gates are all in good shape; recommend addressing the path-in-artifact nit before the eventual confirmatory follow-up relies on these hashes._

---

PR #3695, opened 2026-07-13
mu_cosine: audit PSD graph geometries with independent embedding controls

s243a:
## Summary

This PR establishes the statistical and numerical groundwork for learning item-correlation geometry before feeding it to the existing joint square-root/QR conditioner.

- Freezes the theory, candidate geometries, confidence gates, leakage rules, and an append-only record of alternatives rejected or deferred.
- Adds PSD-by-construction graph kernels: closed-neighborhood Gram, finite cumulative-walk Gram, exact heat and regularized-Laplacian references, descendant gating, role-aware pair features, convex mixtures, and Schur products.
- Adds deterministic, revision-pinned Nomic and MiniLM embedding caches, with e5 retained as the shared-input redundancy control.
- Adds two synthetic mechanism audits and an outcome-blind campaign geometry inventory.
- Documents the one-kernel structure `R = I_n ⊗ B0 + K_theta ⊗ Bg`, whose item eigenbasis reduces conditioning to `n` channel blocks. CUDA/QR timing remains gated on real covariance evidence.

## What the audits found

The original v1 comparison **failed**: using a common path amplitude did not match covariance strength across kernels. That failure is retained rather than hidden.

The frozen v2 comparison matches maximum off-item coupling. At `rho_max=0.20`, every non-deranged planted geometry passes the preregistered strong-effect gates:

- 90.5–98.5% nonzero selection;
- 86.0–97.5% predictive-equivalence selection;
- 97.0–98.5% truth-over-derangement win rate;
- positive selected held NLL gain for every non-deranged truth.

Power at `rho_max=0.10` remains mixed, so this is a mechanism result, not a real-data deployment claim. Block-null nonzero selection is 5.5%.

The outcome-blind campaign inventory found:

- the initial same-hop walk feature is nearly identity-like and does not encode adjacency well, so it is retained only as a negative control;
- cumulative walk restores the intended adjacency ordering and is the scalable graph candidate;
- Nomic/e5 distance Spearman correlation is 0.776 exploratory and 0.838 fresh;
- MiniLM/e5 is still more redundant (0.810 / 0.868);
- cumulative-walk/Nomic correlation is moderate (0.590 / 0.493), but natural disagreement cells are sparse.

Accordingly, Nomic is the primary external semantic covariate, MiniLM is a sensitivity comparator, and neither is treated as independent ground truth or as an extra independent vote.

## Safety boundary and next experiment

No judge residual or score is used by the real-data inventory. This PR does **not** unlock real structured covariance, adjacency batching, QR deployment, or CUDA performance claims.

The next gate is a purpose-built repeated-judge campaign with more than 400 independent components, at least three calls per judge family, component-disjoint calibration/evaluation, and deliberate oversampling of graph/Nomic disagreement cells. Only after simultaneous lower confidence bounds and held predictive gates pass should the selected covariance enter dense QR and the single-kernel eigenmode/CUDA comparison.

## Review hardening

- Result provenance now fingerprints every direct loader/feature dependency.
- Singular kernels report numerical rank and a clearly named positive-spectrum condition number; zero eigenvalues are not silently presented as full-matrix conditioning.
- Independent embedding manifests validate model revision, text order/content, normalization, and array hashes.

## Validation

- Focused graph-geometry suite: 34 passed.
- Full relevant inherited + new suites: **157 passed, 9 skipped**.
- Both synthetic audits and the outcome-blind inventory were rerun after the final provenance/diagnostic fixes.
- Recorded SHA-256 values are same-environment integrity checks, not cross-stack byte-reproducibility promises. NumPy/BLAS drift can move the weakest `walk_decay` result by roughly 1/200 without changing any frozen gate; cross-environment reruns should compare scientific fields and gate booleans within the report's stated precision.

### DIFF
--- a/prototypes/mu_cosine/REPORT_graph_geometry_confirmatory.md
+++ b/prototypes/mu_cosine/REPORT_graph_geometry_confirmatory.md
@@ -241,11 +241,17 @@ python3 prototypes/mu_cosine/run_graph_geometry_inventory.py \
   --out /tmp/graph_geometry_inventory_cumulative_final.json
 ```
 
-Portable output SHA-256:
+Same-environment output-integrity SHA-256:
 
 - v1 mechanism: `9cd6c0b6f6e7b4efdbf6e8afcff88607b871348fc3a8c212c421a8c8e5d2ee51`;
 - v2 mechanism: `24333d08941ac2c356bb33fdd602a0009317de0daa6d0f7788e176c6bcc33969`;
 - outcome-blind inventory: `81c35eaa8f9fa82afbbcf6a9040f4905f97140b4b91d7134ecfb69957b9fb9a1`.
+
+These hashes exclude wall-clock timing and absolute paths, but they are not a promise of byte-identical output
+across NumPy/BLAS stacks.  Floating-point drift can change the weakest `walk_decay` selection count by about one
+replicate in 200 while leaving every frozen gate unchanged.  Cross-environment reproduction should therefore
+compare the reported scientific fields and gate booleans within their stated precision; use the hashes only to
+check artifact integrity within a matched numerical environment.
 
 The graph-geometry focused suite has 34 passing tests.  Including the inherited adjacent-residual,
 structured-covariance, covariance-sensitivity, and NumPy/Torch square-root-conditioner suites gives
--- a/prototypes/mu_cosine/REPORT_graph_geometry_confirmatory.md
+++ b/prototypes/mu_cosine/REPORT_graph_geometry_confirmatory.md
@@ -1,0 +1,252 @@
+# Graph geometry for residual covariance — theory, mechanism audit, and outcome-blind inventory
+
+## Bottom line
+
+The graph-geometry infrastructure is ready, but no real covariance or QR deployment is unlocked.
+
+- PSD graph geometry is now structural: closed-neighborhood, sparse walk-feature, cumulative-walk, exact heat,
+  and exact regularized-Laplacian references all produce unit-diagonal Gram matrices without eigenvalue repair.
+- The original synthetic comparison failed honestly because common path amplitude `alpha` did not mean common
+  covariance strength across kernels.  Its family-wise null control passed, but selected held likelihood was
+  usually worse.
+- A preregistered v2 matched maximum off-item coupling.  At `rho_max=0.20`, every non-deranged truth had positive
+  held gain, 86.0--97.5% predictive-class recovery, and 97.0--98.5% wins over equal-energy topology controls.
+  Block-null nonzero selection was 5.5%.  At `rho_max=0.10`, power was mixed and the heat case had essentially
+  zero selected gain.
+- The first outcome-blind campaign inventory rejected the original same-hop walk concatenation: it was nearly
+  identity and did not make adjacent roots closer.  Cumulative walk fixes that while remaining a sparse PSD
+  feature Gram.
+- Nomic and MiniLM are different models from the e5-based operating pipeline, but their geometry is not
+  statistically independent of e5.  Nomic is modestly less redundant and remains the primary external
+  semantic candidate; MiniLM is a lower-cost sensitivity comparator.
+
+The next scientific step is the repeated-judge campaign described below.  The current independent/block item
+covariance stays in production.
+
+## Theory implemented
+
+### PSD comes before interpretability
+
+An arbitrary graph distance is useful for sampling but is not automatically a covariance kernel.  In
+particular, a Gaussian RBF over general graph shortest-path distances is not guaranteed PSD.  The accepted
+constructions are:
+
+```text
+explicit Gram:       K = Phi Phi.T
+heat reference:      K = exp(-t L)
+resolvent reference: K = (I + tau L)^-1
+convex mixture:      K = sum_g theta_g K_g, theta_g >= 0
+interaction:         K = K_graph * K_embed       # Schur product
+```
+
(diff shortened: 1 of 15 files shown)


prototypes/mu_cosine/REPORT_graph_geometry_confirmatory.md
```diff
@@ -0,0 +1,252 @@
...
+
+python3 prototypes/mu_cosine/run_graph_geometry_inventory.py \
+  --artifact-repo /home/s243a/Projects/UnifyWeaver \
+  --campaign /tmp/mu_data/campaign_scored.tsv \
+  --minilm-prefix /tmp/campaign_minilm_geometry \
+  --nomic-prefix /tmp/campaign_nomic_geometry --lmdb-no-lock \
+  --out /tmp/graph_geometry_inventory_cumulative_final.json
+```
+
+Portable output SHA-256:
```
s243a: **Documentation nit (VERY LOW, non-blocking).** These hashes are genuinely "portable" in the sense that matters — no `wall_seconds`, no absolute paths in the payload (both prior-PR defects are fixed here; `_content_record` stores only `{size_bytes, sha256}`). But they remain **BLAS/numpy-float-dependent**: independent reruns on numpy 2.4.6 produced different hashes (`5ad6e40…` / `1a53ef00…` here vs `9cd6c0b…` / `24333d08…`), because the weakest kernel (`walk_decay`, off-diag RMS 0.072) drifts by ~1 replicate of 200 across stacks. So the hash is a same-stack integrity check, not a cross-machine reproducible artifact.

Suggest adding the same caveat the #3685 report carried — "compare scientific fields within tolerance rather than assuming byte-identical output" — so a reader on a different stack isn't surprised when `sha256sum` differs while every gate still passes. The strong families (closed/heat/resolvent/block) reproduced to the digit.

---

s243a (commented):
## Review — audit PSD graph geometries with independent embedding controls

**Verdict: the cleanest of the series — approve the direction. No correctness defects across all 15 files, and both recurring reproducibility defects from #3675/#3685 are now fixed.** One documentation-level nit on the "portable" hashes (inline). Draft PR; all deployment gates correctly hard-`False`.

I read the PSD-kernel library line-by-line, reproduced both synthetic audits, verified the `rho_max` normalization, and ran two independent deep reviews (synthetic runners; embedding cache + inventory). All three reviews agree, down to the same 1-replicate drift.

### Scope
Single commit `f74d416a` — 15 files, +3105 (theory/decisions/design/report, two libraries, three runners, four test files, tracked summary).

### Verified
**`graph_geometry.py` — PSD-by-construction is genuinely structural.** Every kernel is a Gram of an explicit real feature map (closed-neighborhood, walk, cumulative-walk) or a PSD-preserving operation (`normalize_psd_kernel`, convex mixture, Schur product, `exp(−tL)`, `(I+τL)⁻¹`), each validated by an eigenvalue floor in `_kernel`. Descendant gating and the RBF item kernel are PSD by the Schur product theorem. `kernel_diagnostics` reports a correctly-named `positive_spectrum_condition_number` over positive eigenvalues only.

**The methodological fix is correct.** v1 genuinely uses the flawed common-amplitude path `(1−α)I + α·kernel` (single α across kernels with ~3.3× different off-diagonal magnitudes) and **fails for real** (`all_mechanism_gates_pass: False`, retained not hidden). v2's `amplitude = rho / max_off_diagonal(kernel)` scales each kernel so its maximum off-item coupling equals `rho` exactly — a fair coupling-matched comparison, PD-guarded by `amplitude < 0.95`.

- **34 focused tests pass**; all CLIs parse `--help`.
- **v2 reproduces essentially exactly:** closed@0.20 nonzero 98.5% / gain +0.014632; heat@0.20 94.5% / +0.008939; resolvent@0.20 93.0% / +0.009957; truth-over-derangement 97.0–98.5%; **block-null 5.5% exact**. Only the weakest kernel (`walk_decay`, near-identity) drifts by 1/200 (90.0%/41.0% vs 90.5%/40.0%) — expected float sensitivity; every gate boolean still passes. `rho_max=0.10` is honestly mixed.
- **All deployment gates hard-`False`** (`real_covariance_gate_unlocked`, `qr_deployment_unlocked`, `batching_gate_unlocked`); v1 fails as claimed.
- **Embedding cache** detects staleness (fault-injected wrong-revision + swapped-array both raise), byte-deterministic and portable across prefixes; the **inventory is outcome-blind** (no score/label/residual reaches it, verified through `semantic_pair_features`); Spearman matches `scipy.stats.spearmanr` exactly incl. ties.
- No train/test leakage; power measured against the correct familywise block-null; estimators are unbiased indicator means; the equal-energy derangement control preserves coupling energy (fair matched control).

**Both prior reproducibility defects are fixed** — `wall_seconds` is stdout-only (fixes #3675), and provenance is content-addressed with no absolute paths (fixes #3685's residual). Nice progression across the three PRs.

### Finding (VERY LOW — documentation nit, non-blocking)
The "Portable output SHA-256" hashes correctly drop timing/paths but remain BLAS/numpy-float-dependent, so they can't reproduce byte-for-byte across stacks even though the science does. See the inline note on the report's hash line — suggest re-adding the "compare scientific fields within tolerance" caveat.

- `_rbf_item_kernel` in the inventory skips the explicit `_kernel` PSD assertion, but it's PSD by construction and only feeds distances — harmless.
- v1 `train_fields=12` vs v2 `48` — intended stabilization change.

_Automated review posted as a comment (not a formal approval vote). Theory, PSD rigor, honest v1-failure retention, matched-coupling v2 fix, outcome-blindness, and the now-clean reproducibility story are all in good shape._

---

PR #3701, opened 2026-07-13
mu_cosine: preregister batched repeated-judge covariance campaign

s243a:
## Summary

This is a **no-spend preregistration and tooling PR**, not a covariance result or live-campaign authorization.

- Freezes the planned repeated-judge campaign: endpoint-disjoint anchor/adjacent/distant components, 32 balanced cells per corpus, source-safe nested folds, repeated judge waves, estimands, nulls, alternatives, and fail-closed decision gates.
- Amortizes the LLM system prompt over stable blocks of up to 10 same-role rows while explicitly modeling shared-request covariance and using prompt blocks as inference clusters.
- Adds outcome-blind sampling/scheduling plus keyed response validation: exact request bytes are hashed, stable row IDs are submitted, retries are preserved, and provider call IDs cannot be reused as nominally fresh repeats.
- Adds a repeat-aware synthetic sizing harness with block-null and smooth-mean false-promotion controls, residual and latent-state posterior NLL endpoints, topology derangements, and full nested refitting.
- Adds an exact compound-symmetry eigenmode path for exchangeable prompt schedules, with dense overlap-aware fallback and dense decision-boundary rechecks.
- Adds deterministic process workers, one BLAS thread per worker, atomic checkpoint/resume, a hard null-calibration barrier, canonical result reconstruction, and immutable provenance checks.

## Why

Scoring one item per LLM request repays the system prompt for every item. Batching is materially cheaper, but rows in one prompt can share request and list-position effects. This design keeps batching economical while making the induced dependence explicit instead of treating batched rows as independent.

The repeated calls are needed to separate row-call noise from persistent item/judge covariance before any graph/Nomic geometry can be considered for the joint square-root/QR conditioner.

## Important boundaries

- No endpoint was selected and no judge/model/API call was made.
- The full preregistered simulation was **not** run, so there is no sample-size recommendation.
- Smoke/custom runs are diagnostic-only and cannot set sizing fields.
- Final campaign G, deployment, independent batching, QR specialization, and CUDA remain false/null.
- The repository-owned candidate builder, verified historical inventory, and approved exact prompt/model/settings contract are still prerequisites.
- List-position effects are mitigated by independent role-by-judge orderings and spread repeat rotations, but remain a required pilot/train-only-adjustment/power-sensitivity gate before live scoring.
- The two synthetic corpus labels are independent copies of a generic DGP; they enforce multiplicity, not empirical corpus heterogeneity.

See `prototypes/mu_cosine/REPORT_repeated_judge_campaign_power.md` for the full claim boundary, math, compute audit, and next steps.

## Performance notes

The original dense implementation projected roughly 230–300 serial CPU-hours. Exact prompt-block diagonalization reduced local 10×10-block scoring by 28.43× (residual) and 22.24× (posterior), projecting roughly 40–45 CPU-hours / 11–15 hours on four physical cores before a full run.

A diagnostic two-worker smoke completed in 40.15 s; checkpoint resume reproduced byte-identical scientific JSON in 0.02 s. A G=160 diagnostic with five null draws and one replicate across all 14 scenarios completed in 30.55 s. These are compute checks, not scientific results.

## Validation

- Repeated campaign/sampler/power/runner: **68 passed**
- Relevant inherited graph geometry, synthetic-v2, structured covariance, JointPosterior, and square-root conditioner suites: **97 passed, 9 optional CUDA skips**
- Dense/eigenmode maximum observed differences: residual **2.22e-16**, posterior **3.33e-16**
- Python compilation and `git diff --check`: clean

## Review focus

1. Prompt-block membership, position scheduling, and request-covariance assumptions.
2. The two zero-coupling false-promotion gates and synthetic-primary-event boundary.
3. Collective/contrast eigenmode algebra and dense fallback conditions.
4. Checkpoint fingerprinting, null barrier, and deterministic reconstruction.
5. Whether the remaining live-campaign prerequisites are sufficiently fail-closed.

### DIFF
--- a/prototypes/mu_cosine/run_repeated_judge_power.py
+++ b/prototypes/mu_cosine/run_repeated_judge_power.py
@@ -697,19 +697,24 @@ def _file_digest(path):
     return {"bytes": size, "sha256": digest.hexdigest()}
 
 
+def _portable_blas_runtime(records):
+    """Keep numerical BLAS identity while excluding host-specific paths."""
+    fields = (
+        "user_api", "internal_api", "prefix", "version",
+        "threading_layer", "architecture",
+    )
+    return sorted(
+        ({field: record.get(field) for field in fields} for record in records),
+        key=lambda record: tuple(str(record[field]) for field in fields),
+    )
+
+
 def _blas_runtime_identity():
     try:
         from threadpoolctl import threadpool_info
     except ImportError:
         return []
-    fields = (
-        "user_api", "internal_api", "prefix", "filepath", "version",
-        "threading_layer", "architecture",
-    )
-    return sorted(
-        ({field: record.get(field) for field in fields} for record in threadpool_info()),
-        key=lambda record: tuple(str(record[field]) for field in fields),
-    )
+    return _portable_blas_runtime(threadpool_info())
 
 
 def _provenance():
@@ -723,7 +728,10 @@ def _provenance():
         "python_version": sys.version,
         "numpy_version": np.__version__,
         "blas_runtime": _blas_runtime_identity(),
-        "worker_blas_threads": 1,
+        "blas_thread_limit_policy": (
+            "child workers request one thread via environment; serial execution "
+            "uses threadpoolctl when available"
+        ),
         "seed_derivation": "sha256(base,G,R,scenario,replicate); null namespace is disjoint",
     }
 
--- a/prototypes/mu_cosine/run_repeated_judge_power.py
+++ b/prototypes/mu_cosine/run_repeated_judge_power.py
@@ -10,9 +10,12 @@ not part of the unit-test suite.
 from __future__ import annotations
 
 import argparse
+from concurrent.futures import ProcessPoolExecutor, as_completed
+from contextlib import contextmanager
 import hashlib
 import json
 import math
+import multiprocessing
 import os
 import sys
 import tempfile
@@ -24,17 +27,23 @@ import numpy as np
 HERE = os.path.dirname(os.path.abspath(__file__))
 sys.path.insert(0, HERE)
 from repeated_judge_power import (  # noqa: E402
+    Candidate,
     DEFAULT_GAMMAS,
     DEFAULT_MEAN_RIDGES,
     DEFAULT_RHOS,
     MAX_PROMPT_ROWS,
     PRIMARY_ENDPOINTS,
+    PowerReplicate,
     SCENARIOS,
     SCENARIO_BY_NAME,
     SYNTHETIC_CORPORA,
+    _replicate_geometry_and_splits,
     aggregate_power_records,
     calibrate_synthetic_selector_null,
     derive_seed,
+    draw_repeated_field,
+    finite_null_maximum_threshold,
+    inner_candidate_search,
     run_power_replicate,
     summary,
 )
@@ -66,6 +75,17 @@ FULL_INNER_FOLDS = 3
 FULL_SHRINKAGE = 0.05
 FULL_MISSING_RATE = 0.02
 FULL_SEED = 1207000
+
+CHECKPOINT_SCHEMA_VERSION = 1
+_BLAS_ENVIRONMENT_VARIABLES = (
+    "OPENBLAS_NUM_THREADS",
+    "OMP_NUM_THREADS",
+    "MKL_NUM_THREADS",
+    "BLIS_NUM_THREADS",
+    "VECLIB_MAXIMUM_THREADS",
+    "NUMEXPR_NUM_THREADS",
+)
+_WORKER_THREADPOOL_LIMIT = None
 
 SCIENCE_ARGUMENTS = (
     "config",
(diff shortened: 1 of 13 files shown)


prototypes/mu_cosine/run_repeated_judge_power.py
```diff
@@ -0,0 +1,1183 @@
...
+    return {"bytes": size, "sha256": digest.hexdigest()}
+
+
+def _blas_runtime_identity():
+    try:
+        from threadpoolctl import threadpool_info
+    except ImportError:
+        return []
+    fields = (
+        "user_api", "internal_api", "prefix", "filepath", "version",
```
s243a: **Reproducibility (LOW, recurring theme).** `_blas_runtime_identity` includes the absolute filesystem path to the BLAS shared library (`"filepath"`), and this is serialized into `payload.provenance.blas_runtime` and folded into the run fingerprint (~line 751). It's empty (`[]`) only when `threadpoolctl` is absent (as in my review env); on a machine that has it — likely your four-core runs — the output JSON embeds a machine-specific `.so` path, so the artifact is non-reproducible across machines and checkpoints become non-portable (fingerprint mismatch after a BLAS reinstall at a different path, even when numerically identical).

This is exactly the "absolute path in a provenance block" class that #3695 eliminated. Credit where due — the two prior defects are correctly avoided here: `wall_seconds` is stdout-only, and the file provenance stores `{bytes, sha256}` with no runner path. Only this BLAS `filepath` slipped through.

**Suggested fix:** drop `filepath`; keep `version` / `threading_layer` / `architecture`. (Minor adjacent note: `worker_blas_threads: 1` in provenance is hard-coded even on the serial path when `threadpoolctl` is absent, where the main process isn't actually pinned — the field slightly overstates enforced threading.)

---

s243a (commented):
## Review — preregister batched repeated-judge covariance campaign

**Verdict: the strongest of the series — approve. No correctness or safety defects across ~5,000 lines of code. One LOW reproducibility finding (a BLAS library path in a provenance block) that echoes the theme from #3695.** Draft, no-spend preregistration + tooling PR; all deployment gates verified false/null.

I verified the eigenmode algebra by hand and empirically, reproduced the tests and checkpoint determinism, and ran two independent deep reviews (campaign+sampler; runner+gates). All converge.

### Scope
10 commits, 13 files, +7783 (three design docs; campaign harness 1464; power/eigenmode library 1869; runner 1183; sampler 439; ~2000 lines of tests).

### Verified
**The compound-symmetry eigenmode algebra is mathematically exact.** For an exchangeable prompt block, `Σ = I_m⊗A + (J_m−I_m)⊗B` block-diagonalizes into a collective mode `A+(m−1)B` and `m−1` contrast modes `A−B`. The code builds exactly these (`marginal±shared_request`), maps `w₀=√m·mean` and centered residuals through the correct factors, and sums `logdet` correctly — for **both** the Gaussian and latent-state-posterior paths. Empirically the dense-vs-eigenmode max diff is **0.0** (Gaussian) and **4.4e-16** (posterior), matching the claimed 2.22e-16 / 3.33e-16. The exchangeability guard (identical `observed`+`counts`) is the correct sufficient condition, with dense fallback otherwise, plus a **dense recheck within 1e-12 of the strict `>0` boundary** so a last-bit difference can't flip a promotion.

**Statistical core is sound:** full-procedure family-wise null (re-runs the entire selector under block-null across both corpora/all folds — the #3675 lesson), strict `observed > threshold` fail-closed promotion, prompt-block-clustered wild-multiplier inference.

- **67 focused tests pass** (agent subsets: 19 + 24).
- **Byte-identical checkpoint resume reproduced** (`cmp` identical), a mismatched checkpoint (seed+1) is **rejected** with a fingerprint error, and **1-vs-2-worker output is byte-identical** (multiprocessing determinism).
- **Fail-closed gates confirmed:** the two zero-coupling false-promotion gates (block-null, smooth-mean) use a one-sided exact-binomial control and drive `pass=False` on over-promotion; smoke/custom runs **cannot** set sizing fields (all gated on `--full-prereg`, which rejects scientific overrides); the null barrier is enforced by **data dependency** (threshold feeds the power tasks) and is resume-safe.
- **All deployment gates false/null** (`final_campaign_G_recommendation: null`, `deployment_pass: false`, `real_covariance_deployment: false`; QR/batching/CUDA in `unsimulated_required_gates`).
- **Keyed response validation** is robust with no bypass: `request_id` binds the exact request bytes + row set, row-id/request join enforced, retries preserved, and **provider call-id replay rejected**.

### Finding (LOW — reproducibility, recurring theme)
`_blas_runtime_identity` (`run_repeated_judge_power.py:706`) serializes the absolute BLAS shared-library path into `provenance.blas_runtime` and the run fingerprint — the same "absolute path in a provenance block" class #3695 eliminated. Empty here only because `threadpoolctl` is absent; present on a typical multi-core host. Details + fix in the inline note (drop `filepath`, keep version/threading_layer/architecture). Credit: `wall_seconds` is correctly stdout-only and file provenance is `{bytes, sha256}` with no runner path — only this one field slipped through.

- `worker_blas_threads: 1` in provenance is hard-coded even on the serial path when `threadpoolctl` is absent (main process not actually pinned) — slightly overstates enforced threading.
- `batch_size > 10` is a *soft* gate (downgrades to smoke-only) rather than a hard cap — safe now (no live-call path exists), but should become a hard reject before any live scoring.

_Automated review posted as a comment (not a formal approval vote). The eigenmode math, fail-closed gating, keyed-response integrity, and checkpoint/determinism engineering are all in excellent shape; recommend addressing the BLAS-path provenance nit for cross-machine artifact/checkpoint portability._

---

PR #3707, opened 2026-07-13
mu_cosine: fail closed on repeated-judge candidate capacity

s243a:
## Summary

This is a no-spend structural preflight for the repeated-judge graph-covariance campaign.

The preregistered 10% connected-component concentration cap makes the repository's frozen candidate design infeasible at every registered per-corpus size before candidate enumeration, historical filtering, Nomic embedding, or judge calls.

The primary necessary bound deliberately charges only one endpoint to each declared source component:

```text
U_1(G) = sum_s min(|V_s|, max(1, floor(0.10 G)))
```

That conservative proof avoids depending on how the preregistration's disconnected distant comparator is interpreted. The sharper same-source four-endpoint bound is retained only as a sensitivity.

## Frozen-graph result

| G per corpus | cap/source | exploratory U1 | fresh U1 | joint gate |
|---:|---:|---:|---:|---|
| 160 | 16 | 143 | 16 | fail |
| 320 | 32 | 194 | 32 | fail |
| 512 | 51 | 251 | 51 | fail |
| 800 | 80 | 335 | 80 | fail |

- exploratory: 84,136 nodes, 196,876 unique undirected edges, 34 components
- fresh: 75,901 nodes, 99,971 unique undirected edges, one component
- U4 sensitivity, exploratory: 58/93/131/189
- U4 sensitivity, fresh: 16/32/51/80

Because both corpora are required, no registered campaign can proceed under the current source definition. Later hop/cell/agreement/history/packing constraints can only remove capacity while the audited source components remain frozen.

## What changed

- adds a deterministic, path-portable capacity preflight
- makes an infeasible gate write its audit and exit 2 by default, so shell callers fail closed without remembering an extra flag
- provides explicit `--audit-only` reporting mode; it changes only the exit status and never unlocks an authorization
- records content provenance for the frozen graphs, dataset identity, and all loader/config implementation dependencies
- commits the byte-identical 14,322-byte reproduction artifact (SHA-256 `403b1a59b6e53c87f881dad65ab6af2f401e046c0c7009a1efd220eb3899cdd8`)
- documents the blocker, rejected alternatives, and the outcome-blind source-group amendment boundary
- corrects the frozen Nomic task-prefix bytes from `"clustering:"` to `"clustering: "` (trailing U+0020), matching the cache builder
- leaves candidate, Nomic, judge, campaign, covariance, QR, and CUDA authorizations false

No endpoint was selected, no embedding/model was loaded, and no provider was contacted.

## Review follow-up

Claude's official review found no correctness or safety defect and noted one optional orchestration risk: exit 2 previously required `--require-feasible`. This is now inverted:

- blocked default: writes JSON, then exits 2
- blocked `--audit-only`: writes byte-identical JSON, then exits 0
- feasible default: exits 0
- load, validation, and write failures still propagate

Pipeline automation should not use `--audit-only`. Exit 2 is also argparse's usage-error status, so artifact presence plus `decision.candidate_builder_must_stop` distinguishes a completed capacity block from invalid invocation.

## Validation

- 24 focused capacity tests
- 92 repeated-judge regression tests
- 111 inherited geometry/posterior/square-root tests passed; 9 expected skips
- real default replay writes the tracked artifact and exits 2
- default, `--audit-only`, and `--lmdb-no-lock` outputs are byte-identical
- `py_compile` and `git diff --check` pass
- independent post-review audit found no remaining actionable issue

## Next decision

Amend the dependency unit outcome-blindly—most plausibly deterministic exclusive branch/source groups with multi-branch assignment, concentration caps, fold containment, and group-aware inference—then rerun this necessary gate. Do not silently relabel branches as connected components or relax the cap after seeing capacity.

### DIFF
--- a/prototypes/mu_cosine/run_repeated_judge_candidate_capacity.py
+++ b/prototypes/mu_cosine/run_repeated_judge_candidate_capacity.py
@@ -4,6 +4,9 @@
 The command uses the repository's canonical graph loaders and emits a portable,
 content-addressed JSON result.  It consumes no scores, residuals, embeddings, or
 judge responses and cannot authorize a live call.
+
+An infeasible gate exits with status 2 after writing the audit.  ``--audit-only``
+is the explicit reporting-mode escape hatch when a zero exit is required.
 """
 
 from __future__ import annotations
@@ -293,9 +296,12 @@ def build_arg_parser():
     parser.add_argument("--out", required=True)
     parser.add_argument("--lmdb-no-lock", action="store_true")
     parser.add_argument(
-        "--require-feasible",
+        "--audit-only",
         action="store_true",
-        help="return status 2 when the necessary structural capacity gate fails",
+        help=(
+            "return zero after a completed audit even when infeasible; "
+            "does not unlock anything"
+        ),
     )
     return parser
 
@@ -312,7 +318,7 @@ def main(argv=None):
         "artifact": artifact,
         "candidate_builder_must_stop": payload["decision"]["candidate_builder_must_stop"],
     }, indent=2, sort_keys=True))
-    if args.require_feasible and payload["decision"]["candidate_builder_must_stop"]:
+    if not args.audit_only and payload["decision"]["candidate_builder_must_stop"]:
         return 2
     return 0
 
--- a/prototypes/mu_cosine/run_repeated_judge_candidate_capacity.py
+++ b/prototypes/mu_cosine/run_repeated_judge_candidate_capacity.py
@@ -1,0 +1,321 @@
+#!/usr/bin/env python3
+"""Audit repeated-judge structural capacity before candidates or Nomic.
+
+The command uses the repository's canonical graph loaders and emits a portable,
+content-addressed JSON result.  It consumes no scores, residuals, embeddings, or
+judge responses and cannot authorize a live call.
+"""
+
+from __future__ import annotations
+
+import argparse
+import json
+import os
+from pathlib import Path
+import sys
+
+
+HERE = Path(__file__).resolve().parent
+sys.path.insert(0, str(HERE))
+
+from repeated_judge_campaign import content_record
+from repeated_judge_candidate_capacity import (
+    ENDPOINTS_PER_COMPONENT,
+    GUARANTEED_SOURCE_ENDPOINTS_PER_COMPONENT,
+    SOURCE_COMPONENT_CAP_FRACTION,
+    audit_graph_capacity,
+)
+
+
+SCHEMA_VERSION = 1
+ALGORITHM = "repeated-judge-candidate-capacity-preflight-v1"
+REGISTERED_COMPONENT_SIZES = (160, 320, 512, 800)
+IMPLEMENTATION_FILES = (
+    "repeated_judge_candidate_capacity.py",
+    "run_repeated_judge_candidate_capacity.py",
+    "repeated_judge_campaign.py",
+    "sample_sigma_hop_fresh_corpus.py",
+    "sigma_hop_confirmatory.py",
+    "run_product_kalman_realdata.py",
+    "run_structured_residual_covariance.py",
+    "mu_attention.py",
+    "lmdb_id.py",
+)
+
+
+def _json_bytes(value):
+    return (
+        json.dumps(
+            value,
+            ensure_ascii=False,
+            indent=2,
+            sort_keys=True,
+            allow_nan=False,
+        )
+        + "\n"
+    ).encode("utf-8")
+
+
+def _implementation_records(root=HERE):
+    root = Path(root)
+    return {
+        name: content_record((root / name).read_bytes())
+        for name in IMPLEMENTATION_FILES
+    }
+
+
+def _graph_bundle_record(graph_inputs):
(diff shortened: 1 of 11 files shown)


prototypes/mu_cosine/run_repeated_judge_candidate_capacity.py
```diff
@@ -0,0 +1,321 @@
...
+        "output": os.path.abspath(args.out),
+        "artifact": artifact,
+        "candidate_builder_must_stop": payload["decision"]["candidate_builder_must_stop"],
+    }, indent=2, sort_keys=True))
+    if args.require_feasible and payload["decision"]["candidate_builder_must_stop"]:
+        return 2
+    return 0
+
+
+if __name__ == "__main__":
```
s243a, line 326: This gate is correct and fails closed: `candidate_builder_must_stop = not all_feasible` over a validated-non-empty `registered_sizes`, so `all()` can't be vacuously true, and invalid/missing inputs raise before reaching either `return`. 👍

Minor observation (not a defect): the exit-2 signal is **opt-in** — without `--require-feasible`, an infeasible design still exits 0. That's fine here because the written JSON always carries `decision.candidate_builder_must_stop: true` and all 8 `authorization.*` flags false, so any JSON-reading consumer stays blocked regardless of exit code. Only a shell caller that keys solely on `$?` and forgets the flag would miss the block — worth keeping in mind for whatever orchestrates the eventual pipeline.

---

s243a (commented):
## Review — fail closed on repeated-judge candidate capacity

**Verdict: correct and cleanly fail-closed — approve. No correctness or safety defects, and notably the first PR in this series whose claimed artifact SHA-256 reproduces byte-for-byte.** Draft, no-spend preflight; all downstream authorizations false.

I verified the capacity math by hand, recomputed the bounds independently from the tracked component sizes, checked the Nomic-prefix fix against the cache builder, and ran an independent deep review of the runner + reproduction. All converge.

### Scope
Single commit `57dc2076` — 11 files, +1460 (capacity library 206, runner 321, tests 265, report/decisions/prereg edits, tracked `summary.json`, and a 4-line `repeated_judge_campaign.py` constant fix).

### Verified
**The necessary-bound math is rigorous.** `U₁(G) = Σ_s min(|V_s|, cap)` with `cap = max(1, floor(0.10·G))` is a valid *necessary* upper bound: endpoint-disjoint selection caps each source component at `|V_s|` candidates and the 10% concentration cap at `floor(0.10·G)`. Charging only **one** endpoint per component (not four) makes it deliberately loose, so `U₁(G) < G` is a conclusive infeasibility proof independent of how the "distant/disconnected" comparator is interpreted — and every later constraint (hop cells, degree quartiles, agreement, history, packing) can only reduce capacity. The `+1e-12` floor guard is correct. `U₄` (four-endpoint) is retained only as a sharper sensitivity.

**The result is decisive and I reproduced it exactly.** Recomputing from the tracked component sizes: exploratory `U₁ = 143/194/251/335`, fresh `U₁ = 16/32/51/80`, `U₄` sensitivity `58/93/131/189` and `16/32/51/80` — all matching the report, with node/edge/component counts (84136/196876/34; 75901/99971/1) matching too. The single-component **fresh** corpus is the clincher: with one component, `U₁ = cap = floor(0.10·G) < G` for every G, so no registered size can ever pass while the source definition stays frozen.

**Empirical/safety checks:**
- **20 focused tests pass**; `--require-feasible` **fails closed** (exit 2, non-vacuous — see inline note), and edge-case/missing inputs raise before any `return 0`.
- **All 8 authorizations false** (`candidate_enumeration`, `nomic_embedding`, `candidate_builder`, `judge_calls`, `live_campaign`, `covariance_deployment`, `qr_specialization`, `cuda_claim`); only `candidate_enumeration` is feasibility-gated, and it's still false since infeasible.
- **Provenance is complete:** `content_record` `{size_bytes, sha256}` for the frozen graph inputs + a canonical graph-bundle hash + fingerprints of all 9 loader/config dependencies — any changed input is detected.

**The recurring reproducibility theme is fully resolved here.** The tracked `summary.json` is byte-identical to the claimed artifact — SHA-256 `0524769b…`, 14322 bytes — with **no** `wall_seconds`, **no** absolute paths, and **no** BLAS-library path (the `os.path.abspath` is stdout-only). This is the first PR in the #3675→#3685→#3695→#3701 arc where I could reproduce the claimed hash exactly.

**Genuine correctness fix:** `FROZEN_NOMIC_PREFIX` `"clustering:"` → `"clustering: "` (trailing U+0020) correctly matches the cache builder's `MODEL_SPECS["nomic"].task_prefix` (`independent_embedding_cache.py:43`). The old value would have mismatched the actual embedding bytes — a real latent bug caught here.

### Minor / optional (not a defect)
- The `--require-feasible` exit-2 signal is opt-in; the always-present `candidate_builder_must_stop: true` + false authorizations keep JSON-reading consumers blocked regardless. Detail in the inline note.

_Automated review posted as a comment (not a formal approval vote). The proof, the byte-exact reproducible artifact, the fail-closed gate, complete provenance, and the Nomic-prefix fix are all in excellent shape — this cleanly establishes the blocker and correctly refuses to proceed._

---

PR #3726, opened 2026-07-14
mu_cosine: audit topology-only source regions and fail closed

s243a:
## Summary

- add a deterministic topology-only source-region partitioner that returns exactly `K` exclusive induced-connected regions across true weak components;
- define the exact three-hop core as `core(R) = {v in R : B_3(v) is a subset of R}`, retain `weak_component_id` separately, and evaluate the four-endpoint 10% cap bound `U4`;
- add a portable, fail-closed two-corpus audit runner for the frozen `K={64,96,128}` and `G={160,320,512,800}` grids;
- bind the frozen cumulative-walk weights and feature implementation, recheck implementation identity before output, and commit the content-addressed reproduction artifact;
- amend the preregistration and append the theory, rejected alternatives, result, non-claims, and next dependence-aware fork.

## Result — compute-only / no-spend

No source-region count passes jointly.

| corpus | K | core fraction | U4(800) |
|---|---:|---:|---:|
| exploratory | 64 / 96 / 128 | 2.95% / 2.48% / 2.42% | 391 / 314 / 295 |
| fresh | 64 / 96 / 128 | 32.66% / 31.28% / 30.66% | 4,056 / 4,805 / 5,003 |

Exploratory fails both the frozen 50% core-retention gate and larger-`G` four-endpoint capacity. Fresh passes capacity but fails retention. Increasing `K` does not repair the joint design.

Accordingly, the selected partition is `null`; history, candidate enumeration, Nomic, builder verification, judge calls, covariance deployment, QR specialization, and CUDA all remain locked.

This does **not** show that useful cross-item correlation is absent. It shows that demanding hard three-hop feature-support isolation is too destructive on these hub/small-world graphs. Source regions are design/sensitivity units, not independent observations.

## Correctness and provenance

- The first distance-threshold splitter was discarded after a star-graph regression showed it could return more regions than requested.
- The reported implementation uses deterministic metric-rooted spanning-tree cuts; exact count, connectivity, ordering, weak-component containment, ball containment, and hash-seed invariance are tested.
- Implementation identity is snapshotted before the run and rechecked before output. The graph inputs are content-addressed and required to remain read-only during the audit.
- Reproduction artifact: 64,449 bytes, SHA-256 `66e3fd0deb4f63a3c20e52d04ef1210490bb17f3dc3f33794156dc4e67d3a46f`.

## Validation

- 25 focused source-region/runner tests passed
- 117 repeated-judge regressions passed
- 15 graph-geometry tests passed
- Python compilation and `git diff --check` passed
- independent final review matched every table value, gate, authorization, and embedded implementation hash to the artifact

The repeated-judge and graph-geometry suites were run in separate pytest processes so test-order BLAS loading does not alter the inherited runtime-provenance check.

## Non-claims and next step

This PR reads no outcomes, historical inventory, candidate pool, Nomic cache, or judge response; it runs no embedding model, conditioner, QR specialization, or CUDA kernel. It deliberately leaves the v2 `source_component` selector unchanged because no usable v3 partition passed.

The recommended next PR is a prospectively powered dependence-aware source design: keep an exclusive concentration label, quantify cross-region graph exposure/effective information, and carry statistically promoted covariance into dense joint QR instead of pretending hard regional independence.


prototypes/mu_cosine/repeated_judge_source_regions.py
```diff
@@ -0,0 +1,600 @@
...
+                cut_edges.append((node, neighbor))
+                boundary.update((node, neighbor))
+    distances = {node: 0 for node in boundary}
+    queue = deque(sorted(boundary, key=_node_key))
+    while queue:
+        node = queue.popleft()
+        # A node whose nearest cut-edge endpoint is at distance d has a node
+        # from another region at distance at most d + 1.  Its radius-r ball is
+        # therefore safe exactly when d >= r.  Thus the halo cutoff is r - 1.
+        if distances[node] >= max(support_radius - 1, 0):
```
s243a, line 343: Verified this subtle halo/core cutoff is exact. For `v ∈ R`, any shortest path to another region crosses a cut edge `(a,c)` with `a ∈ R`; since `a` is a boundary node on that path, it's a *closer* boundary node than the other-region endpoint `c`. So the nearest other-region node sits at **exactly** `d(v)+1` (never closer), which makes `B_r(v) ⊆ R ⟺ d(v) ≥ r` — i.e. the `distance < support_radius` halo is precisely right, and including the c-side boundary nodes as BFS sources is harmless because they're always farther. Nicely, this also yields `certified_minimum_cross_region_core_distance = 2·halo_hops+1`, so with `halo_hops = support_radius = 3` two different-core nodes are ≥7 apart and their radius-3 cumulative-walk feature balls are provably disjoint (7 > 2·3). Clean construction — no issue here.

---

s243a (commented):
## Review — audit topology-only source regions and fail closed

**Verdict: correct, mathematically careful, and cleanly fail-closed — approve. No correctness or safety defects; the artifact SHA-256 reproduces byte-for-byte and the recurring provenance theme stays resolved (2nd clean PR running).** Draft, no-spend preflight; all downstream authorizations locked.

I verified the partitioner and core-isolation math by hand, recomputed every headline number independently from the tracked artifact, and ran an independent deep review of the runner + provenance. All converge.

_(CI note: the `test` job is green; `mergeable_state: unstable` is only an unrelated "C# Query Runtime Smoke" check still in progress — not a failure and unrelated to this Python-only change.)_

### Scope
Single commit `9ceb63ae` — 10 files, +4332 (partitioner library 600, runner 339, tests 240+214, tracked `summary.json`, report/decisions/prereg edits).

### Verified
**The partition construction is correct and guarantees exactly K regions.** `_allocate_counts` apportions K across weak components (each ≥1, ≤size) via exact integer ratio comparison, and `_metric_tree_cut` splits each piece by cutting one metric-rooted spanning-tree edge — both sides stay connected — with an integer-proportional part allocation. This structurally fixes the discarded distance-threshold splitter that over-produced regions on hubs/stars (enforced by an assertion at `:442-443` and the `exact_requested_region_count` gate, tested on a star graph).

**The three-hop core/halo cutoff is exact** — the subtle part, and it's right (detail in the inline note): for `v∈R` the nearest other-region node is *exactly* `d(v)+1`, so `B_3(v)⊆R ⟺ d(v)≥3`, and `certified_minimum_cross_region_core_distance = 2·halo_hops+1 = 7` provably makes different-core radius-3 walk-feature supports disjoint.

**The result is decisive and I reproduced it exactly** from the tracked core sizes: exploratory core fraction 2.95%/2.48%/2.42% and U4(800) 391/314/295 (fails both the 50% retention gate and large-G capacity); fresh core fraction 32.66%/31.28%/30.66% and U4(800) 4056/4805/5003 (passes capacity, fails 50% retention). `all_topology_gates_pass=False` for all six cells; every K fails the joint two-corpus gate with `passing_corpora=[]`; `selected_region_count=null`. Increasing K doesn't repair it.

**Safety / provenance:**
- **25 focused tests pass**; runner `--help` parses; blocked audit exits 2 (`--audit-only` exits 0).
- **Fail-closed:** all 9 authorizations false; only `historical_inventory_unlocked` is data-driven and requires *both* required corpora to pass (hardcoded `("exploratory","fresh")`, non-vacuous) — no path flips a downstream authorization true.
- **Implementation-identity recheck is real, not a no-op:** the runner snapshots 12 loader/config file hashes, re-reads them fresh after the computation and before writing, and raises on any diff (tested). Graph inputs are content-addressed.
- **Recurring provenance theme absent:** the written artifact has no `wall_seconds`, no absolute paths, and no BLAS-library path (the `os.path.abspath` is stdout-only). Tracked `summary.json` = **64449 bytes, SHA-256 `66e3fd0d…`**, byte-identical to the claim, with all 12 embedded implementation hashes matching the committed source. This is the second consecutive PR (after #3707) whose claimed hash reproduces exactly.

### Framing
The report is honest about scope: this doesn't show cross-item correlation is absent — it shows that demanding hard three-hop feature-support isolation is too destructive on these hub/small-world graphs, and that source regions are design/sensitivity units, not independent observations. The recommended next step (a powered dependence-aware design that carries statistically promoted covariance into dense joint QR rather than pretending hard regional independence) is the right call.

No defects found. _Automated review posted as a comment (not a formal approval vote)._

---

PR #3735, opened 2026-07-14
mu_cosine: define dependence-aware source-region topology bridge

s243a:
## Summary

- replace the failed three-hop-core campaign construction with full exclusive source regions plus a PSD topology-only exposure Gram,
  `E = normalize(Z) normalize(Z)^T`, from region-average cumulative-walk landing profiles;
- add deterministic cap-constrained exposure-aware allocations, exact equal-mean design-effect/ESS diagnostics, and numerically guarded allocation-free ESS floors over the frozen `K/G/rho` grid;
- add a portable fail-closed runner with start/end implementation and external-graph provenance checks, path-free numerical-runtime identity, full and tracked-summary output modes, and complete tests;
- amend the preregistration, decision log, earlier source-region/power reports, and document the Stage A / Stage B authorization order.

## Result

All `K={64,96,128}` pass the narrow structural gates in both frozen corpora at every `G={160,320,512,800}`. No `K` is selected.

At `G=800, rho=.20`, the prospective-allocation effective component counts are:

| corpus | K=64 | K=96 | K=128 |
|---|---:|---:|---:|
| exploratory | 103.3 | 174.2 | 235.3 |
| fresh | 224.7 | 302.6 | 365.3 |

The allocation-free guarded floors are 19–25 for exploratory and 33–37 for fresh. These are sensitivity diagnostics under a stipulated source-correlation path, not empirical residual covariance or campaign power.

The two final computations produced the same complete payload record: 2,767,735 bytes, SHA-256 `bf9a09c35e54bd36c2e7efea19c432ccf1e9105ff67c4154cfc1c6e744a843b2`. The tracked review projection is 308,090 bytes and content-addresses every omitted exact matrix/quota.

## Claim and authorization boundary

This PR is no-spend and topology-only. It reads no history, candidates, embeddings, labels, or judge responses. It does not select a region count, estimate covariance, claim end-to-end power, or authorize history, candidate construction, Nomic, judge calls, covariance deployment, independent batching, QR specialization, or CUDA. Every authorization remains false.

The next step is Stage A: extend the complete repeated-judge null/power procedure with the registered source matrices, source-atomic folds, synthetic prompt incidence, and two-way prompt/source inference. Only a passing family-wise null / >=80% power gate may unlock identity-only history. Stage B then performs exact post-exclusion candidate packing and realized-design revalidation.

`JointPosterior` remains the calibrated decision comparator; it is not this exposure kernel or the square-root/QR factorization.

## Validation

[...]

PR #3742, opened 2026-07-14
mu_cosine: preregister corpus-specific source-dependence power harness

s243a:
## Summary

- Freeze the no-history Stage-A source-power procedure over all 25 ordered `(exploratory eta_source, fresh eta_source)` pairs from `{0, .025, .05, .10, .20}²`.
- Add source-atomic outer/inner folds, split-contained source-spreading prompt blocks, an equal-component marginal quasi-NLL selector, and graph-aware prompt-plus-source max-t inference.
- Calibrate the selector against two null families in every source-strength pair, then require exact one-sided Clopper–Pearson gates for discovery and a seed-disjoint fixed-pair confirmation.
- Add a content-addressed exact source-design bundle, deterministic/checkpointable bounded-spawn runner, compact checkpoint shards, path-free provenance, and full science/bundle/runner tests.
- Record the exact prompt-by-source incidence matrices and fold diagnostics in the result payload so the frozen split gates are independently auditable.
- Document the Stage-B order, rejected alternatives, and the separation between this covariance study, `JointPosterior`, and the downstream square-root/QR conditioner.

## Why

The previous bridge treated source dependence on a shared diagonal of equal-strength corpus worlds. That is insufficient: for every frozen design with `G > K`, `H E Hᵀ - I` is indefinite, so `.20` is not a universal Loewner upper envelope, and the exploratory and fresh corpora need not have the same source coupling. This harness therefore calibrates and evaluates the complete corpus-specific nuisance family instead of relying on one “worst” scalar.

## Frozen decision contract

- Null barrier: 2 nulls × 25 source-strength pairs × 1,999 draws; one-based rank 1,900 with a strict comparison, using the maximum cell threshold.
- Discovery: 200 replicates per scenario/pair; choose the smallest passing `G`, then the coarsest/smallest passing `K`.
- Confirmation: fresh calibration plus 200 seed-disjoint replicates at the fixed selected pair.
- Family: `M=425` Bonferroni-adjusted exact binomial gates. At `alpha=.05/425`, primary/topology gates require at least 180/200 successes and null/deranged gates permit at most 5/200 promotions.
- Failure is terminal under this preregistration unless a future prospective amendment accounts for the failed attempt through cross-attempt alpha spending or closed testing.
- Only passing confirmation may unlock the attempted-input **identity** inventory. Nomic data, candidate enumeration, judge calls, deployment, QR specialization, and CUDA remain locked.

## Scope and result

This is a tooling/protocol PR, **not a power result**. The immutable discovery and confirmation runs have not been launched. No attempted-input history, real Nomic embedding, candidate identity, label, judge response, or model output was read, and no judge/model call was made. Every downstream authorization remains false.

The exact full workload is large—up to 1,291,620 expensive corpus worlds—so the report deliberately gives auditable task counts rather than extrapolating an ETA from a one-replicate pilot.

## Validation

- 56 source-power science/bundle/runner tests passed.
- 84 inherited source-dependence/source-region/capacity tests passed.
- 49 inherited baseline science/runner tests passed.
- 189 related tests total across three fresh pytest processes; the inherited baseline spawn suite is isolated because it fingerprints the BLAS runtimes loaded in its process.
- `py_compile` and `git diff --check` passed.
- Final provenance-matching `K=128, G=800, R=3` audit-only pilot with the full 5×5 pair grid and 999 multiplier draws passed:
  - 17.92 s GNU wall time; 78,800 KiB maximum RSS;
  - 620,435-byte payload, SHA-256 `3b74736d8444e9c68b17621ac4f4e6f95c57c61b9adbfa339fee4a922677c195`;
  - five checkpoint files / 89,546 bytes;
  - every authorization false, as required.

The full-output SHA is an integrity check for the recorded Python/NumPy/BLAS stack, not a portable scientific digest. Across numerical stacks, compare scientific fields within numerical tolerance and require the registered gate booleans to agree; require exact SHA equality only when the numerical stack and scientific inputs match.

## Next step after review

Run the immutable Stage-A discovery using durable checkpoint storage. If—and only if—its fixed-pair seed-disjoint confirmation passes, proceed to the identity-only Stage-B inventory and realized-design rerun in the preregistered order.


prototypes/mu_cosine/run_repeated_judge_source_power.py
```diff
@@ -0,0 +1,2172 @@
...
+        "source_design_bundle": source_design_bundle_identity(bundle),
+        "source_dependence_full_payload": copy.deepcopy(
+            bundle["parent"]["full_payload_record"]
+        ),
+        "reviewed_source_summary": copy.deepcopy(
+            bundle["parent"]["reviewed_summary_record"]
+        ),
+        "python_version": sys.version,
+        "numpy_version": np.__version__,
+        "blas_runtime": _blas_runtime_identity(),
```
s243a, line 795: Provenance is correctly path-free here — `_portable_blas_runtime` reuses the #3735 fix and drops `filepath`/`num_threads`, and `wall_seconds` stays stdout-only. 👍

One observation on the reported pilot SHA (`3b74736d…`), not a defect: unlike the #3707/#3726/#3735 artifacts, this full payload embeds the *environment* — the portable `blas_runtime` (OpenBLAS **version**/threading_layer), `python_version`, and numpy float decimal-serialization — so the hash is bit-deterministic **within** a stack but won't cross-reproduce across stacks. On my box (numpy 2.4.6, `threadpoolctl` absent → `blas_runtime=[]`) the pilot came out `9d385667…`/620,142 B with identical *structure* (5 checkpoints, all authorizations false) and byte-identical across two runs. That's honest provenance-binding, but since the report presents the SHA as a reproduction target, consider adding the same "compare scientific fields within tolerance across numerical stacks" caveat the #3695 report carried, so a reviewer on a different stack isn't surprised the hash differs while every gate still matches.

---

s243a (commented):
## Review — preregister corpus-specific source-dependence power harness

**Verdict: statistically rigorous, correctly fail-closed, and well-engineered — approve. No correctness or safety defects.** Draft, no-spend Stage-A tooling/protocol PR; the immutable runs aren't launched and every downstream authorization is false. One low-severity documentation note on the pilot SHA's cross-stack reproducibility (inline).

I verified the statistical decision core by hand and by execution, and ran two independent deep reviews (science library; runner + bundle). All converge.

### Scope
Single commit `d08774b1` — 13 files, +7828 (science library 1959, runner 2172, bundle 882, tests 931+333+786, a 969,914-byte content-addressed `source_design.json`, and doc/prereg updates).

### Verified
**The corpus-specific covariance model is correct and the motivation is sound.** `source_component_factor` builds `A = [√(1−η)·I | √η·H·L_E]`, so `A Aᵀ = (1−η)I + η·H E Hᵀ` — PSD by construction. The "Why" holds: since `H E Hᵀ − I` is indefinite for G>K, no single scalar (`.20`) is a universal Loewner envelope, which justifies calibrating the full corpus-specific `(η_exploratory, η_fresh)` family rather than a shared diagonal.

**The decision contract is exactly as specified** (verified by execution):
- Null threshold = order statistic at one-based rank `ceil(0.95·2000) = 1900` of 1999 draws, from a **full-procedure family-wise** null that re-runs the whole selector across all folds/both corpora (not a fixed-path shortcut).
- `confirmation_binomial_family_size() = (2·5 + 2·6·5 + 3·5)·5 = 85·5 = 425`; α = 0.05/425.
- **Exact one-sided Clopper–Pearson gates map to the claimed cutoffs precisely:** primary/topology require `CP-lower(successes,200,α) ≥ 0.80` ⟺ **≥180/200** (180→0.8008 passes, 179→0.7944 fails); null/deranged require `CP-upper ≤ 0.10` ⟺ **≤5/200** (5→0.0934 passes, 6→0.1014 fails). Bisection on exact `math.comb`/`math.fsum` tails; directions are fail-closed.
- Discovery uses α=0.05 as a selection screen; the strict M=425 Bonferroni applies at the seed-disjoint fixed-pair confirmation — a sound discovery/confirmation split.

**Science library (independently reviewed):** source-atomic folds hold whole regions out together with leak-free outer/inner nesting and split-contained prompt blocks (asserted); source-spreading prompt blocks are outcome-blind and deterministic; the equal-component marginal quasi-NLL selector is an exact dense Gaussian composite with strict `>0` eligibility; the graph-aware max-t is **genuinely two-way** (source + prompt clusters), sweeps the full η-grid, and promotes on the worst-case `min` over η so dependence is never ignored; the unidentified-multiplier case **fails closed** (`MultiplierNotIdentifiedError` → `promoted=False`), and the omitted intersection term only *inflates* SE (conservative).

**Runner + bundle (independently reviewed):** all 8 downstream authorizations hardcoded false; the only unlockable field requires exact `--full-prereg` **and** a passing seed-disjoint confirmation (`provisional_design_unlocks_nothing`), proven even against a worker monkeypatched to claim a pass. Checkpoints are atomic, resume is **byte-identical**, and a changed config/code/seed is rejected by fingerprint; a hard null-calibration barrier can't be skipped by resume. Bounded-spawn workers use one BLAS thread each with order-independent index reassembly and content-derived, phase-disjoint seeds. The bundle is content-addressed (deterministic canonical JSON) with tested change detection.

**Provenance** is path-free (reuses the `_portable_blas_runtime` fix from #3735; `wall_seconds` stdout-only) — the recurring theme stays resolved.

**Empirical:** 55 focused tests pass; `--help` parses; the audit-only pilot reproduced exactly in structure (5 checkpoints, ~620 KB payload, all authorizations false), bit-deterministic within-stack.

### Notes (non-blocking)
- **Pilot SHA is environment-bound** (inline): the full payload records the numerical-runtime identity (BLAS version, `python_version`, numpy float serialization), so the reported hash is deterministic within a stack but not across stacks — suggest the #3695-style "compare scientific fields within tolerance" caveat.
- **Test-coverage suggestion:** a dedicated `CheckpointStore` fingerprint-mismatch reopen test (config/code/seed) would lock in the manifest-rejection path the second reviewer verified manually.

No defects found. _Automated review posted as a comment (not a formal approval vote)._

---

PR #3770, opened 2026-07-14
KT-HEAP-SNAPSHOT-OPT-2: peel T4 get_constant so append recursion wins

s243a:
## Summary

Eliminate the residual per-recursive-entry T4 `_t4` map copy so lowered `append` under `emit_mode(functions)` beats the interpreter — not just ~0.85× after KT-DISPATCH-SNAPSHOT-OPT.

## Profile first (append_500, functions, 80×5, `WAM_KT_PROFILE=1`)

Instrumentation now times **all** `snapshotForNative` calls (including T4 `_t4`):

| config | min_ms | snap_fraction | snap_count | native_entries | max_register_map_size |
|---|---:|---:|---:|---:|---:|
| **BEFORE** (entry `_t4` every hop) | 305.6 | **0.481** | 200 800 | 200 400 | **508** |
| **AFTER** (peel leading `get_constant`) | 17.0 | **0.086** | 800 | 200 400 | 508 |

Map grows ~O(depth) (`H<n>`) → prior copy was ≈O(depth²) total. After peel: snaps ≈ 2× queries (top-level tryRun + base-case nil path only).

## Fix

When a T4 predicate’s first clause starts with `get_constant` / `get_nil` / `get_integer`, emit a **peel**: closed discriminant miss (ground mismatch) jumps to later clauses **without** an entry snapshot. Bindable cases (`null` / `Var` / exact match) still snapshot before committing clause 1. Last remaining clause needs no snapshot.

Append’s recursive cons path is therefore snap-free.

**Skipped:** (a) heap-outside-map reclaim, (b) trail-with-old-values — peel removed the hot-path copy. Documented in `docs/design/WAM_KOTLIN_OPTIMIZATION_HISTORY.md`.

## Hardened bench (5 warmup / 15 batches, min speedup)

| program | before (KT-DISPATCH) | **after** |
|---|---:|---:|
| append_100 | ~1.03× | **7.21×** |
| append_500 | ~0.85× | **30.33×** |
| member_100 / 500 | ~1.4× | ~1.37–1.40× (unchanged; first instr `get_list`) |

## Acceptance

- [x] Unit suite (`tests/test_wam_kotlin_target.pl`) green
- [x] `CONFORMANCE_TARGETS=kotlin,kotlin_functions` green
- [x] `append_100` / `append_500` ≥1.0× with large margin
- [x] Docs: history + bench + fleet card `KT-HEAP-SNAPSHOT-OPT-2` + perf log + cross-target case study

## Env note

[...]

PR #3781, opened 2026-07-15
EMIT-KOTLIN-5: lower deterministic mid-body call + arithmetic builtins

s243a:
## Summary

Lower mid-body deterministic `call` and arithmetic `builtin_call` (`is/2`, compares, `=/2`) so **fib/ack** lower under `emit_mode(functions)` instead of declining.

## Determinism gate (hard correctness constraint)

Inline `if (!dispatch("P/N", state)) return false` takes only the **first** solution and cannot backtrack into the callee. Mid-body `call` lowers **only** when every target is:

- **self-recursion**, or
- a **single-clause deterministic** callee (same gate recursively)

Multi-clause callees decline. Test: `kt5_nd_ok :- choice(X), X = b` stays on the interpreter (would wrong-fail if lowered). When unsure → decline.

## Arithmetic

`kotlinLoBuiltinCall` + top-level `wamEvalArith` — same evaluation as the interpreter `builtin_call` path (no re-implementation).

## Stack ceiling

Linear mid-body-call recursion → `StackOverflowError` around **~750–780** frames on the default JVM stack. Fib/ack need O(n) frames; conformance `fib(10)` / `ack(2,3)` are safe. Prefer decline over wrong answers if a workload would overflow.

Classic non-tail `reverse` via mid-body `append` still declines (`append` is multi-clause). Acc-reverse already lowers via `execute`.

## Bench (hardened 5/15, min speedup)

| program | speedup_min |
|---|---:|
| fib_15 | **1.85×** |
| ack_23 | **1.78×** |
| append_500 | ~28× (unchanged class) |

## Acceptance

- [x] fib/ack LOWER + match interpreter (gradle differential; `registerNative`)
- [x] nondet mid-body caller DECLINES
- [x] unit suite + `CONFORMANCE_TARGETS=kotlin,kotlin_functions` green
- [x] BENCH table extended with fib/ack
- [x] docs: fleet card, STATUS, optimization history (boundary + stack ceiling)

## Env

`LANG=C.UTF-8`; gradle stderr may include daemon/`JAVA_TOOL_OPTIONS` noise — assert no `error:` only.

<div><a href="https://cursor.com/agents/bc-9748a055-a632-4d75-9b1b-6d3cd9dbf421"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-9748a055-a632-4d75-9b1b-6d3cd9dbf421"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/docs/WAM_FLEET_GAP_TASKS.md
+++ b/docs/WAM_FLEET_GAP_TASKS.md
@@ -57,7 +57,7 @@ self-contained so a single coding agent can pick it up in isolation.
 | EMIT-KOTLIN-2 ✅ | Lowered emitter (structures) | Kotlin | M | done — write-mode structures (`cursor/emit-kotlin-structures-f421`) |
 | EMIT-KOTLIN-3 ✅ | Multi-clause deterministic | Kotlin | M | done — T5/T4 no call (`cursor/emit-kotlin-multi-clause-f421`) |
 | EMIT-KOTLIN-4 ✅ | Last-call `execute` | Kotlin | M | done — tail execute (`cursor/emit-kotlin-execute-f421`) |
-| EMIT-KOTLIN-5 | Mid-body `call` | Kotlin | M | fib/ack / non-tail continuation |
+| EMIT-KOTLIN-5 ✅ | Mid-body `call` + arith builtins | Kotlin | M | done — det-only mid-body + is/2 (`cursor/emit-kotlin-5-f421`) |
 | BENCH-KOTLIN ✅ | Lowered vs interpreter timing | Kotlin | S | done — recursion regresses; short cases noise (`cursor/bench-kotlin-f421`) |
 | KT-DISPATCH-SNAPSHOT-OPT ✅ | Perf: cheapen recursive dispatch | Kotlin | M | done — skip recursive tryRun snap (`cursor/kt-dispatch-snapshot-opt-f421`) |
 | KT-HEAP-SNAPSHOT-OPT-2 ✅ | Perf: eliminate residual T4 `_t4` copy | Kotlin | M | done — peel leading get_constant (`cursor/kt-heap-snapshot-opt-2-f421`) |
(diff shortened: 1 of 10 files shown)


s243a:
## Review: approve ✅ — verified correct; one nuanced soundness note (not a blocker)

I focused on the determinism gate, since the whole risk is a mid-body goal missing backtracking.

### Verified
- **fib/ack lower and match the interpreter** (full differential: `fib(10,55)`→T, `fib(10,54)`→F, `ack(2,3,9)`→T, `ack(2,3,8)`→F).
- **The gate's core safety works:** `nd(X) :- choice(X), X=b` (mid-body call to the multi-clause `choice/1`) correctly **DECLINES** and stays on the interpreter.
- **Conformance clean in both modes** (exit 0, 0 tolerated-failures/xfails/errors). **Unit suite green** (and the `CompileErr` hardening from #3770 that you merged is now doing its job — no daemon-noise false failures).
- **Bench reproduced:** fib/ack ~1.8× — mid-body call is a win now that dispatch is cheap.
- Arithmetic reuses the interpreter's `evalArith` (correct by construction). Nice.

### Nuanced soundness note (worth a follow-up, not a blocker)
`kotlin_safe_midbody_callee` exempts **self-recursion unconditionally** — it never checks the self-recursive predicate is deterministic. I built the adversarial case (a *non-deterministic* self-recursive predicate whose post-recursive-call goal needs to backtrack into the recursion):
```prolog
p(z,a). p(z,b).
p(s(N),R) :- p(N,R), R=b.     % first p(z,R)=a → R=b fails → needs backtrack to R=b
```
This **lowers**, and in principle the inline `if (!dispatch) return false` would miss the backtrack. **But it does not produce a wrong observable answer**, because:
1. Missed mid-body backtracking yields a wrong-**false**, never a wrong-true; and
2. the **top-level `tryRun` native→bytecode fallback** (kept from EMIT-KOTLIN-4, only skipped on recursive hops) re-runs the query on the interpreter and gets the right answer. My adversarial queries all MATCHED.
3. Any predicate that could *observe* a wrong internal binding is a multi-clause caller, which the gate declines.

So it's **sound in observable behavior**, but that soundness is **load-bearing on the top-level fallback** — an undocumented coupling. If a future optimization ever drops that fallback (e.g. "lowering is proven, remove it"), these non-det self-recursive cases would silently break. **Recommend a follow-up** to either (a) tighten the exemption to require the self-recursive predicate be deterministic, or (b) explicitly document the "self-recursion safety depends on the top-level fallback" invariant + add a regression test that would catch its removal. I can file that card.

**Good to merge.** fib/ack complete the lowering coverage; correctness is verified for all reachable cases.

---

PR #3929, opened 2026-07-22
plawk: support field comparison patterns

s243a:
## Summary

- parse positive-index bare field comparisons such as `$1 > $2` as `field_cmp2(I, Op, J)`
- lower the new pattern through both the single-rule and multi-rule/combinator guard paths
- project both operands with the active `FS`, including regex FS, and apply POSIX strnum comparison semantics
- add a bounded slice comparator that NUL-terminates exact field spans without persistently interning per-record values
- keep BINFMT use as a clean compiler exit-3 rejection
- document the expression-pattern surface and add system-awk parity coverage

## Semantics

Both operands are input-derived strnums. Comparison is numeric when both field texts look numeric; otherwise it is lexical. Missing fields compare as empty strings.

All six comparison operators are supported, and the pattern composes with `!`, `&&`, `||`, and existing base patterns.

The runtime helper uses one reusable geometrically growing scratch buffer containing the two length-delimited slices. This preserves streaming memory behavior: distinct field values are not retained in the atom registry. Invalid slices or allocation failure return a reserved result that codegen masks to a false guard.

## Validation

- `tests/test_plawk_field_cmp_pattern.pl`: 18/18 passed, with every textual runtime case checked against system awk
- `tests/test_plawk_expr_pattern.pl`: 14/14 passed
- `tests/test_plawk_surface_pattern_combinators.pl`: 19/19 passed
- `tests/test_plawk_strnum.pl`: 29/29 passed
- `tests/test_plawk_regex_fs.pl`: 11/11 passed
- `tests/test_plawk_binary_records.pl`: 14/14 passed
- runtime helper assembled and passed LLVM verification plus direct numeric, lexical, empty, invalid, and overflow probes
- `git diff --check`
- independent review completed with no remaining findings

## Coordination

Adds a `base_pattern` alternative and pattern-guard clauses next to the existing `field_cmp` / `special_cmp` family. NR-pattern work touches the same `special_cmp` area and counter wiring, but this uses a different functor, so conflict risk is low.


s243a:
Superceeded by 3931

PR #3935, opened 2026-07-22
BENCH-R: scale-300 effective-distance matrix row for hybrid WAM R

s243a:
## Summary

Adds the hybrid WAM **R** target to the scale-300 effective-distance benchmark matrix (BENCH-R).

### Reused existing R functionality
- Auto-detected `category_ancestor/4` recursive-kernel plumbing
- `write_wam_r_project/3` with `emit_mode(functions)` and `r_fact_sources`
- Existing accumulated effective-distance helpers from `prolog_target`

### New work
- Fleet hops contract for R's CA kernel: `(+Cat,+Root,-Hops,+Visited)`, with one result per path
- Kind-aware aggregate finalization, including `sum`
- `grouped_by_first_atoms` FactSource so numeric-looking Wikipedia IDs remain atoms
- R benchmark generator and readable Mustache companion runner
- Guarded R block in the cross-target harness
- Focused structural and Rscript end-to-end tests in both functions and interpreter modes
- Permanent hosted-CI R effective-distance regression gate

### Scale-300 result

| Metric | Value |
|---|---:|
| query_ms (3-rep median) | **62595** |
| samples | 65368, 62595, 62559 |
| total_ms | **63424** |
| rows | **271** |
| reference parity | **match** (canonical sort, 6-decimal normalization) |

Measured 2026-07-22 on hosted Ubuntu 24.04 with SWI-Prolog 10.0.2 and R 4.3.3, single-core, `kernels_on` + `emit_mode(functions)`. `total_ms` is process-start setup plus one median query, not the wall time of all three repetitions.

### Review follow-up
- Fixed `total_ms`: the original runner started its timer after sourcing the generated program and FactSource; it now measures setup from R process start.
- Guarded large integer sums so R's signed-32-bit conversion cannot turn a valid aggregate into an `NA` condition.
- Added assertions that `total_ms >= query_ms`.
- Added the complete R effective-distance generator/runtime suite to hosted CI.
- Removed the temporary scale-300 CI measurement after recording the corrected result.

### Deliberately out of scope
- No host-side transitive-closure or effective-distance traversal in the runner
- No unrelated Go/Python harness refactor
- No CI performance threshold

### Size

Current diff: **+954 / -116** across generator, runner/runtime integration, tests, harness wiring, and documentation. The two largest additions are the 271-line generator and 220-line Mustache runner; the runner composes the generated predicates and does not duplicate traversal logic.

### Verification
- [x] Generator and contract suite
- [x] Rscript dev-scale e2e: functions and interpreter
- [x] Scale-300 271-row reference parity
- [x] Corrected hosted scale-300 measurement
- [x] Full R classic conformance
- [x] Hosted R effective-distance regression gate
- [x] `git diff --check`


s243a:
Review complete. The BENCH-R implementation is contract-correct and the large diff is mostly justified by the standalone generator, Mustache companion runner, focused tests, and documentation. The runner calls the generated `effective_distance_sum_selected/3`; it does not reimplement transitive closure or effective-distance traversal on the host side.

I found and fixed three issues on the PR branch:

1. `total_ms` started after `generated_program.R` and its FactSource were sourced, so it did not include the setup it claimed to measure. It now uses R elapsed time from process start, with structural and e2e assertions.
2. Aggregate `sum` could crash for values outside R's signed-32-bit integer range because `as.integer(total)` becomes `NA`. The conversion is now range/finite/integrality guarded.
3. The new benchmark tests were not part of hosted CI. The complete dev-scale R effective-distance generator/runtime suite is now a permanent R job step.

A temporary hosted scale-300 run passed full R conformance and reference parity (271 rows), producing corrected results: `query_ms=62595`, `total_ms=63424`, samples `65368,62595,62559`. I recorded those values and removed the temporary scale-300 step so normal CI retains correctness coverage without a ~3-minute benchmark cost.

No blocking findings remain. Final head `b370357` is mergeable and the complete UnifyWeaver Tests workflow passed, including the permanent R effective-distance suite and full R classic conformance.

PR #3940, opened 2026-07-23
plawk: bare scalar-variable patterns (NAME OP int)

s243a:
## Summary

**Stack 3 of 4** — based on #3939 (float division). First prerequisite in decomposing **field-vs-scalar** (`$1 > threshold`) into landable pieces.

Adds `NAME OP int { … }` as a bare rule pattern (`n > 2`, reversed `2 < n`), where `NAME` is a user scalar compared to an integer literal. This establishes the **scalar-read-in-guard plumbing** that the next PR (`$1 > n`) builds on.

## The plumbing (the crux)

The scalar's current value is already in scope at the guard block (rule 0: the loop-head phi `%slot_M`; rule N: the per-rule input phi `%rule_N_in_slot_M`, emitted just above the guard) — exactly as the working scalar-`if` path reads it. Rather than thread slot state through the entire `plawk_pattern_guard_ir` API (a wide signature change the combinator recursion would have to forward), a **pattern-AST rewrite at the rule-chain call site** (`plawk_resolve_scalar_cmp/4`) resolves `scalar_cmp(Name,Op,V)` → `scalar_cmp_resolved(SSARef,Op,V)` using the `StatePlan` + rule index. The new guard clause then just emits an i64 `icmp`. The rewrite recurses through `!`/`&&`/`||` (and `range`), so scalar patterns compose with the combinators **for free** — zero changes to the combinator code.

## Scope (v1)

The scalar must be an i64 (counter) slot — i.e. assigned somewhere so it has a slot. A never-assigned scalar (no slot) or a double/string scalar is **cleanly declined at compile time** (`scalar_counter/1` fails to unify → the driver backtracks to the standard unsupported diagnostic) rather than mis-lowered. (awk treats an unset scalar as 0; auto-creating a zero-valued read-only slot, and double/string scalar compares, are follow-ons.)

## Changes
- **parser** (`plawk_parser.pl`): `scalar_i64_cmp_pattern` → `scalar_cmp(Name, Op, Value)`, both operand orders; `scalar_cmp_reserved_name/1` keeps `NR`/`NF`/`FNR`/specials and reserved words on their own paths. Registered after `special_i64_cmp_pattern`.
- **codegen** (`plawk_native_codegen.pl`): `plawk_resolve_scalar_cmp/4` rewrite wired into the scalar rule chain; new `plawk_pattern_guard_ir/5` clause for `scalar_cmp_resolved`.
- **tests**: new `test_plawk_scalar_pattern` (12 tests) — parse (incl. specials/calls not captured), runtime (counter, reversed, window, negation, field combine, scalar→END). Verified against system `awk`.
- **audit**: row 25 updated.

## Verification
`plawk_scalar_pattern` — **All 12 tests passed**. No regression: `plawk_scalar_if` (8) and `plawk_surface_pattern_combinators` (19) green. Confirmed clean-decline (exit 3, no binary) for a never-assigned scalar and a string scalar.

## Stack
Base chain: #3938 → #3939 → **this**. Next on top: **field-vs-scalar pattern** (`$1 > n`), reusing this resolution machinery.

---
_Generated by [Claude Code](https://claude.ai/code/session_01HJci66LfkyrStXJPxmMksn)_


s243a:
Superseded by #3942. The stacked bases tangled on merge (this PR targeted the `floatdiv` branch, which never reached the feature line). #3942 carries this scalar-pattern work — plus the float-division changes from #3939 that also didn't land — as a clean 3-commit diff directly on top of the feature line. Closing in favor of #3942.

---

PR #3943, opened 2026-07-23
PERF-R-CA-DIRECT: capability-gated arg1 lookup for R recursive kernels

s243a:
## Summary

Adds a capability-gated bound-arg1 lookup for the R `category_ancestor/4` kernel. Indexed FactSources now avoid repeated `iterate_goal`/fact dispatch inside DFS while unsupported sources retain the existing semantic fallback.

## Existing-functionality audit

- R previously had indexed fact dispatch, but recursive kernels could not access its arg1 index directly.
- Rust, F#, and Haskell already pass an explicit parent lookup into their native kernels.
- This implementation follows that model through `shared_program$arg1_lookups`; kernels do not guess generated variable names.

## Confirmed bottleneck

Baseline instrumentation found:

| Metric | Value |
|---|---:|
| `collect_parents` / `iterate_goal` calls | 516,522 |
| time in `collect_parents` | 49.6 s of 57.7 s |
| share of query | 86.1% |

After capability registration, the benchmark uses the direct path and makes zero `collect_parents` calls.

## Design

- Registry key: `"predicate/arity"`
- Capability: bound arg1 TermValue → ordered list of atom-valued arg2 TermValues
- In-memory sources reuse `build_fact_indexes`; no second index is created.
- Lazy LMDB uses on-demand `lmdb_arg1_v1_lookup`.
- Cached LMDB uses its existing bounded cache.
- Eager LMDB uses the materialized index.
- Missing capability retains `category_ancestor_collect_parents` via `iterate_goal`.
- Numeric-looking atoms and per-path visited semantics are preserved.

## Supported sources

| Source | Direct behavior |
|---|---|
| grouped TSV, CSV, legacy eager LMDB, inline facts | existing in-memory arg1 index |
| `lmdb_arg1_v1` eager | materialized index |
| `lmdb_arg1_v1` lazy | on-demand lookup; no materialization |
| `lmdb_arg1_v1` cached | cached on-demand lookup |
| unregistered source | generic WAM fallback |

## Comparable hosted benchmark

Scale 300, `kernels_on` + `functions`, Ubuntu 24.04 hosted runner, R 4.3.3:

| Metric | BENCH-R baseline | PERF-R-CA-DIRECT |
|---|---:|---:|
| query_ms samples | 65368, 62595, 62559 | 11038, 10307, 10296 |
| median query_ms | 62595 | 10307 |
| total_ms | 63424 | 11103 |
| query speedup | — | **6.07×** |
| rows | 271 | 271 |
| reference | match | match |

The updated result is recorded in `WAM_CROSS_TARGET_BENCHMARK_RESULTS.md`. The temporary scale-300 CI step was removed after measurement; no permanent performance threshold was added.

## Review follow-up

- Wired the complete 348-line direct/fallback contract suite into the permanent hosted R job. The original green workflow only ran a smaller structural assertion from the existing effective-distance test.
- Re-ran the new suite and full R classic conformance successfully in hosted CI.
- Recorded a comparable hosted benchmark rather than replacing the matrix with numbers from a different local machine.

## Size

Current diff: **+544 / -28** across eight files.

Production lookup implementation remains **+174 / -20**. Most remaining additions are contract tests and benchmark documentation.

## Verification

- [x] direct lookup selected for `grouped_by_first_atoms`
- [x] fallback when capability is absent
- [x] direct/fallback result parity
- [x] multipath, cycle, non-empty Visited, depth-boundary, numeric-atom cases
- [x] functions and interpreter modes
- [x] lazy/eager/cached LMDB generation and existing runtime regressions
- [x] permanent hosted direct-lookup contract suite
- [x] full R classic conformance
- [x] scale-300 271-row reference parity
- [x] `git diff --check`

The effective-distance runner still only loads inputs and calls generated WAM helpers. It contains no host-side graph traversal.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -242,6 +242,19 @@ jobs:
         run: |
           swipl -q -f init.pl -s tests/test_wam_r_effective_distance_generator.pl -g "run_tests, halt" -t "halt(1)"
 
+      - name: Run R direct arg1 lookup contract suite
+        run: |
+          swipl -q -f init.pl -s tests/test_wam_r_ca_direct_lookup.pl -g "run_tests, halt" -t "halt(1)"
+
+      # Review-only measurement on the same hosted environment as the
+      # documented BENCH-R baseline. Remove after recording the result.
+      - name: Measure PERF-R-CA-DIRECT scale-300 timing
+        run: |
+          OUT="$RUNNER_TEMP/wam-r-ed-300"
+          swipl -q -s examples/benchmark/generate_wam_r_effective_distance_benchmark.pl -- data/benchmark/300/facts.pl "$OUT" kernels_on functions
+          ( cd "$OUT/R" && Rscript run_effective_distance.R "$GITHUB_WORKSPACE/data/benchmark/300" 3 ) > "$RUNNER_TEMP/wam-r-ed-300.tsv" 2> "$RUNNER_TEMP/wam-r-ed-300.metrics"
+          cat "$RUNNER_TEMP/wam-r-ed-300.metrics"
+
       - name: Run full R classic conformance
         run: |
           swipl -q -f init.pl -s tests/test_wam_cross_target_conformance.pl -g "run_tests" -t halt
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -246,15 +246,6 @@ jobs:
         run: |
           swipl -q -f init.pl -s tests/test_wam_r_ca_direct_lookup.pl -g "run_tests, halt" -t "halt(1)"
 
-      # Review-only measurement on the same hosted environment as the
-      # documented BENCH-R baseline. Remove after recording the result.
-      - name: Measure PERF-R-CA-DIRECT scale-300 timing
-        run: |
-          OUT="$RUNNER_TEMP/wam-r-ed-300"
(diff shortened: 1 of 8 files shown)


s243a:
Review complete. The capability design is sound: the R kernel receives an explicit lookup from `shared_program$arg1_lookups`, reuses existing indexes, preserves ordered/multipath results, and retains `iterate_goal` as the unsupported-source fallback. Lazy and cached LMDB remain on-demand; no runner-side graph traversal was introduced.

I found one blocking coverage/documentation gap and fixed it on the PR branch:

- The new 348-line direct/fallback runtime contract suite was not wired into hosted CI, so the original green run only exercised the smaller structural assertion in the existing effective-distance test. It now runs permanently in the R job.
- The cross-target matrix still reported the pre-optimization 62.595 s result. I remeasured scale 300 on the same hosted runner class, recorded the comparable result, and removed the temporary measurement step.

Hosted comparison:
- baseline query median: 62,595 ms
- optimized query median: 10,307 ms (samples 11,038 / 10,307 / 10,296)
- query speedup: 6.07×
- total_ms: 11,103
- rows: 271
- reference parity: match

The added contract suite and full R classic conformance pass on final head `5be631b`. No blocking implementation findings remain; the PR is mergeable. The remaining general workflow jobs are still completing.

PR #3961, opened 2026-07-23
PERF-R-CA-IDDFS: integer-ID traversal and downward hops for R CA kernel

s243a:
## Summary

Adds optional integer-ID traversal and downward hop accumulation to the R `category_ancestor/4` kernel, layered on the existing PERF-R-CA-DIRECT registry without adding another global registry.

## Existing-functionality audit

- Before this PR, R capabilities returned parent TermValues and CA used string-keyed visited state, upward hop rewrites, and repeated list growth.
- Rust already carries depth downward over integer IDs; F# accepts `int -> int list`.
- R kernels still receive explicit capabilities and never guess generated variable names.

## Capability and fallback design

Registry entries normalize to:

```r
list(
  lookup     = function(key_term) -> list(atom TermValues) | NULL,
  lookup_ids = function(atom_id)  -> integer parent IDs    | NULL
)
```

Selection remains:

1. `lookup_ids`: integer-ID DFS, integer visited vector, downward depth, buffered integer results.
2. `lookup`: existing TermValue DFS with upward rewrite.
3. No capability: existing `iterate_goal` fallback.

Lazy/cached LMDB converts on-demand lookup results to IDs without materializing the database. Indexed in-memory and eager sources reuse their existing indexes; no duplicate adjacency map is introduced. Legacy bare-function capabilities remain TermValue-only.

## Review finding and fix

The initial ID path compacted `Visited` to atom entries and used that compacted count as the depth budget. The existing TermValue and `iterate_goal` paths count every Prolog list element for depth while only atoms participate in cycle membership. A `Visited` list containing a non-atom could therefore traverse deeper only on the new fast path.

The fix preserves the full list length and stores non-atoms as a nonmatching zero sentinel in the integer visited vector.

The contract suite now also checks:

- non-atom `Visited` depth parity across all three paths
- distinguishable DFS order using unequal hop counts, rather than equal values that could not prove ordering

## Profiling

Post-#3943 profiling identified:

| Hot section | Evidence |
|---|---|
| TermValue parent capability | 516,522 calls; roughly 843k TermValues |
| string/index operations | `exists`, `as.character`, and `paste0` remained material |
| original R IDDFS profile | memory attribution fell from roughly 5.2 GB to 3.8 GB |

The ID path now traverses intern IDs, carries depth downward, grows an integer result buffer, and creates `IntTerm` values only at the WAM boundary.

## Comparable hosted benchmark

Scale 300, `kernels_on` + `functions`, Ubuntu 24.04 hosted runner, R 4.3.3:

| Metric | Post-#3943 | IDDFS |
|---|---:|---:|
| query samples | 11038, 10307, 10296 | 8260, 7509, 7521 |
| median query_ms | 10307 | 7521 |
| total_ms | 11103 | 8341 |
| incremental speedup | — | **1.37×** |
| initial BENCH-R cumulative speedup | — | **8.32×** |
| rows | 271 | 271 |
| reference | match | match |

The hosted result is recorded in the benchmark matrix. The temporary measurement step was removed; no permanent scale-300 run or performance threshold remains.

## Size

Current diff: **+273 / -72** across five files.

- Production runtime/templates: **+193 / -41**
- Extended existing contract suite: **+68 / -20**
- Benchmark documentation: **+12 / -11**

No second large test harness was added.

## Verification

- [x] ID-native, TermValue-capability, and `iterate_goal` three-way parity
- [x] distinguishable DFS emission-order parity
- [x] atom and non-atom initial `Visited`
- [x] multipath, cycles, root-in-Visited, boundary, unreachable, numeric atoms
- [x] functions and interpreter modes
- [x] indexed, eager, lazy, and cached source shapes
- [x] permanent hosted direct-lookup/IDDFS contract suite
- [x] full R classic conformance
- [x] hosted scale-300 271-row reference parity
- [x] `git diff --check`

The effective-distance runner remains unchanged and contains no graph traversal.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -246,15 +246,6 @@ jobs:
         run: |
           swipl -q -f init.pl -s tests/test_wam_r_ca_direct_lookup.pl -g "run_tests, halt" -t "halt(1)"
 
-      # Review-only measurement on the same hosted runner class as the
-      # documented post-PERF-R-CA-DIRECT baseline. Remove after recording.
-      - name: Measure PERF-R-CA-IDDFS scale-300 timing
-        run: |
-          OUT="$RUNNER_TEMP/wam-r-ed-300"
-          swipl -q -s examples/benchmark/generate_wam_r_effective_distance_benchmark.pl -- data/benchmark/300/facts.pl "$OUT" kernels_on functions
-          ( cd "$OUT/R" && Rscript run_effective_distance.R "$GITHUB_WORKSPACE/data/benchmark/300" 3 ) > "$RUNNER_TEMP/wam-r-ed-300.tsv" 2> "$RUNNER_TEMP/wam-r-ed-300.metrics"
-          cat "$RUNNER_TEMP/wam-r-ed-300.metrics"
-
       - name: Run full R classic conformance
         run: |
           swipl -q -f init.pl -s tests/test_wam_cross_target_conformance.pl -g "run_tests" -t halt
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -246,6 +246,15 @@ jobs:
         run: |
           swipl -q -f init.pl -s tests/test_wam_r_ca_direct_lookup.pl -g "run_tests, halt" -t "halt(1)"
 
+      # Review-only measurement on the same hosted runner class as the
+      # documented post-PERF-R-CA-DIRECT baseline. Remove after recording.
+      - name: Measure PERF-R-CA-IDDFS scale-300 timing
+        run: |
+          OUT="$RUNNER_TEMP/wam-r-ed-300"
+          swipl -q -s examples/benchmark/generate_wam_r_effective_distance_benchmark.pl -- data/benchmark/300/facts.pl "$OUT" kernels_on functions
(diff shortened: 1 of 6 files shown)


s243a:
Review complete. The overall integer-ID capability/fallback design is sound, but I found and fixed one semantic mismatch before merge:

- The new ID path counted only atom entries in `Visited` toward the depth bound. The existing TermValue and `iterate_goal` paths count every list element for depth while only atoms participate in cycle membership. A mixed/non-atom `Visited` list could therefore traverse farther only on the ID path.
- `runtime.R.mustache` now preserves the full `Visited` length and uses zero sentinels for non-atoms, restoring three-path parity.
- The test helper now accepts explicit TermValue lists and covers this case across ID, TermValue, and generic fallback paths.
- I also replaced the previous equal-hop DFS-order check (which could not detect reordering) with a graph whose observable hop sequence is `c(3L, 2L)`.

Hosted R validation passed with the strengthened suite, and a temporary scale-300 measurement recorded median `query_ms=7521`, `total_ms=8341` (`8260, 7509, 7521`). That is about 1.37× faster than the comparable post-DIRECT hosted baseline (`10307 ms`) and 8.32× faster cumulatively than the initial R row (`62595 ms`). The benchmark/fleet docs were updated and the temporary measurement step was removed.

Final head: `967d69262566ba113a4e1c82a2d7c9125408ed4a`. The cleanup workflow run is queued awaiting GitHub runner capacity; it contains no new runtime/test change beyond the already-passing measured head (only benchmark documentation and removal of the temporary measurement step). I see no remaining code-level blocker.

PR #3968, opened 2026-07-23
PERF-R-CA-IDCACHE: memoized numeric-key parent buckets for R CA

s243a:
## Summary

Memoize ordered parent-ID vectors by numeric atom id inside the in-memory indexed `lookup_ids` capability used by the R common-ancestor ID-DFS kernel.

Based on `main` after PR #3961.

## Profiling

The post-#3961 scale-300 path made 516,522 `lookup_ids` calls for only 2,273 distinct atom IDs (99.56% cache-hit opportunity).

Instrumented uncached lookup time was about 4.31 s:

| Component | Time |
|---|---:|
| `paste0` key construction | 1.373 s |
| index environment lookup | 1.179 s |
| parent-ID extraction | 1.503 s |

A numeric-list adjacency image for the scale-300 workload is about 151 KiB (2,165 keys, 6,008 parent integers, maximum ID 2,548), versus about 283 KiB for a named list.

## Implementation

- Closure-private lazy numeric-list memo for the common dense-ID path.
- Dense growth bounded to `max(4096, length(facts))`.
- Rare large/sparse IDs use a closure-private hashed overflow, so an arbitrary high-ID miss cannot force an enormous list allocation.
- No separate global adjacency registry.
- Lazy LMDB remains on-demand and uncached here.
- Cached LMDB retains its existing bounded cache with no second cache.
- Eager LMDB materializes indexed facts and uses this in-memory capability.
- TermValue and `iterate_goal` fallbacks are unchanged.

## Contract coverage

The existing focused suite covers:

- cache miss then hit, including cached empty results
- ordered parents, shared parents, duplicate/mixed values
- missing and non-atom-only buckets
- numeric-looking atoms
- large sparse valid and missing IDs without dense-list growth
- sticky memo semantics over immutable indexed facts
- `Visited` depth/cycle behavior and distinguishable DFS order
- parity among ID, TermValue, and generic fallback paths
- unchanged lazy/cached/eager LMDB registration

## Hosted benchmark

GitHub Actions `ubuntu-24.04`, SWI-Prolog 10.0.2, R 4.3.3, scale 300, three query repetitions:

| Stage | query_ms median | total_ms | samples |
|---|---:|---:|---|
| Post-#3961 IDDFS | 7521 | 8341 | 8260, 7509, 7521 |
| **IDCACHE** | **4812** | **5610** | **5573, 4812, 4795** |

That is a **1.56× query improvement** on the same hosted runner class. An independent same-host cloud-agent comparison (4612 → 2940 ms) measured the same 1.57× relative improvement.

Output retained exact reference parity for 271 rows. The temporary hosted measurement step was removed after the benchmark and the fleet documents now use the hosted result.

## Size

The final production change is approximately +44 net lines in `runtime.R.mustache`, with focused additions to the existing test suite.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -246,15 +246,6 @@ jobs:
         run: |
           swipl -q -f init.pl -s tests/test_wam_r_ca_direct_lookup.pl -g "run_tests, halt" -t "halt(1)"
 
-      # Review-only measurement on the same hosted runner class as the
-      # documented post-PERF-R-CA-IDDFS baseline. Remove after recording.
-      - name: Measure PERF-R-CA-IDCACHE scale-300 timing
-        run: |
-          OUT="$RUNNER_TEMP/wam-r-ed-300"
-          swipl -q -s examples/benchmark/generate_wam_r_effective_distance_benchmark.pl -- data/benchmark/300/facts.pl "$OUT" kernels_on functions
-          ( cd "$OUT/R" && Rscript run_effective_distance.R "$GITHUB_WORKSPACE/data/benchmark/300" 3 ) > "$RUNNER_TEMP/wam-r-ed-300.tsv" 2> "$RUNNER_TEMP/wam-r-ed-300.metrics"
-          cat "$RUNNER_TEMP/wam-r-ed-300.metrics"
-
       - name: Run full R classic conformance
         run: |
           swipl -q -f init.pl -s tests/test_wam_cross_target_conformance.pl -g "run_tests" -t halt
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -246,6 +246,15 @@ jobs:
         run: |
           swipl -q -f init.pl -s tests/test_wam_r_ca_direct_lookup.pl -g "run_tests, halt" -t "halt(1)"
 
+      # Review-only measurement on the same hosted runner class as the
+      # documented post-PERF-R-CA-IDDFS baseline. Remove after recording.
+      - name: Measure PERF-R-CA-IDCACHE scale-300 timing
+        run: |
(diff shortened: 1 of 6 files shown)


s243a:
Review complete. The core optimization is well targeted: the profile shows 516,522 `lookup_ids` calls over only 2,273 distinct IDs (99.56% reuse opportunity), and the closure-private memo preserves tuple/DFS order without adding a second cache to lazy or cached LMDB sources.

I made two review updates before merge:

1. **Bound sparse-ID memory growth.** The original numeric list executed `length(cache) <- aid` for every positive queried ID, including misses. A call with a large sparse ID could therefore allocate a huge list. The dense numeric fast path is now bounded to `max(4096, length(facts))`, with a closure-private hashed overflow for rare larger IDs. The focused R suite now covers both a valid parent bucket at ID 1,000,000,000 and a cached miss at `.Machine$integer.max`.
2. **Use comparable hosted benchmark data.** The PR originally replaced the fleet row with a Cursor-host number. I ran a temporary measurement in the same GitHub-hosted Ubuntu 24.04 job class as #3961, recorded the result, updated the fleet docs, and removed the temporary workflow step.

Hosted scale-300 result:

- post-#3961 IDDFS: query 7521 ms / total 8341 ms
- IDCACHE: query 4812 ms / total 5610 ms
- samples: 5573, 4812, 4795
- query improvement: **1.56×**
- 271-row reference parity retained

The independent Cursor same-host comparison (4612 → 2940 ms) also measured 1.57×, so the relative gain is well corroborated.

The strengthened direct-lookup suite, effective-distance smoke, full R classic conformance, and all sampled WAM target jobs are green on final head `f5d6f54c495c6b952911ac906c646b1a51447e1e`. I see no remaining code-level blocker.

PR #3990, opened 2026-07-25
PERF-R-CA-STACK: iterative integer-ID DFS frame stack for R CA

s243a:
## Summary

Replace the recursive R `category_ancestor` integer-ID helper with an explicit iterative DFS frame stack. This removes recursive call overhead while preserving the existing CA contract: path-local cycle detection, exact parent/DFS order, root handling, max-depth behavior, and mixed/non-atom `Visited` semantics.

## Motivation and profile

The scale-300 effective-distance workload traverses 516,522 CA nodes and 843,891 parent edges (maximum/average depth 9/7.575). Before this change, recursive traversal accounted for 26% self / 70% total time in the CA profile and grew visited vectors 59,187 times.

## Implementation

- Keep a single pre-sized visited-path vector and restore its logical length on unwind.
- Store each active DFS frame's ordered parent vector, next parent position, depth, and saved visited length.
- Remove the recursive production helper.
- Keep all helpers inline in the existing R runtime template because this is a small, cohesive runtime fragment.
- Preserve the LMDB and generic fallback paths unchanged.

Review follow-up: stack unwind now clears its preallocated parent slot with `st_parents[sp] <- list(NULL)`. Using `[[sp]] <- NULL` deletes and shifts an R list element, introducing avoidable allocation/capacity churn.

## Validation

- R direct arg1 lookup suite: pass.
- Effective-distance smoke: pass.
- Full R classic conformance: pass.
- Added exact unsorted-order and three-path parity coverage across ID, TermValue, and generic paths, including cycles, root cases, depth boundaries, and non-atom `Visited` entries.
- Rebasing/merging current `main` produced no conflicts; the PR branch now includes `main` at `c74dcf7`.

## Hosted scale-300 result

Ubuntu 24.04, three query samples:

- STACK: `4475, 3679, 3554 ms`; median query/total `3679 / 4344 ms`; 271 rows.
- Previous IDCACHE: `4812 / 5610 ms`.
- Improvement: **1.31× query** and **1.29× total**.
- Cumulative query improvement over the original R row (`62595 ms`): about **17.0×**.

The temporary measurement step was removed after recording the result in the benchmark/status documents.

## Change size

- Production runtime: +92 / -30 (net +62).
- Existing focused test harness: +132 / -9 (net +123).
- Remaining changes are benchmark/status documentation.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -246,15 +246,6 @@ jobs:
         run: |
           swipl -q -f init.pl -s tests/test_wam_r_ca_direct_lookup.pl -g "run_tests, halt" -t "halt(1)"
 
-      # Review-only measurement on the same hosted runner class as the
-      # documented PERF-R-CA-IDCACHE baseline. Remove after recording.
-      - name: Measure PERF-R-CA-STACK scale-300 timing
-        run: |
-          OUT="$RUNNER_TEMP/wam-r-ed-300"
-          swipl -q -s examples/benchmark/generate_wam_r_effective_distance_benchmark.pl -- data/benchmark/300/facts.pl "$OUT" kernels_on functions
-          ( cd "$OUT/R" && Rscript run_effective_distance.R "$GITHUB_WORKSPACE/data/benchmark/300" 3 ) > "$RUNNER_TEMP/wam-r-ed-300.tsv" 2> "$RUNNER_TEMP/wam-r-ed-300.metrics"
-          cat "$RUNNER_TEMP/wam-r-ed-300.metrics"
-
       - name: Run full R classic conformance
         run: |
           swipl -q -f init.pl -s tests/test_wam_cross_target_conformance.pl -g "run_tests" -t halt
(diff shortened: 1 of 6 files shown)


s243a:
Review complete. The iterative frame-stack design preserves the required CA semantics: path-local cycle handling, exact ordered DFS output, root behavior, depth boundaries, and mixed/non-atom `Visited` handling. The focused differential/order coverage is appropriately kept in the existing R CA harness.

I made one small runtime correction before merge: `st_parents[[sp]] <- NULL` deletes and shifts an R list element. The unwind path now uses `st_parents[sp] <- list(NULL)`, which clears the preallocated slot without defeating the stack's capacity strategy.

I also incorporated current `main`, removed the temporary benchmark workflow step, and recorded the hosted result in the status docs. Final hosted R validation on `2dee37cb` is green: focused R regressions, ISO smoke, effective-distance smoke, direct arg1 contract suite, and full R classic conformance all pass.

Hosted scale-300 result: 271 rows; query samples `4475, 3679, 3554 ms`; median query/total `3679/4344 ms`. That is 1.31× faster in query time and 1.29× faster overall than IDCACHE (`4812/5610 ms`), and about 17× faster than the original R query row.

The PR is mergeable against current `main`. I found no remaining code-level blocker; the rest of the repository workflow is still completing.

PR #3995, opened 2026-07-25
docs(mu_cosine): harden process-expression and filing-decoder handoffs

s243a:
## Summary

Turn the deferred process-expression encoder sketch into an implementable handoff:

keep canonical AST identity outside the lossy latent;

define resolved typed-AST tokens and complete role paths;

separate exact numeric reconstruction from smooth numeric conditioning;

specify view-conditioned alignment and a bottleneck-only grammar decoder;

require sealed structural splits and an identity-safe NameFunctionCond adapter.

## Correct the tree-position design:

give exact operator equations and parameter counts;

clarify the limits of convolution/HRR unbinding;

distinguish complete paths from lossy depth/local-role features;

restrict architecture screening to inner data before freezing one held-test arm.

Add a filing path-decision handoff that extends—not duplicates—the existing routed-task.v2 authority:

preserve catalog-recorded breadcrumbs for existing folders;

define SELECT_EXISTING, PROPOSE_NEW, and ABSTAIN;

separate acceptable-set simulated absence from genuine prospective novelty;

require content-bound privacy and external-authorization receipts;

retain deterministic best_so_far rollback and prohibit graph mutation.

Mark the older lineage-decoder sketch as historical and cross-link the implementation ladder.

## Scientific boundaries

Latents and position vectors are conditioning features, never identity or cache keys.

The evidence that activates encoder development is already exposed; a fresh post-activation population is required for an untouched outer comparison.

Artificially withholding acceptable folders does not establish that no suitable folder naturally exists.

Simulated absence may support only a labeled, owner-reviewed research pilot; ordinary live proposals require prospective naturally absent examples.

Private external naming requires an exact content-bound authorization receipt. Unknown privacy remains local-only.

This PR changes documentation only. It performs no training, scoring, provider calls, or graph mutation.

## Validation

45 passed:

test_process_cards.py

test_process_expression_p1_protocol.py

Relative Markdown links verified.

Staged and rebased diffs pass whitespace checks.

## Review focus

Full process-identity and internal residual-lookup contract.

Exact typed role-path schema and corrected position-operator mathematics.

Additive integration with routed-task.v2.

Boundary between simulated catalog absence and prospective live novelty.

Privacy derivation and external-naming authorization receipts.

### DIFF
--- a/prototypes/mu_cosine/DESIGN_expression_encoder_future.md
+++ b/prototypes/mu_cosine/DESIGN_expression_encoder_future.md
@@ -1,75 +1,445 @@
-# Expression Encoder–Decoder (FUTURE WORK — do not schedule before the P1 gate)
+# Process-Expression Encoder and Reconstruction Decoder — Deferred Implementation Handoff
 
-**Status: future-work design capture, 2026-07-24.** Not part of the current P0–P4 ladder and not
-to interrupt the frozen P1 protocol (#3982's rigor contract). Sits at "P3.5": pursue only if
-P1 passes (expression conditioning earns its keep) AND P3's deterministic composition plateaus.
-Self-contained enough to hand to a separate agent with spare inference; the handoff needs only
-this doc, the companion `DESIGN_tree_position_encoding_theory.md`, and the current-art references
-below.
+**Status: future work, 2026-07-24; amended as an implementation handoff.** This remains outside
+the active P0–P4 ladder. Do not schedule it unless both activation gates hold:
 
-## Current art (what exists today, and its limits)
+1. the frozen P1 primary shows that structured expression conditioning earns its keep over flat
+   process tokens; and
+2. P3's deterministic AST composition has been implemented and has plateaued under its frozen
+   multi-process and cross-corpus LOCO evaluation.
 
-- `DESIGN_process_expression_{language,philosophy,implementation}.md` + `process_cards.py` (P0,
-  PR #3980): the process-expression grammar with registry-driven lexing. **Today the grammar is
-  used ONLY for canonicalization and token identification** — lossless identity (canonical AST +
-  ast_sha) and deterministic card rendering (V0–V3). "Self-parsing" in #3980 means only that the
-  parser must parse the spec's own examples (acceptance test), not a formal novelty.
-- Embedding today: rendered card string → frozen e5 → NameFunctionCond residual. The
-  implementation doc flags e5-on-formal-strings as a leap of faith at V2/V3; P2 adds an
-  NL-template arm to test it.
-- P3 (planned): DETERMINISTIC composition — node embedding = card-e5(operator) + Σ fixed
-  position-weighted child embeddings; stochastic elements seeded by SHA-256 of the canonical AST.
-  No learning anywhere in the composition.
-- `ARCHITECTURE_filing_engine.md` rows this feeds: FUSE-003 (name-conditioned identity — the
-  motivation), OPENQ-012 (cross-corpus transfer of process conditioning — the forest question).
+If either gate fails, archive this design. A learned encoder is not a way to rescue a failed
+conditioning premise. The P1/P3 results used to activate the design are thereby exposed design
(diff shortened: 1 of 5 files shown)


s243a:
## Review

Docs-only, +1384/−145 across 5 files. This is real hardening rather than a restatement — the corrections to the position-encoding math and the identity/latent separation both fix things the earlier sketches got wrong. I checked the load-bearing claims against the landed code rather than reading them on their own terms.

### Claims verified against code

- **Identity formula matches.** `sha256(REGISTRY_VERSION + "|" + canonical_identity_string)` is exactly what `process_expression_p1_protocol._full_process_digest` computes.
- **The `ast_sha` criticism is accurate.** `process_cards.py:458` really does truncate to `[:16]` and does not bind the factory fingerprint, so requiring the full 64-hex digest for new artifacts is justified rather than defensive.
- **Circular-convolution correction is right.** `P_(t,(i,j)) = 1 iff i+j ≡ t (mod D)` applied to the outer feature `x_(i,j) = d_i b_j` does reproduce `Σ_i d_i b_((t-i) mod D)`. Worth stating explicitly that the formula as given is unnormalized, which the doc does.
- **Parameter counts all check out** — factorized `R(D_d+D_b+D_o)+D_o`, dense `D_o·D_d·D_b+D_o`, scalar `D_d·D_b+1`.
- **The HRR section is the most important fix.** Replacing "approximately invertible" with the known-key-and-conditioned-codebook statement is correct: given only `z = d ⊛ b`, the pair is not recoverable, and the doc now says so.
- **Cited architecture rows exist** — RANK-001, FUSE-004, CORP-002, DR-001/DR-002 all support the §0 framing.
- All 10 relative Markdown links resolve.

### One finding

**`DESIGN_filing_path_decoder_handoff.md` §0, bullet 3** lists "an existing folder is identified by a typed stable ID, never by its title" among the standing decisions in `ARCHITECTURE_filing_engine.md`, but there is no atomic row asserting it. The other three bullets map cleanly onto RANK-001/DR-006, FUSE-004, and CORP-002/DR-001.

The rule is real *practice* — `routed_queries.py:287` sets `primary_grading: exact_destination_id` with `best_title_equivalent_destination` only as a sensitivity arm — it just isn't a row. Since the architecture doc's own premise is that every element is one atomic row with a stable ID, citing an unrowed rule slightly overstates the source. Either add an `IDENT-00x` row or soften the citation to point at the grading contract instead.

### Note on sequencing and data

One practical point in favour of the ordering this PR sets up: the encoder/decoder work in `DESIGN_expression_encoder_future.md` needs **no private corpus at all**, which the doc asserts in §10 and which holds up in practice. The generator in §6 enumerates expressions from the checked-in registry (18 entries at `REGISTRY_VERSION v0.3`), and the §4.2 teacher targets are computed locally by applying the pinned e5 to rendered cards.

Confirmed end to end on a clean machine — parse → canonical identity → V1/V2/V3 cards → `intfloat/e5-small-v2` at the pinned revision `ffb93f3` → `(3, 384)` normalized float32 teacher vectors, with nothing private touched. So when the P1/P3 activation gates do open, the pretraining and reconstruction work is unblocked by data availability; what genuinely needs real rows is the §8 downstream study, and separately the tree-position envelope scan item 2 ("every process in the frozen downstream ledger") — and that ledger currently has no real labels, as `p1_ledger_builder.py` notes.

That is an argument for the gates as written, not against them: the encoder cannot be scheduled on convenience grounds when the evidence that would activate it is already exposed.

### Related

I implemented the Stage A slice that §11 authorizes — the `filing-path-request.v1` / `filing-path-decision.v1` schemas, parent re-derivation with an exact QID join, typed-ID resolution, verbatim principal-path breadcrumbs, `SELECT_EXISTING`/`ABSTAIN` only with no reachable `PROPOSE_NEW` path, and 53 acceptance tests — on `claude/mu-cosine-hardening-pr-zlrsuu`, based on this branch. Nothing in it depends on the encoder ladder.

---

PR #4004, opened 2026-07-25
mu_cosine: closeness objective for the process-expression decoder

s243a:
## Summary

Revises §5 of the step-2 generator specification after review of #4003. Exactness on decoder output is relaxed to closeness, the relaxation is expressed in the **objective** rather than only in prose, and the two review nits are fixed.

Encoder-only, as before. 179 passed (4 new tests).

## Why relaxing exactness is safe

Because identity already lives outside the bottleneck. `process_identity.py` retains canonical bytes and the full digest alongside, and no pipeline recomputes a digest, cache key, or residual lookup from decoded text — so a decoder rendering `0.84` for `0.85` breaks nothing identity-shaped. Exactness would only be load-bearing if decoded text flowed back into a digest, which the architecture forbids.

The spec now states the three layers separately, because conflating them is how an unnecessary exactness requirement gets imposed on the model:

```text
tokenizer   exact by construction   (reversibility is a serialization property)
decoder     close by objective      (a learned approximation, graded by tolerance)
identity    independent of both     (canonical bytes retained alongside)
```

The load-bearing rule is now explicit: **decoder output is candidate generation and conditioning, never identity.** The tokenizer's §9 round-trip guarantee is untouched — "close is good" must not leak downward into it.

## Stating tolerance is not enough

Plain byte-token cross-entropy *is* an exactness objective: it penalizes `0.84` for `0.85` exactly as hard as `0.31` for `0.85`. A run that declares "close is good" and then trains on unmodified token CE quietly optimizes spelling anyway. So §5.3 specifies **stochastic exact verification**:

```text
L_numeric = p * verifier_reward + (1 - p) * tolerance_loss
p = 0.75    # provisional; ablated over {0, 0.5, 0.75, 1}
```

Two properties make this the right shape rather than a compromise:

- **The verifier is free.** Decode, re-parse, canonicalize, compare bytes. The tokenizer's exact reversibility supplies a deterministic oracle at zero cost — no teacher model, no learned similarity, nothing to calibrate. Most "close is good" regimes stall on defining similarity; this one can check correctness whenever it likes.
- **It behaves like masking.** The decoder cannot tell which rows will be strictly graded, so its optimal policy is exactness wherever exactness is affordable. The equilibrium is the profile actually wanted: exact in the production region where values are well trained and cheap to spell, gracefully close in the tail where exactness costs more capacity than it is worth. `p` is the dial between exact-always and close-always.

Three implementation requirements are recorded, each guarding a specific failure:

1. **Verify per field, not per expression.** Whole-expression canonical equality is all-or-nothing — one wrong digit in a 120-token row fails the entire verification and the training signal goes sparse.
2. **The unverified fraction still needs a loss.** Falling back to plain byte CE on the `(1 − p)` rows silently reimposes exactness and undoes the mechanism.
3. **Watch for collapse onto common values.** A large exact-hit reward can be maximized by emitting the most frequent canonical value forever — high verified-exact rate, dead numerics. Reporting verified-exact rate *alongside* latent numeric sensitivity makes that visible instead of mysterious.

The same verifier is recorded as an inference-time reranker: sample `k` decodings, prefer the ones that verify. No training change, and consumers needing only closeness ignore it.

## Also revised

- **Per-field tolerance, integers exempt.** `menus=10` vs `11` and `depth=3` vs `4` are behaviorally different processes, not near misses. `number` fields grade within tolerance; `int` fields grade exact after rounding; strings exact. Tolerances are spec numbers reproduced by a test.
- **The digit floor is now a *resolution* floor, not a correctness floor.** A decoder can render `0.47` as `0.45` from trained glyphs, so an unseen digit is no longer catastrophic. But in the production threshold range the forced error (`0.07` → `0.06`, i.e. 0.01) exceeds the routing tolerance of 0.005 — so the widened grid stays, with its justification corrected. There is a test asserting exactly that inequality.
- **Arm evaluation** switches from exact-match string comparison to tolerance-based reconstruction (structure exact, numbers within tolerance, ints exact after rounding), with verified-exact rate reported separately. The preregistered prediction is revised accordingly: under a closeness metric N2's "clean exactness story" stops being an advantage and its scalar head becomes the natural fit, while N0's weakness moves from tail robustness to plain resolution.

## Review nits from #4003

- The tautological assertion `len(s.modifiers) >= 0` is replaced by one showing the cap actually bites: `luna` and `llm` each declare two modifiers, so the one-per-node restriction is real rather than vacuously satisfied.
- The new spec numbers (`p = 0.75`, its ablation grid, the tolerance table) land as tests, like every other number in this track.

## Still open

- **The witnessed / ring / floor mass split remains unset.** Noted in review that it must be frozen into `generator_spec_sha256` before the generator is, so "unset" cannot drift into "decided in code". Agreed — but the numbers should come from the P1 run list rather than from me, so it stays open here rather than being invented.
- The `validate_decision` forward-compat test from #4001 — filing track, separate PR.

https://claude.ai/code/session_017xCXhV5sHDsMiDyU5q7Hqg

---
_Generated by [Claude Code](https://claude.ai/code/session_017xCXhV5sHDsMiDyU5q7Hqg)_


prototypes/mu_cosine/DESIGN_process_expression_generator.md
```diff
@@ -222,17 +262,73 @@ weakness.
...
+declares "close is good" and then trains on unmodified token CE quietly optimizes spelling anyway.
+Closeness has to be expressed in the loss or it does not exist.
+
+#### Stochastic exact verification
+
+The mechanism: on a sampled fraction `p` of training rows, grade the decoded output with an
+**exact verifier**; on the rest, grade it with a tolerance loss.
+
+```text
+L_numeric = p * verifier_reward + (1 - p) * tolerance_loss
```
s243a: **[P1] Make the verifier term trainable.** Decode, parse, and exact-byte comparison happen after discrete token selection, so ordinary backpropagation provides no gradient; at `p = 1` there is no differentiable term. `+ verifier_reward` is also sign-ambiguous in a minimized loss. Specify a differentiable exact surrogate or a complete policy-gradient/expected-risk estimator.

prototypes/mu_cosine/DESIGN_process_expression_generator.md
```diff
@@ -192,12 +192,52 @@ near slice, it is dropped.
...
   features plus FiLM earn their place.
 
+### 5.0 Exactness is layered, and only one layer is mandatory
+
+The three layers have different obligations, and conflating them is how an unnecessary exactness
+requirement gets imposed on the model:
+
+```text
+tokenizer   exact by construction   (reversibility is a serialization property)
+decoder     close by objective      (a learned approximation, graded by tolerance)
```
s243a, line 231: **[P1] Reconcile the inherited decoder contract.** This changes reconstruction from exact to tolerance-based, but `DESIGN_expression_encoder_future.md` still mandates exact canonical bytes/full digest, grammar-masked token CE, and exact-AST reporting. Amend that authoritative handoff and its gates in the same change, explicitly define precedence, or retain exact reconstruction.

prototypes/mu_cosine/DESIGN_process_expression_generator.md
```diff
@@ -192,12 +192,52 @@ near slice, it is dropped.
...
+|---|---|
+| `number`, `number_list` | within tolerance |
+| `int`, `int_list` | exact after rounding — no tolerance |
+| `string` | exact |
+
+Tolerances are spec numbers reproduced by a test, revised when the production distribution moves —
+the same rule as the caps. Provisional starting values, absolute unless stated:
+
+```text
+tolerance.routing.t      = 0.005     # thresholds live in 0.01-0.1; must not blur a margin band
```
s243a: **[P1] Prevent routing-tier collapse.** The registered thresholds `0.02` and `0.03` are only `0.01` apart. Independent tolerance of `0.005` lets their decodes converge arbitrarily close to `0.025`, effectively erasing a routing tier while passing reconstruction. Require a safety margin below half the minimum operational gap plus ordering/minimum-separation validation, or grade route-decision equivalence.

prototypes/mu_cosine/test_process_expression_envelope.py
```diff
@@ -287,3 +290,87 @@ def test_specification_records_the_measured_numbers():
...
+    assert ("menu", "n") in int_fields
+    assert ("lineage", "depth") in int_fields
+    number_fields = {
+        (name, key)
+        for name, sig in REGISTRY.items()
+        for key, spec in sig.kwargs.items()
+        if spec.kind in ("number", "number_list")
+    }
+    # The tolerance table covers exactly the number-valued fields.
+    assert number_fields == {
```
s243a, line 440: **[P1] Preserve declared numeric types in the stream.** This assumes every registry `number` field receives tolerance grading, but the frozen resolver derives scalar type from the Python value: valid `margin(t=1)` emits `<INT>` despite `margin.t` being registered as `number`. A stream-driven loss would exact-grade it. Preserve the declared registry kind and add an integer-spelled `number` fixture.

prototypes/mu_cosine/test_process_expression_envelope.py
```diff
@@ -287,3 +290,87 @@ def test_specification_records_the_measured_numbers():
...
+    """
+
+    nearest_trained = 0.06  # 0.07 unreachable: digit 7 is absent
+    forced_error = abs(0.07 - nearest_trained)
+    routing_tolerance = 0.005
+    assert forced_error > routing_tolerance
+    assert ABSENT_DIGITS == {"4", "6", "7", "9"}
+
+
+def test_decoder_output_is_never_an_identity_source():
```
s243a: **[P1] This does not test the decoder/identity boundary.** The function never passes decoder output through identity, cache, or residual lookup; current APIs can digest expression text and mint identity from any parsed node. It only proves different ASTs hash differently, so the premise supporting relaxed exactness remains unenforced. Add an origin-bound interface/integration guard or make decoder output explicitly diagnostic-only.

prototypes/mu_cosine/DESIGN_process_expression_generator.md
```diff
@@ -222,17 +262,73 @@ weakness.
...
+Three requirements on the implementation:
+
+1. **Verify per field, not per expression.** Whole-expression canonical equality is all-or-nothing:
+   one wrong digit in a 120-token row fails the whole verification and the signal goes sparse.
+   Parse the decoded candidate and compare field by field, which also matches the per-field
+   tolerance structure of §5.0.1. Integer fields verify exactly, always.
+2. **The unverified fraction still needs a loss.** On the `(1 - p)` rows, falling back to plain
+   byte CE silently reimposes exactness and undoes the whole mechanism. The tolerance loss is the
+   N2 scalar head or a digit-distance-weighted cross-entropy — never unmodified CE.
+3. **Watch for collapse onto common values.** A large exact-hit reward can be maximized by always
```
s243a: **[P2] Measure decoder collapse directly.** Latent sensitivity does not prove the decoder uses that coordinate. An encoder can separate `0.85` from `0.5` while a decoder always emits the majority value, yielding both high sensitivity and high aggregate exact rate. Add macro/per-value reconstruction, field-level confusion, output variance, or another decoded-target dependence metric.

prototypes/mu_cosine/DESIGN_process_expression_generator.md
```diff
@@ -222,17 +262,73 @@ weakness.
...
+   emitting the most frequent canonical value — `0.85` forever — yielding a high verified-exact
+   rate with dead numerics. The diagnostic is already in the metric list: report verified-exact
+   rate *alongside* latent numeric sensitivity. High exact rate with flat sensitivity is collapse,
+   and seeing both numbers makes it visible instead of mysterious.
+
+#### Inference-time reuse
+
+The same verifier reranks without any training change: sample `k` decodings and prefer the ones
```
s243a: **[P2] Define the inference-time oracle.** For reconstruction, callers already retain the reference canonical bytes and can use them directly. For novel generation, parsing and canonicalizing establish validity rather than exactness because no target digest exists. Define the reference, `k`, and no-match fallback, or remove this reranking claim.

prototypes/mu_cosine/test_process_expression_envelope.py
```diff
@@ -287,3 +290,87 @@ def test_specification_records_the_measured_numbers():
...
+
+
+# --------------------------------------------------------------------------
+# §5.0 the closeness objective — its numbers are spec, so they are tests too
+# --------------------------------------------------------------------------
+
+
+def test_specification_records_the_closeness_numbers():
+    text = SPEC.read_text(encoding="utf-8")
+    for needle in (
```
s243a, line 302: **[P2] Pin the complete tolerance table.** This drift test checks `routing.t` and `lineage.decay`, but not `margin.t` or `blend.w`; those values can change or disappear while every test passes. Parse and compare the complete tolerance mapping against all registry `number` and `number_list` fields.

s243a (commented):
**Recommendation: request changes.** GitHub does not allow the PR author to submit a formal `REQUEST_CHANGES` review on their own PR, so this is posted as a comment review instead.

The focused process-expression suites and GitHub Actions pass, but the proposed objective is not trainable as written and the governing specifications disagree. The inline comments identify five blocking issues: the nondifferentiable verifier term, the inherited exact-reconstruction contract, routing-threshold collapse at the stated tolerance, registry/stream numeric-type disagreement, and the unenforced decoder/identity boundary. Three additional comments cover the collapse diagnostic, inference reranker, and incomplete tolerance drift test.

Validation at `ca0da4057e671d4aa8035b62b1a225f799ab529c`: 64 focused tests passed, `git diff --check` passed, Python compilation passed, and GitHub Actions was green.

s243a:
Took seven of the eight findings; pushed as `984a4d9`. **193 passed.** Two were real defects, and finding 3 is the sharpest catch in the review.

### 1 — the objective had no gradient ✅

Correct, and the fix improves the design rather than patching it. `p * verifier_reward + (1-p) * tolerance_loss` evaluated the exact branch only *after* discrete decoding, parsing and canonicalization, so nothing backpropagated; at `p = 1` there was no differentiable term at all.

Reframed as grading-mode **selection**: each row is assigned by a seeded draw to exact token cross-entropy or to the tolerance loss — both ordinary teacher-forced losses over known targets. No policy gradient, no sign ambiguity, no variance control needed, because at training time the target is known and "grade this row exactly" is just exact CE.

This is also the better reading of the masking analogy that motivated the idea: masking selects which positions are graded, it does not add a reward. Exact verification survives as a free *metric*.

### 2 — precedence over the parent handoff ✅

Correct and worth stating plainly. New §5.0.1: the handoff's exact-AST gate stays **primary and blocking**, tolerance grading is an additional reported metric and a registered arm, and promotion requires amending §4.3/§7/§8 in one change. Noting one thing that makes the interim coherent: at `p = 1` the arm reduces to exactly the handoff's own loss, so only `p < 1` needs the amendment.

### 3 — the tolerance could erase a routing tier ✅

The best finding here. Registered thresholds are `0.01` apart, so independent absolute tolerance of `0.005` admits `[0.025, 0.025]` — both elements "within tolerance" and the two-tier policy destroyed while the primary metric reports success.

Verified the enabling condition rather than assuming it: the parser accepts both `t=[0.025,0.025]` and `t=[0.03,0.02]` today, so no ordering or separation is enforced. Threshold tolerance drops to `0.002` (strictly below half the gap), and ordered fields now carry a structural ordering/separation predicate that must hold *in addition* to field-wise tolerance. There's a test asserting `tolerance < gap/2` and demonstrating both degenerate parses.

### 4 — the stream discarded the declared registry kind ✅

Confirmed against the code before acting: `margin.t` is registered `number`, but `margin(t=1)` resolved as `value_type="int"` and emitted `<INT>`, with role path `KWARG(t,int)`. That violates the handoff's rule that derived tokens agree with the pinned registry, and a stream-driven grader would have exact-graded a tolerance field.

Declared kind now wins; runtime inference remains only as a fallback. **No pec-v1 golden row was affected** — which is precisely why pec-v2 adds `int-spelled-number` and `int-spelled-number-list` rows. Contract bumped to pec-v2 with a new golden bundle; pec-v1 is retained on disk and now fails closed, per the rule that a contract change creates a new bundle rather than mutating a sealed one. A test asserts that.

### 5 — identity separation ⚠️ taken, but not as specified

The overclaim was real and is removed. Nothing mechanically stops a decoded candidate entering `full_ast_digest_for_expression` or `deployed_identity` — those APIs cannot tell where a parsed node came from. The spec now says the separation is a **discipline, not an invariant**, records an origin-tagged wrapper as a **step-3 obligation blocking on the decoder** with its test named, and keeps decoded output diagnostic-only until then. The existing test is renamed to what it actually proves, and a companion test asserts the gap so it cannot be quietly forgotten.

I did not build the origin-bound interface now. There is no decoder to tag, so the wrapper would be speculative API design against an unwritten consumer — the same over-building the track has avoided elsewhere. Recording the obligation with a named test seemed the honest middle. Happy to be overruled if you think the interface should land ahead of its consumer.

### 6 — collapse diagnostic ✅

Correct: latent sensitivity is an *encoder* property, so a decoder can ignore the coordinate and emit the majority value while both aggregates look healthy. Now requires target-conditioned decoder-level metrics — macro-averaged reconstruction over distinct targets, field-level confusion, decoded-output variance.

### 7 — inference reranker ✅

Correct, and my original claim was muddled. For reconstruction the caller already holds the reference bytes, so reranking is redundant; for novel generation no reference exists, so parsing establishes **validity and canonical stability**, not exactness. Claim narrowed, with `k` and the fallback pushed to whatever consumer wants it rather than asserted here.

### 8 — incomplete tolerance table ✅

Correct. The test now parses the whole table and asserts its keys equal exactly the registry's `number`/`number_list` fields, so `margin.t` or `blend.w` cannot vanish silently.

---

Still open and unchanged: the witnessed/ring/floor mass split, which needs to come from the P1 run list rather than be invented, and must be frozen into `generator_spec_sha256` before the generator is.

---

prototypes/mu_cosine/process_expression_contract.py
```diff
@@ -49,7 +49,12 @@
 
 #: Bumped whenever the structural stream or role-path serialization changes.
 #: Versioned independently of the grammar, per §3.1.
-CONTRACT_VERSION = "pec-v1"
+#:
+#: pec-v2: scalar literals carry their *declared registry kind* rather than a
+#: type inferred from the Python runtime value.  ``margin(t=1)`` now emits
+#: ``<NUMBER>`` instead of ``<INT>``.  No pec-v1 fixture row was affected — that
+#: is precisely why pec-v2 adds a row covering an integer-spelled ``number``.
+CONTRACT_VERSION = "pec-v2"
```
s243a, line 58: **[P1] Move the authoritative current-bundle pointers to v2.** This makes `pec-v2` the only contract accepted by the current loader, but `DESIGN_expression_encoder_future.md` §11 still directs step 2 to reproduce `PROCESS_EXPRESSION_GOLDEN_v1.json`, and the track README advertises only v1. Following the authoritative parent now either fails closed or misses the integer-spelled-number coverage added here. Keep v1 as audit-only provenance, but update the parent and README atomically to name v2 as current.

prototypes/mu_cosine/test_process_expression_contract.py
```diff
@@ -474,6 +477,40 @@ def test_golden_detects_a_row_recomputed_differently():
         verify_golden(document)
 
 
+def test_superseded_v1_bundle_is_retained_and_fails_closed():
+    """A contract change creates a new bundle; it never mutates a sealed one.
+
+    pec-v1 stays on disk as the sealed artifact it was.  Because scalar literals
```
s243a, line 483: **[P2] Enforce the sealed-bundle lifecycle rather than only describing it.** `load_golden(v1)` rejects on contract version before checking its checksum, after which this test checks only the version string and missing row names; a token-corrupted v1 therefore passes. The generator also defaults to 9 rows while committed v2 has 20, permits `--extra` to replace registered names, and truncates an existing `--out`. Pin/verify the historical v1 integrity, put the required v2 cases in a canonical source manifest, reject overrides, and refuse overwriting a sealed target.

prototypes/mu_cosine/DESIGN_process_expression_generator.md
```diff
@@ -222,17 +321,95 @@ weakness.
...
+### 5.3 The objective is where closeness lives or dies
+
+Stating tolerance in the spec is not enough. Plain byte-token cross-entropy is an *exactness*
+objective — it penalizes `0.84` for `0.85` exactly as hard as `0.31` for `0.85` — so a run that
+declares "close is good" and then trains on unmodified token CE quietly optimizes spelling anyway.
+Closeness has to be expressed in the loss or it does not exist.
+
+#### Stochastic exact grading
+
+The mechanism is **grading-mode selection**, not a reward term: each training row is assigned, by a
```
s243a: **[P2] Choose one stochastic-grading unit and schedule.** This paragraph and formula assign one mode per row, while requirement 1 below assigns a draw per numeric field. Those produce different estimators and correlations. Specify the per-field draw explicitly and say whether it is fixed forever or resampled per step/epoch; a permanent row-level assignment also weakens the later claim that the decoder cannot infer which examples are exact-graded.

prototypes/mu_cosine/test_process_expression_envelope.py
```diff
@@ -287,3 +290,179 @@ def test_specification_records_the_measured_numbers():
...
+    """§5.2: the digit floor is about resolution, not catastrophic failure.
+
+    A decoder can render 0.47 as 0.45 using only trained glyphs, so an unseen
+    digit is not fatal.  But in the production threshold range the forced error
+    exceeds the routing tolerance, which is why the grid still widens.
+    """
+
+    nearest_trained = 0.06  # 0.07 unreachable: digit 7 is absent
+    forced_error = abs(0.07 - nearest_trained)
+    routing_tolerance = 0.005
```
s243a: **[P3] Use the current routing tolerance here.** The specification and the full-table test now pin `routing.t = 0.002`, but this scenario still hard-codes the superseded `0.005`. The inequality happens to pass under either value, so this test no longer demonstrates what its name/docstring claim. Read `_spec_tolerances()["routing.t"]` instead.

s243a (commented):
**Re-review at `984a4d91bc017c0ce3cf1a74504eda017a7c65e0`: changes are still needed.** GitHub does not allow the PR author to submit a formal `REQUEST_CHANGES` review on their own PR, so this is posted as a comment review.

All eight original findings are resolved for this PR's current scope. In particular, finding 5 is accepted: the original requested remedy explicitly allowed decoded output to be made diagnostic-only, and the revision now does that while honestly recording the missing origin guard as work that blocks the decoder. Building a speculative wrapper before a decoder exists is not required here. Before step 3/adapter integration lands, however, the authoritative handoff must carry that gate and the implementation must prevent provenance laundering through `wrapper.text -> parse -> deployed_identity`.

The new blocker is the `pec-v2` authority split described inline: the current loader/tests reject v1 while the authoritative handoff and track README still direct step 2 to v1. The v1 artifact should remain for audit, but v2 must be named as current everywhere.

Two cleanup points in addition to the inline comments:

- The precedence fix is acceptable because this document is explicitly non-authorizing and says `p < 1` requires a later parent amendment. Still, `p = 1` is the *compatible endpoint*, not the “default endpoint” (the provisional default below is `p = 0.75`).
- The PR description is stale: it still presents the nondifferentiable `verifier_reward`, tolerance `0.005`, the old identity-invariant claim, and exactness reranking. Please refresh it to match the revised head.

Validation: 70 focused process-expression tests passed; `git diff --check` and Python compilation passed; the worktree is clean; GitHub Actions run 30177621328 is green.

s243a:
All four re-review findings addressed in `2e0c6e2`. **199 passed.**

### [P1] Bundle pointers moved to v2 — the blocker

Correct and now fixed atomically. `DESIGN_expression_encoder_future.md` §11 step 1 and the README two-track table both name `PROCESS_EXPRESSION_GOLDEN_v2.json` as current, with v1 described as audit-only provenance. A test asserts every authoritative document names the current bundle, and that a superseded bundle is never mentioned without the current one beside it.

The deeper cause was that the handoff §9 states the supersession *rule* without saying which pointers move with it — which is exactly how the drift happened. That procedure is now written down as §0 of the generator spec: bump the version, seal a **new** bundle, add coverage to the canonical case set, move `CURRENT_GOLDEN_BUNDLE` and every document naming a bundle, record the superseded one with its file digest.

It also records the inference that was easy to miss here: **if no existing row changed, the new coverage case is missing — not the change harmless.** That was true of pec-v2 and I nearly took it as reassurance.

### [P2] Sealed-bundle lifecycle now enforced, not described

All four sub-points verified against the code before fixing:

- **Canonical source.** `REQUIRED_COVERAGE_CASES` lives in the module, so the committed bundle is reproducible from code rather than from flags typed once — the generator default really did produce 9 rows against a committed 20. A test rebuilds from the module and compares `golden_sha256`; regeneration is byte-identical.
- **No-clobber.** The generator refuses to overwrite an existing target (`--force` to override), since rewriting a sealed bundle is the precise failure the rule exists to prevent.
- **No silent override.** `--extra` may no longer replace a registered name.
- **Independent integrity.** Confirmed the gap: `verify_golden` rejects on `contract_version` before reaching `golden_sha256`, so a token-corrupted v1 raised the identical error to an intact one. Superseded bundles are now pinned by file digest, with a test that demonstrates the gap rather than just closing it.

### [P2] Grading unit and schedule

Correct — the formula said per row while requirement 1 said per field, which are different estimators. Now stated once: **per numeric field, resampled every step.**

The schedule point is the one worth keeping visible. A permanent assignment is *learnable* — the decoder could discover which examples are never exactly graded and relax on those specifically, which defeats the mechanism's central property. Fresh draws keep "cannot tell which will be strictly graded" true throughout training rather than only at initialization. Both choices are now argued in the spec rather than asserted.

### [P3] Hard-coded tolerance

Correct: `test_missing_digits_cap_resolution_below_the_stated_tolerance` still used `0.005`, and since the inequality holds under either value the test no longer demonstrated its own claim. It now reads `_spec_tolerances()["routing.t"]`.

---

Thanks for the ruling on finding 5 — the diagnostic-only deferral with a named step-3 obligation is what I'd hoped was the right balance, and it's useful to have that confirmed rather than assumed.

---

PR #4005, opened 2026-07-25
mu_cosine: freeze SM-FS lineage ranking and retention-transfer protocols

s243a:
## Summary

Freeze the two statistical protocols required to advance the first SM-FS positive-only training diagnostic:

1. Outcome-blind graded-negative construction for a genuine lineage-ranking evaluation.
2. Separate retention and transfer experiments using matched arms derived from frozen checkpoints.

This PR contains protocols, content-bound preregistrations, executable validators, deterministic samplers, and enforcement tests. It contains no new ranking or transfer result.

## Ranking protocol

- Uses only the 361-map SM-FS exploration split.
- Leaves all 1,481 reserved rows untouched and unscored.
- Freezes 126,807 exhaustive non-ancestor candidates.
- Defines exact graph-derived targets and hard, medium, and easy negative strata.
- Freezes five lineage-family folds.
- Defines a deterministic sampler with known-answer test vectors.
- Compares graded-negative training against a retrained positive-only arm.
- Uses held MRR with a paired lineage-block bootstrap.
- Requires at least a `+0.01` MRR gain with the 95% interval excluding zero.

## Retention and transfer protocol

Retention and transfer remain separate questions:

- **Retention:** Does adding SM-FS supervision preserve the existing Pearltrees behavior?
- **Transfer:** Starting before Pearltrees training, does SM-FS improve performance on a fresh Pearltrees cohort?

The protocol:

- introduces a matched structure-scrambled supervision null;
- derives each comparison arm from the same applicable base checkpoint;
- requires a fresh, certified-public Pearltrees cohort for confirmatory transfer;
- excludes historical examples by both pearl identity and canonical resource hash;
- clusters duplicate resources in typed resource/folder bootstrap blocks;
- serially gates total transfer, SM-FS-specific transfer, and noninferiority to e5.

## Pilot interpretation correction

The `+0.82` LINEAGE validation correlation is a successful onboarding diagnostic: the model learned shared SM-FS hierarchy structure on map-disjoint examples.

It is not yet:

- ranking evidence;
- unseen-lineage generalization;
- cross-corpus transfer evidence.

Most validation rows reuse ancestor structures observed during training, and an ancestor-lookup baseline reaches approximately `+0.63`.

The previously reported `+0.37` warm-start correlation is not treated as supported evidence because the run receipt did not hash or evaluate that warm start. Evaluation of the identified current warm-start checkpoint instead produced `−0.112`.

## Execution boundary

Negative-ledger construction may proceed after merge. Model fitting remains blocked until an amendment content-binds:

- exact starting checkpoints and parameter-growth procedure;
- positive, negative, and scrambled-null ledgers;
- candidate catalog and text/embedding tables;
- training code and configuration;
- optimizer, trainable parameters, anchoring, and augmentation;
- numerical environment and seed schedule;
- fresh transfer cohort and retention behavior panel.

This prevents implementation choices from being selected after observing ranking or transfer outcomes.

## Validation

- 170 relevant tests passed.
- Both preregistrations pass standalone hash verification.
- Fold assignment and sampling are deterministic.
- Sampler known-answer vectors are enforced.
- Python compilation and staged-diff checks pass.
- Independent statistical, implementation, and source-grounding reviews found no blocking issues.

### DIFF
--- a/prototypes/mu_cosine/PROTOCOL_sm_fs_lineage_ranking.md
+++ b/prototypes/mu_cosine/PROTOCOL_sm_fs_lineage_ranking.md
@@ -1,0 +1,336 @@
+# Prospective protocol: SM-FS LINEAGE negative construction and ranking
+
+**Status:** frozen after the positive-only onboarding run and before constructing any negative
+ledger, fitting a negative-trained checkpoint, or scoring a ranking metric. This is an
+**exploratory, catalog-transductive mechanism study** over the 361 certified exploration maps. It
+does not open or score the 1,481 reserved maps. A material change requires a new protocol version
+and preregistration ID before an affected score is computed.
+
+The positive-only run remains useful, but its evidence is narrower than initially reported. The
+trained checkpoint reproduces a validation correlation near `+0.82`; the exact currently available
+warm start instead reproduces `-0.112`, not the reported `+0.37`, and the run receipt did not hash
+or evaluate its initialization. Moreover, 356 of 384 validation rows reuse an ancestor path seen
+in training and an ancestor-mean lookup reaches correlation `+0.630`. The result therefore shows
+shared-hierarchy interpolation on unseen maps. It is not a ranking, unseen-lineage, or transfer
+result and supplies no decision-bearing observation for this protocol.
+
+## 1. Decision and estimand
+
+The question is whether topology-derived non-ancestor supervision improves exact-destination
+ranking relative to the same model trained on positive ancestors only.
+
+For held exploration map \(q\) and training seed \(s\), let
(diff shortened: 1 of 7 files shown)


s243a:
Cross-lane verification (fable): I independently reproduced both audit claims about the positive-only pilot before endorsing — true warm-start (0-step) val corr **−0.112** (my reported +0.37 was contaminated by the 5-step smoke run; accepted), shared-ancestor val rows **356/384**, ancestor-mean lookup corr **+0.630** (exact match). The +0.82 downgrade to shared-hierarchy interpolation is correct and accepted. Protocol tests 40/40 under py3.10/numpy2. Recommend merge as-is.

Follow-up on my side (post-merge, non-blocking): patch fine_tune_sm_fs.py so the .run.json receipt hashes the initialization checkpoint and records a step-0 evaluation, per §2's binding rules — closing the gap this audit exposed.

PR #4014, opened 2026-07-26
PERF-R-CA-IDTABLE: direct in-memory integer adjacency table for R CA

s243a:
## Summary

Add an optional integer adjacency table to in-memory indexed FactSource capabilities so the R category-ancestor loop reads ordered parent-ID vectors directly instead of invoking `lookup_ids` about 516,000 times per scale-300 query.

The table is materialized **lazily on first CA use**. Registration is emitted for every in-memory FactSource with arity ≥2, so unrelated predicates do not pay the table's setup or memory cost. Lazy/cached LMDB remains on demand and receives no `id_table`. TermValue and generic `iterate_goal` fallbacks are unchanged.

## Profiling

Post-STACK/PATHMARK:

- approximately 516,522 DFS node visits;
- approximately 2,273 distinct IDs;
- approximately 99.56% lookup-cache-hit opportunity;
- cached `lookup_ids` approximately 6.9% self / 19.9% total in Rprof;
- 516k-call microbenchmark: cached closure approximately 0.41 s, direct list reads approximately 0.055 s;
- wrapping the table in a per-node closure remained approximately 0.35 s, confirming closure dispatch as the target.

PATHMARK membership was not revisited.

## Design

- `make_indexed_arg1_parent_id_table`: bounded dense list plus sparse overflow.
- `make_lazy_indexed_arg1_parent_id_table`: one-time materialization state.
- `category_ancestor` resolves the table once per kernel invocation.
- `category_ancestor_hops_ids` uses direct list access on dense hits.
- Existing `lookup` and lazy-memo `lookup_ids` APIs remain available.
- Capability normalization is additive; legacy bare functions receive `id_table = NULL`.
- Dense capacity is derived from fact atom IDs and capped at 65,536, never from a query-time ID.
- Helpers remain inline in `runtime.R.mustache`; no new template partial.

Review follow-up: the original branch built a table eagerly for every registered arity≥2 in-memory FactSource. The corrected implementation leaves the capability unbuilt after registration and materializes it only if CA requests it. The focused test asserts both states.

## Performance

Same-host Cursor cloud agent, R 4.3.3, three repetitions:

| Stage | query median | total median | query samples |
|---|---:|---:|---|
| STACK | 2729 ms | 3255 ms | 3299, 2729, 2661 |
| IDTABLE | 2607 ms | 3186 ms | 3158, 2607, 2601 |

Hosted Ubuntu 24.04, R 4.3.3:

- IDTABLE samples: `4078, 3412, 3268 ms`;
- median query/total: **3412 / 4076 ms**;
- hosted STACK baseline: `3679 / 4344 ms`;
- improvement: **1.08× query**, **1.07× total**;
- 271 rows with canonical six-decimal reference parity;
- about **18.3×** query improvement over the original hosted R result of 62,595 ms.

The table build is approximately 0.03 s and approximately 151–154 KiB at scale 300. Post-change Rprof removes `lookup_ids` from the hot list; `hops_ids` is approximately 25% self / 32% total.

The temporary hosted measurement step was removed after recording the result.

## Compatibility and tests

The existing focused R CA harness covers:

- table/`lookup_ids` parity;
- lazy table materialization;
- duplicates and exact parent order;
- cycles, diamonds, roots, depth boundaries, and mixed/non-atom `Visited`;
- empty buckets and sparse/high IDs;
- legacy capability normalization;
- LMDB laziness markers;
- TermValue and generic fallbacks;
- exact unsorted three-path differential parity.

Validation:

```bash
swipl -g run_tests -t halt tests/test_wam_r_ca_direct_lookup.pl
```

The hosted R lane passes the focused regressions, ISO smoke, effective-distance smoke, direct arg1 suite, benchmark parity, and full classic conformance.

## Change size

- Production runtime: +137 / −28, net +109.
- Focused existing test harness: +55 / −11, net +44.
- Remaining changes are benchmark/status documentation.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -265,6 +265,14 @@ jobs:
         run: |
           swipl -q -f init.pl -s tests/test_wam_r_ca_direct_lookup.pl -g "run_tests, halt" -t "halt(1)"
 
+      - name: Measure PERF-R-CA-IDTABLE scale-300 timing
+        run: |
+          mkdir -p /tmp/wam-bench/r-ed-300
+          swipl -q -s examples/benchmark/generate_wam_r_effective_distance_benchmark.pl -- \
+            data/benchmark/300/facts.pl /tmp/wam-bench/r-ed-300 kernels_on functions
+          ( cd /tmp/wam-bench/r-ed-300/R && \
+            Rscript run_effective_distance.R "$GITHUB_WORKSPACE/data/benchmark/300" 3 )
+
       - name: Run full R classic conformance
         run: |
           swipl -q -f init.pl -s tests/test_wam_cross_target_conformance.pl -g "run_tests" -t halt
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -265,14 +265,6 @@ jobs:
         run: |
           swipl -q -f init.pl -s tests/test_wam_r_ca_direct_lookup.pl -g "run_tests, halt" -t "halt(1)"
 
-      - name: Measure PERF-R-CA-IDTABLE scale-300 timing
-        run: |
-          mkdir -p /tmp/wam-bench/r-ed-300
-          swipl -q -s examples/benchmark/generate_wam_r_effective_distance_benchmark.pl -- \
-            data/benchmark/300/facts.pl /tmp/wam-bench/r-ed-300 kernels_on functions
-          ( cd /tmp/wam-bench/r-ed-300/R && \
-            Rscript run_effective_distance.R "$GITHUB_WORKSPACE/data/benchmark/300" 3 )
-
       - name: Run full R classic conformance
         run: |
           swipl -q -f init.pl -s tests/test_wam_cross_target_conformance.pl -g "run_tests" -t halt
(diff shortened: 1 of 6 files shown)


s243a:
Review complete. The direct-table optimization is supported by the profile and by both same-host and hosted measurements, but I found one scope issue that needed correction before merge.

The emitter registers an arg1 capability for every in-memory FactSource with arity >= 2. The original branch eagerly built an adjacency table during each registration, so unrelated predicates would have paid the table's setup and memory cost even if no CA kernel referenced them. I changed `id_table` to a lazy state: registration remains cheap, `category_ancestor` resolves/materializes it once on first CA use, and subsequent DFS calls retain direct list access. The existing lazy-memo `lookup_ids` API remains unchanged for non-CA callers. The focused test now asserts that the table is unbuilt immediately after registration and built after the first explicit request.

I also:
- merged current `main` (the intervening changes were unrelated);
- removed an unused thin table-reader helper and a duplicate cached-miss assertion;
- removed the temporary benchmark workflow step after measurement;
- promoted the hosted IDTABLE numbers into the R fleet/status/benchmark docs;
- updated the PR description and line counts.

Hosted result on the corrected runtime/test head:
- query samples: `4078, 3412, 3268 ms`;
- median query/total: `3412 / 4076 ms`;
- 271-row canonical parity;
- 1.08x query and 1.07x total improvement over hosted STACK `3679 / 4344 ms`;
- about 18.3x query improvement over the original hosted R result.

The complete hosted R lane is green: focused regressions, ISO smoke, effective-distance smoke, direct arg1 contract suite, hosted benchmark, and full R classic conformance. The final cleanup head only removes the temporary measurement and updates documentation/comments; its repository-wide workflow is queued.

The PR is mergeable. I found no remaining code-level blocker.

PR #4019, opened 2026-07-26
mu_cosine: experimental vNext process-expression frontend kernel

s243a:
## Summary

First executable milestone of [`DESIGN_process_expression_patterns.md`](../blob/main/prototypes/mu_cosine/DESIGN_process_expression_patterns.md):

```text
functional expression
    -> elaboration-preserving source AST  (functional_parser)
    -> registry-driven elaboration        (elaborator)
    -> in-memory typed ground term        (ast)
```

**This is an experimental frontend kernel, not vNext activation.** It mints no identity, emits no canonical bytes, implements no `pe-typed-ast-v1` serialization, and exposes no resolution, verification, or deployment surface. Registry `v0.3`, `pec-v2`, and both sealed golden bundles are frozen legacy contracts — verified byte-identical.

The Prolog adapter is deliberately **not** implemented here; cross-surface equivalence should follow once frontend semantics and the wire schema are stable.

## Architecture and state boundary

The parse/elaborate seam is the point of the milestone, so the halves are kept strictly apart:

- **`functional_parser.py`** — an **elaboration-preserving** source AST: spans, positional order, *every* named-field occurrence in source order including duplicates, exact numeric lexemes, unchecked annotations. It does **not** promise exact source reconstruction — grouping, trivia and original literal spellings are not retained — and says so, because a guarantee the implementation cannot meet is worse than an accurate one. It sorts and de-duplicates nothing, since either would hide the duplicate-field error elaboration must raise.
- **`elaborator.py`** — resolves against an injected registry; validates arity and named fields; rejects duplicate, unknown, missing and ignored fields; inserts defaults; unifies indexed ownership; checks `::` as an assertion, never a conversion.
- **`ast.py`** — two distinct trees. Typed nodes are immutable with structural equality; spans and successfully-checked annotations are absent, so `pearltrees` and `pearltrees::corpus` are the same term. Value indices (`ReferenceIndex`, experimental `TermIndex`) are a **separate hierarchy from types**, so `substrate[pearltrees]` is not a type named `pearltrees`. `debug_repr` is labelled noncanonical and non-identity-bearing.
- **`numerics.py`** — lexemes stay text until the declared type decides (§3.6). Converting through `float` during parsing would round a decimal the specification requires to survive *and* erase the signed-zero distinction `float64` must preserve.
- **`registry.py`** — loads only an injected fixture, never falling back to v0.3. Borrowing v0.3 signatures would make this frontend look more validated than it is and would smuggle back the retired `source` catch-all that §3.2 drops.

Public seam:

```python
surface = parse_functional(text, registry)
typed   = elaborate(surface, registry)
```

A test asserts the package exports nothing whose name contains `digest`, `hash`, `identity`, `deploy`, `resolve`, `verify`, or `receipt`.

## Changed files

Seven new package/test paths, plus one modification:

```
A  process_expression_vnext/{__init__,ast,functional_parser,registry,numerics,elaborator}.py
A  process_expression_vnext/testdata/frontend_registry_fixture.json
A  test_process_expression_vnext_frontend.py
M  .github/workflows/test.yml          # runs the frontend + lightweight legacy suites
```

## Registry fixture

`status: experimental-test-fixture`, `production: false`, and **no** `registry_version` — versions belong to sealed release registries, and the loader actively rejects anything resembling one.

Signatures follow the specification rather than convenience: `lineage_at` is the `relation_family` reference, `lineage_op(S,…)` is `abstract_lineage_process[S]`, and `hop_decay_targets` is `target_factory[S,hop_decay]`. All four of §4.1's coarse-to-precise examples are executable, and the coarse form inserts **no** defaults, because hop-decay-shaped defaults would silently answer the interpretation question `lineage_op(S)` exists to leave open. Synthetic test mechanics live in explicitly named `default_probe` / `list_probe` / `order_probe` / `rebinding_probe` entries so they cannot be mistaken for lineage semantics.

## Tests and sealed hashes

```
python3 -m pytest -q \
  test_process_expression_vnext_frontend.py \
  test_process_cards.py \
  test_process_expression_contract.py \
  test_process_expression_envelope.py \
  test_process_expression_p1_protocol.py
-> 184 passed        (75 frontend)
```

`git diff --check` and `compileall` clean.

| file | sha256 | |
|---|---|---|
| `PROCESS_EXPRESSION_GOLDEN_v1.json` | `b053351a2a419ac58b7ab644afe15c60543846ce8b9d5a3d9bcbc332ca24db29` | ✅ |
| `PROCESS_EXPRESSION_GOLDEN_v2.json` | `85e6421f5a1347fca5937d1243dc01500a9aa5b7221571b4918248e57ece6344` | ✅ |

Behaviors worth singling out:

- `margin(t=1)` elaborates to **`real`**, not `int` — the declared type wins over the spelling;
- `0.1234567890123456789012345` survives exactly, and the test proves the `float`-rounded form differs;
- `-0.0` normalizes to zero as a `real` but keeps `8000000000000000` as a `float64`, distinct from `0000000000000000`;
- `01` is **accepted** and normalizes to `1`, matching v0.3 — forbidding it would be a specification decision;
- `cross_substrate(principal_tree(pearltrees), principal_tree(simplewiki))` fails because `C` binds once, and a `binds` declaration cannot overwrite a constraint unification already established;
- resource ceilings are controlled errors: `1e1000000000` is refused without allocating it, the host integer-conversion limit is translated rather than leaked, and debug rendering of a real is bounded;
- variables, modifiers, pins and function types are rejected with *"not implemented in this milestone"* rather than misparsed.

## Deferred contract decisions

Recorded rather than silently resolved. Each would change type meaning, numeric normalization, source-to-semantic mapping, or future canonical bytes.

**1. Wire form of an expression-valued index.** `abstract_lineage_process[S]` needs `S` to be a substrate *expression*. `TermIndex` carries it in memory so the correct signature is expressible, but it has no serialization and cannot be written in an annotation — that would presume the very decision this milestone must not make. *Smallest example:* `lineage_op(principal_tree(pearltrees))`.

**2. Canonical `real` string form.** §2.1 rule 8 requires "canonical exact-decimal strings" without defining one. Equality here is on the normalized `(sign, digits, exponent)` triple; rendering is explicitly noncanonical and bounded. *Recommendation:* plain decimal with trailing zeros stripped, plus an exponent threshold. *Smallest example:* `1e0`, `1`, `1.0`.

**3. How an index binds to an argument.** The specification shows the result type, not the mechanism. Resolved here by an explicit fixture field `binds: {"C": 0}`. *Open:* whether a nested `Call` may serve as an index — currently yes, via the experimental `TermIndex`.

**4. May a positional argument follow a named field?** The EBNF permits interleaving; this rejects it. *Smallest example:* `lineage_op(decay=0.5, principal_tree(pearltrees))`.

**5. Are registry defaults surface text or pre-typed values?** Text here, elaborated through the same path as authored values, so explicit and elided provably cannot diverge — and so a JSON number cannot round before elaboration sees it.

**6. Is `list` a reserved constructor?** Reserved here. *Smallest example:* a registry entry named `list`.

## Non-goals honored

No changes to `process_cards.py` or v0.3; no variables, `PatternAST`, grounding, or alpha-normalization; no Prolog parsing or execution; no `pe-typed-ast-v1`, AST hashes, or identity; no interpretation/representation rules; no candidate collection, selectors, or aggregation; no resolution or verification receipts; no factory realization; no migration manifests; no `pec-v3` or v3 golden bundle; no encoders, decoders, training, or deployment.

https://claude.ai/code/session_017xCXhV5sHDsMiDyU5q7Hqg

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -117,7 +117,10 @@ jobs:
         # corpus either way.
         run: |
           python3 -m pytest -q prototypes/mu_cosine/test_process_expression_contract.py \
-            prototypes/mu_cosine/test_process_expression_envelope.py
+            prototypes/mu_cosine/test_process_expression_envelope.py \
+            prototypes/mu_cosine/test_process_cards.py \
+            prototypes/mu_cosine/test_process_expression_p1_protocol.py \
+            prototypes/mu_cosine/test_process_expression_vnext_frontend.py
 
       - name: Run F# WAM target tests (codegen + dotnet smoke + LMDB conformance)
         env:
--- a/prototypes/mu_cosine/process_expression_vnext/__init__.py
+++ b/prototypes/mu_cosine/process_expression_vnext/__init__.py
@@ -1,0 +1,38 @@
+#!/usr/bin/env python3
+"""Experimental vNext process-expression frontend kernel.
+
+Proves the parser/elaborator seam described in
+``DESIGN_process_expression_patterns.md``:
+
+```text
+functional expression
+    -> lossless source AST        (functional_parser)
+    -> registry-driven elaboration (elaborator)
+    -> in-memory typed ground term (ast)
+```
+
+**This is not vNext activation.** It creates no identity-bearing bytes, no
+canonical ``pe-typed-ast-v1`` serialization, no digests, no receipts, and no
+deployed processes. Registry ``v0.3``, ``pec-v2``, and both sealed golden
+bundles are untouched frozen contracts, not scaffolding this package extends.
+
+The public seam is deliberately two functions. Hashing, identity, deployment,
+resolution, and verification are *not* exported, because exporting them would
+invite exactly the premature identity-minting the state machine in §1 forbids.
+"""
+from __future__ import annotations
+
+from .elaborator import ElaborationError, elaborate
+from .functional_parser import NotImplementedInMilestone, ParseError, parse_functional
+from .registry import Registry, RegistryError, load_registry
+
+__all__ = [
+    "parse_functional",
+    "elaborate",
+    "load_registry",
+    "Registry",
+    "ParseError",
+    "NotImplementedInMilestone",
+    "ElaborationError",
+    "RegistryError",
+]
--- a/prototypes/mu_cosine/process_expression_vnext/__init__.py
+++ b/prototypes/mu_cosine/process_expression_vnext/__init__.py
@@ -6,10 +6,13 @@ Proves the parser/elaborator seam described in
 
 ```text
 functional expression
-    -> lossless source AST        (functional_parser)
-    -> registry-driven elaboration (elaborator)
-    -> in-memory typed ground term (ast)
+    -> elaboration-preserving source AST  (functional_parser)
+    -> registry-driven elaboration        (elaborator)
+    -> in-memory typed ground term        (ast)
 ```
+
+The source tree keeps what elaboration and diagnostics need; it does not
+promise exact source reconstruction.
 
 **This is not vNext activation.** It creates no identity-bearing bytes, no
 canonical ``pe-typed-ast-v1`` serialization, no digests, no receipts, and no
--- a/prototypes/mu_cosine/process_expression_vnext/ast.py
+++ b/prototypes/mu_cosine/process_expression_vnext/ast.py
@@ -3,14 +3,21 @@
 
 Two distinct trees, deliberately:
 
-* **Source AST** — lossless with respect to the text.  Keeps spans, positional
-  order, *every* named-field occurrence including duplicates, exact numeric
-  lexemes, and unchecked annotations.  Sorting or de-duplicating fields here
-  would hide the duplicate-field error the elaborator must raise, so the parser
-  never does either.
+* **Source AST** — elaboration-preserving.  Keeps spans, positional order,
+  *every* named-field occurrence including duplicates, exact numeric lexemes,
+  and unchecked annotations.  Sorting or de-duplicating fields here would hide
+  the duplicate-field error the elaborator must raise, so the parser never does
+  either.
 * **Typed AST** — immutable semantic nodes with structural equality.  Spans and
   successfully-checked annotations are absent, so ``pearltrees`` and
   ``pearltrees::corpus`` are the same term (§3.5).
+
+The source tree is **elaboration-preserving**, not lossless: it keeps
+everything elaboration and diagnostics need, but it does not promise exact
+source reconstruction.  Grouping parentheses, whitespace and comments
+are not retained, and atom/string nodes hold decoded values rather than original
+spellings.  The narrower claim is deliberate — a "lossless" guarantee this
+implementation cannot satisfy would be worse than an accurate one.
 
 Nothing here is identity-bearing.  ``DESIGN_process_expression_patterns.md``
 §2.1 freezes the canonical wire schema ``pe-typed-ast-v1`` separately, and this
@@ -40,7 +47,7 @@ class Span:
 
 
 # --------------------------------------------------------------------------
-# source AST — lossless, span-carrying, unchecked
+# source AST — elaboration-preserving, span-carrying, unchecked
 # --------------------------------------------------------------------------
 
 
@@ -129,6 +136,17 @@ class SourceFunctionType(SourceType):
 
 
 @dataclass(frozen=True)
+class SourceReferenceIndex(SourceType):
+    """A registered reference occupying an index position, e.g. ``[pearltrees]``.
+
+    Kept distinct from :class:`SourceTypeName` so the elaborator can produce a
+    :class:`ReferenceIndex` rather than conflating it with a type.
+    """
+
+    name: str
+
+
+@dataclass(frozen=True)
 class Type:
     """Semantic type.  Structural equality; no spans."""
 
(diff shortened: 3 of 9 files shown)


s243a:
Updated in place as `daa66f8`. All eleven findings **fixed**; none deliberately rejected, none blocked. 63 frontend tests, **172 passed** across the required command.

### 1. Value indices separated from types ✅

New `ValueIndex` hierarchy: `ReferenceIndex` for a registered reference, plus an explicitly `EXPERIMENTAL` `TermIndex` for an expression-valued index. `substrate[pearltrees]` now carries `ReferenceIndex("pearltrees")`, not a type named `pearltrees`.

Index positions use the same longest-match lexing as expressions, so both required cases work:

```
substrate[pearltrees]           -> ReferenceIndex("pearltrees")
judge_scope[gpt-5.5-low]        -> ReferenceIndex("gpt-5.5-low")   # one name, not -5.5-low
```

Rather than fake the expression case, `TermIndex` is in-memory only with no serialization, and an expression-valued index **cannot be written in an annotation** — that would presume a wire form nobody has approved. It raises "open specification decision" instead.

### 2. Fixture semantic types corrected ✅

The conflation is gone. Per §6/line 675:

```
lineage_at          -> relation_family                       (reference)
lineage_op(S,...)   -> abstract_lineage_process[S]
hop_decay_targets   -> target_factory[S,hop_decay]
```

Both indexed results are now *expressible* because `S` is carried as a `TermIndex`, so the representative signatures survive rather than being replaced by neutral stand-ins. `lineage_op(haiku)` remains a negative test and no longer asserts a false result type.

### 3. Ownership bindings preserved ✅

A `binds` declaration can no longer overwrite a constraint argument unification already established. Added a `rebinding_probe` fixture entry that binds `C` from argument 0 by unification, then attempts to rebind from argument 1:

```
rebinding_probe(principal_tree(pearltrees), pearltrees)   ok
rebinding_probe(principal_tree(pearltrees), simplewiki)   "would rebind index C"
```

### 4. Defaults type-checked ✅

Inserted defaults go through the same expected-type path and the same `_unify` call as authored fields. A default whose type is still unresolved is **rejected** rather than made order-dependent — no dependency solver, per your allowance. Three tests: explicit ≡ elided, ill-typed default rejected, unresolved-type default rejected.

### 5. No `TypeVar` escapes ✅

Both belts: references declaring unbindable variables are rejected at **registry load** (`entity[S]` on a reference has no arguments to bind `S`), and the final term is validated **recursively** on the way out of `elaborate()`.

### 6. Registry loading fails closed ✅

Duplicate-aware decoding via `object_pairs_hook`, unknown-key rejection at top level / entry / field / bind, string type declarations, list `args`, boolean `required`, and defaults restricted to JSON string or null.

That last one earns its place: accepting a JSON number and calling `str()` on the resulting Python float would round `0.1234567890123456789012345` *before elaboration ever saw it* — silently defeating the exact-decimal guarantee. Both your named negative cases are tested.

### 7. Lone surrogates rejected ✅

Checked after JSON decoding, before `SourceString` construction, so every typed `String` stays UTF-8 encodable. Valid pairs survive (`"😀"` → 😀); lone high, lone low, and embedded lone surrogates all raise.

### 8. Leading-zero rule removed ✅

You were right, and I verified it: `process_cards.parse("margin(t=01)")` canonicalizes to `margin(t=1)`. v0.3 accepts it, the vNext grammar says only "a finite decimal numeric token", and forbidding it is a specification decision. The test now asserts `01` is **accepted** and normalizes identically to `1`.

### 9. Integer resource use bounded ✅

A digit ceiling (`MAX_INT_DIGITS`, an experimental frontend limit, not a language contract) is checked **before** the value is materialized, so `1e1000000000` is refused without allocating it. CPython's integer-string conversion limit can no longer escape as a raw `ValueError`. Tested for both a long literal and a short lexeme with a catastrophic exponent.

### 10. Span and losslessness contract corrected ✅

Parenthesized expressions span their full syntactic extent; trailing trivia no longer lands inside a name token.

```
pearltrees          [0:10]
(pearltrees)        [0:12]
(pearltrees)::corpus[0:20]
"  pearltrees  "    ends at 12, not 14
```

The public claim is narrowed from "lossless source AST" to **elaboration-preserving**, stated in both module docstrings, with a test asserting the phrase "lossless source AST" no longer appears. Grouping, trivia and original literal spellings genuinely are not retained, so the accurate claim is the right one.

### 11. CI runs the suite ✅

The process-expression step now runs the frontend suite plus `test_process_cards.py` and `test_process_expression_p1_protocol.py` alongside the two already there.

---

### Validation

```
pytest -q  (the five required files)      -> 172 passed
git diff --check                          -> clean
compileall                                -> clean
```

| sealed bundle | sha256 | |
|---|---|---|
| `..._GOLDEN_v1.json` | `b053351a…ca24db29` | ✅ unchanged |
| `..._GOLDEN_v2.json` | `85e6421f…7ece6344` | ✅ unchanged |

### Deferred decisions, updated

The wire form of an **expression-valued index** is now the sharpest open item — `TermIndex` exists so the correct signatures are expressible in memory, but its serialization is exactly the decision this milestone must not make. The other five from the original PR body stand, minus the index-representation one, which finding 1 resolves for the reference case.

All non-goals held: no Prolog adapter, canonical bytes, identity, resolution, receipts, migration, `pec-v3`, encoder, decoder, or deployment work.

---

s243a:
Pushed `f961675`. All ten findings **fixed**; none rejected, none blocked. **184 passed** (75 frontend). PR description rewritten — it was stale on all four counts you listed.

### 1. Registry and surface agree on a value index ✅ [P1]

The registry type parser now accepts punctuated index names, and a **second pass after all entries load** rewrites index-position type names that actually name a registered reference into `ReferenceIndex`.

The second pass is necessary rather than stylistic: whether a bare index token denotes a type (`hop_decay`) or a value (`pearltrees`) cannot be known while an individual signature is parsed, because the full name set does not exist yet. Your `fixed_tree` case is now a regression test — it loads *and* annotates, so the self-contradictory `expected substrate[pearltrees], found substrate[pearltrees]` is gone. `parse_type("judge_scope[gpt-5.5-low]")` works.

### 2. `ValueIndex` can no longer enter `ListType.element` ✅ [P1]

Two gaps, both closed: `list[T]` requires a genuine `Type`, and a `SourceReferenceIndex` naming a **callable** is rejected outright. Both of your cases now fail:

```
[]::list[pearltrees]        -> "list element type must be a type, not the value pearltrees"
[]::list[principal_tree]    -> "'principal_tree' is a callable and cannot be used as a value index"
```

### 3. Default resolution is order-independent ✅ [P1]

Authored fields are processed in a first pass, defaults in a second. The `order_probe` fixture entry declares the defaulted `x` **before** the required `s` that binds `S` — the awkward order deliberately — and the test additionally swaps the JSON members and asserts an identical result.

### 4. Debug rendering bounded ✅ [P1]

`plain_string` fell back from fixed-point to scientific past a ceiling. Verified against your probe values: `1e100000` and `1e-100000` were 100,001 and 100,002 characters and are now short, while `0.1234567890123456789012345` still renders plainly — so exactness stays visible where it matters.

### 5. Fixture matches the specification ✅

`lineage_op` and `hop_decay_targets` now carry the §4.1/§13 fields, so all four coarse-to-precise examples execute. `weights` moved out. The coarse form inserts **no** defaults — your sharpest point here, since hop-decay-shaped defaults would silently answer the interpretation question `lineage_op(S)` exists to leave open. Synthetic mechanics live in `default_probe` / `list_probe` / `order_probe` / `rebinding_probe`, named so they cannot be read as lineage semantics.

### 6. References cannot carry discarded declarations ✅

Presence, not truthiness, and `binds` included. `{"args": [], "fields": {}, "binds": "garbage"}` now fails with "cannot declare call-only key(s)".

### 7. Atom surrogates ✅

Same Unicode-scalar guard as strings. Canonical encoding covers atom values, so the asymmetry was a real hole.

### 8. Host `ValueError` translated ✅

You were right that the 4096 ceiling only *happened* to sit below CPython's 4300 default. With `sys.set_int_max_str_digits(640)` a 700-digit integer now raises `NumericError`, not a raw host error.

### 9. Span/documentation cleanup completed ✅

A simple type name closes its span before lookahead whitespace, so `pearltrees::corpus  ` ends at 18. The package overview and the exported `parse_functional` docstring no longer promise a lossless AST, and the test now covers **all four** public docstrings instead of two — that gap is exactly why the two remaining claims survived the first pass.

### 10. Trivia no longer downgrades the deferral ✅

The expression-index lookahead uses ordinary trivia handling, so `abstract_lineage_process[principal_tree (pearltrees)]` gets the "open specification decision" diagnostic rather than a generic `expected ']'`.

---

### Validation

```
required five-file suite   -> 184 passed  (75 frontend)
git diff --check           -> clean
compileall                 -> clean
```

| sealed bundle | | |
|---|---|---|
| `..._GOLDEN_v1.json` | `b053351a…ca24db29` | ✅ unchanged |
| `..._GOLDEN_v2.json` | `85e6421f…7ece6344` | ✅ unchanged |

### PR description

Rewritten. It no longer calls the AST lossless, no longer claims all files are additions (CI is modified), drops the retired `TypeName` index workaround, and reports 184/75. The deferred list is re-ranked: the **wire form of an expression-valued index** is now the leading open decision, since `TermIndex` exists precisely to keep that question open rather than answer it.

All non-goals held.

---

